PR #1, opened 2020-06-03
Move proposals from strimzi-kafka-operators repo to new proposals repo

scholzj:
As part of https://github.com/strimzi/strimzi-kafka-operator/blob/master/design/github-repository-restructuring.md, this PR moves the proposals into the new repository.


scholzj:
@ppatierno Ok, I added `00` as padding ... would that be enough? ;-)

PR #2, opened 2020-06-12
Added proposal/discussion about HTTP bridge metrics

ppatierno:
Signed-off-by: Paolo Patierno <ppatierno@live.com>


005-http-bridge-metrics.md
```diff
@@ -0,0 +1,75 @@
...
+* Created consumers
+    * Could bring labels like “consumergroup”, “instanceid”, “base_uri” (not sure if this one is really useful). It could increase a consumers counter on the dashboard.
+* Subscription (requests)
+    * Could bring labels as “topics” but it should be a list, how to leverage that? What when the client uses “topic_pattern” and not specific topics in the request?
+* Deleted consumer
+    * Could bring labels like “consumergroup”, “instanceid”. It could decrease the previous consumers counter on the dashboard.
+* Poll (requests)
+    * Could bring labels like “consumergroup”, “instanceid”, number of “records” polled
+* Send (requests)
+    * Could bring labels like “topic” and number of “records” sent
```
scholzj, line 67: Do we have to add these? Can't we just expose the metrics from the Kafka client?

ppatierno: That could be a way but then we need to expose them to Prometheus. The way we do this for Kafka, ZK etc is using JMX exporter which starts an HTTP server by itself. The bridge has its own HTTP server which already exposes the HTTP related metrics on the `/metrics` endpoint. It sounds to me that JMX exporter is not usable (if we don't want to have another running HTTP server listening on different port ... but then it means Prometheus scraping the bridge metrics on two HTTP servers ... sounds weird and not sure if possible). Maybe we could access the JMX beans programmatically and exposing them on the bridge HTTP server in a Prometheus format.

ppatierno: I had a try using the `JmxCollector` library provided by the official Prometheus JMX exporter and I was able to get Kafka clients related metrics and then publishing them through the HTTP server `/metrics` endpoint avoiding the need for JMX exporter as an agent or as a standalone HTTP server. Without any configuration it exposes ALL the metrics but I guess that at this point it could be useful providing the bridge a YAML JMX config file to filter metrics. It would do in a `metrics` section in the `KafkaBridge` custom resource as well. @scholzj wdyt?

scholzj: So, does it provide the Kafka metrics together with the Vert.x based metrics?

ppatierno: I get metrics from two different sources, the `PrometheusMeterRegistry` from Vert.x with the HTTP server ones and my JMX to Prometheus stuff for the Kafka side and I can chain them together.

scholzj: Ok ... so they are all in the same file / endpoint, right? i wonder how weird it would be if the `.spec.metrics` section in `KafkaBridge` configures only half of them. That is probably my only concern.

ppatierno: That could be a concern but we can work in this way. The `.spec.metrics` is related to JMX stuff as it is for `Kafka.spec.kafka` while we could have something under `.spec.http.metrics` for HTTP related metrics configuration; anyway, right now there is nothing to be configurable there. The same could be in `.spec.amqp.metrics`.

ppatierno:
> At the moment this proposal presents two alternatives but doesn't really enumerate the pros and cons of either

I thought that they were clear enough in the proposals and you also listed them in the comment. I would also add the "customizable" pros. Today, a user can customize the `metrics` section (which is actually the configuration of the JMX Exporter agent) for filtering, renaming, hiding metrics.

It seems to me that in the current status we are aiming for Prometheus exposed metrics (it fits well in the CNCF ecosystem as well) using Grafana for related dashboards.
We use the JMX Exporter agent as the only way to go from JMX to Prometheus.
In this case, we should go only in the Prometheus direction but what about the JMXTrans we have right now?
Prometheus metrics allow better integration with other CNCF projects as well, see KEDA for autoscaling. Using the Prometheus scaler it will be possible to scale the bridge in this case.

Other than that, the actual metrics configuration in the CRs is actually JMX related.

Concretely speaking about the bridge, we have two kinds of metrics: HTTP and Kafka.
While we have the flexibility for the first one, we don't have for the second one exposed via JMX.
Both metrics go to Prometheus, because we are supporting it everywhere even with dashboards so I think we should not change our direction on this.
Currently, the HTTP metrics go to Prometheus directly and right now I don't see any possibility to configure them. It implies that a possible `metrics` in the `KafkaBridge` allows to configure Kafka metrics only and not the HTTP ones, maybe generating confusion in users.

If we think we don't need JMX everywhere but we have to cope with them due to Kafka itself, then:

* how we handle the JMXTrans support?
* how we provide consistency in CRs for the best UX?

tombentley:
> Concretely speaking about the bridge, we have two kinds of metrics: HTTP and Kafka.
While we have the flexibility for the first one, we don't have for the second one exposed via JMX.

You can provide your own `MetricsReporter`.

> * how we handle the JMXTrans support?

One of my points is that you're presupposing that we should do this and I'm asking, why should we do this? Who is asking for JMXTrans support in the bridge or anywhere else?

ppatierno:
> You can provide your own MetricsReporter.

The current implementation has its `MetricsReporter` has a central place from where both kind of metrics are taken (the HTTP related from Vert.x, the Kafka related from JMX) so that the scraping happens on one component.
It could be ok but my concern is about more configurability in the `KafkaBridge` custom resource.
The Kafka JMX part is configurable because it uses the same component as the JMX Exporter. We should provide an example configuration as we do for Kafka with the related dashboard. Using `metrics` would leave the freedom to users to change that. Or we can decide to have a "closed not customizable" configuration avoiding to put it into `metrics` field.
Anyway it's only about Kafka part. The HTTP part is not customizable as the JMX one (which is customizable just because it's about bridging from JMX to Prometheus).

> One of my points is that you're presupposing that we should do this and I'm asking, why should we do this? Who is asking for JMXTrans support in the bridge or anywhere else?

You should know me, now :-) I am quite often for consistency. If we support JMXTrans and users use it for Kafka, Kafka Connect and so on maybe they could ask why not for the bridge as well.
Right now, I don't know users asking for this anyway.

ppatierno:
Anyway, after the tests I did yesterday, having the HTTP metrics exported via JMX as well using Vert.x Micrometer, I had the confirmation of my feeling. The JMX part in Vert.x metrics is not touched for since long time and due to some limitation in Dropwizard (which is used under the hood), the JMX mbean are so clean as the Kafka ones.
They are no multi-key but just one key "name" with the name of metrics and the key-value pairs (so labels) just wired one after the other so that the JMX configuration for exporting in Prometheus is not so great.
Maybe it could be a good reason for not using it?

At this point, if we leave as it is today what about the integration into the CO with the `metrics` in `KafkaBridge`. It's going to configure only the JMX part, isn't it confusing for users?
Tbh there is nothing to configure on the HTTP part because the metrics are pretty clean. The JMX configuration is mostly for filtering the ones we are interested in for the dashboard and for some renaming and adding labels.

scholzj:
My view is that our commitment is to provide metrics in the Prometheus format:
* while there are alternatives it is a dominant solution right now
* it is part of the CNCF and Kubernetes ecosystem

For Kafka (+ Connect etc.), we do not live in vacuum and we only provide one of the options how to run these. That is for me the reason why we:
* Let users to extensively configure the Prometheus metrics export instead of having our own values set automatically (value for users: they can use existing dashabords and monitoring if they already have it)
* Expose the JMX interface in Kafka (value for users: it helps existing users to connect their existing JMX tooling they used with Kafka before they moved to Strimzi)

I do not think any of these apply to the Bridge:
* It has no legacy users with JMX
* It has no legacy users with different Prometheus dashboards who need to differently configure the metrics

So IMHO we should approach it like with the operators - provide Prometheus endpoint with the metrics we see relevant, use all or some of them in our dashabord, so not give unnecessary controls over the exports. I particularly feel like the idea of commiting to JMX on our own compoenents without previous JMX legacy is a bad call given the proprietarry nature of JMX and its connection to Java.

But none of these are really something I would necessarily block, I just have different opion that is all.

ppatierno:
I totally agree with the Prometheus choice for the reason you mentioned.

Regarding the HTTP part exposed by Vert.x, we are fine (it's like the ones exported by the operators).
Regarding the Kafka part, we are already using JMX on the bridge because the Kafka clients provide a bunch of useful metrics for the dashboard I have developed. Using these metrics was one of your suggestions and I agree on that because it would be a pity not leveraging them but writing our own custom metrics when anyway they are already there (when you start the bridge, you can attach a JMX tool and see the Kafka metrics).

So you are suggesting to ...

1. just locking down the YAML configuration from translating the JMX metrics to Prometheus for our dashboard? or ...
2. not using the JMX metrics, disabling them somehow and writing our own custom metrics for the Kafka part?

Tbh I would be for 1.
It would mean no other needed steps about the integration with CO (no more `metrics` field in the `KafkaBridge`) but just providing the dashboard (which in this case would not be just an example but THE bridge dashboard provided by Strimzi).

scholzj:
I think it is fine to use the JMX metrics from Kafka clients if that is the best solution. I just see it as an implementation detail rather than a feature. So I do not think it is needed to expose the Vert.x metrics through JMX (I'm fine if that is the easiest implementation - I just don't think it is needed) to have everything available also through JMX. 

And similarly I do not necessarily see the need to give users so big space to customize them (if others think differently I'm fine with it).

One question I guess is whether the should be some `metrics` configuration in the `KafkaBridge` to merely enable / disable the metrics rather than configure them. I guess that depends on the performance impact which I do not dare to judgewith my Bridge expertise.

Again ... just my view which does not have to be how we do it if you want to do it differently.

ppatierno:
I had the same point about allowing to disable/enable. In Kafka we don't disable JMX but just the agent which in the bridge case would mean disabling fge JMX Prometheus collector. I agree that making it configurable via an env var would be useful, of course even for the HTTP metrics.
At this point it seems that we agree on leaving metrics with current implementation.

ppatierno:
@tombentley @scholzj I also added a proposal for enabling/disabling metrics exposure in the `KafkaBridge` custom resource through the cluster operator.

005-http-bridge-metrics.md
```diff
@@ -0,0 +1,126 @@
...
+
+### HTTP server metrics
+
+The HTTP metrics could be:
+
+* The number of opened HTTP connections to the bridge.
+* Requests being processed.
+* Requests processed/secs grouped by HTTP method.
+* The requests rate in total and even grouped by response codes (2XX, 4XX, 5XX).
+* Bytes received /sec and Bytes sent /sec.
```
scholzj, line 39: Is there a way to also show some latecies for the different requests etc.?

005-http-bridge-metrics.md
```diff
@@ -0,0 +1,126 @@
...
+
+* Number of consumers, producers and related opened connections
+* Producer:
+    * The average number of records sent per second (grouped by topic)
+    * The number of outgoing bytes sent to all brokers per second (grouped by topic)
+    * The averate number of records per second that resulted in errors (grouped by topic)
+* Consumer:
+    * The average number of records consumed per second (grouped by clientId-topic)
+    * The average number of bytes consumed per second (grouped by clientId-topic)
+    * Partitions assigned (grouped by clientId)
```
scholzj, line 65: Should we also expose some error metrics such as failed connection attempts etc? Or do you think that it is not needed?

ppatierno:
@scholzj I already have a Grafana dashboard with a bunch of panels with useful information on HTTP and Kafka side. For now, I would avoid to add more tbh but I am not saying that the ones you are proposing are not useful. Maybe we could start in this way and then improve it later? 
@tombentley any more comments? 
Despite the fact this PR is not approved yet, all the stuff described is already implemented and I have a PR ready to be opened :-( ... so I would love to have this one sorted out soon. Thanks!

scholzj:
@ppatierno So, I wonder what is the sense of this proposal. You already merged most of the stuff and reject comments because you have built the stuff. So maybe this should be just closed because it lost its purpose?

ppatierno:
Tbh this proposal was opened too long without any further comments and I decided to implement part of it and maintainers approved the ralated PRs (if the

scholzj:
> Then which comments I am rejecting? The ones related to the latencies and connections errors? From one side I said that we could add them later on the other side the comments are not there for being always accepted.

The point of a proposal is to discuss the future direction and agree on it. So rejecting a comment because you already implemented something else is IMHO not a valid reason.

> Do we want to count how many of my comments and feedback were rejected? I don't think what you said is fair. Anyway if you think it should be closed then I will do it. You can update your latest proposal to be the 5th instead of the 6th.

Don't get upset and personal. I'm not saying it is your fault. I'm just saying that this proposal does not seem to make much sense anymore if it is all implemented and mostly even already merged.

ppatierno:
> The point of a proposal is to discuss the future direction and agree on it. So rejecting a comment because you already implemented something else is IMHO not a valid reason.

I can always add the metrics you are proposing because it's an addition and not a total refactoring of what I did but I don't agree that them are necessary right now, because as I said the dashboard is already so huge. 

> Don't get upset and personal. I'm not saying it is your fault. I'm just saying that this proposal does not seem to make much sense anymore if it is all implemented and mostly even already merged.

At this point we should review the process for proposals review because this proposal https://github.com/strimzi/proposals/pull/3 was review and approved in just 6 days. My proposal was lacking comments for a too long time and this can block people on the development. Maybe I am taking it personal, maybe not.
Anyway I will close this. Please change your above proposal to be the 5th.

scholzj:
> At this point we should review the process for proposals review because this proposal #3 was review and approved in just 6 days. My proposal was lacking comments for a too long time and this can block people on the development. Maybe I am taking it personal, maybe not.

I commented on it rght on the first day. Maybe it covers too niche area? To be honest until seeing what Kafka client metrics are available on Thursday as part of other work, I had for example no idea there are some metrics about failed connections etc. So maybe people don't understand it enough to comment? Maybe it is a learning process to figure out what deserves and doesn't deserve a proposal.

PR #3, opened 2020-06-21
Improve configurability of Kafka listeners

scholzj:
This proposal suggests possible improvements to the Kafka custom resource - to the way how listeners are configured. It should give us more flexibility and freedom when configuring listeners, increase the number of configured listeners etc.


006-improving-configurability-of-kafka-listeners.md
```diff
@@ -0,0 +1,240 @@
...
+
+Some of the use cases raised by out users in the past are listed here:
+
+### Multiple external listeners
+
+In some situations, having multiple external listener could be very useful.
+Apart from the obvious such as multiple authentication mechanisms or listeners with and without encryption, multiple listeners can be also useful to handle access from different networks outside the Kubernetes cluster.
+
+For example in AWS, you can have two types of load balancers: public and internal.
+Public load balancers are exposed tot he Internet and can be used by applications running anywhere online.
```
tombentley, line 80:
```suggestion
Public load balancers are exposed to the Internet and can be used by applications running anywhere online.
```

006-improving-configurability-of-kafka-listeners.md
```diff
@@ -0,0 +1,240 @@
...
+      - name: tls
+        port: 9093
+        type: service
+```
+
+Being easily able to convert the old and new structures makes it easy to have single code in our operator while being able to easily work with the old format as well.
+
+### Code changes
+
+Since both the new and the old structure will use the same `listeners` property, the CRD generator will need to be enhanced to support the OpenAPI `oneOf` for different types.
```
tombentley, line 202: I'm sorry to have to tell you that Kubernetes does not allow CRD schema objects (subschemas if you like) to be one of a set of JSON types, see https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#schemaObject. 

To do this for v1beta1 resources we'd have to have a different property than `listeners` and use a `oneOf` constraint to force only one of them being set. This option is relatively easily done, but means we need a different name than `listeners`, and I can't think of a good one.

An alternative would be to have a `v1beta2` CRD in which `listeners` was a list as described here. That comes with a bunch of other considerations like having a upgrade webhook to convert `v1beta1` to `v1beta2` and would also require support in the `CrdGenerator` for generating different schemas. It's also predicated on dropping support for Kubernetes v1.11. This option is a *lot* more work, but results in a nicer API.

scholzj: I'm not sure I understand what do you suggest and how does the link relates. The example I listed there:

```yaml
listeners:
  oneOf:
    - type: object
        properties:
          # ...
    - type: array
        items:
          # ...
```

seemed to work fine for me.

tombentley: Schema validation is (supposedly) based on openapi v3, see https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation. Are you saying you've tried this with Kubernetes and the CRD was accepted? And it was correctly able to validate CRs using that schema?

tombentley: Oh, right I see you're not using an array-typed `type`. Maybe it would work then.

scholzj: I based it on this: https://swagger.io/docs/specification/data-models/data-types/ (section mixed types). That should refer to OAPIv3. I have a protype for the `api` and `crd-generator` changes for this. I can share it with you tomorrow. It seemed to work fine within our v1beta1 CRDs on Kube 1.18.

tombentley: OK, well if it works, that would be great, because this is something I'd really like to see.

006-improving-configurability-of-kafka-listeners.md
```diff
@@ -0,0 +1,240 @@
...
+            - broker: 1
+                advertisedHost: my-cluster-brokers-kafka-1.myns.cluster.local
+            - broker: 2
+                advertisedHost: my-cluster-brokers-kafka-2.myns.cluster.local
+      - name: encrypted2
+        port: 9093
+        type: service
+        tls: true
+        authentication:
+          type: acram-sha-512
```
Frawless, line 121:
```suggestion
          type: scram-sha-512
```

006-improving-configurability-of-kafka-listeners.md
```diff
@@ -0,0 +1,240 @@
...
+          type: tls
+```
+
+The listeners will be configurable as array.
+Each listener will have its own unique name and will specify the port and type as required values.
+The port can be set to anything apart from `9091` (internal replication listener) and `9404` (Prometheus).
+A new type service will be introduced for the internal listeners designed for apps running inside the same Kubernetes cluster.
+Additionally, all the types from the existing external listeners will be supported as well.
+
+together with this change, I suggest to change the default value of the listener `tls` flag from `true` to `false`.
```
Frawless, line 142:
```suggestion
Together with this change, I suggest to change the default value of the listener `tls` flag from `true` to `false`.
```

006-improving-configurability-of-kafka-listeners.md
ppatierno: I used the `oneOf` for the bridge as well and the validation works (of course it's not related to Kubernetes). @tombentley out of curiosity, why it shouldn't have worked with K8S?

tombentley: I just forgot that `oneOf` (and `@OneOf`, which I implemented) didn't violate the "`type` must be scalar" rule.

006-improving-configurability-of-kafka-listeners.md
```diff
@@ -0,0 +1,240 @@
...
+  oneOf:
+    - type: object
+        properties:
+          # ...
+    - type: array
+        items:
+          # ...
+```
+
+In the `api` module, the setter and getter for the `listeners` field will use generic `JsonNode` type and inside the `api` module decide whether it should be decoded into the new or old format when the getter or setter is called.
```
tombentley, line 216: I think you could also use a `JsonSerializer` which would allow you to have a more strongly typed POJO with `listenerList` and `listenerObject` properties. You can see how that would work by looking at the fabric8 `IntOrString` class.

006-improving-configurability-of-kafka-listeners.md
```diff
@@ -0,0 +1,240 @@
...
+But our current implementation makes that hard because each listener has its own class in the `api` module with its own configs, settings etc.
+
+## Motivation for change
+
+The current implementation has many limitations
+* Only 3 listeners can be configured
+* 2 out of the 3 listeners have fixed roles in terms of encryption, so when you want to use for example encrypted listeners only, you have effectively only 2 available listeners.
+* The `plain` and `tls` listeners do not offer some additional configuration options such as overriding advertised hosts etc.
+* Only one listener can be used for outside of the Kubernetes cluster.
+* Since Kafka listeners can have each only single authentication mechanism, possibilities for offering different authentication to different clients are limited.
```
samuel-hawker, line 69: Since each Kafka listener can have only a single

006-improving-configurability-of-kafka-listeners.md
```diff
@@ -0,0 +1,240 @@
...
+* Port numbers are assigned and are not configurable.
+
+Some of the use cases raised by out users in the past are listed here:
```
samuel-hawker, line 72: *our

006-improving-configurability-of-kafka-listeners.md
```diff
@@ -0,0 +1,240 @@
...
+For example in AWS, you can have two types of load balancers: public and internal.
+Public load balancers are exposed tot he Internet and can be used by applications running anywhere online.
+On the other hand internal load balancers are exposed only within the VPC (Virtual Private Cloud) - and accessible only to the applications running outside of Kubernetes but inside the private network of the user.
+
+### Configurability of internal listeners
+
+Some users heavily customize their Kubernetes network.
+One of the examples is joining their Kubernetes network with their network outside of Kubernetes.
+In such case, a common problem is using or not using the cluster service DNS suffix (`.cluster.local` by default) which is in some cases needed and in some not.
+Having the ability to override the advertised hostnames on internal listeners similarly to how they can be treated on the external listener or being able to have ore internal listeners could help in these situations.
```
samuel-hawker, line 88:
```suggestion
Having the ability to override the advertised hostnames on internal listeners similarly to how they can be treated on the external listener or being able to have one internal listeners could help in these situations.
```

006-improving-configurability-of-kafka-listeners.md
```diff
@@ -0,0 +1,240 @@
...
+      - name: nodeports
+        port: 9095
+        type: nodeport
+        tls: true
+        authentication:
+          type: tls
+```
+
+The listeners will be configurable as array.
+Each listener will have its own unique name and will specify the port and type as required values.
```
samuel-hawker, line 137:
```suggestion
Each listener will have its own unique name and will specify a type and unique port as required values.
```

Since we are banning 9091 and 9094, I don't think it will be much extra work to iterate through the defined listeners and ensure there are no duplicate port usages.

006-improving-configurability-of-kafka-listeners.md
```diff
@@ -0,0 +1,240 @@
...
+
+In the `api` module, the setter and getter for the `listeners` field will use generic `JsonNode` type and inside the `api` module decide whether it should be decoded into the new or old format when the getter or setter is called.
+The `api` module will keep both the old and new format to be able to also serialize back to the original CRD.
+When the old format is used, it will be converted in the `fromCrd` method inside the `KafkaCluster` class in the `cluster-operator` module and from there on, only the new format will be used to configure the pods, services etc.
+Since all listeners will be part of the same array, it should be possible to simplify the code configuring the listeners and just pass all of them through a single loop.
+
+### Multiple external listeners
+
+Having multiple external listeners would require to have multiple sets of external services.
+For backwards compatibility, when the name of the external listener is `external`, the current names of the services will be used.
```
samuel-hawker, line 224: `current names of the services and routes` ?

006-improving-configurability-of-kafka-listeners.md
```diff
@@ -0,0 +1,240 @@
...
+
+```yaml
+    listeners:
+      - name: plain
+        port: 9092
+        type: service
+        tls: false
+        authentication:
+          type: scram-sha-512
+        networkPolicyPeers:
```
samuel-hawker, line 158: Why do the network policy peers and authentication get removed in the example below?

scholzj: The second example show how would 
```yaml
    listeners:
      plain: {}
      tls: {}
```
be converted. So the network policy peer arrays are empty nd not used.

006-improving-configurability-of-kafka-listeners.md
scholzj: I changed it to cover both options so that we can decide during implementation. I used the `JsonNode` in my prototype and know it can work. But maybe the custom (de)serializer is a bit cleaner and more elegant approach. So maybe we I try it later.

006-improving-configurability-of-kafka-listeners.md
scholzj: @tombentley In case you are interested ... these are the changes I have done as as part of the prototyping this (the code is of course not PR worthy, but it should give the idea): https://github.com/strimzi/strimzi-kafka-operator/compare/master...scholzj:make-it-possible-to-alternate-multiple-types-in-api-schema?expand=1

PR #4, opened 2020-07-01
Add proposal to secure the Kafka Connect REST API endpoint

ajborley:
This proposal suggests possible improvements to the KafkaConnect (and
KafkaMirrorMaker2) custom resource to allow users to secure the Kafka
Connect REST API endpoint.

Co-authored-by: Samuel Hawker <samuel.hawker@ibm.com>

### DIFF
--- a/006-securing-kafka-connect-rest-endpoint.md
+++ b/006-securing-kafka-connect-rest-endpoint.md
@@ -78,7 +78,65 @@ This proposal is backwards compatible - the `spec.rest` property is optional and
 It is also worth noting that this opens Strimzi up to further extensions in the future, such as:
  - adding support for client authentication at the Kafka Connect REST API endpoint.
  - adding external listeners (such as new routes explicitly for Kafka Connect REST calls).
+ - adding `networkPolicyPeers` support to the `spec.rest.tls` property, allowing users to define additional network policies as part of the CR. The default network policy that only permits access from the operator pod would continue to exist.
 
+### Proposed schema
+```
+openAPIV3Schema:
+  type: object
+  properties:
+    spec:
+      type: object
+      properties:
+        rest:
+          type: array
+          items:
+            type: object
+            properties:
+              port:
+                type: integer
+                minimum: 1024
+                description: The port number of the REST listener.
+              tls:
+                type: object
+                properties:
+                  configuration:
+                    type: object
+                    properties:
+                      brokerCertChainAndKey:
+                        type: object
+                        properties:
+                          certificate:
+                            type: string
+                            description: The name of the file certificate in
+                              the Secret.
+                          key:
+                            type: string
+                            description: The name of the private key in the
+                              Secret.
+                          secretName:
+                            type: string
+                            description: The name of the Secret containing the
+                              certificate.
+                        required:
+                        - certificate
+                        - key
+                        - secretName
+                        description: Reference to the `Secret` which holds the
+                          certificate and private key pair. The certificate
+                          can optionally contain the whole chain. If this field is
+                          missing, the operator generates keys and certificates
+                          stored in a secret.
+                    description: Configuration of TLS listener.
+                description: Configures TLS on the REST listener.
+            required:
+            - port
+            description: Configuration of a REST listener.
+          description: List of configurations for Kafka Connect REST API listeners.  
+            If this field is empty or missing, the default Kafka Connect REST API
+            listener on port 8083 without TLS is created.
+        ...
+```
 
 ## Rejected alternatives
 
@@ -129,15 +187,8 @@ spec:
     config.providers: file
     config.providers.file.class: org.apache.kafka.common.config.provider.FileConfigProvider
 
-    # non-HTTPS related config
-    group.id: connect-cluster-2
-    offset.storage.topic: connect-cluster-offsets
-    config.storage.topic: connect-cluster-configs
-    status.storage.topic: connect-cluster-status
-
-    config.storage.replication.factor: 1
-    offset.storage.replication.factor: 1
-    status.storage.replication.factor: 1
+    # other config
+    ...
 ```
 
 This works to secure the Kafka Connect REST API endpoint, but requires several modifications to the operator to allow it to send REST API requests:
--- a/006-securing-kafka-connect-rest-endpoint.md
+++ b/006-securing-kafka-connect-rest-endpoint.md
@@ -1,0 +1,146 @@
+# Securing the Kafka Connect REST API endpoint
+
+## Current situation
+
+Currently, instances of Kafka Connect that are deployed by the Strimzi operators are configured with the default REST API endpoint settings. This means that the Kafka Connect REST API endpoint uses HTTP on port 8083 and that the Strimzi KafkaConnect, KafkaConnector and KafkaMirrorMaker2 operators make unsecured REST client calls based on this default configuration.
+
+The default network policies created by the Strimzi operators restrict incoming REST API calls to only allow access from the operator pod. If required, further network policies can be provided to override the default policy and allow wider access.
+
+## Motivation for change
+
+The Kafka Connect REST API endpoint is perhaps the last key endpoint that cannot currently be secured using Strimzi. Many users have requested the ability to secure the Kafka Connect REST API endpoint, for example:
+ - https://github.com/strimzi/strimzi-kafka-operator/issues/3229
+ - https://github.com/strimzi/strimzi-kafka-operator/issues/2161
+ - https://cloud-native.slack.com/archives/CMH3Q3SNP/p1577722052135000
+
+## Proposed changes
+
+This proposal adds a `rest` property to the KafkaConnect spec, allowing users to configure Connect REST API  endpoints that use HTTP or HTTPS. Client authentication is not addressed here, but there is scope for future proposals to address it.
+
+Simple example CR:
+
+```
+apiVersion: kafka.strimzi.io/v1beta1
+kind: KafkaConnect
+spec:
+  rest:
+  - tls: {}
+  ...
+```
+
+This CR would generate a cert and key in a secret (for use by the operator clients), create SSL truststores and keystores from the cert and key and set the following properties in the generated connect configuration to enable an HTTPS connection:
+
+```
+listeners: https://:8443
+rest.advertised.listener: https
+rest.advertised.port: 8443
+
+listeners.https.ssl.client.auth: none
+listeners.https.ssl.truststore.location: /tmp/kafka/kafka-connect-rest.truststore.p12
+listeners.https.ssl.truststore.password: ***generated password***
+listeners.https.ssl.truststore.type: PKCS12
+listeners.https.ssl.keystore.location: /tmp/kafka/kafka-connect-rest.keystore.p12
+listeners.https.ssl.keystore.password: ***generated password***
+listeners.https.ssl.keystore.type: PKCS12
+
+```
+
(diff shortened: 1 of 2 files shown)


006-securing-kafka-connect-rest-endpoint.md
```diff
@@ -0,0 +1,146 @@
...
+
+```
+
+A more complex example CR:
+
+```
+apiVersion: kafka.strimzi.io/v1beta1
+kind: KafkaConnect
+spec:
+  rest:
```
tombentley, line 54: I think it would be good to explicitly describe the  schema of the `rest`. I'm guessing it has `port` (defaulting to ???) and `tls` properties.

006-securing-kafka-connect-rest-endpoint.md
```diff
@@ -0,0 +1,146 @@
...
+
+The operator REST API client code will prefer HTTPS over HTTP when both are configured, in a similar way to how the Kafka Connect workers coordinate via HTTPS when it is available (see [KIP 507](https://cwiki.apache.org/confluence/display/KAFKA/KIP-507%3A+Securing+Internal+Connect+REST+Endpoints)).
+
+The above proposal will require modifications to the `AbstractConnectOperator` to correctly read and use the additional CR configuration, apply the TLS certs as keystore credentials to a WebClient truststore and then the webclient (or its configuration) will have to be passed into the `KafkaConnectApi` interface and `Impl` so that each call can optionally use this. This should mean for regular http calls it will continue to work as is, but the calls will also be able to speak to the secured endpoint if configured.
+
```
tombentley, line 71: Can a user configure a different network policy for the different ports? Presumably using the same mechanism as Kafka CR `spec.listeners.tls`, so `tls` has a `networkPolicyPeers`, right? Again, I think it would be good to describe the schema in a bit more detail, especially as you said 

>  This uses a _similar_ interface as used for the Kafka CR `spec.listeners.tls` property.

(my emphasis) a couple of paragraphs above.

006-securing-kafka-connect-rest-endpoint.md
ajborley: I have added a `Proposed schema` section in OpenAPI format. Hope it looks OK.

006-securing-kafka-connect-rest-endpoint.md
ajborley: This proposal wasn't planning to address additional network policies beyond the default network policy that denies access to the REST port from anything other than the operator. As is currently the case, users can add network policies in k8s to allow wider access. It's a good idea though - I have added a note to the `Future extensions` section. Do you feel it's a necessary function we should be adding?

tombentley: No, I don't think it is necessary, but thanks for clarifying.

006-securing-kafka-connect-rest-endpoint.md
```diff
@@ -0,0 +1,197 @@
...
+```
+apiVersion: kafka.strimzi.io/v1beta1
+kind: KafkaConnect
+spec:
+  rest:
+  - tls: {}
+  ...
+```
+
+This CR would generate a cert and key in a secret (for use by the operator clients), create SSL truststores and keystores from the cert and key and set the following properties in the generated connect configuration to enable an HTTPS connection:
```
ppatierno, line 31: I would add that the cert is signed by the cluster CA certificate right?

ajborley: That's right - I have updated that paragraph to clarify.

006-securing-kafka-connect-rest-endpoint.md
```diff
@@ -0,0 +1,197 @@
+# Securing the Kafka Connect REST API endpoint
+
+## Current situation
+
+Currently, instances of Kafka Connect that are deployed by the Strimzi operators are configured with the default REST API endpoint settings. This means that the Kafka Connect REST API endpoint uses HTTP on port 8083 and that the Strimzi KafkaConnect, KafkaConnector and KafkaMirrorMaker2 operators make unsecured REST client calls based on this default configuration.
+
+The default network policies created by the Strimzi operators restrict incoming REST API calls to only allow access from the operator pod. If required, further network policies can be provided to override the default policy and allow wider access.
```
scholzj, line 7: It might be worth mentioning that at this point, this is not the intention but a side effect of how Network Policies work and how they are discoverable.

006-securing-kafka-connect-rest-endpoint.md
```diff
@@ -0,0 +1,197 @@
...
+```
+apiVersion: kafka.strimzi.io/v1beta1
+kind: KafkaConnect
+spec:
+  rest:
+  - tls: {}
+  ...
+```
+
+This CR would cause the operator to generate a certificate (signed by the cluster CA certificate) and a key, both stored in a secret for use by the operator clients and the Kafka Connect pods. The certificate and key would be mounted in each Kafka Connect pod, which would then create SSL truststores and keystores from the cert and key and set the following properties in the generated connect configuration to enable an HTTPS connection:
```
scholzj, line 31: The KafkaConnect has not hard coupling to any particular Kafka cluster. It can be used even with clusters which are not Strimzi based. So I do not think this can rely on cluster CA certificate. It would need to have its own CA.

006-securing-kafka-connect-rest-endpoint.md
```diff
@@ -0,0 +1,197 @@
+# Securing the Kafka Connect REST API endpoint
```
scholzj, line 1: I do not think securing is the right word given this covers only TLS encryption. It should be clear from the name that this is encryption only.

ajborley:
@scholzj - Thank you for the frank and clear comments! :) 

> My (personal) view is for a long time that we should not go this way. We should instead:
> 
> - Finish the remaining features of the Connector operator (restarts of connectors are the main missing point IMHO)
> - Lock down the Connect API to make it accessible only view the Connector operator
> - This would solve the whole security including authN and authZ which would be offloaded to Kubernetes RBAC.

I agree that this is a simpler route to follow and I'm happy to go down this route for the first steps here. A few questions about the approach:
 - Re. the connector restart API: I can imagine using an annotation on the KafkaConnect or KafkaConnector CR to start the connector restart action, in a similar way to how optimization proposals are refreshed and approved on the KafkaRebalance CR (https://strimzi.io/docs/operators/latest/using.html#proc-approving-optimization-proposal-str) via an annotation that is subsequently removed by the operator - does that make sense or were you thinking of a different approach? 
 - I expect there will be users who would prefer to continue to access the Connect REST API directly rather than indirectly via CRs, even when using Connector CRs. Do you think it would be worth adding a label or annotation to disable the locking-down, reverting back to the default Connect HTTP listener?
 - Shall I raise new proposal PRs for the connector restart action and how the REST API is locked down? 

Responding to some of your other comments:
> I think the way suggested here has several issues:
> - It requires huge amount of work (although some parts of it such as TLS would probably be needed also for the solution I prefer as well)
>    - TLS certificate management (including things such as renewals etc. which are very painful), custom certificates, support for different public CAs etc.

Am I right that TLS cert renewals would still need to be handled for the locking-down proposal? Would you expect that the operator just manages renewals silently?

>    - Different methods woth authN and authZ which will be required etc.

Kafka currently only includes BasicAuth support for the Connect REST API (in the `BasicAuthSecurityRestExtension` class) - are you imagining adding REST extensions that implement Strimzi's own authN options (Scram SHA 512, Mutual TLS, OAuth,  etc) here?

> It questions the value of the Connector operator (what is it for if users need to use the REST API? It actually fights against the things done via the REST API so locking down the REST API or at least switching it to read only mode is IMHO required for the connector operator in long term)

Personally, I don't see the REST API conflicting with the Connector operator - we would just expect the user to select between them, as is the case with other Strimzi function e.g. using the KafkaUser operator vs managing ACLs via the Kafka tools. As you say, both ways could be available in parallel - when the connector operator is being used, the REST API is locked-down and unavailable, and when the connector operator is not in use, allow access to the REST API for connector management.

> It requires a commitment to actually execute on all of this

This is a good point, and one that I doubt I can answer. Starting with your locking-down approach sounds like a reasonable way forward right now, so I'm not sure if there is value in extending this proposal to cover TLS cert management, authN, authZ, etc. 

I expect we will find that users will continue to want to use the Connect REST API, as it's a known Kafka management API, and therefore would like to be able to secure the endpoint whether they use the Connector operator or not.

Thanks again for the review :)

scholzj:
> Re. the connector restart API: I can imagine using an annotation on the KafkaConnect or KafkaConnector CR to start the connector restart action, in a similar way to how optimization proposals are refreshed and approved on the KafkaRebalance CR (https://strimzi.io/docs/operators/latest/using.html#proc-approving-optimization-proposal-str) via an annotation that is subsequently removed by the operator - does that make sense or were you thinking of a different approach?

I guess there are two options which at the end might need to be combined:
* Restarting using an annotation as you mentioned
* Automatic restarting (operator sees a failed task / connector, it will try to restart it)

I know of some users preferring one and others the other and I can imagine that the annotation will be always available versus the automatic restarts will be enabled / disabled in the connector `.spec` section.

> I expect there will be users who would prefer to continue to access the Connect REST API directly rather than indirectly via CRs, even when using Connector CRs. Do you think it would be worth adding a label or annotation to disable the locking-down, reverting back to the default Connect HTTP listener?

Right. I do not have necessarily an intention to block it - even if only for backwards compatibility reasons. But I can imagine it being done in a way that when the connector operator is enabled the REST API is disabled or by some separate switch ... I can live with both.

> Shall I raise new proposal PRs for the connector restart action and how the REST API is locked down?

Feel free to do so if you want. But right now this is my personal preference / opinion which might not be shared by others. So feel free to wait on how they see it.

> Personally, I don't see the REST API conflicting with the Connector operator - we would just expect the user to select between them, as is the case with other Strimzi function e.g. using the KafkaUser operator vs managing ACLs via the Kafka tools. As you say, both ways could be available in parallel - when the connector operator is being used, the REST API is locked-down and unavailable, and when the connector operator is not in use, allow access to the REST API for connector management.

Maybe I phrased it badly ... I think that using both at the same time is weird and leads to problems with users creating connectors over POST and wondering why they were deleted by the connector operator etc. But I can see why someone would choose one over the other etc. I can see that we support both

I'm just trying to keep in mind all the effort which might be behind it for development, testing and maintenance - that is why I'm a bit careful about doing both. If you want to work on securing the API for users and take care of it I'm fine with it (and I have of course no reason not to trust you to deliver on it given the MM2 work you did). 

> Am I right that TLS cert renewals would still need to be handled for the locking-down proposal? Would you expect that the operator just manages renewals silently?

That is right. Although it might be a bit simpler when it is internal only ... maybe you don't need a CA and just self signed cert etc. 

I was for a long time against it. But maybe we need to look more at things like Cert Manager to offload this to someone else :-o. Because this is lot of effort but at the same time it is also not really our _core business_ so to say.

> Kafka currently only includes BasicAuth support for the Connect REST API (in the BasicAuthSecurityRestExtension class) - are you imagining adding REST extensions that implement Strimzi's own authN options (Scram SHA 512, Mutual TLS, OAuth, etc) here?

I think that ideally we should have one mechanism supported everywhere. OAuth seems to be one common option here. But it can be something else. Either way it would be good to understand what the plan is and how will it be handled (e.g. for mutual TLS you would need a way to manage user certificates etc.).

ajborley:
This proposal has been superseded by #8 and #10 and can be closed.

PR #5, opened 2020-07-06
Proposal: Design documentation

samuel-hawker:
This pull request puts forward two different proposed ways of doing design documentation, each with its own merits and implementation.


007-design-docs.md
```diff
@@ -0,0 +1,97 @@
...
+
+It would be appropriate to add a design document when a significant new feature is being added, examples of this being:
+- A new custom resource type being added, a design document explaining how its configuration would be implemented.
+- A new feature to a pre-existing custom resource type, e.g. new component
+- A reworking of an old feature, deprecating the old feature, including if applicable of how a user would migrate.
+
+### Where to write a design document
+
+A design document should be a markdown file located in the `design` directory of the `strimzi-kafka-operator` project.
+Design documents for other Strimzi projects such as Oauth should also be located here as the intent is that the design documents are all colocated and that all components are a part of 'the operator' project.
```
scholzj, line 32: I do not think this makes sense - we have several projects with their own repos. We should either have the design folder in each of them with their own designs or we should have a central repo for all designs.

007-design-docs.md
```diff
@@ -0,0 +1,97 @@
...
+
+### When is a design document complete
+
+A design document is considered complete when the reviewers and maintainers believe the document to accurately reflect the current implementation of the components behaviour. Once approved the document is merged and is understood to be correct, any mistakes or changes to behaviour should be corrected in the form of a pull request.
+
+### How to write a design document
+
+The design documents would live in the strimzi-kafka-operator repo and would match the current code state, so that code changes could be audited with their corresponding design doc changes.
```
scholzj, line 40: How will you enforce that it matches the current state of the code?

007-design-docs.md
```diff
@@ -0,0 +1,97 @@
...
+A description of the problem the program/code tackles, why the program/code is necessary, what people need to know to assess this project, and how it fits into the technical strategy of Strimzi.
+
+**Goals and Non-Goals**
+
+The Goals section should describe the user-driven impact of the code and specify how to measure success using metrics, normally that would be tested.
+Non-Goals are equally important to describe which problems you won’t be fixing so everyone is on the same page.
+
+**Milestones**
+
+A list of measurable checkpoints, if there is future progress planned, including a milestone for the future in the design doc maye be useful for a reader to understand where the project is heading.
```
scholzj, line 63: How would the progress through the milestones be tracked?

007-design-docs.md
```diff
@@ -0,0 +1,97 @@
...
+- A reworking of an old feature, deprecating the old feature, including if applicable of how a user would migrate.
+
+### Where to write a design document
+
+A design document should be a markdown file located in the `design` directory of the `strimzi-kafka-operator` project.
+Design documents for other Strimzi projects such as Oauth should also be located here as the intent is that the design documents are all colocated and that all components are a part of 'the operator' project.
+
+### When is a design document complete
+
+A design document is considered complete when the reviewers and maintainers believe the document to accurately reflect the current implementation of the components behaviour. Once approved the document is merged and is understood to be correct, any mistakes or changes to behaviour should be corrected in the form of a pull request.
```
scholzj, line 36: My experience tells me that writing tech designs before the design is complicated:
* You either end up changing it significantly later during the implementation in which case the initial design was useless
* Or you implement it during the design phase in which case you risk lot of wasted effort during the review when the design is modified or rejected by reviewers.

It also makes the process very weird:
1) Create a proposal and get it approved
2) Write design docs and get it approved
3) Write the code (+ edit the design docs to correspond to reality???)

I think I would prefer to have the proposal to approve the direction and have the design docs part of the PRs delivering the code (i.e. without a dedicated approval - approved together with the implementation). That should make it also easier to have it up to date.

007-design-docs.md
```diff
@@ -0,0 +1,97 @@
...
+The Goals section should describe the user-driven impact of the code and specify how to measure success using metrics, normally that would be tested.
+Non-Goals are equally important to describe which problems you won’t be fixing so everyone is on the same page.
+
+**Milestones**
+
+A list of measurable checkpoints, if there is future progress planned, including a milestone for the future in the design doc maye be useful for a reader to understand where the project is heading.
+
+**High Level Description**
+
+High level description describing the current implementation as well as example flow(s) to illustrate how users interact with this system and/or how data flow through it. This should build on the context.
```
scholzj, line 67: I think the user part should be part of proposal in the first place.

007-design-docs.md
```diff
@@ -0,0 +1,97 @@
...
+
+**Technical Architecture section**
+
+Walk through the flows in great detail to concertise the implementation. Feel free to include diagrams where relevant and sub-sections for edge-cases and implementation details.
+
+## Summary of changes
+
+Alongside code changes, behavioural changes of the operator will also be documented in the above described design documentation format.
+
+Newly contributed proposals or issues would describe how their changes might alter the design documentation and are expected to be delivered alongside their code changes.
```
scholzj, line 77: This is hard given no previous design docs for the old features.

007-design-docs.md
```diff
@@ -0,0 +1,97 @@
...
+
+### When is a design document complete
+
+A design document is considered complete when the reviewers and maintainers believe the document to accurately reflect the current implementation of the components behaviour. Once approved the document is merged and is understood to be correct, any mistakes or changes to behaviour should be corrected in the form of a pull request.
+
+### How to write a design document
+
+The design documents would live in the strimzi-kafka-operator repo and would match the current code state, so that code changes could be audited with their corresponding design doc changes.
+
+A design document would be split into `X` parts
```
scholzj, line 42: I think the structure of the document is only one very small part which is actually irrelevant. More important is the structure of the design docs. Because that defines how easily contributors will find the right piece. Once they find the right piece the structure is less important. So IMHO you should add how should the individual documenta be organized.

Additionally, my another major issue is that template suggests a set of delta documents describing the design changes for different features. That is IMHO not friendly for how you want to use it. Contributors will need to go through all the different deltas to understand design of a single component. What we need instead is to have a document describing the technical design and is continuously updated instead of a document describing individual changes wchih will get outdated by new changes. This will also distinguish it from the proposals.

007-design-docs.md
```diff
@@ -0,0 +1,97 @@
...
+
+Newly contributed proposals or issues would describe how their changes might alter the design documentation and are expected to be delivered alongside their code changes.
+
+Design documents accumulate and form a body of design information for the project, and should be kept up to date with design changes. 
+
+
+
+## Alternative implementation - Strimzi Improvement Proposal (SIP)
+
+Inspired and closely resembling [Kafka Improvement Proposals (KIP)](https://cwiki.apache.org/confluence/display/KAFKA/Kafka+Improvement+Proposals), we merge the concept of design documentation and proposals together, meaning that each proposal is a point in time descriptor of current Strimzi behaviour and a proposal as to how it should change.
```
scholzj, line 85: As described in other comments. I think tech design docs and proposals are two very different things. They should (in the right format) ideally coexist next to each other.

007-design-docs.md
samuel-hawker: I think that is a fair comment. I will amend this.

007-design-docs.md
ppatierno: I agree with @scholzj on the fact that a lot of things/issues/problems are clear only during the development so you should change your design doc completely because it's not valid anymore. Having it in the PR allows us to evolve the design while developing and during the review, you can tweak the design, for current discussion and future reading, as well as the code accordingly. These two things should go in parallel, it's quite difficult to handle them sequentially.

007-design-docs.md
tombentley, line 42: A very narrow point: If you don't know the number, then avoid mentioning the number:
```suggestion
A design document would be split into the following parts:
```

Much wider point: there are quite a few sections there. I think you could merge overview and context, since it's difficult to describe both in a readable way while still having linear text. I'm not sure we'd always need the high-level description, since for simpler changes it would be repeating the context/overview. Overall I would observe that to (especially first time) contributors each section is a barrier that has to be crossed, so the fewer the better. We can always ask people to elaborate as part of the process.

007-design-docs.md
```diff
@@ -0,0 +1,97 @@
...
+
+The title of the design doc, normally this would pertain to a single component or one larger aspect of the whole of the Strimzi project. New pages would be discussed and decided on as part of a proposal where needed.
+
+**Overview**
+
+A high level summary that any user of Strimzi should be able understand and use to decide if it’s useful for them to read the rest of the doc. It should be 3 paragraphs max.
+
+**Context**
+
+A description of the problem the program/code tackles, why the program/code is necessary, what people need to know to assess this project, and how it fits into the technical strategy of Strimzi.
```
tombentley, line 54: "assess this project", do you really mean project?

"technical strategy of Strimzi", unless that's actually written down somewhere, and linked to, I think it makes it harder for a new contributor to know how to address that point.

007-design-docs.md
```diff
@@ -0,0 +1,97 @@
...
+
+
+## Proposal
+
+### When to write a design document
+
+It would be appropriate to add a design document when a significant new feature is being added, examples of this being:
+- A new custom resource type being added, a design document explaining how its configuration would be implemented.
+- A new feature to a pre-existing custom resource type, e.g. new component
+- A reworking of an old feature, deprecating the old feature, including if applicable of how a user would migrate.
```
tombentley, line 27: These are all a bit subjective. The very fact that they're examples means that something which is not congruent to one of them is a case that's in doubt.

While it's not perfect, the test used for KIPs in the Kafka community is pretty clear: Anything which touches a public API, with a bunch of things defined as "public API".

So I would suggest we use something similar, with our own definition of "public API":

* Anything which changes the structure of any of the project's CRD (this definition excludes documentation, which we might want to argue about), or how they're used or function (e.g. annotations)
* Anything which would add, remove or change any other public Strimzi APIs (e.g. new endpoints in the bridge, or new REST APIs).
* Anything which changes the accessibility of any of Kafka's public APIs (e.g. changes to how metrics get exposed, changing which broker configs are reserved to the operators, or locking/exposing functionality either via the network, or Kafka protocol, such as authorization)

I don't think we should _mandate_ design documentation for (significant) refactorings, though we might chose to do so.

007-design-docs.md
tombentley: True, but you can deploy that argument against any attempt to improve the status quo.

I think all we really want is a way to improve from where we are (and asymptotically reach design documentation nirvana). That doesn't require placing the entire burden on the first person to come along who wants to make a change in a given area. I would avoid including this language here, since it's another barrier, and instead trust that as part of the process we remember to ask people nicely to improve on what we have.

007-design-docs.md
tombentley: While I agree that there are two purposes to be served, I think by making there be two separate processes we're in danger of requiring more and more work to be done to actually change anything. That's not to say that there aren't benefits to this approach, but I'd rather we erred on the side of having lighter weight processes and requirements until such time as we've shown that such an approach doesn't work.

samuel-hawker:
Hello all, thanks so much for the review 🙏 

This was a big learning experience and it really helped my understanding of our requirements and the problem we are trying to solve. I have cut a lot of sections and made it a lot simpler.
I also 'rejected' the `SIP` suggestion.

I will be working to update it a bit more when I can but I hope this addresses a majority of the concerns.
I am really trying to trim the process as much as I can and make clear this is optional, but encouraged.

I think some discussion into the exact specifics of design doc (file) structure would be useful and I took a first pass at it.

007-design-docs.md
```diff
@@ -0,0 +1,124 @@
...
+
+#### Strimzi Design Document Organization
+Design documents should be splt into logical areas, for repositories that aren't `strimiz-kafka-operator` I suspect this would start as a single or just a few succinct documents.
+For `strimzi-kafka-operator` I would suggest splitting design documentation from 'public API' perspective, that is to say all documents relating to the `Kafka` custom resource would be separated into its own folder called `Kafka`, similarly for `KafkaConnect` etc.
+For design documentation that would cover the whole operator and not just any single custom resource I would suggest a folder named `operator` this would describe behaviours not unique to any one specific reconcile loop or verticle.
+
+
+### When is a design document complete
+
+A design document is considered complete when the reviewers and maintainers believe the document to accurately reflect the current implementation of the components behaviour. Once approved the document is merged and is understood to be correct, any mistakes or changes to behaviour should be corrected in the form of a pull request.
```
scholzj, line 44: I do not think this really addressed my previous comment. I think this has again very confusing boundaries between proposals and your design docs. As I read it, I would need to:
1) Write a proposal and get it approved
2) Write a design docs and get it approved
3) Do the actual code (this is where you figure out the proposal and design docs do not work :-o)

This would totally stop any progress. As I suggested before, I think the way this can be done is that the design docs are technical design documentation written as part of the code PR. But I guess there might be other options. But I do not think it is possible to build another approval barrier.

As an example, we need to for example learn from @ppatierno's proposal around bridge metrics. Which shows that we maybe didn't yet learn properly how to handle the proposals before introducing some new hurdle.

007-design-docs.md
```diff
@@ -0,0 +1,124 @@
...
+Then the cluster will be rebalanced to 'X' state
+When the KafkaRebalance is refreshed
+Then the KafkaRebalance has no proposal
+And the KafkaRebalance will be in the Ready state
+```
+
+
+### How to write a design document
+
+The design documents would live in the related repository in the Strimzi Organization and would match the current code state, so that code changes could be audited with their corresponding design doc changes.
```
scholzj, line 70: Would _compared_, _undersstood_ or something else be a nicer word here than _audited_? I do not think we are planning to do any code audits at this point, or?

007-design-docs.md
```diff
@@ -0,0 +1,124 @@
...
+**Technical Architecture section**
+
+Walk through the flows in great detail of the implementation. Feel free to include diagrams where relevant and sub-sections for edge-cases and implementation details.
+This section would contain test scenarios in the form of behaviour driven tests in the [Gherkin Syntax](https://cucumber.io/docs/bdd/better-gherkin/) if applicable.
+
+
+## Summary of changes
+
+In an effort to improve where we are, the above guidelines are suggestions for future contributors but no enforced or mandated. This means that we will not place the entire burden on the first person to come along who wants to make a change in a given area. 
+Instead with larger coontributions and proposed changes we would request contributors to add to whatever framework of design documentation already exists.
```
scholzj, line 98: I'm afraid I do not understand what does this mean. Is this trying to say that given missing design docs, you should not need to write them from zero when fixing some smaller bugs and issues?

007-design-docs.md
```diff
@@ -0,0 +1,124 @@
...
+
+Alongside code changes, behavioural changes of the operator will also be documented in the above described design documentation format.
+
+Newly contributed proposals or issues would preferably describe how their changes might alter the design documentation and would be requested to be delivered alongside their code changes.
```
scholzj, line 102: I think this sounds weird ... if you deliver it together with code changes, you are not anymore describing _how your changes might alter the code_. They already exist and alter the code.

katheris:
This proposal seems a bit muddled to me in terms of what the aim is.

The goal at the beginning states "to make contributors more effective by forcing all involved to think through the design and gather feedback from others", to me that would imply that the design doc should be used as a mechanism to discuss new features and make sure they will not cause upgrade or usage problems. This ties in to later in the proposal where it states "It would be appropriate to add a design document when anything which touches a 'Public API'."

However the main benefits also include making it easier for "new contributors to get into the project and understand behaviours of certain components" and "capturing conventions and behaviour enforcements that normally only get picked up and tweaked at review time". This then makes sense with the later sections about having design docs live somewhere slightly separate to the code.

As someone who would be interested in contributing more to Strimzi there are two things that I could see as a benefit from some of the proposed approached:
1. Having a documented mechanism for proposing a new feature. For this I think a PR into this repo seems the current recommended mechanism?
2. Documentation of the high level design of Strimzi. For this I don't think it needs to cover the details of every change, but just clearly stating things like "the CRD is generated by Java annotations" or "the topic and user operator should be optional" would help a new contributor.

From the current proposal as a new contributor I don't think I would understand exactly how to find out about existing design without a lot of reading or how to propose a new feature.

samuel-hawker:
Updated again, would like a re-review  when convenient. Hoping this proposal is a lot more palattable

007-design-docs.md
```diff
@@ -0,0 +1,86 @@
...
+This information will be invaluable to a reader to understand where the configuration points are and how to interact with the system/component.
+
+**Dependencies and Scope**
+Is this intended to be a standalone component, are other components dependent on it, or is it dependent on other components.
+Is this intended for a single Kafka cluster or to interact with multiple.
+
+
+### When is a design document complete
+
+A design document is considered complete when the reviewers and maintainers believe the document to accurately reflect the implementation of the components behaviour.
```
scholzj, line 76: I would probably remove this sentence. I think that at the end the document is never complete because the code evolves and the document has to evolve together with it.

samuel-hawker:
Thanks so much Jakub! Resolved your comment now.

katheris:
@samuel-hawker yeah this makes things a lot clearer and I think contributors would definitely find it useful

007-design-docs.md
```diff
@@ -0,0 +1,87 @@
...
+
+Design document change us considered complete when the reviewers and maintainers believe the updated document to accurately reflect the implementation of the components behaviour. 
+Once approved the document is merged and is understood to be correct, any mistakes or changes to behaviour should be corrected in the form of a pull request alongside the code that changes the behavior.
+A design document is never truly 'completed', as the code evolves the document(s) evolve with it.
+
+A lot of implementation issues, problems and compromises are only clear during or after development of a feature or change, so a design doc is recommended to be written alongside the code development, this helps in avoiding the writing of invalid design documents and help highlight potentially problematic elements of implementation or important holes in the testing strategy. 
+
+Having the design document within the pull request allows contributors and reviewers to evolve the design while developing and during the review. This allows tweaks to the design, for current discussion and future reading, as well as to the code accordingly. This ensures the code and design docs evolve in parallel.
```
ppatierno, line 82: I guess that the PR mentioned here is about the changes/new features are under development so the design doc should be together with them. Isn't it better to clarify it?

samuel-hawker:
@ppatierno I have re-worded that section to be more clear, I hope this addresses your comment. Do let me know if anything is unclear or undesirable.

samuel-hawker:
Did we need any more approvals on this?

@tombentley @Frawless any other comments or suggestions?

scholzj:
@samuel-hawker As per the governance, I guess this would be _other change_ requiring _a ⅔ majority vote_. So I guess you need at least one more approval.

tombentley:
Looking

007-design-docs.md
```diff
@@ -0,0 +1,87 @@
...
+- [KafkaUser Operator](../kafka-user-operator.md)
+(Continued)
+```
+
+The above is just an example but I think it demonstrates the value of a deeper context for how the Strimzi project is written but without describing implementation details. Moreover this should be fairly simple to maintain as the overall architecture and flow doesn't change often and should need little more than a single sentence change when they do.
+
+
+### Where to write a design document
+
+A design document should be a markdown file located in the `design` directory of the repository the component code is located in, for instance all design documents relating to the strimzi operator(s) should be located in the `strimzi-kafka-operator` repository under the `design` folder.
```
tombentley, line 42:
```suggestion
A design document should be a markdown file located in the `design` directory of the repository the component code is located in, for instance all design documents relating to the strimzi operator(s) should be located in the `strimzi-kafka-operator` repository under the `design` directory.
```

007-design-docs.md
```diff
@@ -0,0 +1,87 @@
...
+This information will be invaluable to a reader to understand where the configuration points are and how to interact with the system/component.
+
+**Dependencies and Scope**
+Is this intended to be a standalone component, are other components dependent on it, or is it dependent on other components.
+Is this intended for a single Kafka cluster or to interact with multiple.
+
+
+### When is a design document complete
+
+Design document change us considered complete when the reviewers and maintainers believe the updated document to accurately reflect the implementation of the components behaviour.
```
tombentley, line 76:
```suggestion
Design document change is considered complete when the reviewers and maintainers believe the updated document to accurately reflect the implementation of the components behaviour. 
```

007-design-docs.md
```diff
@@ -0,0 +1,87 @@
...
+
+**Dependencies and Scope**
+Is this intended to be a standalone component, are other components dependent on it, or is it dependent on other components.
+Is this intended for a single Kafka cluster or to interact with multiple.
+
+
+### When is a design document complete
+
+Design document change us considered complete when the reviewers and maintainers believe the updated document to accurately reflect the implementation of the components behaviour. 
+Once approved the document is merged and is understood to be correct, any mistakes or changes to behaviour should be corrected in the form of a pull request alongside the code that changes the behavior.
```
tombentley, line 77:
```suggestion
Once approved, the document is merged and is understood to be correct, any mistakes or changes to behaviour should be corrected in the form of a pull request alongside the code that changes the behavior.
```

PR #6, opened 2020-07-08
feat: write Strimzi ui proposal

matthew-chirgwin:
- initial proposal for a Strimzi ui
- covers high level implementation, goals, suggested deliverables and
points of discussion
~~adds a template for future proposals, based on existing proposals in
the repo (happy to contribute this separately if desired)~~ - moved to #7 
- more detail to be added in following commits as required

Signed-off-by: Matthew Chirgwin <chirmatt@uk.ibm.com>
Signed-off-by: Nic Townsend <nictownsend@uk.ibm.com>


000-template.md
```diff
@@ -0,0 +1,31 @@
+# <Title>
```
ppatierno, line 1: I see it as a proposal to have a template for proposals. Because it's not part of the proposal itself I would prefer to remove it from this PR (focused on the Strimzi UI) and opening a different PR for it.

008-strimzi-ui.md
```diff
@@ -0,0 +1,119 @@
...
+- Uses a [GraphQL](https://graphql.org/) API
+- Should be fully translatable, and accessible
+
+I would suggest that the UI implementation to follow the Model View Controller (MVC) pattern, with clear separation between business logic, and the view logic that renders the state of the UI. This would enable/allow the view logic to be modified/swapped in/out as required, but keep the same business logic. I would suggest using the [Carbon](https://www.carbondesignsystem.com/) design system as the view layer for the contributed UI, given it is open source, is supported, and I can provide both the design and implementation using it. As an example, the following is a mock up of how a Topics page could look using the Carbon design system, which would allow a user to view, create, edit and delete topics in a given cluster:
+
+![Topics mock up](./008-topicsdesign.png)
+
+By maintaining and enforcing a thin view layer, this should not stop other view layers being designed, implemented and contributed in other frameworks, such as [PatternFly](https://www.patternfly.org/v4/) or [Material](https://material.io/) for example.
```
ppatierno, line 27: I am not a UI expert and I really don't know about these frameworks but what does it mean for users who want to change the colors, fonts, and so on. It's not just like changing a CSS, is it more?

008-strimzi-ui.md
```diff
@@ -0,0 +1,119 @@
...
+  - State design
+  - Configuration/capability discovery from the server
+  - Routing and navigation logic
+  - Maintenance/tracing capabilities
+
+I am more than happy to elaborate and collaborate on any of these points here, and would suggest these (and any follow items identified while discussing this proposal) are completed before any client side UI work begins.
+
+#### Topic list page
+
+Given the prerequisites are satisfied, the first user task I would suggest is implemented is a listing of all the topics in a given cluster. This will exercise all the prerequisites, as well as offer a user of Strimzi a new way to interact with their deployment. Using a [gherkin](https://cucumber.io/docs/gherkin/reference/) behavioural syntax to describe the capability, it would:
```
ppatierno, line 70: is it limited just to listing the topic names or even showing basic information like number of partitions and replication factor?

008-strimzi-ui.md
```diff
@@ -0,0 +1,119 @@
...
+
+In all of these cases, capabilities can be added in a prioritised order, and should be added in a phased manner themselves (for example, add a view of the brokers and their configuration, then the ability to modify select configuration).
+
+## Affected projects
+
+I would expect that the main development effort for a UI will be in a new repository in the Strimzi Github organisation. However, given the discussion in https://github.com/strimzi/strimzi-kafka-operator/issues/2540 , one option to source backend Kafka data would be via the [Bridge](https://github.com/strimzi/strimzi-kafka-bridge). As required, this could be extended to support the various Kafka backend calls needed to surface any required information in a UI. In addition, depending on how and where the UI is provided in Strimzi, changes may be needed in the [strimzi-kafka-operator](https://github.com/strimzi/strimzi-kafka-operator) to either define a UI CRD, or deploy a UI deployment if configured to do so via the operator. I would very much welcome discussion around this, and what would make the most sense for the Strimzi project.
```
ppatierno, line 109: I would say that at this point the backend will be developed via a new admin server as per this PR https://github.com/strimzi/strimzi-admin/pull/4 even because providing GraphQL API is not part of the bridge roadmap which is more about adding different protocol heads (HTTP, AMQP, ..) to the Kafka protocol.

README.md
```diff
@@ -5,10 +5,11 @@
 [![License](https://img.shields.io/badge/license-Apache--2.0-blue.svg)](http://www.apache.org/licenses/LICENSE-2.0)
 [![Twitter Follow](https://img.shields.io/twitter/follow/strimziio.svg?style=social&label=Follow&style=for-the-badge)](https://twitter.com/strimziio)
 
-This repository list of proposals for the Strimzi project
+This repository list of proposals for the Strimzi project. The template for new proposals can be found [here](./000-template.md)
 
 |  #  | Title                                                                 |
 | :-: |:----------------------------------------------------------------------|
+|  8  | [Strimzi UI](./008-strimzi-ui.md) |
```
ppatierno, line 12: shouldn't be 6?

000-template.md
matthew-chirgwin: Of course - I will remove it from this PR and raise it in another (with the readme change to support it)

README.md
samuel-hawker: I assume he saw the other two PRs and thought to follow on after those.
I agree this should be 6 and the numbering of other proposal PRs should change prior to merge if they go in after this one.

000-template.md
samuel-hawker: +1

008-strimzi-ui.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Proposal
+
+This proposal will describe a Kafka administration UI capability. It will describe example capabilities which could be offered, cover at a high level how the UI code itself could be implemented (with more low level detail to be provided as required), how this could be delivered, as well as possible future work and how this could integrate with Strimzi today.
+
+### UI implementation
+
+At a high level, I would propose a UI implemented as follows:
+
+- It would be a [React](https://reactjs.org/) based UI
```
samuel-hawker, line 17: I know it might be heavily implied, but it might make sense to clarify that it is written mostly in Javascript due to its dependency on React?

008-strimzi-ui.md
matthew-chirgwin: It will vary by framework. Carbon for example has a theming capability, allowing for changing of colours and such. The different frameworks will of course go about and provide things differently. What I am suggesting here is to keep the view layer (using any framework in theory) seperate from a core set of business logic. This allows for (if needed) the ability for different framework UI implementations to be provided if one enables a particular visual capability over another.

008-strimzi-ui.md
```diff
@@ -0,0 +1,119 @@
...
+- This UI is hosted and provided via an [Express](https://expressjs.com/) server
+- This UI is built in a modular/metadata driven manner - allowing for easy extensibility, modification, and dynamic behaviour at runtime
+- Uses a [GraphQL](https://graphql.org/) API
+- Should be fully translatable, and accessible
+
+I would suggest that the UI implementation to follow the Model View Controller (MVC) pattern, with clear separation between business logic, and the view logic that renders the state of the UI. This would enable/allow the view logic to be modified/swapped in/out as required, but keep the same business logic. I would suggest using the [Carbon](https://www.carbondesignsystem.com/) design system as the view layer for the contributed UI, given it is open source, is supported, and I can provide both the design and implementation using it. As an example, the following is a mock up of how a Topics page could look using the Carbon design system, which would allow a user to view, create, edit and delete topics in a given cluster:
```
samuel-hawker, line 23: Since you've added so many links to other technologies and concepts, is it worth throwing one in for [MVC](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) or is it widely known enough not to merit that?

008-strimzi-ui.md
matthew-chirgwin: I would suggest in the first phase we only fetch and show data. Your correct, a name is a bit meaningless on its own! More than happy to expand this first deliverable with other items, such as partition count and replication factor (as shown in the mockup).

README.md
matthew-chirgwin: I saw two other PRs using 6 and 7 - so I went for 8. Happy to change this to 6.

008-strimzi-ui.md
matthew-chirgwin: Can do - I see no harm in being explicit.

008-strimzi-ui.md
matthew-chirgwin: Good point! I will add one (the Wikipedia description does a nice job of introducing it)

000-template.md
matthew-chirgwin: #7 opened for this

008-strimzi-ui.md
```diff
@@ -0,0 +1,119 @@
...
+
+![Topics mock up](./008-topicsdesign.png)
+
+By maintaining and enforcing a thin view layer, this should not stop other view layers being designed, implemented and contributed in other frameworks, such as [PatternFly](https://www.patternfly.org/v4/) or [Material](https://material.io/) for example.
+
+I would also suggest using Express as a server for this UI. This is due to it's small footprint, modularity and available 3rd party modules, such as [helmet for http security](https://helmetjs.github.io/), [Passport.js for authentication and authorization support](http://www.passportjs.org/) or [the graphql and express-graphql modules for GraphQL server support](https://graphql.org/graphql-js/running-an-express-graphql-server/).
```
samuel-hawker:
```suggestion
I would also suggest using Express as a server for this UI. This is due to its small footprint, modularity and available 3rd party modules, such as [helmet for http security](https://helmetjs.github.io/), [Passport.js for authentication and authorization support](http://www.passportjs.org/) or [the graphql and express-graphql modules for GraphQL server support](https://graphql.org/graphql-js/running-an-express-graphql-server/).
```

006-strimzi-ui.md
```diff
@@ -0,0 +1,120 @@
...
+
+This UI would also be provided with a full set of supporting elements - such as end to end tests, automation (examples of which are [in this section](#a-ui-repository-and-ways-of-working-in-the-repository)) to manage common tasks, and documentation around implementation approach, UI best practise and so on.
+
+### A UI repository, and ways of working in the repository
+
+In addition to the implementation detailed above, I would propose that any UI contributed to Strimzi be managed and developed as follows:
+
+- Be developed in a dedicated repository inside the Strimzi Github organisation
+- Both the server and client are contained in this repo - worked as sub modules in a monorepo
+- That design/discussion/implementation and defect issues are kept solely in this repository
```
samuel-hawker, line 39: If an issue pertaining to some bug in the UI is opened in `strimzi-kafka-operator` is it expected to be closed and moved to the UI, or is a new one opened in the UI repo and would be a blocker/dependency to the operator repo bug being fixed?

006-strimzi-ui.md
```diff
@@ -0,0 +1,120 @@
...
+I would also suggest using Express as a server for this UI. This is due to it's small footprint, modularity and available 3rd party modules, such as [helmet for http security](https://helmetjs.github.io/), [Passport.js for authentication and authorization support](http://www.passportjs.org/) or [the graphql and express-graphql modules for GraphQL server support](https://graphql.org/graphql-js/running-an-express-graphql-server/).
+
+This UI would also be provided with a full set of supporting elements - such as end to end tests, automation (examples of which are [in this section](#a-ui-repository-and-ways-of-working-in-the-repository)) to manage common tasks, and documentation around implementation approach, UI best practise and so on.
+
+### A UI repository, and ways of working in the repository
+
+In addition to the implementation detailed above, I would propose that any UI contributed to Strimzi be managed and developed as follows:
+
+- Be developed in a dedicated repository inside the Strimzi Github organisation
+- Both the server and client are contained in this repo - worked as sub modules in a monorepo
```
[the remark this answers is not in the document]
ppatierno: I had same doubt reading this line the first time but I then realize that the admin API is referred as "backend" in the proposal. Anyway making it clearer is for sure better.

006-strimzi-ui.md
matthew-chirgwin: > If an issue pertaining to some bug in the UI is opened in `strimzi-kafka-operator` is it expected to be closed and moved to the UI, or is a new one opened in the UI repo and would be a blocker/dependency to the operator repo bug being fixed?

The former I would suggest. My reasons for adding this point are two fold: it keeps the issues/context with the code they relate to, and that typically UIs can have lots of issues open at any one time for things like edge cases, feature suggestions and so on.

006-strimzi-ui.md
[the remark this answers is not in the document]
matthew-chirgwin: > I assume this refers to the Espress based server and not to the APIs used by the UI/UI backend which will be in the `strimzi-api` project, right? It might be useful to explain it more explicitely.

It does - I will update the wording, add add a diagram to illustrate the proposed topology

006-strimzi-ui.md
```diff
@@ -0,0 +1,120 @@
...
+
+![Topics mock up](./006-topicsdesign.png)
+
+By maintaining and enforcing a thin view layer, this should not stop other view layers being designed, implemented and contributed in other frameworks, such as [PatternFly](https://www.patternfly.org/v4/) or [Material](https://material.io/) for example.
+
+I would also suggest using Express as a server for this UI. This is due to it's small footprint, modularity and available 3rd party modules, such as [helmet for http security](https://helmetjs.github.io/), [Passport.js for authentication and authorization support](http://www.passportjs.org/) or [the graphql and express-graphql modules for GraphQL server support](https://graphql.org/graphql-js/running-an-express-graphql-server/).
```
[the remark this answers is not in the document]
matthew-chirgwin: > It might be interesting to list the licenses of the main dependencies and check their compatibility with out Apache Software License 2.0.

Absolutley - as it happens most I mention are either Apache 2 themselves, or similar (such as MIT). Any dependency added should be pedegree checked.

matthew-chirgwin:
> I left some rather formal comments. I think one point to discuss is how will the UI be deployed and how will it relate to the different clusters:
> 
>     * Will it be deployed independently on any particular Kafka cluster and be able to handle multiple Kafka clusters at the same time?
> 
>     * Will it be deployed per Kafka cluster? In which case how will it deal with Connect or MIrror Maker if we add such features in the future?
> 
> 
> Did you had any thoughts about this?
> 
> This basically applies for the API server as well I guess, which is bit out of scope for this proposal, but it is related.

I was curious to see what the community would suggest is the best fit/of most value. The ability to manage multiple clusters from one UI I personally think would be quite desirable, but does have additional considerations attached (compatability, user management and so on).

In the case where the UI is deployed per Kafka cluster, the UI server could configured to be made aware of/proxy to 0 to n Mirror Maker/Kafka Connect/other supporting elements related to that cluster, as well as the backend api server for that particular cluster.

006-strimzi-ui.md
```diff
@@ -0,0 +1,129 @@
...
+
+This UI would also be provided with a full set of supporting elements - such as end to end tests, automation (examples of which are [in this section](#a-ui-repository-and-ways-of-working-in-the-repository)) to manage common tasks, and documentation around implementation approach, UI best practise and so on.
+
+The below shows how these pieces could integrate.
+
+![Suggested topology](./006-topology.png)
+
+Where:
+
+- 'Strimzi-api' is a backend server for Kafka data/requests
```
scholzj, line 39:
```suggestion
- 'strimzi-api' is a backend server for Kafka data/requests
```

scholzj:
> I was curious to see what the community would suggest is the best fit/of most value.

Yeah ... I added it to the agenda for the next community meeting.

006-strimzi-ui.md
```diff
@@ -0,0 +1,129 @@
...
+- 'Supporting elements' could be other Kafka related deployments such as Mirror Maker, Kafka Connect and Cruise control. These could also be non Kafka related deployments (not deployed as a part of Strimzi), such as LDAP servers, metric stores etc.
+
+### A UI repository, and ways of working in the repository
+
+In addition to the implementation detailed above, I would propose that any UI contributed to Strimzi be managed and developed as follows:
+
+- Be developed in a dedicated repository inside the Strimzi Github organisation
+- Both the UI server and client are contained in this repo - worked as sub modules in a monorepo
+- That design/discussion/implementation and defect issues are kept solely in this repository
+- All development is done in a behavioural manner - focusing on the end user, and the task they are trying to achieve (see [topic list](#topic-list-page) for an example)
```
tombentley, line 49: What personae are we targeting with this UI. e.g. which of the following does it cater to:

* developers of Kafka applications
* operators of Kafka applications
* operators of the Kafka cluster (and that can potentially be split into someone providing a Kafka service, and someone consuming that Kafka service who is separate from the operator for a Kafka application running on it).

006-strimzi-ui.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Phased delivery
+
+A UI is more than just what a user sees. It also needs to be supported by backend data, services, and resources. To this end, and in bearing in mind the cloud development model mentioned above, all UI capability (both front and backend) should be provided in a phased manner. [I would suggest initially that a topic listing page is the first deliverable](#topic-list-page), but before that can happen, [prerequisites](#prerequisites) need to be discussed, agreed and implemented.
+
+#### Prerequisites
+
+In order to provide any UI, the whole stack which will support it will need to be discussed and considered. This will include (but not be limited to):
+
+- Setting up the required UI build/packaging for deployment to Strimzi
```
tombentley, line 61: Currently the `strimzi-kafka-operators` repo has the build for both the operator code (essentially a maven build), the operator images and the packaging into zipped YAML, Helm charts etc. I suppose we might integrate this as follows:

I can see the `strimzi-ui` repo building the docker images, what we need to figure out is whether the installation bits also belong in `strimzi-ui`, or in `strimzi-kafka-operators`, and what the UX for installing the UI is. E.g. is it just a separate installation, or does it get installed with the CO, or is it installed by the CO. (I'm assuming you get a separate UI per CO, but maybe I'm wrong about that).

008-strimzi-ui.md
nictownsend: One thought we had was to add new HTTP endpoints to bridge if/when needed, then the UI backend server provides a GraphQL layer that fetches from bridge.

If however the admin api proposed provides GraphQL, then the UI bridge endpoints can instead move to the admin api (and either go via kafka admin client, or via bridge).

006-strimzi-ui.md
```diff
@@ -0,0 +1,129 @@
...
+This UI would also be provided with a full set of supporting elements - such as end to end tests, automation (examples of which are [in this section](#a-ui-repository-and-ways-of-working-in-the-repository)) to manage common tasks, and documentation around implementation approach, UI best practise and so on.
+
+The below shows how these pieces could integrate.
+
+![Suggested topology](./006-topology.png)
+
+Where:
+
+- 'strimzi-api' is a backend server for Kafka data/requests
+- 'Supporting elements' could be other Kafka related deployments such as Mirror Maker, Kafka Connect and Cruise control. These could also be non Kafka related deployments (not deployed as a part of Strimzi), such as LDAP servers, metric stores etc.
```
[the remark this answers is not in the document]
ppatierno: Yes, that's the idea. I guess that from a JavaScript UI client perspective is always better/simple interacting with a REST HTTP or GraphQL API than a Kubernetes API server (even if it's HTTP at the end). It means that we are not going to rely on Strimzi custom resource for getting cluster and topic information but using the Kafka Admin client in the backend.
Do you see any specific reason why we could need some interaction with the Kubernetes API?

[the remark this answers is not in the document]
matthew-chirgwin: It will interact with a backend for Kafka data - hence the explicit mentioning of it. Given the discussion it will be the implementation discussed in strimzi/strimzi-admin#4 I imagine. And as required, other data backend APIs could be added, using the UI server as a layer of abstraction/configuration point to enable them.

006-strimzi-ui.md
matthew-chirgwin: In time I think it could be all of the above. I would suggest initially we focus on the operator(s) of a Kafka cluster - covering the exact usages you mention above.

006-strimzi-ui.md
matthew-chirgwin: > Currently the `strimzi-kafka-operators` repo has the build for both the operator code (essentially a maven build), the operator images and the packaging into zipped YAML, Helm charts etc. I suppose we might integrate this as follows:
> 
> I can see the `strimzi-ui` repo building the docker images, what we need to figure out is whether the installation bits also belong in `strimzi-ui`, or in `strimzi-kafka-operators`, and what the UX for installing the UI is. E.g. is it just a separate installation, or does it get installed with the CO, or is it installed by the CO. (I'm assuming you get a separate UI per CO, but maybe I'm wrong about that).

Absolutley - I imagine where the UI is deployed will be the key part of how the implementation will need to be done, but also the UX. As per a previous comment, I am open to what the community thinks is the best place to deploy a UI, and thus also the build/deploy implementation/UX for it as well.

006-strimzi-ui.md
ppatierno: I guess that the users of this UI should authenticate somehow. Are we going to use Kafka ACLs for that or some other system? Using ACLs provides authorization for administrative operations like altering cluster config or altering topics but it even provides auth mechanism for clients using the topics which could be anyway configured somehow using this UI (i.e. listing the topics and specifying who can write/read on topics).

006-strimzi-ui.md
```diff
@@ -0,0 +1,129 @@
...
+- This UI is hosted and provided via an [Express](https://expressjs.com/) server
+- This UI is built in a modular/metadata driven manner - allowing for easy extensibility, modification, and dynamic behaviour at runtime
+- Uses a [GraphQL](https://graphql.org/) API
+- Should be fully translatable, and accessible
+
+I would suggest that the UI implementation to follow the [Model View Controller](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) (MVC) pattern, with clear separation between business logic, and the view logic that renders the state of the UI. This would enable/allow the view logic to be modified/swapped in/out as required, but keep the same business logic. I would suggest using the [Carbon](https://www.carbondesignsystem.com/) design system as the view layer for the contributed UI, given it is open source, is supported, and I can provide both the design and implementation using it. As an example, the following is a mock up of how a Topics page could look using the Carbon design system, which would allow a user to view, create, edit and delete topics in a given cluster:
+
+![Topics mock up](./006-topicsdesign.png)
```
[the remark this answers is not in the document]
matthew-chirgwin: Sure - I will add one now/move images to it

006-strimzi-ui.md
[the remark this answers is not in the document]
matthew-chirgwin: And to your comment @ppatierno - it might be interesting for example for some degree of pod status of a deployed Strimzi instance to be surfaced in a 'live status' type UI (eg X of Y Kafka pods running, Z of Y crashloopbackoff etc). At which point the UI of course would need a service to interact with kube to get pod status.

006-strimzi-ui.md
ppatierno: I would expect one UI per CO but maybe able to handling more Kafka clusters deployed by the same CO in different namespaces. Maybe the UI server should be configured to connect to different backend API; I am assuming that each Kafka cluster deployed in its own namespace will get its own backend server.

006-strimzi-ui.md
```diff
@@ -0,0 +1,129 @@
...
+Additional Kafka capabilities which could be added later include (but would not be limited to):
+
+- Create, update and delete topics on the topics list page (recontextualised as a Topics page - as per mock up)
+- View Consumer group status for a given topic, or the whole cluster
+- View the Brokers in a given cluster, their configuration, and where appropriate, allow modification of broker configuration
+- Provide details of Kafka listener/bootstrap addresses, along with sample configuration, to allow streamline client connectivity
+
+If desired, additional Strimzi provided capabilities, such as Kafka Users, could be leveraged/added/managed through this UI also (for example, allowing access to the UI and its capabilities, general CRUD updates to them etc). 
+
+Finally, one capability which may be of interest (which may have bearing on how this is surfaced in Strimzi itself) is having a single UI, which can manage multiple Kafka clusters deployed via the Strimzi operator.
```
ppatierno, line 113: with this comment are you assuming that if the Strimzi operator deploys N Kafka cluster in N different namespaces, each Kafka cluster gets its own UI server + backend server? It's quite related to the comment I made above.

008-strimzi-ui.md
ppatierno: Yes, that could be a possibility as well. In the end, with my pretty low knowledge about GraphQL, the GQL server can connect to any kind of REST API service using different endpoints for fathering information with one model response to the GraphQL client.
The concern could be that at this point the bridge becomes a mandatory component for the clusters where people want the UI and that at same time it's used for both consumers/producers + UI administrative operations.
I was wondering if the bridge Admin API endpoint is more for administrative operations made by HTTP Kafka clients and not an admin UI. It's a topic for discussion. @strimzi/maintainers wdyt?

006-strimzi-ui.md
matthew-chirgwin: It should indeed (with the option to enable/disable it as required). I think the UI should maintain its own model of auth/authz on resources, and that this then mapped onto a backing implementation. This backing implementation could be swapped in/out as users require. What do you think of this concept, and using Kafka users (for example) to enable this for Strimzi users?

006-strimzi-ui.md
matthew-chirgwin: I am just about to push a commit with a proposed deployment approach which demonstrates how this could be implemented. My thought is there is one UI (and UI server), configured to be aware of N Kafka clusters in N namespaces.

matthew-chirgwin:
> > I was curious to see what the community would suggest is the best fit/of most value.
> 
> Yeah ... I added it to the agenda for the next community meeting.

I have just added in my latest commit a proposed deployment, given some of the discussion so far. Hopefully it will help in this discussion.

FYI @samuel-hawker @ppatierno @tombentley

006-strimzi-ui.md
ppatierno: And what about the backend? Is it one for namespace where a Kafka cluster lives?

006-strimzi-ui.md
```diff
@@ -114,6 +114,28 @@ Finally, one capability which may be of interest (which may have bearing on how
...
+![Suggested deployment](./images/006-deployment.png)
+
+Where:
+
+- The UI is deployed standalone, akin to how the Kafka bridge, Kafka Connect are currently for example (so a new CRD would be required, with configuration needed to reference which cluster(s) to interact with).
+- The UI's spec would contain 0 to N Kafka clusters, which: 
+  - Contain 1 or more 'backend' entries. These represent backend sources of data which can be surfaced in the UI.
+  - One of these entries will be the `strimzi-api` (ie each Kafka kind will contain its own api server).
+  - Each of these entries will define items such as the address to use to connect to it, any auth credentials required, Transport security required, etc.
+- The UI's spec will contain general configuration for the UI - such as certificates to expose to clients etc.
```
ppatierno, line 130: @matthew-chirgwin can you provide an example of these things configured in the new Web UI resource you are mentioning and if needed the additions in the Kafka one, please?

006-strimzi-ui.md
```diff
@@ -114,6 +114,28 @@ Finally, one capability which may be of interest (which may have bearing on how
...
+This UI could be deployed as a part of Strimzi as follows:
+
+![Suggested deployment](./images/006-deployment.png)
+
+Where:
+
+- The UI is deployed standalone, akin to how the Kafka bridge, Kafka Connect are currently for example (so a new CRD would be required, with configuration needed to reference which cluster(s) to interact with).
+- The UI's spec would contain 0 to N Kafka clusters, which: 
+  - Contain 1 or more 'backend' entries. These represent backend sources of data which can be surfaced in the UI.
+  - One of these entries will be the `strimzi-api` (ie each Kafka kind will contain its own api server).
```
ppatierno, line 128: how the backend is deployed? Adding some fields in the `Kafka` resource or always starting from the backend entries in the new Web UI resource? I am thinking about the user experience in configuring the UI where they would need to create a new resource for the actual Web UI (on which I could agree) but even adding some more information to the `Kafka` resource as far as I understood in your proposal.

006-strimzi-ui.md
matthew-chirgwin: Hi @ppatierno - sorry for the delay in replying. Of course - I will come up with some pseudo-yaml to represent the above in context and add it to the proposal.

006-strimzi-ui.md
matthew-chirgwin: I am thinking the `kafka` resource would deploy (if configured - ie it could be optional, but would defer to the ongoing api proposal/discussions) a deployment which is the strimzi-api. It could be as a part of this deployment (say as a piece of status information) output the configuration needed for the UI proposed here. A user could then take that, and insert it at UI deploy time to their CR. I do see the backend needing to exist/be deployed before the UI could be deployed myself - but this could be achieved in a number of different ways. Would you think a more automated approach would be preferable?

006-strimzi-ui.md
matthew-chirgwin: @ppatierno I have just pushed a new commit with some more details - please let me know if there is more you would like me to elaborate on

006-strimzi-ui.md
```diff
@@ -0,0 +1,178 @@
...
+
+This could look as follows in a CR configured by a user (note that common fields such as image, or readiness and liveness probes have been omitted here for conciseness, but would be present in a full CR). Comments are included inline for clarification of intent/type:
+
+```
+...
+spec:
+  clusters: # Array/list
+  - name: 'dev-cluster' # string - metadata, but could be used in the UI (eg for cluster selection if more than one cluster) in the clusters array (order of which respected by the UI)
+    clientCert: <mounted TLS cert to serve to clients> # optional - if omitted, UI server will serve via http rather than https
+    uiConfig: <stringified JSON, containing general UI config - eg features to disable/enable, logging> # string - optional - if provided will override default configuration values
```
ppatierno, line 140: what about a reference to a ConfigMap (containing the JSON) instead of having a potential huge JSON here?

006-strimzi-ui.md
```diff
@@ -0,0 +1,178 @@
...
+- The UI's spec will contain general configuration for the UI - such as certificates to expose to clients etc.
+
+This could look as follows in a CR configured by a user (note that common fields such as image, or readiness and liveness probes have been omitted here for conciseness, but would be present in a full CR). Comments are included inline for clarification of intent/type:
+
+```
+...
+spec:
+  clusters: # Array/list
+  - name: 'dev-cluster' # string - metadata, but could be used in the UI (eg for cluster selection if more than one cluster) in the clusters array (order of which respected by the UI)
+    clientCert: <mounted TLS cert to serve to clients> # optional - if omitted, UI server will serve via http rather than https
```
ppatierno, line 139: so AFAIU this is the certificate for TLS encryption between a client (browser) and the UI server.
Why is this a field of a specific cluster? The UI server is just one (for showing more clusters data) so I guess it should be a root field directly under `spec`. It doesn't bring "information" related to connection between the UI server and a specific cluster.

006-strimzi-ui.md
```diff
@@ -0,0 +1,178 @@
...
+  - Each of these entries will define items such as the address to use to connect to it, any auth credentials required, Transport security required, etc.
+- The UI's spec will contain general configuration for the UI - such as certificates to expose to clients etc.
+
+This could look as follows in a CR configured by a user (note that common fields such as image, or readiness and liveness probes have been omitted here for conciseness, but would be present in a full CR). Comments are included inline for clarification of intent/type:
+
+```
+...
+spec:
+  clusters: # Array/list
+  - name: 'dev-cluster' # string - metadata, but could be used in the UI (eg for cluster selection if more than one cluster) in the clusters array (order of which respected by the UI)
```
ppatierno, line 138: I would say that this one should be the same as the `metadata.name` of the `Kafka` resource describing the cluster to handle.

006-strimzi-ui.md
```diff
@@ -0,0 +1,178 @@
...
+    clientCert: <mounted TLS cert to serve to clients> # optional - if omitted, UI server will serve via http rather than https
+    uiConfig: <stringified JSON, containing general UI config - eg features to disable/enable, logging> # string - optional - if provided will override default configuration values
+    authAuthzMechanism: KafkaUser # string (although perhaps a custom type may be better). 'none' to mean disabled (ie no authentication or authorisation) - see assumptions below.
+    useAuthentication: true # boolean - only valid if authAuthzMechanism is not 'none'. Should the UI check a user's identity - see assumptions below 
+    useAuthorization: true # boolean - only valid if useAuthentication is true - should resources be protected/operated on by authoised users only. If false all users can access/modify all resources (eg topics) - see assumptions below
```
ppatierno, line 143: are these three fields related to an authn/authz mechanism between the UI server and cluster API backend or between a client (browser) and the UI server?

006-strimzi-ui.md
```diff
@@ -0,0 +1,178 @@
...
+spec:
+  clusters: # Array/list
+  - name: 'dev-cluster' # string - metadata, but could be used in the UI (eg for cluster selection if more than one cluster) in the clusters array (order of which respected by the UI)
+    clientCert: <mounted TLS cert to serve to clients> # optional - if omitted, UI server will serve via http rather than https
+    uiConfig: <stringified JSON, containing general UI config - eg features to disable/enable, logging> # string - optional - if provided will override default configuration values
+    authAuthzMechanism: KafkaUser # string (although perhaps a custom type may be better). 'none' to mean disabled (ie no authentication or authorisation) - see assumptions below.
+    useAuthentication: true # boolean - only valid if authAuthzMechanism is not 'none'. Should the UI check a user's identity - see assumptions below 
+    useAuthorization: true # boolean - only valid if useAuthentication is true - should resources be protected/operated on by authoised users only. If false all users can access/modify all resources (eg topics) - see assumptions below 
+    backend: # Array/list
+    - name: 'strimzi-api' # string - metadata
```
ppatierno, line 145: is this information really needed?

006-strimzi-ui.md
```diff
@@ -0,0 +1,178 @@
...
+      address: 'https://route-to-strimzi-api-for-dev-cluster:port' # string - endpoint address for this backend
+      version: 1 # integer - the version of the API this UI will use
+      useTls: true # boolean - should TLS transport security be used
```
ppatierno, line 148: do we really need a flag for saying that TLS is enabled? isn't it enough having the following `tlsXXX` fields to be there?

006-strimzi-ui.md
```diff
@@ -0,0 +1,178 @@
...
+      tlsCert: <mounted TLS certificate for backend server> # required if useTls true - certificate used in SSL handshake with backend
+      tlsKey: <mounted TLS key for tlsCert> # required if useTls true - key used in SSL handshake with backend
```
ppatierno, line 150: what part of TLS stuff do you want to achieve here? If it's just TLS encryption between UI server and backend, you don't need a TLS key, you just need the "CA/root" certificate used for signing the backend API certificate. It will be used by the UI server to validate the backend API certificate itself during the handshake.

006-strimzi-ui.md
```diff
@@ -0,0 +1,178 @@
...
+- The UI's spec would contain 0 to N Kafka clusters the UI will operate against, which will: 
+  - Contain 1 or more 'backend' entries. These represent backend sources of data which can be surfaced in the UI.
+    - One of these entries will be the `strimzi-api` (ie each Kafka kind will contain its own api server).
+  - Each of these entries will define items such as the address to use to connect to it, any auth credentials required, Transport security required, etc.
+- The UI's spec will contain general configuration for the UI - such as certificates to expose to clients etc.
+
+This could look as follows in a CR configured by a user (note that common fields such as image, or readiness and liveness probes have been omitted here for conciseness, but would be present in a full CR). Comments are included inline for clarification of intent/type:
+
+```
+...
```
ppatierno, line 135: I guess this should be the new Kafka web UI custom resource, can you be more explicit with the lines related to apiVersion, kind and so on please?

006-strimzi-ui.md
```diff
@@ -0,0 +1,178 @@
...
+    useAuthorization: true # boolean - only valid if useAuthentication is true - should resources be protected/operated on by authoised users only. If false all users can access/modify all resources (eg topics) - see assumptions below 
+    backend: # Array/list
+    - name: 'strimzi-api' # string - metadata
+      address: 'https://route-to-strimzi-api-for-dev-cluster:port' # string - endpoint address for this backend
+      version: 1 # integer - the version of the API this UI will use
+      useTls: true # boolean - should TLS transport security be used
+      tlsCert: <mounted TLS certificate for backend server> # required if useTls true - certificate used in SSL handshake with backend
+      tlsKey: <mounted TLS key for tlsCert> # required if useTls true - key used in SSL handshake with backend
+...
+```
```
ppatierno, line 152: what about the example showing the additions in the `Kafka` resource for deploying the backend API?

006-strimzi-ui.md
```diff
@@ -0,0 +1,178 @@
...
+- Should be fully translatable, and accessible
+
+I would suggest that the UI implementation to follow the [Model View Controller](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) (MVC) pattern, with clear separation between business logic, and the view logic that renders the state of the UI. This would enable/allow the view logic to be modified/swapped in/out as required, but keep the same business logic. I would suggest using the [Carbon](https://www.carbondesignsystem.com/) design system as the view layer for the contributed UI, given it is open source, is supported, and I can provide both the design and implementation using it. As an example, the following is a mock up of how a Topics page could look using the Carbon design system, which would allow a user to view, create, edit and delete topics in a given cluster:
+
+![Topics mock up](./images/006-topicsdesign.png)
+
+By maintaining and enforcing a thin view layer, this should not stop other view layers being designed, implemented and contributed in other frameworks, such as [PatternFly](https://www.patternfly.org/v4/) or [Material](https://material.io/) for example.
```
ppatierno, line 27: @matthew-chirgwin can you explain me what exactly the "view layer" will contain referring to the MVC pattern? Is it JS, CSS, ..?

006-strimzi-ui.md
samuel-hawker: > I would expect one UI per CO

Is that just an expectation or a suggested enforcement @ppatierno ? I could see a use-case for multiple UIs, one for each Kafka.

006-strimzi-ui.md
samuel-hawker: Sort of similar in use to the `strimzi.io/cluster` label in use, I think that is a good idea.

006-strimzi-ui.md
samuel-hawker: Tying into this, how would we want to limit some users from seeing details from one cluster and not another? Is this the purpose of the field or should this this just be at the root as Paolo suggests?

006-strimzi-ui.md
samuel-hawker: Perhaps we allow for an inline JSON string, or a ConfigMap + Key reference.

006-strimzi-ui.md
samuel-hawker, line 143: I think we should avoid booleans here and instead opt for configuration, i.e. an `authentication` and `authorization` as separate fields that are implicitly `true/on` if you have configured them, meaning something like
```
authentication:
  type: someType
  mechanism: someMechanism
  someCerts: orSomething
```
would read better

006-strimzi-ui.md
samuel-hawker: I suppose this could be used for indentifying the tab/page associated with the backend in the UI?

Also for uniquely identifying elements in a list it is best we have some sort of unique identifier , that could already be fulfilled by the `address` field, but thought needs to be put into whether that is sufficent, and how this affects behaviour if this changes.

006-strimzi-ui.md
ppatierno: @samuel-hawker What's the use-case you see? A kind of multitenancy? Can you elaborate a little bit more please?

006-strimzi-ui.md
samuel-hawker: Agreed, preferred a `tls` field that specifies the version used. Or an even more generic field that lets you specify encryption type etc

006-strimzi-ui.md
samuel-hawker: Yes I was thinking multi-tenancy. I dont have any sort of data saying this would actually be any specific users requirement. But from a practical engineering point of view if we design this to be installable multiple per cluster to begin with if the requirement ever came up it would be trivial to support as opposed to guaranteeing unique naming etc may be harder if we assume just one per cluster to begin with.  Just wondering what having omly one UI buys us as allowing for multiple configured as any team likes.

matthew-chirgwin:
Hi @ppatierno / @samuel-hawker - thank you for the comments - I will reply to them shortly/update the proposal as needed. To mention it for completeness, I have also been discussing the deployment, authentication and authorisation topics with @gmcrobert as there will be significant overlap between my proposal here, and his proposal for an admin server (#9) . I will look to update this proposal once we have settled on an approach together for both pieces.

006-strimzi-ui.md
matthew-chirgwin: I think a config map makes sense - what I would like to avoid is filling the CRD/CR with UI config options, which normally a user would not need to change. WIll update to reflect this

006-strimzi-ui.md
matthew-chirgwin: Correct @ppatierno - it should be a level up. Will correct this.

@samuel-hawker indeed - we may need to suppress some details from some users. Alot of this depends on how authentication and resource access will work - ie a user would be associated with cluster A and B, but not C say. The `name` could be used here for this.

006-strimzi-ui.md
matthew-chirgwin: Agreed - will update to reflect this.

006-strimzi-ui.md
matthew-chirgwin: When I wrote it, I was thinking between User and UI only. But you are right, this really could/should be for the whole stack (UI, api). As per my comment [here](https://github.com/strimzi/proposals/pull/6#issuecomment-658889455), it is probably worth while working this/aligning aspect across both proposals. I see it is a topic in today's community call as well - it would be good to discuss and explore options for this going forward more broadly.

006-strimzi-ui.md
matthew-chirgwin: It was intended as a unique identifier, one for instance the UI could error on if none were available for a given cluster (if there were no api to provide Kafka data, it makes the UI a bit pointless). If there is a better alternative, I am more than happy to update to it

006-strimzi-ui.md
matthew-chirgwin: Agreed, shall update to reflect this.

006-strimzi-ui.md
matthew-chirgwin: Sure - I will add these in. I wanted to highlight the items a UI would need to know to operate (with the rest as a given). But more than happy to be more explicit

006-strimzi-ui.md
matthew-chirgwin: I will work with @gmcrobert on this, using the outcome from today's community call as a starting point.

006-strimzi-ui.md
matthew-chirgwin: It would be both JS and CSS. What this is really trying to highlight is the UI should be Framework agnostic (while providing one for users of Strimzi), and moreover easily extended by others if needs allow.

matthew-chirgwin:
> Does this already incorporate some of the conclusions from the community meeting from Thu last week? I guess we should try to incorporate the main ones.

Hi @scholzj - generally I have been updating the proposal as discussion has occured. However, I have not yet updated the deployment/security aspects of this proposal since the comunity meeting. My plan is to work with @gmcrobert on these points, and then apply our findings/suggestions to the proposal here. I hope that is ok.

scholzj:
@matthew-chirgwin Yeah, that is ok. I just wanted to understand the state and the next steps (e.g. how much to push others to spend time reviewing / approving it :-))

006-strimzi-ui.md
```diff
@@ -0,0 +1,187 @@
...
+
+### A UI repository, and ways of working in the repository
+
+In addition to the implementation detailed above, I would propose that any UI contributed to Strimzi be managed and developed as follows:
+
+- Be developed in a dedicated repository inside the Strimzi Github organisation
+- Both the UI server and client are contained in this repo - worked as sub modules in a monorepo
+- That design/discussion/implementation and defect issues are kept solely in this repository
+- All development is done in a behavioural manner - focusing on the end user, and the task they are trying to achieve (see [topic list](#topic-list-page) for an example)
+- That the repository operate in a cloud like CI/CD manner; lots of small/little and often deliveries of new function, appropriately sized and gated, allowing for the UI to be shipped at any time
```
seanforyou23, line 50: Any preference on CI solution? CircleCI seems to be what all the cool kids are using these days :)

006-strimzi-ui.md
```diff
@@ -0,0 +1,187 @@
...
+- A Kafka backend server for Kafka data/requests the UI server can call
+- UI server configurability/capability - including aspects such as;
+  - Transport security
+  - User session, authentication and authorization capabilities
+  - Cluster metadata - listener addresses, number of brokers, etc
+  - Maintenance/tracing support
+- UI client side capabilities
+  - State design
+  - Configuration/capability discovery from the server
+  - Routing and navigation logic
```
seanforyou23, line 72: Any preference on a routing lib? [Reach](https://reach.tech/router/) and [remix](https://remix.run/) both look promising, might be worth considering over traditional react-router. I don't have a strong preference but haven't worked with anything beyond react-router yet so would need to POC a bit before diving into one of the newer ones.

006-strimzi-ui.md
```diff
@@ -0,0 +1,187 @@
...
+  - Maintenance/tracing capabilities
```
seanforyou23, line 73: Can you help me understand what this means?

006-strimzi-ui.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Proposal
+
+This proposal will describe a Kafka administration UI capability. It will describe example capabilities which could be offered, cover at a high level how the UI code itself could be implemented (with more low level detail to be provided as required), how this could be delivered, as well as possible future work and how this could integrate with Strimzi today.
+
+### UI implementation
+
+At a high level, I would propose a UI implemented as follows:
+
+- It would be a Javascript based UI (using [babel](https://babeljs.io/) to provide latest ECMAScript capabilities in a cross browser compatible manner), using the [React](https://reactjs.org/) framework
```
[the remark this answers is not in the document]
matthew-chirgwin: Absolutley - I havent used TypeScript in anger yet, but would be very interested to see how a Typescript UI (vs Javascript) would work and be written. Would it be possible to explore a TS version of the prototype we have been looking at?

006-strimzi-ui.md
```diff
@@ -0,0 +1,187 @@
...
+- This UI is hosted and provided via an [Express](https://expressjs.com/) server
```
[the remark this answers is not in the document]
matthew-chirgwin: Yes indeed! Build details and implementation (in more detail) can be found here: https://github.com/strimzi/strimzi-ui/pull/2 - which would have `webpack-dev-server` hosting, with a proxy for 'real' backend data

006-strimzi-ui.md
```diff
@@ -0,0 +1,187 @@
...
+- This UI is built in a modular/metadata driven manner - allowing for easy extensibility, modification, and dynamic behaviour at runtime
+- Uses a [GraphQL](https://graphql.org/) API
+- Should be fully translatable, and accessible
```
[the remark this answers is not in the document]
matthew-chirgwin: Re translation/internationalisation - I think we should build it in from day 0 - its way harder to retrofit. In my view we should provide english strings for all elements at a minimum, and there may be a possibility of providing a translated version of those strings as well - I will need to check.
Re a11y - I was going to suggest [`Section 508`](https://www.section508.gov/) and [`WCAG 2.1`](https://www.w3.org/WAI/standards-guidelines/wcag/glance/) as well. Re the tools used, more than happy to explore any reccomened/suggested tooling you may have. I have worked with a few tools as well, we should go for the most reliable, developer and user friendly IMO.

006-strimzi-ui.md
matthew-chirgwin: I have recently used Github actions in a project (https://github.com/ibm-messaging/kafka-java-vertx-starter), which worked great (even for day to day development/management tasks). Happy to explore options however.

006-strimzi-ui.md
matthew-chirgwin: Agreed - what would be optimal is something Remix esq - no configuration. Only concern there is the fact it hasnt been released yet, and some of the licence considerations. I am familiar and know react-router, and I think with the GraphQL capabilities the backend could provide (Introspection), we may be able to do something similar.

006-strimzi-ui.md
matthew-chirgwin: Items such as audit tracking, logging and tracing (both client and (express) server) and so on. I do have ideas/thoughts around solutions to these (along with the rest of the above), and will go into details into `strimzi-ui` in due course

006-strimzi-ui.md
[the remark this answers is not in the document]
seanforyou23: Ok cool! The good thing is since TS is a superset of JS, valid JS is valid TS, so there's immediately a lot of overlap and all your JS knowledge is 100% relevant. Depending on how strictly we configure `tsconfig.json`, more or less of the typing could be required. Point being, we don't have to jump right into the deep end with full `strict` mode (though its good practice). We can keep it flexy while everyone is learning as not to impede the workflow.

Which prototype were you referring to?

006-strimzi-ui.md
seanforyou23: Ah right, ok that's cool with me!

ppatierno: currently, all the Strimzi projects started using Travis CI but after being accepted in the CNCF, we are using Azure Pipelines (even if Travis CI is still in place). I was wondering if it makes sense to add one more CI/CD solution to the picture. @scholzj @Frawless wdyt?

matthew-chirgwin: More than happy to go with Travis/Azure if that is a better fit for Strimzi. The main thing is to have a set of tasks (linting, test etc) run when we need them. The implementation/triggering of these tasks will be the same regardless of the CI used - eg `npm run test`. The only real difference is what we need to do to integrate with the CI - ie a travis.yaml, (GitHub) actions config etc

006-strimzi-ui.md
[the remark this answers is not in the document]
matthew-chirgwin: Sorry - the view layer one shown at Friday's playback - you should have a copy of it. I would be very interested to see what you would reccomend configuration wise, the syntax required as a result, and discuss pros and cons of TS and JS.

matthew-chirgwin:
> This is good stuff! Just had some general questions. Beyond those, the main other things I didn't see addressed were browser/platform/screen-reader support. Do we have any goals/requirements regarding this? Thx!

For browser support I was going to suggest the latest 2 versions of the major browsers (Safari, Edge, Firefox, Chrome). I will add this to the proposal now for further discussion - thanks for calling it out!

For screen readers, I would suggest `JAWS` at a minimum. I would hope that following Section 508/WCAG 2.1 as above would enable most of what we need to enable any screen reader to operate however.

006-strimzi-ui.md
Frawless: I suggest to use azure or github actions. For azure, we have 40 concurrent builds and we currently use it only for operator repo so it should be fine use it here as well. In case we will reach that limit, we can use github actions which should offers 20 concurrent builds (travis has only 5 if I remember correctly).

006-strimzi-ui.md
```diff
@@ -0,0 +1,192 @@
...
+- It would be a Javascript based UI (using [Babel](https://babeljs.io/) to provide latest ECMAScript capabilities in a cross browser compatible manner), using the [React](https://reactjs.org/) framework
+- That the last 2 major versions of the following browsers are supported (via Babel transpiling/polyfilling):
+  - Google Chrome
+  - Microsoft Edge
+  - Mozilla Firefox
+  - Apple Safari
+- This UI is hosted and provided via an [Express](https://expressjs.com/) server
+- This UI is built in a modular/metadata driven manner - allowing for easy extensibility, modification, and dynamic behaviour at runtime
+- Uses a [GraphQL](https://graphql.org/) API
+- Should be fully translatable, and accessible (using a superset of [US Section 508](https://www.section508.gov/) and [WCAG 2.1](https://www.w3.org/WAI/standards-guidelines/wcag/glance/) rulesets).
```
seanforyou23, line 26: 👍 can we also go ahead and nail down the level of conformance for WCAG? AA seems to be the most practical target with great benefits for end-users.

nictownsend:
@scholzj @tombentley @seanforyou23 @samuel-hawker @ppatierno 

For your reference, I have added to the proposal with a section around UI session & authorization

006-strimzi-ui.md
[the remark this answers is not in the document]
neemiasjnr: First of all thank you @matthew-chirgwin for this initiative, I'm very keen to try this out. I just found the project strimzi-ui today and I might be missing a lot of context, so I would like to ask you a couple of things.

Is there a reason for going in this old school way with webpack, plain html and SASS, instead of Create React App for example? I also don't see why Express is necessary, wouldn't make more sense to deploy the static files served by Nginx?

For me as an user would be really nice if I can deploy the Kafka UI easily as a container in Kubernetes just writing a YAML using some CRDs (as the whole Strimzi project is designed to work with Kubernetes/Openshift), and the UI communicating directly to Kafka Bridge API. Since Kubernetes itself has a lot of security built-ins and I would deploy as a private container accessing it via port forward for example, and I wouldn't care too much about performance or security.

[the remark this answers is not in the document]
matthew-chirgwin: Hi @neemiasjnr! Hopefully I can answer your questions :) 

In my experiece, Create React App (CRA) is awesome for getting something up and running quickly. It however (to enable ease of getting started) hides the configuration and tuning options you may need to change to suit a particular use case. In addition, it also brings a whole number of dependencies and boilerplate code that I do not think we will need to build a UI for Strimzi users.

To keep the UI minimal, and to enable the use cases we propose, I have cherry picked things like React, Webpack, Sass (used by CRA for instance), to build this UI. You can see the start of some of this now in [the Strimzi-ui repo](https://github.com/strimzi/strimzi-ui).

As per the proposal, we intend to use Express to serve the static content, but also own and mantain user session magement and handle client security. There are a number of useful 3rd party tools, such as Helmet, Redis and passport.js ready to use to enable this. It also allows for future expansion. For example, abstracting/combining multiple backend systems into one - enabling one UI managing multiple cluster use cases.

We are still working through the details, but the intent is to have exactly what you describe; a CR which references a Strimzi deployment/[backend](https://github.com/strimzi/proposals/pull/9) with some additional configuration options for items such as [security](https://github.com/strimzi/proposals/pull/13), and enabled/disabled capabilities. This is then reconciled by the cluster operator (for example - to be confirmed) and a deployment is created for you.

Hopefully this answers your questions, please let me know if there is anything more I can allborate on.

009-strimzi-ui.md
```diff
@@ -0,0 +1,261 @@
...
+
+## Proposal
+
+This proposal will describe a Kafka administration UI capability. It will describe example capabilities which could be offered, cover at a high level how the UI code itself could be implemented (with more low level detail to be provided as required), how this could be delivered, as well as possible future work and how this could integrate with Strimzi today.
+
+### UI implementation
+
+At a high level, I would propose a UI implemented as follows:
+
+- It would be a Javascript based UI (using [Babel](https://babeljs.io/) to provide latest ECMAScript capabilities in a cross browser compatible manner), using the [React](https://reactjs.org/) framework
```
dlabaj, line 17: Are we moving to typescript?

dlabaj (approved):
Other then the typescript question this looks good to me.

matthew-chirgwin:
Hi @strimzi/maintainers 

In a meeting I was in with @scholzj last week he mentioned he was happy to work on some of the details (such as how the UI is deployed) in follow on issues outside the scope of the proposal. Given this, and the progression made/work done in the Strimzi-ui repo since this proposal was raised, I have updated the proposal with the latest status, and updated some sections to reflect that discussion/details would follow in other issues worked as a part of the UI’s development. This leaves the proposal in a complete state, one which hopefully can be discussed in the next community call.

Key updates/items to mention are:

- Use TypeScript over JavaScript (see comments in the proposal / https://github.com/strimzi/strimzi-ui/issues/19)
- Links added to both the Admin and Security proposals where appropriate
- Links to lower level/rationale docs now in the Strimzi-ui repo

One question I had re the proposal is around CICD. [There has been discussion around CICD](https://github.com/strimzi/proposals/pull/6#discussion_r460109180) - would the maintainers have any objections if we use/commit to using GitHub actions for the UI?

009-strimzi-ui.md
matthew-chirgwin: Hi @dlabaj - I have just updated the proposal to mention this :)

009-strimzi-ui.md
```diff
@@ -0,0 +1,262 @@
...
+
+Authentication in node/Express can be achieved through by http://www.passportjs.org/ which provides a large collection of “strategies” for authenticating a user. For purposes of the proposal, the below diagrams show an oidc flows for a UI oauth dance, although other authentication mechanisms could be used.
+
+### Sequence diagrams:
+
+Note – the log in flow is assuming an oauth dance to retrieve a user token. Depending on the passport strategy and kafka auth mechanism, the UI may need to render its own basic login page.
+
+#### http flow with valid session
+
+![UI http flow with valid session](./images/009-http-valid-session.svg)
```
ppatierno, line 234: as reported in #13 an alternative is to check the JWT token locally without introspection to the IDP.

009-strimzi-ui.md
```diff
@@ -0,0 +1,262 @@
...
+
+![UI websocket handshake flow with session expiry](./images/009-ws-handshake-session-expiry.svg)
+
+#### websocket message flow with session expiry
+
+![UI websocket message flow with session expiry](./images/009-ws-message-session-expiry.svg)
+
+## Affected projects
+
+I would expect that the main development effort for a UI will be in a new repository in the Strimzi Github organisation. However, given the discussion in https://github.com/strimzi/strimzi-kafka-operator/issues/2540 , one option to source backend Kafka data would be via the [Bridge](https://github.com/strimzi/strimzi-kafka-bridge). As required, this could be extended to support the various Kafka backend calls needed to surface any required information in a UI. In addition, depending on how and where the UI is provided in Strimzi, changes may be needed in the [strimzi-kafka-operator](https://github.com/strimzi/strimzi-kafka-operator) to either define a UI CRD, or deploy a UI deployment if configured to do so via the operator. I would very much welcome discussion around this, and what would make the most sense for the Strimzi project.
```
ppatierno, line 254: it sounds to me we are going into the direction of having a dedicated admin API server to avoid adding too many logic to the bridge itself.

ppatierno (approved):

000-template.md
```diff
@@ -1,29 +0,0 @@
-<!-- This template is provided as an example with sections you may wish to comment on with respect to your proposal. Add or remove sections as required to best articulate the proposal. -->
```
scholzj, line 1: I don't think you wanted to delete this here, or?

scholzj (approved):
One nit about the deleted file. But LGTM otherwise.

009-strimzi-ui.md
matthew-chirgwin: Agreed - will update this section to reference Strimzi-admin

009-strimzi-ui.md
matthew-chirgwin: I will add this for completness also

000-template.md
matthew-chirgwin: Good spot - I did not! Must have messed up a rebase. Will restore it in a new commit

matthew-chirgwin:
@strimzi/maintainers I have rebased this PR now the Secuirty proposal has been merged. If you are happy with this proposal, please feel free to merge it :)

scholzj:
@tombentley @Frawless @samuel-hawker Do you have some more comments to this?

Frawless (approved):

samuel-hawker (approved):
LGTM, well put together @matthew-chirgwin

011-strimzi-ui.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Proposal
+
+This proposal will describe a Kafka administration UI capability. It will describe example capabilities which could be offered, cover at a high level how the UI code itself could be implemented (with more low level detail to be provided as required), how this could be delivered, as well as possible future work and how this could integrate with Strimzi today.
+
+### UI implementation
+
+At a high level, I would propose a UI implemented as follows:
+
+- It would be a TypeScript based UI (using [Babel](https://babeljs.io/) to provide latest ECMAScript capabilities in a cross browser compatible manner), using the [React](https://reactjs.org/) framework
```
tombentley, line 17:
```suggestion
- It would be a TypeScript based UI (using [Babel](https://babeljs.io/) (MIT license) to provide latest ECMAScript capabilities in a cross browser compatible manner), using the [React](https://reactjs.org/) framework (MIT license)
```

011-strimzi-ui.md
```diff
@@ -0,0 +1,262 @@
...
+- That the last 2 major versions of the following browsers are supported (via Babel transpiling/polyfilling):
+  - Google Chrome
+  - Microsoft Edge
+  - Mozilla Firefox
+  - Apple Safari
+- This UI is hosted and provided via an [Express](https://expressjs.com/) server
```
tombentley, line 23:
```suggestion
- This UI is hosted and provided via an [Express](https://expressjs.com/) server (MIT license)
```

011-strimzi-ui.md
```diff
@@ -0,0 +1,262 @@
...
+- This UI is built in a modular/metadata driven manner - allowing for easy extensibility, modification, and dynamic behaviour at runtime
+- Uses a [GraphQL](https://graphql.org/) API
```
tombentley, line 25:
```suggestion
- Uses a [GraphQL](https://graphql.org/) API (MIT license)
```

011-strimzi-ui.md
```diff
@@ -0,0 +1,262 @@
...
+- Should be fully translatable, and accessible (using a superset of [US Section 508](https://www.section508.gov/) and [WCAG 2.1](https://www.w3.org/WAI/standards-guidelines/wcag/glance/) rulesets).
+
+I would suggest that the UI implementation to follow the [Model View Controller](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) (MVC) pattern, with clear separation between business logic, and the view logic that renders the state of the UI. This would enable/allow the view logic to be modified/swapped in/out as required, but keep the same business logic. I would suggest using the [Carbon](https://www.carbondesignsystem.com/) design system as the view layer for the contributed UI, given it is open source, is supported, and I can provide both the design and implementation using it. As an example, the following is a mock up of how a Topics page could look using the Carbon design system, which would allow a user to view, create, edit and delete topics in a given cluster:
```
tombentley, line 28:
```suggestion
I would suggest that the UI implementation to follow the [Model View Controller](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) (MVC) pattern, with clear separation between business logic, and the view logic that renders the state of the UI. This would enable/allow the view logic to be modified/swapped in/out as required, but keep the same business logic. I would suggest using the [Carbon](https://www.carbondesignsystem.com/) design system as the view layer for the contributed UI, given it is open source, is supported, and I can provide both the design and implementation using it. As an example, the following is a mock up of how a Topics page could look using the Carbon design system (Apache 2 license), which would allow a user to view, create, edit and delete topics in a given cluster:
```

011-strimzi-ui.md
```diff
@@ -0,0 +1,262 @@
...
+And Topic 'SampleTopic' has '3' partitions and '2' replicas
+
+Scenario: Viewing a specific topic (when it does not exist in the cluster):
+Given I am a user of the UI
+When I navigate to the topic listing
+And I filter for topic 'SampleTopic'
+Then I am told topic 'SampleTopic' does not exist in the cluster
+```
+
+#### Follow on capabilities
```
tombentley, line 109:
```suggestion
#### Follow-on capabilities
```

011-strimzi-ui.md
```diff
@@ -0,0 +1,262 @@
...
+
+I am suggesting this approach for the following reasons:
+
+- It follows the model of similar supporting capabilities currently available in Strimzi
+- It allows for multiple Kafka clusters to be managed via a single UI (a potential future work item)
+
+This approach does have a few assumptions. These being:
+
+- A `strimzi-api` being deployed as a part of the Kafka cluster (ie one per namespace/cluster). This deployment is then referenced in the UI's CR (as above).
+- A handshake/metadata exchange will be required for the UI to discover and integrate with each backend. I expect this exchange to be the retrieval of a GraphQL schema, so they can be combined/unified by the UI server, as well as any other metadata appropriate to that backend (eg Kakfa version of the cluster a `strimzi-api` is configured to use).
```
tombentley, line 190:
```suggestion
- A handshake/metadata exchange will be required for the UI to discover and integrate with each backend. I expect this exchange to be the retrieval of a GraphQL schema, so they can be combined/unified by the UI server, as well as any other metadata appropriate to that backend (eg Kafka version of the cluster a `strimzi-api` is configured to use).
```

011-strimzi-ui.md
```diff
@@ -0,0 +1,262 @@
...
+- The backend's registered with the UI will need to be version aware and backwards compatible - ie version 2 of an api also needs to support the version 1 api.
```
tombentley, line 191:
```suggestion
- The backends registered with the UI will need to be version-aware and backwards compatible - ie version 2 of an api also needs to support the version 1 api.
```

011-strimzi-ui.md
```diff
@@ -0,0 +1,262 @@
...
+- Express – UI server, handling sessions for the client
+- Client – Browser running the strimzi UI, executing graphql queries/subscriptions
+- Kafka – brokers (connected to via admin clients)
+- Auth – external auth provider, oauth for Kafka (and possible oauth dance for UI)
+
+### Proposed technology:
+
+The UI is being hosted by an Express server. Express has session middleware - https://www.npmjs.com/package/express-session - that can create and persist server-side sessions, using a cookie as a key to hydrate a session into the incoming express request. Proposed session store is a `redis` container as it's a lightweight key/value (and the session will just be storing a token value).
+
+Authentication in node/Express can be achieved through by http://www.passportjs.org/ which provides a large collection of “strategies” for authenticating a user. For purposes of the proposal, the below diagrams show an oidc flows for a UI oauth dance, although other authentication mechanisms could be used.
```
tombentley, line 226:
```suggestion
Authentication in node/Express can be achieved through by http://www.passportjs.org/ (MIT license) which provides a large collection of “strategies” for authenticating a user. For purposes of the proposal, the below diagrams show an oidc flows for a UI oauth dance, although other authentication mechanisms could be used.
```

tombentley (approved):
I few very minor points (as a review it's helpful to know that the licenses of all the proposed dependencies are open source). LGTM.

scholzj:
Thanks for putting the proposal together and working on it @matthew-chirgwin!

PR #8, opened 2020-07-10
Add connector and connector task restart proposal

ajborley:
This proposal adds support for restarting Kafka Connect connectors and
connector tasks via annotations and via the default behaviour of the
`KafkaConnector`, and `KafkaMirrorMaker2` operators.


006-restarting-kafka-connect-connectors-and-tasks.md
```diff
@@ -0,0 +1,116 @@
...
+To disable the new default behaviour a new annotation named `strimzi.io/auto-restart` with a value of `false` can be applied to the `KafkaConnector` and `KafkaMirrorMaker2` CRs. This will be useful when users know that external factors will cause connectors/tasks to be in the `FAILED` state and they do not want the operator attempting automatic restarts. It also means users can keep the current behaviour of the `KafkaConnector` and `KafkaMirrorMaker2` operators if they wish.
+
+For example, the following CR disables the automatic restarting of the connector and it's tasks when they are in the `FAILED` state:
+```
+apiVersion: eventstreams.ibm.com/v1alpha1
+kind: KafkaConnector
+metadata:
+  name: my-source-connector
+  annotations:
+    strimzi.io/auto-restart: "false"
```
samuel-hawker, line 105: Should this necessarily be an annotation or part of the spec instead? Do any of the maintainers have opinions on this?

006-restarting-kafka-connect-connectors-and-tasks.md
```diff
@@ -0,0 +1,116 @@
...
+1. Two new annotations that cause the operator to restart connectors or tasks. They can be applied to the `KafkaConnector`, and the `KafkaMirrorMaker2` custom resources. The annotation acts as a trigger for a single restart call by the operator, and is removed from the CR when the restart REST API call is successfully called.
+
+2. New default behaviour in the `KafkaConnector` and `KafkaMirrorMaker2` operators that automatically attempts to restart connectors and tasks if they are in the `FAILED` state. The `KafkaConnector` and `KafkaMirrorMaker2` CRs can use a new annotation to disable the automatic restarting behaviour if required.
+
+These two features could be delivered separately as they do not rely on each other.
+
+
+### `strimzi.io/restart` annotation
+
+An annotation named `strimzi.io/restart` can be applied to the `KafkaConnector` and `KafkaMirrorMaker2` custom resources. The value of the annotation and the CR it is applied to determines the connector that is restarted. Once the operator successfully calls the restart REST API endpoint, the CR is patched to remove the annotation. The operator attempts to call the restart REST API endpoint once per reconcile cycle.
```
ppatierno, line 40: "The operator attempts to call the restart REST API endpoint once per reconcile cycle." is I got it right, the operator will restart the connector/task only on the next reconcile? why not as soon as the annotation is applied?

006-restarting-kafka-connect-connectors-and-tasks.md
```diff
@@ -0,0 +1,116 @@
...
+
+When applied to a `KafkaConnector` CR, the connector defined by the CR is restarted. Any value can be assigned to the annotation. For example:
+```
+apiVersion: eventstreams.ibm.com/v1alpha1
```
ppatierno, line 44: `kafka.strimzi.io` ? ;-)

006-restarting-kafka-connect-connectors-and-tasks.md
```diff
@@ -0,0 +1,116 @@
...
+metadata:
+  name: my-source-connector
+  annotations:
+    strimzi.io/restart: "true"
+  ...
+```
+
+When applied to a `KafkaMirrorMaker2` CR, the annotation value contains the name of the connector that is restarted. For example:
+```
+apiVersion: eventstreams.ibm.com/v1alpha1
```
ppatierno, line 55: ditto as above ;-)

006-restarting-kafka-connect-connectors-and-tasks.md
```diff
@@ -0,0 +1,116 @@
...
+kind: KafkaMirrorMaker2
+metadata:
+  name: my-mm2-cluster
+  annotations:
+    strimzi.io/restart: "source-cluster->target-cluster.MirrorSourceConnector"
```
ppatierno, line 60: so ... while `strimzi.io/restart` makes a lot of sense to me on a `KafkaConnector` resource because it's clear that you are asking to restart the "resource" so the connector, it's not so clear on a `KafkaMirrorMaker2` resource where it seems that your intention is to restart MM2. Maybe we can change this annotation in `strimzi.io/restart-connector` ?

006-restarting-kafka-connect-connectors-and-tasks.md
```diff
@@ -0,0 +1,116 @@
...
+
+When applied to a `KafkaMirrorMaker2` CR, the annotation value contains the name of the connector and the ID of the task that is restarted, delimited by a `:` character. For example:
+
+```
+apiVersion: eventstreams.ibm.com/v1alpha1
+kind: KafkaMirrorMaker2
+metadata:
+  name: my-mm2-cluster
+  annotations:
+    strimzi.io/restart-task: "source-cluster->target-cluster.MirrorSourceConnector:2"
```
ppatierno, line 87: as for the restart connector annotation, maybe here we should have a clearer `strimzi.io/restart-connector-task`?

006-restarting-kafka-connect-connectors-and-tasks.md
ppatierno: I would agree with you Sam. I think it's compared to the `KafkaConnect` resource where we have `strimzi.io/use-connector-resources: "true"` as an annotation as well ... but tbh I would say that it should be in the spec as well.

ppatierno: @ajborley I have the same comment here for `KafkaMirrorMaker2` that `auto-restart` is not so clear about what is auto restarted so maybe it could be `auto-restart-connector-task` os something like that?

gunnarmorling:
Great effort; (automated) restarts also come up in the Debezium community repeatedly. Two questions after reading the proposal:

* Did you consider to implement this functionality in Kafka Connect itself instead of the operator? Perhaps the upstream Kafka community would be open for a KIP around this?
* Did you consider a way to distinguish between cases where it makes sense to restart and others where it doesn't? E.g. in case of wrong credentials, restarting won't help much until the credentials have been corrected.

006-restarting-kafka-connect-connectors-and-tasks.md
samuel-hawker: I would maybe suggest `auto-restart-task` as I assume it would be clear that it would be a connector task, unless there is still some ambiguity here?

006-restarting-kafka-connect-connectors-and-tasks.md
samuel-hawker: Applying the annotation would trigger a reconcile, no? 

Apply anno -> update in kube resource -> triggers reconcile

006-restarting-kafka-connect-connectors-and-tasks.md
scholzj: I think the `strimzi.io/use-connector-resources` started from the idea to see how it works first before moving it into the `.spec` section. Not uncommon in Kubernetes. I think @tombentley was the main person behind it. So the question is whether to continue this way or not. 

I personally can live with both for the auto restart flag. But the manual restart should be definitely annotation.

ppatierno: I totally agree about the manual restart as an annotation (it's kind of the rebalance annotation for starting the rebalance action in the `KafkaRebalance` resource) but I would like more having `auto-restart` in the spec (so `use-connector-resources` should be moved as well for having consistency for the users).

tombentley: I think the auto-restart should be part of the spec. In general we should try to limit the number and complexity of annotations. They have numerous disadvantages when compared to the spec: They're not really documented systematically, they're less discoverable and they're not validated by the apiserver. I think we will eventually move `strimzi.io/use-connector-resources` into the spec (and change the default), but likely that's with an API bump.

I also have questions about how we should communicate to the user (via the status) that the connector and/or task is being restarted. I think repeated restart is analogous to a pod crashlooping and suggests that the user needs to take action to understand/resolve a problem.

Finally, I'd have to double check, but does a connector/task transitioning to the failed state, and being restarted cause a rebalance of the tasks within connect? If so then doing that repeatedly _ad infinitum_ is not going to be a good idea, and we should have some kind of (configurable?) policy we limit the negative effects on other connectors within the same connect cluster.

006-restarting-kafka-connect-connectors-and-tasks.md
ajborley: 👍 I have added `strimzi.io/restart-connector` for `KafkaMirrorMaker2` CRs

006-restarting-kafka-connect-connectors-and-tasks.md
ajborley: 👍 I have added `strimzi.io/restart-connector-tasks` for `KafkaMirrorMaker2` CRs

006-restarting-kafka-connect-connectors-and-tasks.md
ajborley: Oops :) Have updated

006-restarting-kafka-connect-connectors-and-tasks.md
ajborley: :)

006-restarting-kafka-connect-connectors-and-tasks.md
ppatierno: Yes right, it doesn't say "periodic" reconcile :-)

006-restarting-kafka-connect-connectors-and-tasks.md
ajborley: The latest commit changes the auto-restart to be part of the `KafkaConnector` and `KafkaMirrorMaker2` specs.

@tombentley These are good questions that I've been looking into today. 

I agree that we should add some status info to show the restart count, perhaps something like the following, which would increment the count every time the restart REST API is invoked and record the time when it last restarted:
```
apiVersion: strimzi.io/v1alpha1
kind: KafkaConnector
...
status:
  autoRestart: 
    count: 2
    lastRestartTimestamp: "2020-07-13T21:59:09.832Z"
```

For your second point, since Incremental Cooperative Rebalancing ([KIP-415](https://cwiki.apache.org/confluence/display/KAFKA/KIP-415%3A+Incremental+Cooperative+Rebalancing+in+Kafka+Connect)) went into Kafka 2.3, the cost of connect task rebalancing has much reduced, but I agree that we should limit repeated restarts. We could add a back-off algorithm that watches restarts and increases the number of reconcile loops between them if there have been multiple restarts in a row. What do you think?

ajborley:
@gunnarmorling - thanks for your comments :)

> * Did you consider to implement this functionality in Kafka Connect itself instead of the operator? Perhaps the upstream Kafka community would be open for a KIP around this?

IMO auto-restart functionality fits better in the 'Kubernetes-native' world of Strimzi, where the operator is trying to reach a particular connector state. Conversely, Kafka Connect is driven by the Connect REST API, used to explicitly monitor and configure connectors. If the auto-restart function is found to be particularly useful it may be worth a KIP.

> * Did you consider a way to distinguish between cases where it makes sense to restart and others where it doesn't? E.g. in case of wrong credentials, restarting won't help much until the credentials have been corrected.

This would likely require log monitoring, or updates to the connect REST API to provide extra info. Determining when a connector is not restartable sounds like a hard problem :) Tom raised the point about how repeated restarts are analogous to a pod crashlooping, and that it may be a good idea to limit restarts. I agreed with that and responded with [this comment](https://github.com/strimzi/proposals/pull/8#discussion_r453990339) - if you have thoughts or ideas around handling repeated restarts it would be good to hear them.

tombentley:
Those are good points @gunnarmorling. I think the main thrust of this proposal is to allow use of the existing functionality that Kafka Connect exposes in a Kube native way, that justifies the annotation for triggering a restart. The automatic restart kind-of makes sense by analogy to the way `Pods` work under a `ReplicaSet`. If restarting is found to be a known workaround for some flakiness then it alleviates toil by having the operator do that for you, however that could equally be handled in Kafka Connect itself. 

The current extent of error handling, is somewhat configurable ([KIP-298](https://cwiki.apache.org/confluence/display/KAFKA/KIP-298%3A+Error+Handling+in+Connect) and fixed for `SourceTasks` by [KAFKA-8320](https://issues.apache.org/jira/browse/KAFKA-8320)). 

* @gunnarmorling it sounds like Debezium has use cases KIP-298 still doesn't satisfy, if so could you elaborate?
* @ajborley I guess you need to elaborate on why someone might the auto-restart feature described here instead of (or as well as) the functionality in KIP-298.

gunnarmorling:
> it sounds like Debezium has use cases KIP-298 still doesn't satisfy, if so could you elaborate?

@tombentley, the way KIP-298 implements things, it proved not to be overly helpful for our purposes unfortunately. I hope I remember the details right, but I think it will soley call `poll()` again, if a specific exception type, `RetriableException`, is raised by a connector. Whereas for our case we really need to restart the Debezium connector; E.g. when loosing a DB connection, we'll have to recreate this, we might have to re-do an interrupted snapshot of data, etc. That's why we have implemented a restarting capability internally: if we deem an exception is retriable, we won't propagate it to Connect; if `poll()` is invoked the next time, we check whether there was a failure before, and if so, repeat the same logic which otherwise runs via `SourceConnector#start()`.

Without having thought too much about it, it might help if Kafka Connect had a sibling to `RetriableException`, `RestartableConnection`. If a connector raises that one, KC would stop and re-start the connector and/or task (details tbd.). That also addresses @ajborley's remark on log parsing: handling a restart _within_ Kafka Connect allows for a safe detection of this situation via exception propagation.

You then might be left with just a much smaller need for restarts via the operator, let's say in case of an OOME. But this would be better handled by restarting the entire pod and it might even be handled by the _Connect_ operator already.

006-restarting-kafka-connect-connectors-and-tasks.md
```diff
@@ -0,0 +1,137 @@
...
+apiVersion: strimzi.io/v1alpha1
+kind: KafkaConnector
+metadata:
+  name: my-source-connector
+spec:
+  autoRestart: "false"
+  ...
+```
+
+Similarly, a new field in the `KafkaMirrorMaker2` spec named `autoRestartConnectorsAndTasks` with a value of `false` can be applied, to disable auto-restarting connectors and tasks.
```
scholzj, line 117: I'm not sure I understand why does `KafkaConnector` have `autoRestart` while `KafkaMirrorMaker2` has `autoRestartConnectorsAndTasks`. It seems to do the same in both cases - restart both connectors and tasks - so why not use the same name?

ajborley: I did this to go along with @ppatierno's comment on the annotations (https://github.com/strimzi/proposals/pull/8#discussion_r453168363). It was felt that the `strimzi.io/restart` and `strimzi.io/restart-task` annotations made sense on a KafkaConnector resource because it's clear that users are asking to restart the resource, but it wasn't so clear on a KafkaMirrorMaker2 resource where the intention is to restart the MM2 connectors - hence I updated the annotations to `strimzi.io/restart-connector` and `strimzi.io/restart-task` for KafkaMirrorMaker2 CRs.
When I added the `autoRestart` field to the KafkaConnector spec, I expected the same issue would apply, so I named it `autoRestartConnectorsAndTasks` for KafkaMirrorMaker2. I'm happy to give them the same name if people agree - @ppatierno what do you think?

scholzj: I guess I can see that angle - yeah, it might be confusing if you don't know what it does. So I think I'm ok with the current naming. Thanks for explaining.

gunnarmorling:
Hey, I'd still be very interested in the answer to @tombentley's question here:

> @ajborley I guess you need to elaborate on why someone might the auto-restart feature described here instead of (or as well as) the functionality in KIP-298.

Thanks a lot!

ajborley:
Hi @gunnarmorling 
I would say that the auto-restart function discussed here is at a higher level to the way Kafka Connect handles retries in [KIP-298](https://cwiki.apache.org/confluence/display/KAFKA/KIP-298%3A+Error+Handling+in+Connect). As you say, only certain exceptions are caught for retrying and the focus appears to be on record processing, when failures can also occur on connector or task start-up.
This proposal uses the Connect REST API /restart endpoint to cause a full restart of the task or connector when it is in a `FAILED` state, which may be useful if (as in the case you outlined) the retry logic in KIP-298 is not enough to get the connector running happily again.

I have updated the proposal with some extra info to cover the cases that @tombentley mentioned in [this comment](https://github.com/strimzi/proposals/pull/8#discussion_r453529383)

scholzj:
This seems to be progressing nicely. @Frawless can you review this as well please? Tom Bentley is out this week. So hopefully early next week when he comes back we can close it as approved.

gunnarmorling:
> which may be useful if (as in the case you outlined) the retry logic in KIP-298 is not enough to get the connector running happily again.

Sure, we agree that KIP-298 doesn't go far enough. But again: why don't we try and improve things in Kafka Connect upstream, instead of working around the issue here at the operator level? That way _all_ users of Connect will benefit from this nice improvement.

scholzj:
> Sure, we agree that KIP-298 doesn't go far enough. But again: why don't we try and improve things in Kafka Connect upstream, instead of working around the issue here at the operator level? That way all users of Connect will benefit from this nice improvement.

My view - because it might take too long and it is very unclear whether such KIP will be completely ignored or accepted etc. It would be good if someone starts this effort also in Kafka to have it done there in the long term. But we should not wait for it with our own feature.

ajborley:
> > Sure, we agree that KIP-298 doesn't go far enough. But again: why don't we try and improve things in Kafka Connect upstream, instead of working around the issue here at the operator level? That way all users of Connect will benefit from this nice improvement.
> 
> My view - because it might take too long and it is very unclear whether such KIP will be completely ignored or accepted etc. It would be good if someone starts this effort also in Kafka to have it done there in the long term. But we should not wait for it with our own feature.

I agree with @scholzj here. This proposal fits with the Strimzi 'Kubernetes-native' behaviour using existing Kafka APIs and shouldn't be too complex to code up. I'd be happy to work with you on a Kafka KIP around this if you'd like. I liked your idea for a `RestartableConnection`.

scholzj:
Thanks for the proposal!

PR #9, opened 2020-07-13
Proposal for an admin REST and GraphQL Server

gmcrobert:
for Strimzi.

Signed-off-by: Graeme McRobert <mcrobert@uk.ibm.com>

### DIFF
--- a/011-admin-server.md
+++ b/011-admin-server.md
@@ -7,26 +7,54 @@ The Strimzi Admin repository was setup in December 2019 to hold an implementatio
 An API server will:
 * provide a consolidated backend API for additional interfaces like a browser based UI or a CLI.
 * be capable of supporting both REST and GraphQL interfaces.
-* allow for more sophisticated APIs to be built on top of the existing APIs providing value-add to existing content.
+* allow for more sophisticated APIs to be built on top of the existing APIs like the kubernetes API providing value-add to existing content.
 
 ## Proposal
 
 ### The server
-The proposal is to make a modular server using the Vert.x-Web toolkit which has strong support for creating both REST and GraphQL interfaces. The server contains a single Vert.x verticle which listens for inbound traffic on 1 or more configurable ports. Allowing multiple configurable ports will allow the server to expose different functionality on each port, the simplest example of this being exposing an HTTP API on one port and a separate HTTPS API on a different port. The server module itself does not contain any API implementations but defines a service interface using the Java SPI and will load modules that implement that interface from the classpath. The implementation of the APIs that the server then exposes can be created independently and will be automatically included in the running server if they are available on the classpath when the server bootstraps.
+The proposal is to make a modular server using the Vert.x-Web toolkit which has strong support for creating both REST and GraphQL interfaces.
+The server contains a single Vert.x verticle which listens for inbound traffic on 1 or more configurable ports.
+Allowing multiple configurable ports will allow the server to expose different functionality on each port, the simplest example of this being exposing an HTTP API on one port and a separate HTTPS API on a different port.
+The server module itself does not contain any API implementations but defines a service interface using the Java SPI and will load modules that implement that interface from the classpath.
+The implementation of the APIs that the server then exposes can be created independently and will be automatically included in the running server if they are available on the classpath when the server bootstraps.
 
-The reason for the modular approach is that it allows the strimzi project to create a base set of functionality that can easily be extended by a downstream project. Not only is it easy to extend the API but it can be done in isolation. This removes the requirement of the extensions needing to track the strimzi source repositories. For example, lets assume that strimzi only implements a Kafka administration API which provides a set of APIs based around the Kafka Admin Client. A downstream project then would like to add a Schema Registry to the cluster and needs to add an API in support of the new functionality. With the modular server, they can define the API and its implementation, bundle the new functionality into a JAR file and add the new JAR file to the classpath. A secondary benefit of a modular approach is that it allows the strimzi user to configure the server to only contain the functionality that they would like to use providing for a leaner server.
+The reason for the modular approach is that it allows the strimzi project to create a base set of functionality that can easily be extended by a downstream project.
+Not only is it easy to extend the API but it can be done in isolation.
+This removes the requirement of the extensions needing to track the strimzi source repositories.
+For example, lets assume that strimzi only implements a Kafka administration API which provides a set of APIs based around the Kafka Admin Client.
+A downstream project then would like to add a Schema Registry to the cluster and needs to add an API in support of the new functionality.
+With the modular server, they can define the API and its implementation, bundle the new functionality into a JAR file and add the new JAR file to the classpath.
+A secondary benefit of a modular approach is that it allows the strimzi user to configure the server to only contain the functionality that they would like to use providing for a leaner server.
 
-The classpath of the server can be configured at build time or run time. To configure at build time, the server and dependant modules can be bundled in a fat-JAR and obviously, at run time, the JVM can be configured in multiple ways to include JAR files on the classpath.
+The classpath of the server can be configured at build time or run time.
+To configure at build time, the server and dependant modules can be bundled in a fat-JAR and obviously, at run time, the JVM can be configured in multiple ways to include JAR files on the classpath.
 
 ### REST API Implementations
-REST modules would use the Vert.x Web API Contract support and specifically the OpenAPI3 support to define the shape of the API and bind the operations to a set of handlers. The advantage of using the OpenAPI3 definition is the support it contains for validation which allows you to build a more secure API without having to pollute the implementation with validation code. It leaves the implementation of the operations to focus entirely on the business logic. A secondary advantage of using the OpenAPI3 specification is the numerous tools that are freely available to create generated documentation of the API.
+REST modules would use the Vert.x Web API Contract support and specifically the OpenAPI3 support to define the shape of the API and bind the operations to a set of handlers.
+The advantage of using the OpenAPI3 definition is the support it contains for validation which allows you to build a more secure API without having to pollute the implementation with validation code.
+It leaves the implementation of the operations to focus entirely on the business logic.
+A secondary advantage of using the OpenAPI3 specification is the numerous tools that are freely available to create generated documentation of the API.
  
- The Vert.x OpenAPI3 support will read an OpenAPI3 specification YAML file and allow the developer to map the original operations to handler methods. It creates a Vert.x Router which can be mounted on a Vert.x HTTP/S server. The module would implement the service interface defined by the server which will contain a method that the server can call to load the OpenAPI3 router and mount the router as a subrouter at a specified mount point to the root path. So, for example, if the OpenAPI3 spec defined endpoints `GET /topics` and `POST /topics` and these were mounted on the root path at the mount point `/admin` then the server would listen for incoming requests of `GET /admin/topics` and `POST /admin/topics` and route them to the handlers defined in the module. The mount points create a namespace so should not be empty and should be unique and the server should police that in order to prevent name clashes.  
+The Vert.x OpenAPI3 support will read an OpenAPI3 specification YAML file and allow the developer to map the original operations to handler methods.
+It creates a Vert.x Router which can be mounted on a Vert.x HTTP/S server.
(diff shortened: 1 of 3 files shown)


006-admin-server.md
```diff
@@ -0,0 +1,67 @@
+# Strimzi Admin Server
+
+The Strimzi Admin repository was setup in December 2019 to hold an implementation of an Admin API. The repository has not had any contributions yet so this proposal sets out what the structure of an Admin Server might look like in order to get to the initial implementation of the server.
+
+## Motivation
+
+An API server will:
+* provide a consolidated backend API for additional interfaces like a browser based UI or a CLI.
+* be capable of supporting both REST and GraphQL interfaces.
+* allow for more sophisticated APIs to be built on top of the existing APIs providing value-add to existing content.
```
ppatierno, line 10: what are the "existing" APIs it's referring to?

006-admin-server.md
```diff
@@ -0,0 +1,67 @@
...
+### Security
+Vert.x networking is based on Netty and works in a very similar manner. When a request is received by the server, it is funneled through a pipe and that pipe has a number of interceptors or handlers that can inspect the contents at that point and ignore or modify it before allowing it to proceed through the pipe. Vert.x also creates a routing context that travels through the pipe with the request and is available for inspection and/or modification by all the handlers. A handler can decided whether the request should be passed on to the next downstream handler or the request is complete and the downstream handlers should be bypassed and they have access to the full request through the routing context. The handlers that are called for a particular request is defined by the Vert.x `Router` which maps patterns to particular handlers. To implement a security handler we add a handler that matches all paths, i.e. the root path (`/`) and ensure that it appears in the router at the top of the list of patterns. When the handler is called, it creates a generic security context and attaches the security context to the routing context making it accessible to downstream handlers.
```
ppatierno, line 61: talking about security, how you envisage authn and authz of users for calling REST or GraphQL API?

gmcrobert:
@ppatierno I have changed the file to have each sentence on a separate line. I will elaborate on the security stuff after I have put some more thought into it. Feel free to add some suggestions.

006-admin-server.md
```diff
@@ -0,0 +1,116 @@
...
+* provide a consolidated backend API for additional interfaces like a browser based UI or a CLI.
+* be capable of supporting both REST and GraphQL interfaces.
+* allow for more sophisticated APIs to be built on top of the existing APIs like the kubernetes API providing value-add to existing content.
+
+## Proposal
+
+### The server
+The proposal is to make a modular server using the Vert.x-Web toolkit which has strong support for creating both REST and GraphQL interfaces.
+The server contains a single Vert.x verticle which listens for inbound traffic on 1 or more configurable ports.
+Allowing multiple configurable ports will allow the server to expose different functionality on each port, the simplest example of this being exposing an HTTP API on one port and a separate HTTPS API on a different port.
```
samuel-hawker, line 17: I assume different functionality includes varying methods of authn and authz too? i.e. a mechanism to set different authed endpoints, one scram, one TLS etc?

Or is this handled by the security context piece in the **Security** section?

scholzj, line 17: Maybe you can add some examples for clarification what exactly does this mean and what could be the use-case?

006-admin-server.md
```diff
@@ -0,0 +1,116 @@
...
+
+An API server will:
+* provide a consolidated backend API for additional interfaces like a browser based UI or a CLI.
+* be capable of supporting both REST and GraphQL interfaces.
+* allow for more sophisticated APIs to be built on top of the existing APIs like the kubernetes API providing value-add to existing content.
+
+## Proposal
+
+### The server
+The proposal is to make a modular server using the Vert.x-Web toolkit which has strong support for creating both REST and GraphQL interfaces.
```
scholzj, line 15: What is the value and the cost of supporting both REST and GraphQL? Not just for development, but also QE, maintenance etc.

dlabaj (approved):
lgtm

006-admin-server.md
gmcrobert: I have rewritten this section to clarify the uses and have specifically cited the use of multiple authentication schemes as the main driver for needing to do this.

gmcrobert: I have provided the example of supporting multiple authentication schemes to demonstrate the use case. The idea behind this is to provide something similar to the `advertised listeners` in Kafka.

006-admin-server.md
gmcrobert: I do not think it will add a huge amount of overhead supporting both. Both APIs will be retrieving similar data but in a slightly different way. This allows the collection and aggregation of the information to be done in a way that is common to both and the difference then becomes a small piece on top that formats the results slightly differently. Both apis still use a REST endpoint with the main difference being that the REST api has multiple resource paths and each path represents a resource whereas GraphQL has a single resource path and the different resources to be returned are defined in the payload so similar tools can be used to test both APIs. Both APIs will also be implemented using the Vert.x libraries so keeping the libraries in-sync should be straightforward.

scholzj (approved):
Thanks for the proposal

scholzj:
I added it to the agenda for tomorrows community meeting to push others to review as well ...

011-admin-server.md
```diff
@@ -0,0 +1,139 @@
...
+The server contains a single Vert.x verticle which listens for inbound traffic on 1 or more configurable ports.
+Allowing multiple configurable ports will allow the server to expose different functionality on each port.
+A common use will be to separate traffic using different authentication protocols so, one port can listen for SCRAM users, and a different port can be used for OAuth users on a single admin server.
+All the APIs that are supported on the server will then be available on both of the two ports. 
+The server module itself does not contain any API implementations but defines a service interface using the Java SPI and will load modules that implement that interface using the Java Service Loader.
+The implementation of the APIs that the server then exposes can be created independently and will be automatically included in the running server if they are available to the Service Loader.
+
+The reason for the modular approach is that it allows the strimzi project to create a base set of functionality that can easily be extended by a downstream project.
```
tombentley, line 23: While I don't have a problem with this in principal, I think we need to ensure that the URL structure of the API reflects the identity of the service and also its version. This is to prevent two different plugins (e.g. from different vendors) using the same name and also to provide API evolution between different versions of the same plugin (it's not clear to me whether a service could provide more than one API version).

011-admin-server.md
```diff
@@ -0,0 +1,139 @@
...
+Not only is it easy to extend the API but it can be done in isolation.
+This removes the requirement of the extensions needing to track the strimzi source repositories.
+For example, lets assume that strimzi only implements a Kafka administration API which provides a set of APIs based around the Kafka Admin Client.
```
tombentley, line 26:
```suggestion
For example, let's assume that Strimzi only implements a Kafka administration API which provides a set of APIs based around the Kafka Admin Client.
```

011-admin-server.md
```diff
@@ -0,0 +1,139 @@
...
+A downstream project then would like to add a Schema Registry to the cluster and needs to add an API in support of the new functionality.
+With the modular server, they can define the API and its implementation, bundle the new functionality into a JAR file and add the new JAR file to the classpath.
```
tombentley, line 28: Would this really be classpath-based, or use java modules?

011-admin-server.md
```diff
@@ -0,0 +1,139 @@
...
+A secondary benefit of a modular approach is that it allows the strimzi user to configure the server to only contain the functionality that they would like to use providing for a leaner server.
```
tombentley, line 29:
```suggestion
A secondary benefit of a modular approach is that it allows the strimzi user to configure the server to only contain the functionality that they would like to use, providing for a leaner server.
```

011-admin-server.md
```diff
@@ -0,0 +1,139 @@
...
+It leaves the implementation of the operations to focus entirely on the business logic.
+A secondary advantage of using the OpenAPI3 specification is the numerous tools that are freely available to create generated documentation of the API.
+ 
+The Vert.x OpenAPI3 support will read an OpenAPI3 specification YAML file and allow the developer to map the original operations to handler methods.
+It creates a Vert.x Router which can be mounted on a Vert.x HTTP/S server.
+The module would implement the service interface defined by the server which will contain a method that the server can call to load the OpenAPI3 router and mount the router as a subrouter at a specified mount point to the root path.
+So, for example, if the OpenAPI3 spec defined endpoints `GET /topics` and `POST /topics` and these were mounted on the root path at the mount point `/admin` then the server would listen for incoming requests of `GET /admin/topics` and `POST /admin/topics` and route them to the handlers defined in the module.
```
tombentley, line 43: So with reference to my previous comment, I'm not sure if any plugin should be bindable to the root path. The root path could be used for discovery of the API provided.

tombentley (commented):
My main comment here is about defining the URL structure of the API with respect to plugins from multiple vendors and supporting multiple API versions.

tombentley:
Oh, one other thing: API should be spelt in upper case. "REST api" is a little jarring.

ppatierno (approved):

samuel-hawker (approved):

Frawless (approved):

011-admin-server.md
gmcrobert: @tombentley is it sufficient that the mount point cannot be empty (the root) and must be unique or were there other restrictions you had in mind? The versioning is a little more challenging because the world has moved away from versioning being present on the URI. A common way to version is to use the media type so, e.g. you would have `application/vnd.strimzi.v1+json`, `application/vnd.strimzi.v2+json`, etc and the type without versioning, i.e. `application/json` would always refer to the latest version. Do we want a registration object which allows an added service to specify its mount point, version and anything else that would be useful, e.g. a vendor identifier, but not include the version in the URI? This would of course imply that you couldn't have more than 1 version of the API loaded which makes sense as a true REST API has the same limitation. Handling different version is then the responsibility of the implementor of the API.

011-admin-server.md
gmcrobert: I will correct my colonial english ;-)

011-admin-server.md
gmcrobert: When I originally wrote this, Strimzi had only just made the switch from Java 8 to Java 11. The update I did the other day I was trying to be agnostic of the two versions because I wasn't sure whether the we were ready to exploit Java 11 as a community but I obviously failed. My preference is definitely use the Java 11 modules in which case this statement will need to be updated. If you agree that it should be Java 11 modules then I will update the document to be specific and correct.

011-admin-server.md
gmcrobert: So if the response in the previous comment is sufficient.

011-admin-server.md
tombentley: I don't have a strong preference for including the version in the URL and TBH I wasn't aware of a trend to use media types instead. That said, I think there is _some_ value, for people how know Kubernetes' REST API, in doing something similar with api groups at least. So if it were _me_, I would be envisaging 

* `GET /` would return some description (presumably openapi) of the available endpoints
* The common, strimzi provided plugins might live under `/strimzi.io/...` or even `/core.strimzi.io/...`, to allow for other Strimzi api groups in the future.
* Plugins provided by some vendor might live under `/example.com/...`

scholzj:
@tombentley Is this now OK for you?

012-admin-server.md
```diff
@@ -0,0 +1,139 @@
...
+The schema and wiring are then merged to give a single consolidated GraphQL API.
+This would allow a module to easily create both a REST interface and a GraphQL interface for the module and share a lot of the implementation between the two. 
+
+### Security
+Vert.x networking is based on Netty and works in a very similar manner.
+When a request is received by the server, it is funneled through a pipe and that pipe has a number of interceptors or handlers that can inspect the contents at that point and ignore or modify it before allowing it to proceed through the pipe.
+Vert.x also creates a routing context that travels through the pipe with the request and is available for inspection and/or modification by all the handlers.
+A handler can decided whether the request should be passed on to the next downstream handler or the request is complete and the downstream handlers should be bypassed and they have access to the full request through the routing context.
```
tombentley, line 98:
```suggestion
A handler can decide whether the request should be passed on to the next downstream handler or the request is complete and the downstream handlers should be bypassed and they have access to the full request through the routing context.
```

tombentley (approved):

scholzj:
Thanks for the proposal.

PR #10, opened 2020-07-15
Add proposal to TLS encrypt the Kafka Connect REST API

ajborley:
This proposal reduces access to the Kafka Connect REST API by enabling
TLS encryption on the Kafka Connect REST listener when the connectors
running on the Kafka Connect cluster are managed by the
`KafkaConnector` or `KafkaMirrroMaker2` operators.


006-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,98 @@
...
+#### Annotation alternative to enable a plain HTTP REST API listener
+
+Alternatively, a new annotation `strimzi.io/enable-plain-rest-listener` could be added, to avoid changing the spec.
+Setting the `strimzi.io/enable-plain-rest-listener` annotation value to `true` would add the additional unencrypted HTTP listener on port 8083.
+
+
+### Certificate renewals
+
+The self-signed TLS certificate will have a 1-year expiration.
+One day before the self-signed TLS certificate expires, the operator will automatically renew the certificate, replace the old self-signed certificate, and restart the pods in a controlled manner to ensure the new certificates are in use by Kafka Connect.
```
samuel-hawker, line 93: I assume this is done on the periodic reconcile? Or a reconcile triggered due to the certificate renewal period?

ajborley: Yes, on the periodic reconcile. We could also add in [maintenance time windows](https://strimzi.io/docs/operators/latest/using.html#assembly-maintenance-time-windows-deployment-configuration-kafka), and/or support the `strimzi.io/force-renew` and `strimzi.io/force-replace` annotations to allow users to force new certs and keys to be created. What do you think?

Looking through the Kafka operator code I see that the default renewal period is 30 days before the cert expiry (which has a default of 365 days) - I will update the proposal to change 'One day before..' to '30 days before...'  so that the default periods match. Do you think these certificates need similar fine-grained control that users get in the Kafka `spec.clusterCa` fields?

006-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,98 @@
...
+
+There is currently no way to TLS encrypt the Kafka Connect REST API, leaving the Connect REST listener as perhaps the last key endpoint that cannot currently be restricted using Strimzi.
+
+## Proposal
+
+This proposal changes the behaviour of the `KafkaMirrorMaker2` operator and the `KafkaConnect` operator.
+In the `KafkaConnect` operator case, the behaviour is only changed when the `strimzi.io/use-connector-resources: "true"` annotation is applied in the `KafkaConnect` CR (meaning that all connectors running on the Connect cluster are managed by the `KafkaConnector` operator).
+
+The default behaviour of the `KafkaMirrorMaker2` operator and the `KafkaConnect` operator will be to deploy Kafka Connect clusters with a single TLS encrypted REST API listener.
+A self-signed TLS certificate and key required to enable TLS encryption on the REST API listener are generated by the operator, stored in secrets for use by the operator clients and the Kafka Connect pods.
```
ppatierno, line 23: Why is the TLS certificate self-signed? Can't we use the cluster CA (which is self-signed cluster root certificate) to sign the TLS certificate for each Kafka Connect worker node as we do for each Kafka broker?
In this way, the client (so the operator calling the REST API) che verify the TLS certificate using the cluster CA certificate.

006-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,98 @@
...
+listeners: https://:8443
+rest.advertised.listener: https
+rest.advertised.port: 8443
+listeners.https.ssl.client.auth: none
+listeners.https.ssl.truststore.location: /tmp/kafka/kafka-connect-rest.truststore.p12
+listeners.https.ssl.truststore.password: ***generated password***
+listeners.https.ssl.truststore.type: PKCS12
+listeners.https.ssl.keystore.location: /tmp/kafka/kafka-connect-rest.keystore.p12
+listeners.https.ssl.keystore.password: ***generated password***
+listeners.https.ssl.keystore.type: PKCS12
```
ppatierno, line 36: So it's clear to me about the keystone which contains the TLS certificate and key related to the Kafka Connect worker node for enabling TLS encryption with the client (operator).
It's not clear what's inside the truststore and why it's useful. If client auth would be enabled (which is not in this case) it would contain the CA certificate used to sign the clients certificates in order to validate their identity, but what should contain in this case?

006-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+To communicate with the encrypted Connect REST API listener, the `KafkaMirrorMaker2`, `KafkaConnect` and `KafkaConnector` operators require modifications to the `AbstractConnectOperator` to apply the TLS certs as keystore credentials to a WebClient.
```
ppatierno, line 39: shouldn't it be just a truststore (not keystore) containing the CA cert to verify the TLS server cert got from the Kafka Connect listener on connection?

006-tls-encrypt-the-kafka-connect-rest-api.md
ajborley: The instance of Kafka Connect that the operator is calling may not be connected to a Strimzi Kafka - it can be any Kafka instance, so there might not be a cluster CA available to use. 

We may be able to use the cluster CA if it is available (in the same namespace, etc) or we could generate a Connect-specific CA for this proposal - do you think that generating signed TLS certificates from a CA cert would be better than generating self-signed certificates directly?

006-tls-encrypt-the-kafka-connect-rest-api.md
ajborley: You're right, I think we only need the keystore in this example - I have updated in the latest commit.

006-tls-encrypt-the-kafka-connect-rest-api.md
ajborley: Yep, that's right. I got the terminology mixed up - have updated in the latest commit.

006-tls-encrypt-the-kafka-connect-rest-api.md
ppatierno: Tbh I see a self-signed certificate less secure. Self-signed certs are used for well-recognized CAs which then signs other certificates (or it could be a trusted internal CA in your organization). Are we sure that we want to support an external Kafka Connect not deployed by Strimzi? I thought that this PR was about securing the Kafka Connect instance deployed by Strimzi.

006-tls-encrypt-the-kafka-connect-rest-api.md
samuel-hawker: I'd like to hear the other maintainers opinions on this, but this sounds reasonable to me.

006-tls-encrypt-the-kafka-connect-rest-api.md
ajborley: @ppatierno - Sorry I wasn't clear - I meant an external (non-Strimzi) Kafka cluster that a `KafkaConnect` CR is configured to use. This proposal is about the REST API communication between the operator and a Strimzi-managed Kafka Connect. 

As the cluster CA is only created when you are using a Kafka cluster created via a `Kafka` CR, the cluster CA may not be available for use by the Strimzi-managed Kafka Connect instance. 

We could generate a connect-cluster-CA for sole use by the Kafka Connect instance, and generate certificates using that - would that be your preference here?

scholzj: I agree that the Kafka Cluster CA cannot be used as a basis here because Connect is not bound to any particular Kafka CR.

I can see the concerns about the self-signed term. But the Kafka Cluster CA is also self-signed really. So I'm not sure I can see it make any difference here.

006-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,95 @@
...
+The certificates and keys are mounted in each Kafka Connect pod, which create an SSL keystore from the certificate and key and set the following properties in the generated connect configuration to enable an HTTPS listener:
+
+```
+listeners: https://:8443
+rest.advertised.listener: https
+rest.advertised.port: 8443
+listeners.https.ssl.client.auth: none
+listeners.https.ssl.keystore.location: /tmp/kafka/kafka-connect-rest.keystore.p12
+listeners.https.ssl.keystore.password: ***generated password***
+listeners.https.ssl.keystore.type: PKCS12
```
scholzj, line 33: IIRC, when Connect is running with multiple replicas, it is using the REST interface for communication between workers and their leader node (basically, requests to create connector etc. are forwarded tot he leader). So, since using self-signed certificate, wouldn't you need to also specify a truststore with the public key to allow the workes to trust the leader when forwarding the request?

006-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,95 @@
...
+#### Annotation alternative to enable a plain HTTP REST API listener
+
+Alternatively, a new annotation `strimzi.io/enable-plain-rest-listener` could be added, to avoid changing the spec.
+Setting the `strimzi.io/enable-plain-rest-listener` annotation value to `true` would add the additional unencrypted HTTP listener on port 8083.
+
+
+### Certificate renewals
+
+The self-signed TLS certificate will have a 1-year expiration.
+Thirty day before the self-signed TLS certificate expires, the operator will automatically renew the certificate, replace the old self-signed certificate, and restart the pods in a controlled manner to ensure the new certificates are in use by Kafka Connect.
```
scholzj, line 90: This would need to be a multi-phase process to handle the trust. Not sure if it is important to elaborate on it here, but you will probably need to:

1) Generate new certificate
2) Distribute the public key to all pods and CO truststores
3) Replace the key and roll all Connect pods
4) Remove the old public key from the truststores and roll all Connect pods

006-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,95 @@
...
+                enum:
+                - http
+                description: The protocol of the REST listener.
+            required:
+            - protocol
+          description: List of additional REST listeners.
+        ...
+```
+
+#### Annotation alternative to enable a plain HTTP REST API listener
```
scholzj, line 81: So, is this a rejected alternative? If not, how / when are we going to decide on one or the other?

006-tls-encrypt-the-kafka-connect-rest-api.md
ppatierno: The truststore was there before but I didn't know about this internal communication in a Kafka Connect cluster. If it's the case then we need the truststore. Sorry @ajborley

006-tls-encrypt-the-kafka-connect-rest-api.md
ppatierno: Ok but should at this point allow users to specify their own Kafka Connect CA as we do for the Kafka Cluster CA?

scholzj: If we don't use a CA we cannot let users provide custom CA - that would be only server certificate I guess. Me personally, I'm fine to wait with this if there will be any interest. The Connect API is different from the Kafka listeners so the demand might be different as well.

ppatierno: I was misleading here. I thought the idea was to generate a self-signed Kafka Connect CA used for signing the worker nodes certificates but actually it's just a self-signed certificate (not a CA).

006-tls-encrypt-the-kafka-connect-rest-api.md
ajborley: No worries at all - I had forgotten about the distributed case. I have updated to add the truststore settings back in.

006-tls-encrypt-the-kafka-connect-rest-api.md
ajborley: Thanks for this - I have added these steps to the proposal to assist in the implementation.

006-tls-encrypt-the-kafka-connect-rest-api.md
ajborley: I wasn't sure how folks would feel about the addition to the `spec`, so I added this as a possible alternative. Given that there haven't been objections (so far) to the `spec` addition, I have moved this annotation to the **Rejected Alternatives** section.

006-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Certificate renewals
+
+The self-signed TLS certificate will have a 1-year expiration.
+Thirty day before the self-signed TLS certificate expires, the operator will automatically renew the certificate, replace the old self-signed certificate, and restart the pods in a controlled manner to ensure the new certificates are in use by Kafka Connect.
+
+This would need to be a multi-phase process consisting of the following steps:
+
+1. Generate new certificate
+2. Distribute the public key to all pods and cluster operator truststores
```
tombentley, line 93: Let's use the term certificate rather than public key, since a cert contains more information than just a public key.
```suggestion
2. Distribute the new certificate to all pods and cluster operator truststores
```

006-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,110 @@
...
+            - protocol
+          description: List of additional REST listeners.
+        ...
+```
+
+
+### Certificate renewals
+
+The self-signed TLS certificate will have a 1-year expiration.
+Thirty day before the self-signed TLS certificate expires, the operator will automatically renew the certificate, replace the old self-signed certificate, and restart the pods in a controlled manner to ensure the new certificates are in use by Kafka Connect.
```
tombentley, line 88:
```suggestion
Thirty days before the self-signed TLS certificate expires, the operator will automatically renew the certificate, replace the old self-signed certificate, and restart the pods in a controlled manner to ensure the new certificates are in use by Kafka Connect without losing availability.
```

006-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,110 @@
...
+
+This would need to be a multi-phase process consisting of the following steps:
+
+1. Generate new certificate
+2. Distribute the public key to all pods and cluster operator truststores
+3. Replace the key and roll all Connect pods
+4. Remove the old public key from the truststores and roll all Connect pods
```
tombentley, line 95: Presumably this happens at time (or shortly before) the old cert expires? Best to say so.

006-tls-encrypt-the-kafka-connect-rest-api.md
ajborley: Updated

006-tls-encrypt-the-kafka-connect-rest-api.md
ajborley: Updated

006-tls-encrypt-the-kafka-connect-rest-api.md
ajborley: Updated - thanks for the review :)

scholzj:
Thanks for the proposal!

PR #13, opened 2020-08-10
Security design to support the UI and Admin Server

gmcrobert:
Signed-off-by: Graeme McRobert <mcrobert@uk.ibm.com>


009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+In these environments authentication and authorization can again be delegated to the backend system.
+A discussion and agreement is needed whether we wish to implement authentication and authorization on the Admin Server itself.
+
+Proceeding with the delegated model, I have 3 sets of diagrams showing a SCRAM-SHA-512 solution, an OAuth2 solution and a Mutual TLS solution with the last one only being relevant to the Admin Server.
+
+## SCRAM-SHA-512
+![Kafka Client using SCRAM-SHA512](./images/scram-sha512-client.png)
+
+Strimzi supports SCRAM-SHA512 using native Kafka facilities, so the Kafka client is setup with credentials for the user and these credentials are passed to the broker where the broker authenticates against user credentials stored in Zookeeper.
+When the request is authenticated, the Principal is passed to a Kafka Simple ACL Authorizer which determines whether the user is authorized to access the resource.
```
scholzj, line 28: I think this should work in general as well, or? Regardless of the authorization type.

```suggestion
When the request is authenticated, the Principal is passed to a Kafka Authorizer which determines whether the user is authorized to access the resource.
```

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+Access is defined through ACLs and are stored on Zookeeper.
+
+![Strimzi UI using SCRAM-SHA512](./images/scram-sha512-ui.png)
+
+This Kafka client model translates to a UI and Admin Server in a similar way.
+When the user loads up the UI in a browser, they are presented with a dialogue box requiring them to enter a userId and password.
+The userId and password they enter are pre-defined to Kafka by the creation of a Kafka user through the `User Operator` or the Kafka `bin/kafka-configs.sh ` command line tool.
```
scholzj, line 35:
```suggestion
The userId and password they enter are pre-defined to Kafka by the creation of a Kafka user through the `User Operator` or the Kafka `bin/kafka-configs.sh` command line tool.
```

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+Once the userId and password have been entered, the UI validates the credentials on Zookeeper using `passport.js`.
```
scholzj, line 36: The UI will not have access to Zookeeper. Also, Zookeeper is going away soon. So I think we need to have some other solution here.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+The result that is returned will be marked as valid or invalid.
+If the credentials are valid, they are stored in the session data associated with that browser session.
+When the UI requires data from the Admin Server, the credentials are recovered from the session data and placed as an auth header on the request.
+The Admin Server receives the request and configures the Kafka client in preparation for calling the Kafka broker.
+The SCRAM credentials are retrieved from the auth header and are added to the Kafka client configuration properties before executing the call to the Kafka broker.
+From thereon, the Kafka authentication and authorization mechanism works in an identical way to the description above for the Kafka Clients.
```
scholzj, line 42: I assume this will have to pass the password itself and not the scrambled credentials, right? Maybe we should make that clear for the record.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+
+This mechanism should be straightforward to setup using a `passport.js` strategy.
+The Strimzi UI is served from an Express server and that Express server will be inside the Strimzi cluster allowing the Express Server to talk to the Zookeeper on the secure port 2181.
```
scholzj, line 45: Ok, if this is done from the server, it can have Zookeeper access if deployed as part of the Kafka cluster. But it is still not clear to me how will you validate the scrambled credentials stored in Zookeeper. Also as I mentioned above, the timing is a bit unfortunate with Zookeeper being there today but gone probably in few months. So in the future this would need to talk with Kafka.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+![Kafka Client using Oauth2](./images/oauth-client.png)
+
+In the OAuth 2 scenario, Strimzi currently requires the user to configure the Kafka client with a clientId and secret.
+The client will then obtain an access token from the identity provider which is KeyCloak and that token flows to the broker.
+In the broker, calls to KeyCloak will be made based on the type of token that was requested by the client and passed to the broker.
+
+![Strimzi UI using Oauth2](./images/oauth-ui.png)
+
+In the UI model, the user will be presented with a login dialogue box when the UI is loaded into the browser and requires the user to enter a userId and password.
+The userId needs to be pre-defined to KeyCloak through the KeyCloak user interface.
```
scholzj, line 62: I assume this would work with. generic OAuth 2 server as well?

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+When the user enters their userId and password, the UI will contact KeyCloak to authorise the user and obtain an access token.
+This process will be controlled by `passport.js`.
+KeyCloak will either return an access token or reject the request.
+If an access token is returned, it will be stored in the session data for the session.
+When the UI requires data from the Admin Server, the token is recovered from the session data and placed as an auth header on the request.
+The Admin Server receives the request and configures the Kafka client in preparation for calling the Kafka broker.
+The token is retrieved from the auth header and is added to the Kafka client JAAS configuration properties before executing the call to the Kafka broker.
+The Kafka broker, which is running a custom SASL plugin, retrieves the token and sets up the Kafka Principal for the user.
+The request then passes through the KeyCloak Authorizer for authorization.
```
scholzj, line 71: This is just one of the options. OAuth can still work with simple authZ or with OPA as well.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+![Strimzi UI using Oauth2](./images/oauth-ui.png)
+
+In the UI model, the user will be presented with a login dialogue box when the UI is loaded into the browser and requires the user to enter a userId and password.
+The userId needs to be pre-defined to KeyCloak through the KeyCloak user interface.
+When the user enters their userId and password, the UI will contact KeyCloak to authorise the user and obtain an access token.
+This process will be controlled by `passport.js`.
+KeyCloak will either return an access token or reject the request.
+If an access token is returned, it will be stored in the session data for the session.
+When the UI requires data from the Admin Server, the token is recovered from the session data and placed as an auth header on the request.
+The Admin Server receives the request and configures the Kafka client in preparation for calling the Kafka broker.
```
scholzj, line 68: Will the admin server keep some long held connections? @mstruk can provide more details. But I think there are so re-authentications happening during the lifetime of the connection. So if the API server keep sany long term connections open, it might need refresh token or something to be able to handle them. If they are short connections only ... e.g. connect, list topics, disconnect ... it should be fine I guess.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+The behaviour when the AdminServer received the request would be identical to the flow for the UI calling the Admin Server with a token.
+
+## Mutual TLS
+
+![Admin server using Mutual TLS](./images/mtls-admin.png)
+
+With Mutual TLS, authentication takes place in the TLS handshake between the external web client and the Admin Server.
+The server side can only detect a userId for the client so credentials cannot be propagated to the Kafka Broker that woud allow the broker to authenticate and authorize the request.
+This proposal recommends that the Kafka clients running under the Admin Server can use a valid Kafka User for the connection between the Kafka client and the broker and pass the external web client's userId to the broker login module.
+The Kafka broker then uses a SASL plugin to create a custom authentication module which sets up the Kafka principal which flows through to authorizer.
```
scholzj, line 83: I think this would be quite a lot of additional work and would affect also Kafka connections etc. So I wonder if any alternatives would be better solutions:
* Pass the cert/key from the UI in the login (via a form)
* Not support mutual TLS for the UI / API server at all

009-ui-and-admin-server-security.md
ppatierno: I don't think that it makes sense to set up a ZooKeeper based mechanism right now when things are going to change in the future, that at this point should not be so far.
Maybe the UI can go through the admin server for authenticating the user, setting up a Kafka client for doing a "generic" request (i.e. get Kafka cluster-info) leveraging this request to check if authentication failed or not.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+When the UI requires data from the Admin Server, the credentials are recovered from the session data and placed as an auth header on the request.
+The Admin Server receives the request and configures the Kafka client in preparation for calling the Kafka broker.
+The SCRAM credentials are retrieved from the auth header and are added to the Kafka client configuration properties before executing the call to the Kafka broker.
+From thereon, the Kafka authentication and authorization mechanism works in an identical way to the description above for the Kafka Clients.
+
+This mechanism should be straightforward to setup using a `passport.js` strategy.
+The Strimzi UI is served from an Express server and that Express server will be inside the Strimzi cluster allowing the Express Server to talk to the Zookeeper on the secure port 2181.
+The login would not go through the Admin Server which hides the process from the external network but it may be desirable to add a `login` endpoint to the Admin Server which would allow session based applications to use the Admin Server in a natural way.
```
ppatierno, line 46: maybe it means what I describe above somehow?

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
+# Security Extension
+
+This proposal is supplementary to the security documentation in the [Using Strimzi](https://strimzi.io/docs/operators/latest/using.html#security-str) guide.
+
+## Problem description
+Strimzi currently supports SCRAM-SHA-512, Oauth2 and Mutual TLS for authentication of clients connecting to the Kafka brokers.
```
mstruk, line 6:
```suggestion
Strimzi currently supports SCRAM-SHA-512, OAuth2 and Mutual TLS for authentication of clients connecting to the Kafka brokers.
```

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+The Strimzi cluster operator can be configured to support all 3 mechanisms for authentication for clients inside or outside the Kubernetes cluster.
+The SCRAM-SHA and Mutual TLS support uses the native support in Kafka for authentication and for Oauth2, authentication is delegated to [keyCloak](https://www.keycloak.org/).
```
mstruk, line 8: Correct. I'd suggest:

```suggestion
The SCRAM-SHA and Mutual TLS support use the native authentication support in Kafka. The OAuth2 uses Kafka native SASL OAUTHBEARER support, extending it with [Strimzi Kafka OAuth](https://github.com/strimzi/strimzi-kafka-oauth) library which in principle works with any OAuth2 / OpenID Connect compatible authorization server. 
```

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
+# Security Extension
+
+This proposal is supplementary to the security documentation in the [Using Strimzi](https://strimzi.io/docs/operators/latest/using.html#security-str) guide.
+
+## Problem description
+Strimzi currently supports SCRAM-SHA-512, Oauth2 and Mutual TLS for authentication of clients connecting to the Kafka brokers.
+The Strimzi cluster operator can be configured to support all 3 mechanisms for authentication for clients inside or outside the Kubernetes cluster.
+The SCRAM-SHA and Mutual TLS support uses the native support in Kafka for authentication and for Oauth2, authentication is delegated to [keyCloak](https://www.keycloak.org/). 
+
+Authorization is available through native Kafka ACL support with the use of the Kafka `simpleACLAuthorizer`.
```
mstruk, line 10:
```suggestion
Authorization is available through native Kafka ACL support with the use of the Kafka `SimpleACLAuthorizer`.
```

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+Alternatives are available with Kafka authorizers being supplied for [keyCloak](https://www.keycloak.org/) and [Open Policy](https://www.openpolicyagent.org/docs/latest/).
```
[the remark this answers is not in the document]
mstruk: Please use: 'Keycloak' rather than 'keyCloak'

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+
+The configuration of Oauth2 is described in the [Strimzi documentation](https://strimzi.io/docs/operators/latest/using.html#assembly-oauth-authentication_str) as well as in the [strimzi-kafka-oauth](https://github.com/strimzi/strimzi-kafka-oauth) repository on github. SCRAM and Mutual TLS configuration is documented in the relevant places throughout the [Strimzi documentation](https://strimzi.io/documentation/).
```
mstruk, line 13:
```suggestion
The configuration of OAuth2 is described in the [Strimzi documentation](https://strimzi.io/docs/operators/latest/using.html#assembly-oauth-authentication_str) as well as in the [strimzi-kafka-oauth](https://github.com/strimzi/strimzi-kafka-oauth) repository on github. SCRAM and Mutual TLS configuration is documented in the relevant places throughout the [Strimzi documentation](https://strimzi.io/documentation/).
```

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+
+![Strimzi UI using SCRAM-SHA512](./images/scram-sha512-ui.png)
+
+This Kafka client model translates to a UI and Admin Server in a similar way.
+When the user loads up the UI in a browser, they are presented with a dialogue box requiring them to enter a userId and password.
+The userId and password they enter are pre-defined to Kafka by the creation of a Kafka user through the `User Operator` or the Kafka `bin/kafka-configs.sh ` command line tool.
+Once the userId and password have been entered, the UI validates the credentials on Zookeeper using `passport.js`.
+The result that is returned will be marked as valid or invalid.
+If the credentials are valid, they are stored in the session data associated with that browser session.
+When the UI requires data from the Admin Server, the credentials are recovered from the session data and placed as an auth header on the request.
```
ppatierno, line 39: in which "form" are they set in the auth header? AFAIK in general it's just "user:password" base64 encoded ... should we have TLS encryption between Strimzi UI and admin server?

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+The Strimzi Admin Server API could be called directly by the user creating an auth header on the request.
+Like the UI flow, the credentials would be retrieved from the auth header and placed into the Kafka client properties as SCRAM credentials.
+
+## OAuth 2
+
+![Kafka Client using Oauth2](./images/oauth-client.png)
+
+In the OAuth 2 scenario, Strimzi currently requires the user to configure the Kafka client with a clientId and secret.
+The client will then obtain an access token from the identity provider which is KeyCloak and that token flows to the broker.
+In the broker, calls to KeyCloak will be made based on the type of token that was requested by the client and passed to the broker.
```
mstruk, line 57: The interaction of broker with authorization server is in fact configured ahead of time (whether JWKS keys are used for validation or introspection endpoint is used) and not dependent on how the client obtains an access token.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+The login would not go through the Admin Server which hides the process from the external network but it may be desirable to add a `login` endpoint to the Admin Server which would allow session based applications to use the Admin Server in a natural way.
+
+The Strimzi Admin Server API could be called directly by the user creating an auth header on the request.
+Like the UI flow, the credentials would be retrieved from the auth header and placed into the Kafka client properties as SCRAM credentials.
+
+## OAuth 2
+
+![Kafka Client using Oauth2](./images/oauth-client.png)
+
+In the OAuth 2 scenario, Strimzi currently requires the user to configure the Kafka client with a clientId and secret.
```
mstruk, line 55: There are clients managed by Strimzi, and there are any other clients deployed and managed by users or other automated mechanisms. There are many ways how the client can obtain the access token (clientId + secret, username + password, obtaining and saving a long-lived refresh token and using it to generate a new short-lived access token ...). Not all of these are provided for Strimzi managed clients or implemented in Strimzi Kafka OAuth client library. Kafka Broker only cares about access token fulfilling certain conditions - has a valid signature if it is a JWT token, was issued by expected issuer, has not yet expired ...

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+The client will then obtain an access token from the identity provider which is KeyCloak and that token flows to the broker.
```
mstruk, line 56:
```suggestion
The client will then obtain an access token from the OAuth2 authorization server and send that token to the broker during session authentication.
```

009-ui-and-admin-server-security.md
mstruk, line 57:
```suggestion
On the broker side the passed access token will be validated either by validating the JWT token signature using JWKS keys from configured authorization server JWKS endpoint, or by passing the token to the configured authorization server introspection endpoint.```

File: 009-ui-and-admin-server-security.md (hunk 6 above)
Comment (mstruk) [inline 009-ui-and-admin-server-security.md:62]:
```suggestion
The userId needs to be pre-defined in the identity provider used by the authorization server. In practice the actual identity provider may be another backend server such as an LDAP server used by the authorization server.
```

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+In the broker, calls to KeyCloak will be made based on the type of token that was requested by the client and passed to the broker.
+
+![Strimzi UI using Oauth2](./images/oauth-ui.png)
+
+In the UI model, the user will be presented with a login dialogue box when the UI is loaded into the browser and requires the user to enter a userId and password.
+The userId needs to be pre-defined to KeyCloak through the KeyCloak user interface.
+When the user enters their userId and password, the UI will contact KeyCloak to authorise the user and obtain an access token.
```
mstruk, line 63: How is 'UI will contact Keycloak' to be understood? There's UI as 'the server side app with REST API' and the UI as 'the browser JS app'. Authorization servers like Keycloak are used as Single Sign-On servers where user's browser is completely redirected to a central login page, and after successful login redirected back to web app, so that the applications never see the password, not even the browser JS app - that's the original reason OAuth was developed. None of the applications ever receives the user's password, they only receive the access token, which grants these apps limited access for a limited time, after which the apps redirect user to SSO server to login again. In practice the redirect is often invisible because there is a mechanism to issue a new access token without user interaction in certain circumstances (when using a refresh token).

mstruk, line 63:
```suggestion
When the user enters their userId and password, the UI will contact the authorization server to authenticate the user and obtain an access token.
```

009-ui-and-admin-server-security.md
mstruk: One can always start a new connection if the existing one closes due to a failed re-authentication.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+This process will be controlled by `passport.js`.
+KeyCloak will either return an access token or reject the request.
```
mstruk, line 65:
```suggestion
Authorization server will either return an access token or reject the request.
```

009-ui-and-admin-server-security.md
samuel-hawker, line 35:
```suggestion
The userId and password they enter are pre-defined to Kafka by the creation of a Kafka user through the `User Operator` and `KafkaUser` custom resource or the Kafka `bin/kafka-configs.sh ` command line tool.
```

009-ui-and-admin-server-security.md
samuel-hawker: I agree with Jakub here, is it possible to do this directly via Kafka?

009-ui-and-admin-server-security.md
tombentley: These comments suggest we're not yet all on the same page. In case it's not already obvious:

* KIP-500 will remove ZK from the picture entirely. If it's not there you can't go sneakily looking in there for SCRAM-SHA hashes.
* Under the KIP-500 umbrella, KIP-554 defines the API which will be available for setting the SCRAM-SHA credentials from an `Admin` client instance. In a ZK-less cluster the broker will perform the hashing and write the hashes to the metadata log, defined in [KIP-631](https://cwiki.apache.org/confluence/display/KAFKA/KIP-631%3A+The+Quorum-based+Kafka+Controller#KIP631:TheQuorumbasedKafkaController-RecordFormats). and [KIP-595](https://cwiki.apache.org/confluence/display/KAFKA/KIP-595%3A+A+Raft+Protocol+for+the+Metadata+Quorum). Last I look the metadata log. 595 contains this:

    > At a high level, this proposal treats the metadata log as a topic with a single partition. The benefit of this is that many of the existing APIs can be reused. For example, it allows users to use a consumer to read the contents of the metadata log for _debugging purposes_. 

    (my emphasis). So while it might be possible, post KIP-500 to obtain SCRAM-SHA credientials it would not really be considered best practice from a Kafka PoV. 
* Therefore, we should try really hard to avoid needing to access to the SCRAM-SHA info.
* @ppatierno as mentioned above, using the provided credentials to make an authenticated connection to a broker works in theory, but since you can't tell that the broker actually used the credentials it relies on the whole thing being configured correctly to actually work. In the past someone suggested a `whoAmI()` method on the `Admin` client which, I guess could provide information about how the user was authenticated as well as who they were.

009-ui-and-admin-server-security.md
tombentley: I think this needs elaboration. You can't assume the reader has the same understanding of what "session based applications ... use the Admin Server in a natural way" means as you do, so describe it. In particular, I assume you mean a session on the Admin Server? And maybe this related to my "cache client" question earlier, but TBH I'm not sure.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+The Admin Server receives the request and configures the Kafka client in preparation for calling the Kafka broker.
+The SCRAM credentials are retrieved from the auth header and are added to the Kafka client configuration properties before executing the call to the Kafka broker.
+From thereon, the Kafka authentication and authorization mechanism works in an identical way to the description above for the Kafka Clients.
+
+This mechanism should be straightforward to setup using a `passport.js` strategy.
+The Strimzi UI is served from an Express server and that Express server will be inside the Strimzi cluster allowing the Express Server to talk to the Zookeeper on the secure port 2181.
+The login would not go through the Admin Server which hides the process from the external network but it may be desirable to add a `login` endpoint to the Admin Server which would allow session based applications to use the Admin Server in a natural way.
+
+The Strimzi Admin Server API could be called directly by the user creating an auth header on the request.
```
tombentley, line 48: Is this some standard HTTP header? If so, which? If not, can you describe it in more detail?

009-ui-and-admin-server-security.md
tombentley: Yes, lets talk in terms of "OAuth 2 server" (or "Authorization Server" if that's the term which is used to refer to a server which does OAuth authentication), rather than Keycloak everywhere.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+In the broker, calls to KeyCloak will be made based on the type of token that was requested by the client and passed to the broker.
+
+![Strimzi UI using Oauth2](./images/oauth-ui.png)
+
+In the UI model, the user will be presented with a login dialogue box when the UI is loaded into the browser and requires the user to enter a userId and password.
+The userId needs to be pre-defined to KeyCloak through the KeyCloak user interface.
+When the user enters their userId and password, the UI will contact KeyCloak to authorise the user and obtain an access token.
+This process will be controlled by `passport.js`.
+KeyCloak will either return an access token or reject the request.
+If an access token is returned, it will be stored in the session data for the session.
```
tombentley, line 66: Since there's at least two HTTP servers going on here, let's be explicit about which session we're talking about.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+When the UI requires data from the Admin Server, the token is recovered from the session data and placed as an auth header on the request.
+The Admin Server receives the request and configures the Kafka client in preparation for calling the Kafka broker.
+The token is retrieved from the auth header and is added to the Kafka client JAAS configuration properties before executing the call to the Kafka broker.
+The Kafka broker, which is running a custom SASL plugin, retrieves the token and sets up the Kafka Principal for the user.
+The request then passes through the KeyCloak Authorizer for authorization.
+
+In order to call the Admin Server directly, the user would need to obtain a token from KeyCloak manually and then store that token in an auth header on the call to the Admin Server.
+The behaviour when the AdminServer received the request would be identical to the flow for the UI calling the Admin Server with a token.
```
tombentley, line 74: TBH I read this and I look at the diagram and I think it would be an awful lot simpler if the words referred to the diagram. Even better if the diagram was ascii art because then it would be inline in the text.

````
 OAuthServer  <---------------------------------------+
      ^                                               |
   1. |                                               | 4.
      |        2.                         3.          |
Strimzi UI  -------> Strimzi Admin --------------> Broker

1. UI gets access token as part of OAuth login
2. UI propgates access token via `Xyz` HTTP header
3. Strimzi Admin uses provided token to configure a client
4. Broker validates access token with OAuth server
```

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Mutual TLS
+
+![Admin server using Mutual TLS](./images/mtls-admin.png)
+
+With Mutual TLS, authentication takes place in the TLS handshake between the external web client and the Admin Server.
+The server side can only detect a userId for the client so credentials cannot be propagated to the Kafka Broker that woud allow the broker to authenticate and authorize the request.
```
tombentley, line 81:
```suggestion
The server side can only detect a userId for the client so credentials cannot be propagated to the Kafka Broker that would allow the broker to authenticate and authorize the request.
```

009-ui-and-admin-server-security.md
tombentley: So we're using some custom (EXTERNAL?) SASL mechansism and the broker-side doesn't really authenticate the identity asserted (it trusts it because of the TLS identity it's associated with), but builds a `KafkaPrincipal` using the asserted identity which is then used for authz?

009-ui-and-admin-server-security.md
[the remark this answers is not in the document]
gmcrobert: @scholzj so would it be better to change this to read something more like this: `and for OAuth2, Strimzi provides an [OAuth library](https://github.com/strimzi/strimzi-kafka-oauth) which provides a generic interface allowing you to connect to multiple different OAuth2 servers.`?

009-ui-and-admin-server-security.md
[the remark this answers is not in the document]
gmcrobert: I will correct the spelling and change the name to `Open Policy Agent`.

009-ui-and-admin-server-security.md
gmcrobert: I will remove the use of the specific `SimpleACLAuthorizer` and change this to a general `Authorizer` as suggested. I will also edit the diagram to match it and remove the line connecting the Authorizer to Zookeeper to accommodate for the fact that a general Authorizer may not use a store or use something other than Zookeeper as the store.

009-ui-and-admin-server-security.md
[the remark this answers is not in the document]
scholzj: Yeah, I think that would be perfect.

009-ui-and-admin-server-security.md
gmcrobert: So the summary above seems to be that we need to stay away from talking to Zookeeper. As we would want any UI we create to validate credentials that a user enters, we could potentially issue a call such as `describe cluster` to Kafka from an admin client and test whether it failed authentication. This assumes that the admin client will return an `AuthenticationException` if the credentials are invalid. I am not sure I quite understand why you would not know if the broker used the credentials in this case so I need some help in that. This mechanism seems very artificial but could potentially work. As the Admin Server would be stateless and therefore does not require a login process in order to establish a session, how limiting would it be to say that the UI only supports OAuth2 and SCRAM is restricted to the Admin Server?

scholzj: > So the summary above seems to be that we need to stay away from talking to Zookeeper. As we would want any UI we create to validate credentials that a user enters, we could potentially issue a call such as describe cluster to Kafka from an admin client and test whether it failed authentication. This assumes that the admin client will return an AuthenticationException if the credentials are invalid. I am not sure I quite understand why you would not know if the broker used the credentials in this case so I need some help in that. This mechanism seems very artificial but could potentially work.

Honestly, I'm not sure I see any other way. You could just not validate it at all and show the errors later inside the UI if the credentials don't work. There is no other way to check it. 

I assume that once ZooKeeper is removed, you would still be able to read the credentials somewhere from Kafka as you would today from ZooKeeper. But I think the logic for verifying the password will still be non-trivial. Plus it would not solve other mechanisms.

> how limiting would it be to say that the UI only supports OAuth2 and SCRAM is restricted to the Admin Server?

I think it is fine to decide that some mechanisms are not supported for this. The current TLS authentication has anyway its limitations in not supporting any revocation lists.

Do we want to also support some mode without any authentication? It is obviously not "production ready". But lot of people seem to start that way and it seems like it would be useful to have that mode as well.

gmcrobert: Sorry @scholzj I had always assumed we would have a `security off` mode so will add something about that to the doc. @tombentley , @ppatierno , @mstruk , @samuel-hawker can I get an opinion from you on whether we should drop SCRAM for the UI and support no-auth and OAuth2 only? We can continue to support SCRAM for the Admin Server.

ppatierno: Supporting OAuth2 only means that the user has to deploy an OAuth server so one more component for the UI. 
Maybe he doesn't want to do that because he is happy with using SCRAM.
About doing a "test" call (i.e. `describe cluster`) for SCRAM authentication, it sounds to me that we agree that it could work but @tombentley not. Can you elaborate a little bit more the why, Tom?

tombentley: On second thoughts, I think it probably does work.

1. You configure an `Admin` client with some (list of) addresses and some SCRAM credentials 
2. You `describeCluster()` and you get some result.
3. Have you proved that the credentials you provided were valid?

Only if you know that the client actually connected using SCRAM. I guess if you've set the `security.protocol=SASL_*` and `sasl.mechanism` then that's probably a valid assumption, but I think it's worth having a test for.

009-ui-and-admin-server-security.md
gmcrobert: I have removed all references to Zookeeper and rewritten the proposal in a more general way.

009-ui-and-admin-server-security.md
gmcrobert: I have rewritten the details using OAuth2 as I have a better understanding of how the pieces fit together and the Strimzi OAuth library.

009-ui-and-admin-server-security.md
gmcrobert: I removed the mTLS section altogether because it complicates the picture and, for the first implementation, I think it is sufficient to support no security, SCRAM-SHA and Oauth2. Mutual TLS can be revisited in the future if there is demand for it.

009-ui-and-admin-server-security.md
gmcrobert: I believe I have fixed all uses of the term OAuth2.

009-ui-and-admin-server-security.md
gmcrobert: I have removed all references to the SimpleAclAuthorizer.

009-ui-and-admin-server-security.md
gmcrobert: I have corrected all spellings of Keycloak and OAuth2.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,107 @@
...
+        certKey: "ca.p12"
+
+``` 
+
+#### Proposed libraries
+- https://www.npmjs.com/package/passport - authentication middleware
+- https://www.npmjs.com/package/openid-client - Passport compatible OIDC client for communicating with an OP
+
+#### Outstanding questions
+1. Who is responsible for "authorization" checks that the UI will make to hide features from user. E.g - can user create a topic, can user access the UI, can user produce a message etc.
```
tombentley, line 105: `Admin.describeTopics()` can optionally include the authorized operations on the topic (see `org.apache.kafka.clients.admin.DescribeTopicsOptions#includeAuthorizedOperations(boolean)` and `org.apache.kafka.clients.admin.TopicDescription#authorizedOperations`) if that's what you mean?

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,107 @@
...
+2. If the user authentication is based on a Kafka user - how do we federate/reconcile that user against other applications. E.g, k8s permission to create a KafkaUser or read a Secret, or prometheus access tokens
```
tombentley, line 106: That is an excellent question.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,107 @@
...
+3. Can an access_token be set on an admin client basis; currently docs suggest setting a system property of `oauth.access.token`.
```
tombentley, line 107: @mstruk can you answer this?

tombentley (commented):
This is great, thanks!

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Authorization
+Both security mechanisms allow authorization to be done using the standard Kafka authorizers along with ACL support.
+Alternatives are available with Kafka authorizers being supplied for [KeyCloak](https://www.keycloak.org/) and [Open Policy Agent](https://www.openpolicyagent.org/docs/latest/) as documented in the Strimzi documentation. The proposed authentication mechanisms for the UI and the admin server will allow these documented authorization methods to be used without requiring any additional changes to the documented methods.
+
+## Proposed Authentication Implementations
+### Admin server using SCRAM-SHA-512
+The following describes the flow for authentication and authorization using SCRAM SHA512 as the authentication type when a web client calls a Kafka client API through the admin server.
+In the diagram:
+1. The credentials are encoded in a base64 encoded Basic Auth header and are passed to the admin server.
```
ppatierno, line 27: I would assume that the connection between web client and admin server is TLS encrypted right? Wouldn't it be better to make it clear here?

dlabaj (approved):

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,107 @@
...
+2. The admin server decodes the credentials and populates the Kafka client properties JAAS login string with the userid and password.
+3. The Kafka client api is called from the admin server which authenticates the credentials on the Kafka Broker.
+4. If authentication is successful, the principal is passed to a Kafka authorizer for authorization.
+5. If authorization is successful, the command completes and returns the result to the admin server.
+6. The admin server encodes the response in an HTTP response and returns to the web client.
+
+![Strimzi Admin server using SCRAM-SHA512](images/009-scram-sha512-admin-server.png)
+
+The credentials that are used need to be setup through the Strimzi User Operator.
+The admin server is stateless, so the credentials will need to be provided on every request.
```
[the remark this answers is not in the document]
gmcrobert: It does imply that an AdminClient will be instantiated on each request. This hasn't been benchmarked and we may need to do that. I think the tradeoff here is going to be the latency vs the complication of managing a session which is the argument for making it stateless. I think it is worth bearing in mind that the API is going to be used for Administration purposes and I don't think it is likely that it will have to deal with large numbers of concurrent requests or high volume.

009-ui-and-admin-server-security.md
gmcrobert: I think the answer to that is a community decision but I agree that it should be TLS encrypted as that is a best practise. I guess the one scenario we may not do that is in a "security off" scenario. If the general agreement is that it should be encrypted for all scenarios except the "security off" then I will add specific words to that affect in the proposal.

scholzj (approved):
Thanks for the proposal

009-ui-and-admin-server-security.md
ppatierno: I would say that for admin features we should not consider the "security off" scenario. If we are going to send credentials just base64 encoded we should be sure to have TLS imho.

tombentley (approved):

ppatierno (approved):

samuel-hawker (approved):

Frawless (approved):

scholzj:
@mstruk Do you have any more comments to this? Otherwise I think it will be ready to merge.

009-ui-and-admin-server-security.md
mstruk: It should be - you should be able to set the whole `sasl.jaas.config` directly in `Properties` object passed to `AdminClient` constructor.

E.g.:
```
p.setProperty("sasl.jaas.config", "org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required oauth.access.token=\"YOUR_ACCESS_TOKEN\"  ;");
```

mstruk (approved):

scholzj:
@gmcrobert Could you have a look at the comments from Marko? That seems to be the only thing left before merging this. Thanks.

009-ui-and-admin-server-security.md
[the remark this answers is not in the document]
gmcrobert: So there are benefits in caching the admin clients in a way described by @mstruk (LRU keyed on the Authorization header value so I will add that in. @mstruk as this is an implementation detail, are you happy for me to describe that in the [Admin Server proposal](https://github.com/strimzi/proposals/pull/9) rather than this proposal which is a more generic description of the security mechanisms?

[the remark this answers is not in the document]
mstruk: Sure.

009-ui-and-admin-server-security.md
gmcrobert: @ppatierno I added a statement that the connection is encrypted. I also tried to clarify what I meant by `security off` so hopefully this will satisfy your request.

@mstruk `Basic Auth` was a shortened version of `Basic Authentication` which is the schema on the `Authorization` header. I changed the sentence to be more explicit.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,107 @@
...
+The following diagram shows the flow of requests to the UI:
+
+![Strimzi UI using SCRAM-SHA512](images/009-scram-sha512-ui-request-flow.png)
+
+## OAuth 2
+### Admin server using OAuth2
+In the OAuth2 model, an admin user will register a client with the OpenID Provider.
+The OAuth2 support for the admin server will then be very similar to the SCRAM-SHA-512 model and will take advantage of the [Strimzi Kafka OAuth library](https://github.com/strimzi/strimzi-kafka-oauth).
+Instead of a Basic auth header being used, each request to the admin server will contain a bearer token in an Authorization header.
+The token can be a refresh token or an access token associated with the registered client which is obtained by the user using the [Strimzi cli](https://github.com/strimzi/strimzi-kafka-oauth/blob/master/examples/docker/kafka-oauth-strimzi/kafka/oauth.sh) tool.
```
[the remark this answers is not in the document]
gmcrobert: I removed `refresh token` so it now only contains `access token`.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,107 @@
...
+The admin server will extract the token from the header and configure the client properties before calling the broker using the standard Kafka OAUTHBEARER SASL mechanism.
+
+### UI using OAuth2 (OIDC)
+The OAuth2 flow for the UI is more complicated. The UI is responsible for obtaining the access token or refresh token to put onto the Authorization header on calls to the admin server.
```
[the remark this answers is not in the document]
gmcrobert: I removed `refresh token` so it now only contains `access token`.

009-ui-and-admin-server-security.md
```diff
@@ -0,0 +1,107 @@
...
+The admin server will extract the token from the header and configure the client properties before calling the broker using the standard Kafka OAUTHBEARER SASL mechanism.
+
+### UI using OAuth2 (OIDC)
+The OAuth2 flow for the UI is more complicated. The UI is responsible for obtaining the access token or refresh token to put onto the Authorization header on calls to the admin server. 
+ 
+An admin user will register a client with the OpenID Provider (OP) - to create an OIDC client that provides the `authorization_code` OAuth 2.0 flow. The UI server can then be configured to use the OpenID Provider (OP) for authentication, and generating access tokens.
+
+The UI server will use Passport.js to provide/handle an `authorization_code` interaction to authenticate user and retrieve an access token from the OP. Passport will also handle updating the session for the user to contain the access token, and refreshing/reauthentication when the token expires. 
+
+The access token can then be passed to admin API as a Bearer token Authorization header. Admin API can then use this to issue requests to Kafka via OAUTHBEARER (Kafka will authenticate against the OP).
```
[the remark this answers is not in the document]
gmcrobert: I tidied up the wording to be inline with the suggestion.

gmcrobert:
@scholzj I have responded to the comments and made some minor adjustments to the document to take into account some of the comments that were made.

scholzj:
Thanks for the PR. It is not merged.

PR #14, opened 2020-09-28
A roadmap to using the v1 CRD API

tombentley:


009-crd-v1-roadmap.md
```diff
@@ -0,0 +1,108 @@
...
+1. Install the 0.21.0 CRDs using `kubectl apply|replace -f` like normal.
+2. Touch each of their CRs, changing the apiVersion to v1beta2 (the tool can do this).
+3. Edit the CRD so that `v1alpha1` and `v1beta1` are both served=false (the tool can do this).
+4. Double check that all their CRs are using v1beta2 and all their apps are still functioning (the tool can do this).
+5. Edit the CRD `status.storedVersions` to remove `v1alpha1` and `v1beta1` (the tool can do this).
+At this point they’re all set for Strimzi 0.22.0
+
+### Strimzi "0.22.0"
+
+This should work on Kubernetes 1.16+ (including 1.21+). This will be the first release using CRD v1 API.
```
scholzj, line 66:
```suggestion
This should work on Kubernetes 1.16+ (including 1.22+). This will be the first release using CRD v1 API. 
```

009-crd-v1-roadmap.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Timing
+
+Strimzi 0.21 could be any time before the release of Kubernetes 1.21, but obviously the more time people have to upgrade the better and the more breathing space we have before 1.21 the better to iron out problems people might have in upgrading.
+
+
+
+## Compatibility
+
+Under this proposal Strimzi "0.22" would not support Kubernetes 1.15 or earlier.
```
scholzj, line 92: In 0.20.0 you said _This will be the last version of Strimzi which supports Kubernetes 1.11-1.15._ ... so this should be already under 0.21.0 and not 0.2ě.0?

009-crd-v1-roadmap.md
```diff
@@ -0,0 +1,108 @@
...
+
+Steps 2-5 in the procedure ascribed to Strimzi 0.21, above are technically just preprequisites before installing CRDs which use the v1 API. As such they could equally be done by users as part of the upgrade to Strimzi 0.22. The benefit of doing them sooner is that users would be able to perform those steps during their use of Strimzi 0.21, which might be more convenient.
+
+## Affected/not affected projects
+
+Call out the projects in the Strimzi organisation that are/are not affected by this proposal. 
+
+## Timing
+
+Strimzi 0.21 could be any time before the release of Kubernetes 1.21, but obviously the more time people have to upgrade the better and the more breathing space we have before 1.21 the better to iron out problems people might have in upgrading.
```
scholzj, line 86:
```suggestion
Strimzi 0.21 could be any time before the release of Kubernetes 1.22, but obviously the more time people have to upgrade the better and the more breathing space we have before 1.22 the better to iron out problems people might have in upgrading.
```

009-crd-v1-roadmap.md
```diff
@@ -0,0 +1,108 @@
...
+This release will be a gate release, meaning that all existing Strimzi users will have to pass through 0.21 in order to upgrade to 0.22.
+This should work on Kubernetes 1.16-1.20. It can’t work on lower versions because it relies on having support for per-versions validation schemas in CRDs, a feature that wasn’t available before 1.16. It can’t work on higher versions because that don’t support `v1beta1` of the CRD API.
+
+#### Development
+
+We need to:
+
+* Audit all existing CRD looking for things we need to change in order to have a structural schema. This is basically about checking that the type of each property is not going to need to change.
+* Prime suspects include things which are declared as a `Map<String, Object>` is Java, hence an object in JSON Schema. This includes `Kafka.spec.kafka.jvmOptions`, `Kafka.spec.zookeeper.jvmOptions`, and similar in other CRDs.
+For each of these we should make the analogous change that we did for listeners in 0.20.0.
```
scholzj, line 42: My feeling is that we should hurry up and change this in 0.20.0 or in some version before the v1beta2 move and not in the same version. I guess technically it is not critical, but it will make it IMHO smoother.

009-crd-v1-roadmap.md
```diff
@@ -0,0 +1,108 @@
...
+
+This release will be a gate release, meaning that all existing Strimzi users will have to pass through 0.21 in order to upgrade to 0.22.
+This should work on Kubernetes 1.16-1.20. It can’t work on lower versions because it relies on having support for per-versions validation schemas in CRDs, a feature that wasn’t available before 1.16. It can’t work on higher versions because that don’t support `v1beta1` of the CRD API.
+
+#### Development
+
+We need to:
+
+* Audit all existing CRD looking for things we need to change in order to have a structural schema. This is basically about checking that the type of each property is not going to need to change.
+* Prime suspects include things which are declared as a `Map<String, Object>` is Java, hence an object in JSON Schema. This includes `Kafka.spec.kafka.jvmOptions`, `Kafka.spec.zookeeper.jvmOptions`, and similar in other CRDs.
```
scholzj, line 41: `jvmOptions` is not a Map. Only `-XX` is a map. `jvmOptions` is an object. Not saying this change makes not sense just because of it.

Also, this seems very generic in _and similar in other CRDs_. The main use of maps will be the `config` fields, `metrics` fields or the `labels` and `annotation` fields. Converting these would be rather hard (for users as well as to implement). So we should probably be clear here if these are included or not.

scholzj (commented):
I left some minor comments. I think my main issue with this is the use of versions ... should we use months instead or at least in addition? I think the timing is more important that whether it is 0.21 and 0.22. If we do it in 0.25 but still in time, it will be ok. If we do it in 0.22 but too late it will not work.

009-crd-v1-roadmap.md
tombentley: I agree:

* technically it doesn't matter
* we should work on this as a matter of priority

009-crd-v1-roadmap.md
```diff
@@ -0,0 +1,118 @@
+
+
+# Roadmap to using the CRD v1 API
+
+This describes (roughly) the steps Strimzi users will need to take to get to using v1beta2 Strimzi APIs which, in turn, are using the v1 CRD API. 
+As such it talks about future versions of Strimz. We don't yet know for sure what versions the features described would land in, so we'll use the terms
```
Frawless, line 6:
```suggestion
As such it talks about future versions of Strimzi. We don't yet know for sure what versions the features described would land in, so we'll use the terms
```

Frawless (approved):

009-crd-v1-roadmap.md
```diff
@@ -0,0 +1,118 @@
...
+Strimzi X and Strimzi Z, where X and Z are unknown version numbers > 0.20 with X < Z. We've used Z rather than Y because it's not a given that Z is the version following X.
+The requirement is that what we’re calling Strimzi Z is released around the same time that Kubernetes 1.21 is.
+
+## Current situation
+
+Currently Strimzi defines numerous `CustomResourceDefinitions` using the `apiextensions.k8s.io/v1beta1` API. Those CRDs define the "Strimzi CRs", as `v1alpha1` and `v1beta1` APIS in the `kafka.strimzi.io` API group.
+
+Kubernetes 1.21 is expected to have removed CRD `apiextensions.k8s.io/v1beta1`. Its replacement, `apiextensions.k8s.io/v11`, has been available since Kubernetes 1.16.
```
scholzj, line 14:
```suggestion
Kubernetes 1.22 is expected to have removed CRD `apiextensions.k8s.io/v1beta1`. Its replacement, `apiextensions.k8s.io/v11`, has been available since Kubernetes 1.16.
```

009-crd-v1-roadmap.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+## Motivation
+
+To work on Kubernetes 1.21 Strimzi will need to use the `apiextensions.k8s.io/v1` CRD API. We can't have a single set of CRDs which support both Kubernetes 1.21 or later and also 1.15 or earlier. Furthermore, using `apiextensions.k8s.io/v1` requires that the valiation schemas in the CRD are all "structural schemas".
```
scholzj, line 19:
```suggestion
To work on Kubernetes 1.21 Strimzi will need to use the `apiextensions.k8s.io/v1` CRD API. We can't have a single set of CRDs which support both Kubernetes 1.22 or later and also 1.15 or earlier. Furthermore, using `apiextensions.k8s.io/v1` requires that the valiation schemas in the CRD are all "structural schemas".
```

009-crd-v1-roadmap.md
```diff
@@ -0,0 +1,118 @@
...
+The upgrade will just be a matter of replacing the CRDs as usual.
+
+### Alternative
+
+Steps 2-5 in the procedure ascribed to Strimzi X, above are technically just preprequisites before installing CRDs which use the `apiextensions.k8s.io/v1` API. As such they could equally be done by users as part of the upgrade to Strimzi Z. The benefit of doing them sooner is that users would be able to perform those steps during their use of Strimzi X, which might be more convenient.
+
+
+## Timing
+
+Strimzi X could be any time before the release of Kubernetes 1.22, but obviously the more time people have to upgrade the better and the more breathing space we have before 1.21 the better to iron out problems people might have in upgrading.
```
scholzj, line 96: Should this be Z?

```suggestion
Strimzi Z could be any time before the release of Kubernetes 1.22, but obviously the more time people have to upgrade the better and the more breathing space we have before 1.21 the better to iron out problems people might have in upgrading.
```

009-crd-v1-roadmap.md
```diff
@@ -0,0 +1,118 @@
+
+
+# Roadmap to using the CRD v1 API
+
+This describes (roughly) the steps Strimzi users will need to take to get to using v1beta2 Strimzi APIs which, in turn, are using the v1 CRD API. 
+As such it talks about future versions of Strimz. We don't yet know for sure what versions the features described would land in, so we'll use the terms
+Strimzi X and Strimzi Z, where X and Z are unknown version numbers > 0.20 with X < Z. We've used Z rather than Y because it's not a given that Z is the version following X.
```
ppatierno, line 7: 0.20.0

009-crd-v1-roadmap.md
ppatierno, line 14: typo `v11`

009-crd-v1-roadmap.md
ppatierno, line 19: I would add a link to "structural schemas" definitions.

009-crd-v1-roadmap.md
```diff
@@ -0,0 +1,118 @@
...
+Kubernetes 1.21 is expected to have removed CRD `apiextensions.k8s.io/v1beta1`. Its replacement, `apiextensions.k8s.io/v11`, has been available since Kubernetes 1.16.
+
+
+## Motivation
+
+To work on Kubernetes 1.21 Strimzi will need to use the `apiextensions.k8s.io/v1` CRD API. We can't have a single set of CRDs which support both Kubernetes 1.21 or later and also 1.15 or earlier. Furthermore, using `apiextensions.k8s.io/v1` requires that the valiation schemas in the CRD are all "structural schemas".
+
+## Proposal
+
+### Strimzi 0.20
```
ppatierno, line 23: 0.20.0

009-crd-v1-roadmap.md
```diff
@@ -0,0 +1,118 @@
...
+
+#### Users
+
+Users will need to follow a specific procedure to upgrade:
+
+1. Install the Strimzi X CRDs using `kubectl apply|replace -f` like normal.
+2. Touch each of their CRs, changing the `apiVersion` to `v1beta2` (the tool can do this).
+3. Edit the CRD so that `v1alpha1` and `v1beta1` are both served=false (the tool can do this).
+4. Double check that all their CRs are using v1beta2 and all their apps are still functioning (the tool can do this).
+5. Edit the CRD `status.storedVersions` to remove `v1alpha1` and `v1beta1` (the tool can do this).
```
ppatierno, line 74: AFAIU from the introduction, for upgrading to Strimzi X, users have to upgrade Kubernetes as well choosing a version between 1.16-1.20. Should we make explicit here?

tombentley (commented):
Something that this doesn't cover currently is the upgrade procedure when someone is upgrading their kube cluster from 1.15 or earlier to 1.22 or later. In other words, the Strimzi X gating requirement only works if the user has a kube cluster on which Strimzi X can be installed.

Do we think this scenario is worth discussing here? I think a workable procedure would look like this:

1. Get a copy of all your `Kafka` resources
2. Scale the operator to zero. 
3. Upgrade kube (I've no idea how this works) and install the new CRDs, but don't scale up the operator.
4. Convert the offline resources from step 1 to `v1beta2` resources and create them in kube. 
5. Upgrade the scale up the operator. It sees `v1beta2` resources straight out of the gate.

But there would also be a Strimzi upgrade happening here, which might add complication.

Thoughts?

009-crd-v1-roadmap.md
```diff
@@ -0,0 +1,119 @@
...
+## Current situation
+
+Currently Strimzi defines numerous `CustomResourceDefinitions` using the `apiextensions.k8s.io/v1beta1` API. Those CRDs define the "Strimzi CRs", as `v1alpha1` and `v1beta1` APIS in the `kafka.strimzi.io` API group.
+
+Kubernetes 1.22 is expected to have removed CRD `apiextensions.k8s.io/v1beta1`. Its replacement, `apiextensions.k8s.io/v1`, has been available since Kubernetes 1.16.
+
+
+## Motivation
+
+To work on Kubernetes 1.22 Strimzi will need to use the `apiextensions.k8s.io/v1` CRD API. We can't have a single set of CRDs which support both Kubernetes 1.22 or later and also 1.15 or earlier. Furthermore, using `apiextensions.k8s.io/v1` requires that the valiation schemas in the CRD are all [structural schemas](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema).
```
samuel-hawker, line 19:
```suggestion
To work on Kubernetes 1.22 Strimzi will need to use the `apiextensions.k8s.io/v1` CRD API. We can't have a single set of CRDs which support both Kubernetes 1.22 or later and also 1.15 or earlier. Furthermore, using `apiextensions.k8s.io/v1` requires that the validation schemas in the CRD are all [structural schemas](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema).
```

009-crd-v1-roadmap.md
samuel-hawker: I think this  is  probably already made explicit enough on line  68?

samuel-hawker (approved):
Looks good to me - fairly straightforward!

scholzj (approved):

ppatierno (approved):

PR #15, opened 2020-10-09
Proposal for Strimzi canary

robshelly:


011-kafka-canary.md
```diff
@@ -0,0 +1,40 @@
+# Strmixi Canary
```
scholzj, line 1: Typo

011-kafka-canary.md
```diff
@@ -0,0 +1,40 @@
...
+
+Implement a Kafka canary tool which will act as an indicator of whether Kafka clusters are operating correctly. This would be achieved by creating a canary topic and continously producing and consuming events on the topics.
```
scholzj, line 3: Please explain what continuously means. There is not continuous messaging - it will be in intervals and it is important to understand what kind of intervals are you talking about.

011-kafka-canary.md
```diff
@@ -0,0 +1,40 @@
+# Strmixi Canary
+
+Implement a Kafka canary tool which will act as an indicator of whether Kafka clusters are operating correctly. This would be achieved by creating a canary topic and continously producing and consuming events on the topics. 
+
+## Current situation
+
+Currently  Strimzi does not have any canary style feature. It does have readiness checks for Kafka cluster which can be replaced/augmented with the canary.
```
scholzj, line 7: What readiness checks do you have in mind? The ones in Kube pods? This cannot replace them because they have completely different semantics. For example they do not focus on Kafka cluster at all. They are pod scoped only. It can of course add additional level of health checking, but it would need to be a separate thing run from somewhere else. 

Or did you meant something else with the readiness checks?

011-kafka-canary.md
david-martin, line 3: periodically rather than continously ?

011-kafka-canary.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Motivation
+
+The canary would provide the following:
+  
+  * Improved readiness check for Kafka clusters
```
scholzj, line 13: Please do not call this readiness check since it is too confusing with Kube readiness checks.

011-kafka-canary.md
```diff
@@ -0,0 +1,40 @@
...
+  * SLI(s) for service admins
+  * Metrics:
+    * End-to-end latency
+    * Succesfull/failed events
+    * Disk usage
+    * Disk errors
```
scholzj, line 19: How would you expect this info to be gathered?

011-kafka-canary.md
```diff
@@ -0,0 +1,40 @@
...
+## Motivation
+
+The canary would provide the following:
+  
+  * Improved readiness check for Kafka clusters
+  * SLI(s) for service admins
+  * Metrics:
+    * End-to-end latency
+    * Succesfull/failed events
+    * Disk usage
```
david-martin, line 18: I may be lacking strimzi or kafka knowledge,
but how do you see the canary providing disk metrics?
That seems like more of a host level metric that the product/consumer or even the broker

011-kafka-canary.md
```diff
@@ -0,0 +1,40 @@
...
+Currently  Strimzi does not have any canary style feature. It does have readiness checks for Kafka cluster which can be replaced/augmented with the canary.
+
+## Motivation
+
+The canary would provide the following:
+  
+  * Improved readiness check for Kafka clusters
+  * SLI(s) for service admins
+  * Metrics:
+    * End-to-end latency
```
scholzj, line 16: It is not really end-to-end latency from the customer perspective. It can be measured, but its value is limited. Not sure I would call it end-to-end latency.

david-martin, line 16: If this can also be subdivided to also expose producer & consumer metrics, similar to what's shown in https://www.datadoghq.com/blog/monitoring-kafka-performance-metrics/#kafka-producer-metrics, it would help with deeper troubleshooting of where problems could be coming from

011-kafka-canary.md
```diff
@@ -0,0 +1,40 @@
...
+    * Succesfull/failed events
+    * Disk usage
+    * Disk errors
+
+## Proposal
+
+The canary would be implemented as part of the Kafka Roller component. 
+A single _canary_ topic would be created on creation of a Kafka cluster. The topic would have one partition for Kafka broker. Events would then be continously produced and consumed to/from the canary topic. The internal TLS 9091 port would be used to connect to the brokers.
+
+As this happens the relevant metrics would be collected and exported. Errors such as the inability to produce to the topic can then be used by the KafkaRoller to trigger a restart, acting as healthc check for the Kafka cluster.
```
scholzj, line 26: I think this needs significantly more details on how it will be implemented. Will it be a separate pod deployed by the operator? How will it pass the results to the KafkaRoller? Or will it run inside the operator? If yes, how it will be implemented there with respect to resource consumptions, working in parallel etc.?

It should probably also go into the details how would the canary create and configure its topic and evaluate the result. Would it have topic with replication factor (RF) == min.isr == number of Kafka nodes? If yes, how will it deal with rolling updates etc.? Or will the min.isr be RF-1? How would such topic change the replication factor during scale-ups/scale-downs? If it will have less replicas then nodes, how will it test the whole cluster?

011-kafka-canary.md
```diff
@@ -0,0 +1,40 @@
...
+
+The canary would be implemented as part of the Kafka Roller component. 
+A single _canary_ topic would be created on creation of a Kafka cluster. The topic would have one partition for Kafka broker. Events would then be continously produced and consumed to/from the canary topic. The internal TLS 9091 port would be used to connect to the brokers.
+
+As this happens the relevant metrics would be collected and exported. Errors such as the inability to produce to the topic can then be used by the KafkaRoller to trigger a restart, acting as healthc check for the Kafka cluster. 
+
+
+## Affected/not affected projects
+
+This functionality should not affect other Strmizi projects.
```
scholzj, line 31: Typo. Plus I guess this should say that it affects the Strimzi operators.

011-kafka-canary.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Compatibility
+
+The canary would be fully backwards compatible.
```
scholzj, line 35: I guess saying something like _does not impact any existing functionality_ might be better and more clear. Since there was no _canary_ before, there is no real backwards compatibility.

011-kafka-canary.md
david-martin, line 26: I wonder if this paragraph should be a stretch/future goal & initially deliver the proposal as a passive & optional feature ?

scholzj (changes requested):
Thanks for the proposal. I think it would be an interesting feature. But I think it needs to have a bit more details about how would it work.

011-kafka-canary.md
```diff
@@ -0,0 +1,40 @@
...
+## Current situation
+
+Currently  Strimzi does not have any canary style feature. It does have readiness checks for Kafka cluster which can be replaced/augmented with the canary.
+
+## Motivation
+
+The canary would provide the following:
+  
+  * Improved readiness check for Kafka clusters
+  * SLI(s) for service admins
```
ppatierno, line 14: what kind of SLI(s) you have in mind?

011-kafka-canary.md
```diff
@@ -0,0 +1,40 @@
...
+  * Metrics:
+    * End-to-end latency
+    * Succesfull/failed events
+    * Disk usage
+    * Disk errors
+
+## Proposal
+
+The canary would be implemented as part of the Kafka Roller component.
```
ppatierno, line 23: I would take it separate from the Kafka Roller but used by the Kafka Roller.

011-kafka-canary.md
```diff
@@ -0,0 +1,40 @@
...
+  * Metrics:
+    * End-to-end latency
+    * Succesfull/failed events
+    * Disk usage
+    * Disk errors
+
+## Proposal
+
+The canary would be implemented as part of the Kafka Roller component. 
+A single _canary_ topic would be created on creation of a Kafka cluster. The topic would have one partition for Kafka broker. Events would then be continously produced and consumed to/from the canary topic. The internal TLS 9091 port would be used to connect to the brokers.
```
ppatierno, line 24: Mentioning one partition for Kafka broker, what will be the replication factor (RF)? Maybe it would be just RF = 1?
I see a problem when a Kafka broker goes down for any reason (real actual problem or rolling update).
If, let's say broker A, is hosting partition X, the canary tool should detect that that broker A is down. If RF > 1, a leader for that partition will be elected on another broker, let's say broker B ... but it's not much of interest for us, because broker B is already hosting partition Y for canary test. When broker A comes back we want to have it as leader again for partition X (preferred leader election). Does it really make sense having a RF > 1 for this feature then?

011-kafka-canary.md
ppatierno: Adding more resources consumption to the cluster operator was my first concern but even talking with @tombentley it was clear that we are moving resources consumption somewhere else if the canary is running in a different pod with another JVM and we also need to expose a kind of REST (?) interface to the Kafka Roller to get canary result.
Anyway, we should try to narrow down the resource usage for producer and consumer (buffer sizes, ...) and even the partitions size (maybe even using a compacted topic).

As I mentioned in another comment above what's the value of having RF > 0 if the objective of the canary tool is checking that a specific broker is working or not, when it's hosting one leader partition that would not have sense having a different leader on a different broker because it's already hosting its own "canary" partition. What am I missing?

samuel-hawker: Definitely agree with this - seems like if the Canary proposal were to be accepted, it would be standalone first, and if appropriate then a followup proposal would suggest the use of it for the KafkaRoller

011-kafka-canary.md
```diff
@@ -0,0 +1,40 @@
...
+This functionality should not affect other Strmizi projects.
+
+## Compatibility
+
+The canary would be fully backwards compatible.
+
+
+## Rejected alternatives
+
+Another option would be to create an external canary client to provide much of the above functionality. It could be deployed separately to the Strimzi operator and connect to the brokers thourgh the external route. However the benefit of building this tool into the Strimzi operator is the abilty to use it as a health check for Kafka clusters.
```
samuel-hawker, line 40: I think that the canary as a standalone client would be better personally, and if it were useful for the operator it would be fairly trivial to pull it in as a dependency and run the libraries within the operator.

Another advantage of it being standalone is it could be used with other Kafka clusters, not just strimzi ones.

011-kafka-canary.md
ppatierno: This is the way we use for our HTTP Kafka bridge component because it internally uses native Kafka producer and consumer for bridging the HTTP traffic. We have a JMX collector and then a metrics reporter to provide these metrics in Prometheus format. I think it should work in the same way.
You can also refer to the official Kafka documentation:

https://kafka.apache.org/documentation/#producer_monitoring

https://kafka.apache.org/documentation/#consumer_monitoring

011-kafka-canary.md
ppatierno: Yeah in the current dashboard we get this kind of information from the kubelet volume related metrics.

011-kafka-canary.md
ppatierno: Maybe just "health" check?

011-kafka-canary.md
tombentley: @ppatierno you're right we can get metrics from the producer and consumer themselves, but this would give us a little more insight. For example knowing the producer latency it useful; `request-latency-avg` isn't quite the same thing since it includes non-produce requests and it's also only an average. Measuring directly could yield a histogram.

011-kafka-canary.md
tombentley: I think the point is that getting a successful acknowledgement to a produce request when acks=1 allows us to infer that the broker is able to write to the disk.

011-kafka-canary.md
tombentley: In out-of-band conversation with Rob I suggested we could use a single topic with RF=1 and #partitions = #brokers × #logdirs-per-broker, i.e. with a partition per log dir. This would mean that leadership change would not happen and (outside of rolling restart) would tell us immediately whether there was a problem writing to or reading from the log dir volume. 

The topic would be tuned to have a short retention and small segment size, so it wouldn't require much in the way of disk space.

On reflection this would cause problems when downscaling, since Kafka doesn't support decreasing the number of partitions. This isn't necessarily a problem. We could just ignore the spare partitions, or we could delete the topic and recreate it, since the actual data in the topic isn't of much value, it's the action of producing and consuming it which is useful.

While it would be configurable, I do think for the purposes of the proposal we should be more concrete about the frequency of message production. Are we talking about a 1s, 5s, 10s or 30s period? I suspect 1s or 5s is reasonable.

011-kafka-canary.md
tombentley: @samuel-hawker I don't think that's unreasonable, but I think this is an opportunity to figure out what value the operator could get from this to ensure that whatever got written was actually useful for the operator use case. The value I see is:

* Deeper checking that the broker really is working, by exercising the produce and consume paths, which could be used by the KafkaRoller (we'd need to define exactly what it would do when there was a problem, but until we know the API and the errors which propagate from it that's a little hard to do). 
* Getting this information sooner after something goes wrong, perhaps triggering a reconciliation sooner than the next scheduled one.
* Some better metrics, though I think we should try to define exactly what in this proposal.

To elaborate a little.... the proposal says that it would use the internal listener. Thus any authentication exceptions from the producer would suggest that the internal PKI is broken. 

Another interesting case would be `KafkaStorageException` which is a `RetriableException` so I don't know off the top of my head how that would propagate out of a producer, but we could use that as a signal to check with `Admin.describeLogDirs`, restart the broker and prevent restarting any other brokers in the meantime. 

In both these examples the `Kafka` status could be set with a `Condition` which indicated that there was a problem and whether it was, or was not, being handled automatically.

011-kafka-canary.md
ppatierno: @tombentley doesn't it just mean that the message was written in the page-cache so it's still in memory and not on the disk? The disk I/O error could come later when the operating system will do a flush, or?

011-kafka-canary.md
ppatierno: > In out-of-band conversation with Rob I suggested we could use a single topic with RF=1 and #partitions = #brokers × #logdirs-per-broker, i.e. with a partition per log dir. This would mean that leadership change would not happen and (outside of rolling restart) would tell us immediately whether there was a problem writing to or reading from the log dir volume.

That's exactly what I was explaining with my example ... maybe badly :-)

> ... or we could delete the topic and recreate it ...

Maybe we should pay attention to this because if auto topic creation is enabled (in the broker config) and we delete the topic, because the canary tool is running, we could hit the time period when it's going to produce and the topic is created with the default broker configuration instead of what we need.

011-kafka-canary.md
scholzj: Well, for example with JBOD it would test only one disk out of N. So I think it can discover some disk errors as a side-effect. But it is not really a feature which should be advertised since it does not really check all disks and it really finds only disk errors which affects the particular consumer / producer. It doesn't check any system metrics or anything to find out about disk errors.

011-kafka-canary.md
scholzj: Yes, that is exactly my point. This could be used as an additional health check in KafkaRoller. But not as a readiness check in Kafka pods. Hence the need to properly explain what does it mean and avoid any possible confusion.

011-kafka-canary.md
tombentley: Strictly speaking when changes actually make it to disk depends on a bunch of things including (but probably not limited to):

* The flush settings for the topic (in particular the max number of messages between flushes)
* The configuration of the OS
* And lower level things. (I believe it's not unusual/common for disks to have a battery backed memory cache which allows it to pretend that it wrote to disk while not actually, so even the kernel is being lied to). 

But it's a good point because the better we're tune the topic for disk error inference the less useful it is for metrics from which we can infer anything useful about performance. Kafka recommendation is to use replication rather than OS flush, of course.

ppatierno: > Kafka recommendation is to use replication rather than OS flush, of course.

exactly! because data goes to the page cache and you really don't know when it will be written to disk. So in reality I don't think we can get a good info about disk error from producer.

011-kafka-canary.md
scholzj: In the first place, this needs to be described in the proposal. There are use cases where the CO operates many clusters. In that case running all of this in single pod will not be nice despite the possible memory savings. Especially if you would have it running constantly and sending a message every second or so. So I think a separate pod deployed as part of the cluster would be better here.

011-kafka-canary.md
tombentley: You'd just set `flush.messages = 1`

011-kafka-canary.md
scholzj: > In out-of-band conversation with Rob I suggested we could use a single topic with RF=1 and #partitions = #brokers × #logdirs-per-broker, i.e. with a partition per log dir.

Why not use single partition and RF=<Number of brokers> and min.isr=<RF-1>? That actually tests the rolling update works without disruptions and can measure availability. How would you measure availability if the topics are coming and going during rolling updates etc.?

011-kafka-canary.md
tombentley: I guess it's also worth discussing the merits of just checking `Admin.describeLogDirs()` on every reconciliation (maybe in KafkaRoller). That should tell you if a broker had observed an `IOException` while reading or writing from any segment in a given log directory. It would be less timely (the log directory could have been offline for as much as a reconciliation interval), and it wouldn't validate any of the rest of the broker (e.g. network stack etc). But it would be relatively simple.

ppatierno: That says just to flush every message not ... when to flush. Of course, you know internal Kafka broker behavior better than me but how it reflects on the OS flush, just really calling "fsync" before the ack is sent? or sent the ack and then calling "fsync" ... so you can still lose the message and have I/O disk errors?

011-kafka-canary.md
tombentley: That's not worthless, but it's measuring a different thing. It doesn't really tell you that there's something wrong with _this_ particular broker, which is what KafkaRoller is interested in.

scholzj: That is good point ... so you probably need both to achieve what this proposal promises - one for improved broker health information and one for SLIs.

011-kafka-canary.md
scholzj: Be careful with `flush.messages=1` ... while it is topic level config, I wouldn't be surprised of it to have potential to impact the overall performance and its predictability as well.

011-kafka-canary.md
robshelly: Based this on a comment from @tombentley in another channel but I may have misunderstood. Is this possible @tombentley?

011-kafka-canary.md
robshelly: I'm happy for this functionality to be a stretch goal. However, I believe the main reason for building the canary into the Strimzi (instead of an external one) was to use it in the Kafka Roller. @ppatierno, @tombentley How to you feel about this being a stretch goal?

011-kafka-canary.md
robshelly: Yeah this was a typo. Fixed now!

011-kafka-canary.md
samuel-hawker: I think this sounds great, so the canary would be more of an API right? A standalone library that we would hook the operator straight into. Your suggestion of using it to trigger reconciliations sound useful, but I suppose that would mean spinning up a canary in a separate verticle for each operand (or indeed one canary to manage them all)?

ppatierno:
While digging into some Kafka metrics, I found this project from Linkedin https://github.com/linkedin/kafka-monitor

I had a quick look at the code and the wiki, it would be useful you will have a look as well.
It seems that it can provide what we would need without reinventing the wheel.

It's modular and provides different services. There are the producer and consumer services that do what we need, sending/receiving messages, and exposing some related metrics (via JMX).
The interesting part is:

> Xinfra Monitor can automatically create the monitor topic with the specified config and increase partition count of the monitor topic to ensure partition# >= broker#. It can also reassign partition and trigger preferred leader election to ensure that each broker acts as leader of at least one partition of the monitor topic.

It's the kind of discussion we are having about one partition leader per broker. It's missing the possibility to have one per log dir.

The metrics are exposed via JMX by default but it also supports Statsd, Jolokia, Graphite and if we want Prometheus, there is a YAML configuration file for JMX exporter (that should run alongside it).

@robshelly  @david-martin @tombentley @scholzj @samuel-hawker wdyt?

david-martin:
> I found this project from Linkedin https://github.com/linkedin/kafka-monitor

@ppatierno That looks very useful.
The metrics listed in the readme seem to cover what we're looking for.
The [Motivation](https://github.com/linkedin/kafka-monitor/wiki/Motivation) hits a lot of the same notes as us, and the [Roadmap](https://github.com/linkedin/kafka-monitor/wiki/Roadmap-for-Xinfra-Monitor) looks good too.

1 thing I couldn't gauge is if it's intended to be run against a production cluster with active users.
Either way, it's definitely worth looking into more.

011-kafka-canary.md
```diff
@@ -0,0 +1,69 @@
...
+  * Allows for one (and only one) broker to be down, therefore accommodating rolling updates.
+  * Mimics the availability that a user would expected.
+  * Ensure that all brokers, minus one which can be down, are working as they need to be in-sync
+
+#### Cons
+
+  * Unsure what broker is down. This means if  the tool indicated that a broker is down for a period of time, we are unsure if it is the same broker (indicating a problem) or a different brokers as rolling update occurs (not a problem).
+
+
+This proposal suggest going forward with option 2.
```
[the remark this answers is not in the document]
stanlyDoge: Cannot we specify the partition in the sent canary message? If we would not get messages from the same partition, we could say it is the same broker which is down.

011-kafka-canary.md
```diff
@@ -0,0 +1,69 @@
+# Strimzi Canary
+
+Implement a Kafka canary tool which will act as an indicator of whether Kafka clusters are operating correctly. This would be achieved by creating a canary topic and periodically producing and consuming events on the topics. 
+
+## Current situation
+
+Currently Strimzi does not have any canary style feature. It does have health checks for Kafka cluster which can be replaced/augmented with the canary.
```
tombentley, line 7:
```suggestion
Currently Strimzi does not have any canary style feature. It does have Kubernetes health checks for Kafka cluster which can be replaced/augmented with the canary.
```

011-kafka-canary.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Motivation
+
+The canary would provide the following in indication that the Kafka cluster is operating as expecting from a customers perspective. I.e. messages can successfully be produced and consumed. Some would be exported to reflect the canary's activity. Initial metrics would included:
```
tombentley, line 11:
```suggestion
The canary would provide the following in indication that the Kafka cluster is operating as expecting from a client perspective. I.e. messages can successfully be produced and consumed. Some would be exported to reflect the canary's activity. Initial metrics would included:
```

011-kafka-canary.md
```diff
@@ -0,0 +1,69 @@
...
+
+In future the canary can be used by the Kafka Roller to improve health checks. However this is beyond of scope of the current proposed work.
+
+## Proposal
+
+The canary would be run as separate pod alongside Strimzi operator. It would be specified (optionally) in the Kafka custom resource and deployed by the Strimzi controller. The code will live in a separate repository within the Strimzi project. This is similar setup to the Kafka Bridge. 
+
+The canary will be built using Golang.
+
+There are two main options for how the topic is configured:
```
tombentley, line 27: Are we saying both will be supported?

011-kafka-canary.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Option 1: Partition over replicas
+Create a topic with 1 partition per broker node and a replication factor of zero. The canary would produce/consume from each partition.
```
tombentley, line 30:
```suggestion
Create a topic with 1 partition per broker node and a replication factor of 1. The canary would produce/consume from each partition. 
```

011-kafka-canary.md
```diff
@@ -0,0 +1,69 @@
...
+### Option 1: Partition over replicas
+Create a topic with 1 partition per broker node and a replication factor of zero. The canary would produce/consume from each partition. 
+
+#### Pros
+
+  * Errors will indicate that a specific broker is unavailable.
+
+#### Cons
+
+  * Brokers are expected to be unavailable during rolling updates after cluster configuration changes.
```
tombentley, line 38: Well this is also a pro, since being able to produce and consume again tells you that the broker came back OK (e.g. there weren't problems mounting volumes etc). It's only a con because on its own the canary can't tell that a broker was intentionally being rolled. Presumably this could be dealt with in prometheus alerting by exporting a metric from the CO for each broker which was 1 when intentionally restarting and 0 otherwise, allowing the difference with the canary availability metric to indicate a problem.

011-kafka-canary.md
```diff
@@ -0,0 +1,69 @@
...
+  * Brokers are expected to be unavailable during rolling updates after cluster configuration changes. 
+
+### Option 2: Replicas over partition
+Create a topic with a single parition and a replication factor equal to the number of brokers and a minimum in-sync replicas of one less than the number of brokers.
+
+#### Pros
+
+  * Allows for one (and only one) broker to be down, therefore accommodating rolling updates.
+  * Mimics the availability that a user would expected.
+  * Ensure that all brokers, minus one which can be down, are working as they need to be in-sync
```
tombentley, line 47: This option could be more impactful on the Kafka Roller, which has checks that restarting a broker pod won't affect availability of partitions. So a broker not being in the ISR would prevent another broker being restarted. That's _probably_ the right thing to do, but it might slow down the speed at which the KR could do a rolling restart.

tombentley (commented):
* We should describe whether messages are sent at a constant rate or constant delay after receipt. 
* Can you run >1 instance of the canary?
* I think it would be good to describe the schema for the messages
* It's not described how would you get the metrics out.
    * I know we talked about a REST API on the community call
    * Another possibility would be to write the metrics for the last message received into the next message sent. While this would mean that a separate consumer of the topics couldn't consume metrics if the canary wasn't able to send, it could infer it from the non-receipt of messages.

011-kafka-canary.md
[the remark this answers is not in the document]
ppatierno: In option 2, the partition is only 1, there aren't more partitions but just more replicas of the same only partition.

ppatierno:
> Can you run >1 instance of the canary?

What's the motivation for that @tombentley ?

tombentley:
@ppatierno I'm not saying we need to, but it would seem to be trivial to include in the message the id of the sender, so that multiple canaries could share the same topic. It would allow you to run the canaries in a `Deployment` should you need it.

ppatierno:
I would like to raise some other thoughts about using one topic with 1 partition and RF = N (where N is the number of brokers).
We would need a kind of "reconciler/worker" that looks at the Kafka cluster for creating the topic if it doesn't exist (at the beginning) and reassign replicas in case of scale-up/down of the cluster itself.
The behavior I see, based on Option 2, seems to be:

* if topic doesn't exist --> create it with 1 partition, RF = N and min.insync = N - 1
* check cluster periodically
* if cluster scale up/down (so N becomes != current RF) it has to do a reassignment, updating the topic with the new min.insync = N - 1 (using the "new" N)

I guess the above should be added to the proposal.

Then some concerns/thoughts:

* during a reassignment, is it possible that latency will increase if prod/cons are still running and it can have an impact on the component watching at the metrics? 
* if the topic was used a lot, and the cluster was scaled up ... the new replicas will take time to catch up which means that with acks=all, the producer could not be able to send messages for some time --> is it still valid using a compacted topic or anyway having it pretty short in size?
* Impact of a leader election if the cluster is scaled-down and one of the brokers is the partition leader

In general, I think that we should take into account that the reassignment can have an impact on the latency metrics.

On the other side, going with option 1 (1 topic, partitions = N, RF = 1), should still allow the producer to send to the partitions hosted by the current running brokers and not yet to the ones during a scale-up operation. Also, the latency metrics could be labeled per broker so who is interested in that metric have more information about which brokers are not running properly.

@scholzj @tombentley any thoughts on this?

011-kafka-canary.md
scholzj, line 11: What is the expected frequency of the messages?

011-kafka-canary.md
```diff
@@ -0,0 +1,69 @@
...
+#### Pros
+
+  * Errors will indicate that a specific broker is unavailable.
+
+#### Cons
+
+  * Brokers are expected to be unavailable during rolling updates after cluster configuration changes. 
+
+### Option 2: Replicas over partition
+Create a topic with a single parition and a replication factor equal to the number of brokers and a minimum in-sync replicas of one less than the number of brokers.
```
scholzj, line 41: To make this work with scaling up and down, we will need to be able to handle somewhere the changes to the replication factor and the min-isr. The first one is currently not supported by Strimzi. So that would need to be added tot he CO. I think this is big enough chunk of work to be included in the proposal at least as a sentence or two to make it clear.

scholzj (commented):
Some small comments. But I think this looks mostly good to me.

011-kafka-canary.md
ppatierno: Changing the replication factor means doing a partitions reassignment. Why not having the logic in the canary tool when it discovers that the cluster has a different number of nodes? 
We know it's something we want in Strimzi itself but it should be more related to when a user changes the replication factor on a `KafkaTopic` resource and the TO takes care of it; in this case, the canary tool doesn't interact in this way.
I would expect the canary tool to have a background worker process to check the cluster status and see if a replication factor change (aka partitions reassignment) is needed or not.

scholzj: That might be tricky especially with scale down. Will that not block the KafkaRoller?

ppatierno: This is one of the concerns I wrote here https://github.com/strimzi/proposals/pull/15#issuecomment-718069412 talking about the impact on components watching the metrics (maybe the Kafka Roller).
Anyway, I heard from @tombentley that option 2 could have other issues with Kafka Roller.

scholzj: This is not about watching the metrics. This is about having a topic with 10 replicas and min.isr=9 on 5 node broker.That is why I suggested that this might need to be handled in the CO.

011-kafka-canary.md
```diff
@@ -0,0 +1,69 @@
...
+#### Pros
+
+  * Allows for one (and only one) broker to be down, therefore accommodating rolling updates.
+  * Mimics the availability that a user would expected.
+  * Ensure that all brokers, minus one which can be down, are working as they need to be in-sync
+
+#### Cons
+
+  * Unsure what broker is down. This means if  the tool indicated that a broker is down for a period of time, we are unsure if it is the same broker (indicating a problem) or a different brokers as rolling update occurs (not a problem).
+
```
tombentley, line 52: Another con would be that the latency metric is essentially useless for understanding what the latency for any _real_  topic is. You just see the latency of the 2nd slowest broker. Plenty of real topics (not replicated on the slowest two brokers) would have much better latency. Nor do you know which brokers are the slow ones.

011-kafka-canary.md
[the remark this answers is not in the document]
tombentley: I agree that the proposal should cover only option 2 if that's what being actually going to be implemented under this proposal. option 1 can be mentioned as a Rejected Alternative. 

But TBH I'd like to see a sentence or two explaining why option 2 is seen as preferable. You've enumerated pros and cons of each but you've not explicitly said why option two is better, or precisely for which purposes it's better.

ppatierno:
@david-martin @tombentley @scholzj @samuel-hawker @stanlyDoge can you have another pass on this proposal, please? We would like to reach an agreement as soon as possible, thanks!

tombentley (commented):
I'd like to see a bit more detail on the message schema. It could be useful to run one producer accessing the cluster internally and another accessing it from an external address (e.g. to understand the additional latency through routes etc). We could do this using the same topic if the messages included a "producer id". These could be two producers in the same canary pod or two instances of the canary pod. 

Similarly, it might be useful to run a canary with/without a transactional producer to have visibility into the latency for transaction commit.

011-kafka-canary.md
```diff
@@ -0,0 +1,89 @@
...
+  * consume-error
+  * latency
+
+In future the canary can be used by the Kafka Roller to improve health checks. However this is beyond the scope of the current proposed work.
+Proposal
+The canary would be run as a separate pod alongside Strimzi operator. Once the canary topic is created messages will be produced and consumed from it. Messages would be consumed at a rate not faster than 1 per second and possibly as slow as 10 per second. This rate may be configurable. 
+
+The canary will be built using Golang and metrics will be exposed in Prometheus format though a REST API. Currently Sarama is being considered as the client library.
+
+The current plan is to deploy the canary independently  alongside the Strimzi operator. However, in future, consideration should be given to integrating the canary with Strimzi in a manner similar to the Kafka Bridge. I.e. the canary would be specified (optionally) in the Kafka custom resource and then deployed by the Strimzi operator.
```
scholzj, line 24: Bridge is basically a separate app. I think if we do something, then it needs to be more tightly coupled, so as CC or KE.

```suggestion
The current plan is to deploy the canary independently  alongside the Strimzi operator. However, in future, consideration should be given to integrating the canary with Strimzi in a manner similar to the Kafka Exporter or Cruise Cotnrol. I.e. the canary would be specified (optionally) in the Kafka custom resource and then deployed by the Strimzi operator.
```

011-kafka-canary.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Topic Configuration
+The topic will be configured to have a partition on each broker node and a  well defined replication factor of 3 (which is the most commonly used value).The minimum in-sync replicas should be one less than the number of replicas.
```
scholzj, line 27: Some clarifications to make this work even with small clusters.

```suggestion
The topic will be configured to have a partition on each broker node and a replication factor which will be the minimum of _number of Kafka broker nodes_ and _3 (which is the most commonly used value)_. The minimum in-sync replicas will be either 1 in case replication factor is 1 (there is only one broker in the cluster) or one less than the number of replicas.
```

011-kafka-canary.md
```diff
@@ -0,0 +1,89 @@
...
+The configuration should be something like the following:
+
+* Partitions = N (where N is the number of brokers)
+* Replication Factor = 3
```
scholzj, line 31:
```suggestion
* Replication Factor = Min(<number-of-brokers>, 3)
```

011-kafka-canary.md
```diff
@@ -0,0 +1,89 @@
...
+* Min ISR = Replication Factor - 1 = 2
```
scholzj, line 32:
```suggestion
* Min ISR = Max(1, Replication Factor - 1)
```

scholzj (commented):
Left some nits and some clarifications for smaller clusters with <3 nodes. But looks good overall. (Feel free to your own words instead of my suggestions if you want).

ppatierno:
@tombentley I am going to update the proposal with message scheme idea.
Regarding transactional producer, the Go library we would like to use for this development (the Sarama one) doesn't support transactional producer. Do you see it as a really needed feature for this tool? Maybe something we can add in the future? (but we should contribute to Sarama to add transactional producer or waiting for it at some point). I don't know about other Go libraries but Sarama seems to be the better one anyway.

011-kafka-canary.md
```diff
@@ -0,0 +1,123 @@
...
+  * produce-error
+  * consume-error
+  * latency
+
+In future the canary can be used by the Kafka Roller to improve health checks. 
+However this is beyond the scope of the current proposed work.
+
+## Proposal
+
+The canary would be run as a separate pod alongside Strimzi operator.
```
scholzj, line 26: AFAIU the canary will run per Kafka cluster ... so maybe _alongside the Kafka cluster_ instead of _alongside Strimzi operator_?

011-kafka-canary.md
```diff
@@ -0,0 +1,123 @@
...
+## Proposal
+
+The canary would be run as a separate pod alongside Strimzi operator. 
+Once the canary topic is created messages will be produced and consumed from it. 
+Messages would be consumed at a rate not faster than 1 per second and possibly as slow as 10 per second. This rate may be configurable. 
+
+The canary will be built using Golang and metrics will be exposed in Prometheus format though a REST API. 
+Currently Sarama is being considered as the client library.
+
+The current plan is to deploy the canary independently  alongside the Strimzi operator.
```
scholzj, line 33: Same as above.

011-kafka-canary.md
```diff
@@ -0,0 +1,123 @@
...
+  * consume-error
+  * latency
+
+In future the canary can be used by the Kafka Roller to improve health checks. 
+However this is beyond the scope of the current proposed work.
+
+## Proposal
+
+The canary would be run as a separate pod alongside Strimzi operator. 
+Once the canary topic is created messages will be produced and consumed from it.
```
scholzj, line 27: Should we cover here the configuration of the canary topic? In general, we should probably consider using smaller segment size and retention to make sure this does not unnecessarily fill the disk space. Not sure it needs to be in the proposal, probably not. Just a thought which should definitely not get forgotten in the impl.

scholzj (approved):
Some nits, but basically LGTM.

011-kafka-canary.md
ppatierno: I added a sentence in the "Topic Configuration" section but without numbers about segment size and retention (in bytes). It will come with development and more testing.

tombentley (approved):

ppatierno (approved):

sknot-rh (approved):

samuel-hawker (approved):

david-martin (approved):
Clear and concise.
Approving with a question.
Will the canary be stateless, and capable of rebuilding it's state on restart from reading broker config from the kafka cluster?

The only potential state I see is any counters for exposed metrics, which is fine.
They can be passed into a rate fn in prometheus to avoid resets.

ppatierno:
About metrics, I agree ... I would just rely on Prometheus for rates.
Regarding any other state, I was thinking of scenarios where for example the producer sends a message with id = 36 (just a number) then the canary tool crashes. Would the id restart from 0? Should we discard the latency metric in this case because too long due to a crash? I think they are questions we can address during the development and iterate on it to improve the behavior.

PR #16, opened 2020-10-29
Move Strimzi container images to Quay.io

scholzj:
Given the newly introduced limitations of Docker Hub, I think we need to look for a new container registry. This proposal suggests to start using Quay.io. The rejected alternatives section list the other registries I considered and rejected. If anyone has something better, feel free to let us know in comments.


samuel-hawker (approved):
Sounds like a good idea to me!

For the implementation of this, i assume this would just update references to how we tag, push and pull images in the scripts?
If so we should definitely variabal-ise this to make changing the registry again in the future easier.

scholzj:
@samuel-hawker I think lot of this already is in place in the form of environment variables etc. They just default to `docker.io` currently. So I do not think this should be too much work.

scholzj:
FYI: I did some analysis of the images and clarified which versions should be copied. The proposal now suggests to copy versions 0.12.0 and later since in 0.12.0 we did significant changes to the image structure (number of images and names). This should also help to clear common confusions among users who see the old images on Docker Hub.

Frawless (approved):
LGTM, I was thinking about pushing PR images to some strimzi-ci org so we can do it with quay without digging old credentials from somewhere.

013-move-docker-images-to-quay.io.md
```diff
@@ -0,0 +1,43 @@
+# Move Container Images to Quay.io
+
+Docker is introducing from November 2020 service limits for container images stored in Docker Hub.
+Free accounts will have following rate limits:
+* 100 pulls for anonymous users per 6 hours
+* 200 pulls for authenticated users (on the free plan) per 6 hours
+
+In addition, images will be kept only for 6 months.
```
tombentley, line 8: I don't think this is correct. I think it's only when an image tag has not been pulled or pushed for 6 months that it will be deleted. So long as someone (potentially) pulled images every few months the images would remain indefinitely.

tombentley (approved):
I'm not opposed to this. I wonder if we could continue pushing _releases_ to dockerhub for the time being to give users (with CI pipelines of their own perhaps) notice of this change, rather than dropping it on people in 0.21.

013-move-docker-images-to-quay.io.md
scholzj: I guess that is right, but I'm not sure this makes it any better - just less predictable.

scholzj:
> I wonder if we could continue pushing releases to dockerhub for the time being to give users (with CI pipelines of their own perhaps) notice of this change, rather than dropping it on people in 0.21.

@tombentley I'm not sure how exactly you mean it. Pushing to both repos would mean more work on the CI pipelines but also more testing. User install Strimzi from the YAML files, Helm Charts or from OperatorHub. So they will not need to do anything manually -> they will just install the new files with the new images from new repo. And if someone mirrors the images into some private repo / proxy, they would anyway need to do it manually.

ppatierno (approved):

PR #19, opened 2020-11-13
Add Kafka topic encryption proposal

chris-giblin:


015-kafka-topic-encryption.md
```diff
@@ -0,0 +1,133 @@
...
+This approach has well-known shortcomings, most notably the ability of anyone with appropriate file system permissions, such as system administrators, to read data.  Security best practices and compliance standards
+explicitly call for application- or database-level encryption to protect sensitive information against such exposures.
+This document proposes a technical solution for providing upper-layer encryption-at-rest for Kafka systems.
+
+
+## Proposal
+
+An implementation of topic encryption is proposed whereby the message stream between client and broker is
+intercepted. Incoming data messages (Produce requests) are inspected to determine whether their payload
+should be encrypted according to a policy.  If so, the data portions are encrypted and the modified
```
ppatierno, line 33: what "data portions" means here? just the kafka message payload? what if sensible information are transferred using Kafka headers in the message?

chris-giblin: By "data portions" we are referring to the map of records, keyed by topic, in a Kafka Produce and Fetch messages which are inspected. We have not considered encrypting record headers but this is an interesting suggestion which could be easily accommodated.

As the entire message is navigated to the record level, it would be natural apply encryption to record headers as well. This would require a change to the encryption policy to indicate header names or topic/header combinations to encrypt. The logic change to support header encryption would be minimal.

chris-giblin: fyi @SeanRooooney @zrlvep

015-kafka-topic-encryption.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Key Service
+The Key Service is used in the Encryption Module to handle interactions with a key management system and to perform optional primitives such as those related to envelope encryption. The interface is generic in order to accommodate various key management implementations.
+
+#### Policy Service
+Encrypter-decrypters require a means to lookup encryption requirements and parameters for a topic. As a corresponding standard does not exist, the core component will interact with a generic internal interface for retrieving policy on how a topic is, or not is, to be encrypted. Implementers can develop and configure their own implementation of the policy service, be it a remote microservice, a local pre-configured cache, a cluster registry, etc.
+
+For each topic to be encrypted, a policy exists detailing:
+
+- topic name
```
tombentley, line 63: I can see use cases where it would be convenient to have a single policy governing multiple topics, e.g. all topics beginning with a given prefix should use the same policy.

tombentley (commented):
Thanks for the proposal!

I'd like to see some discussion about what happens when it is necessary to change the cipher or encryption key for a given topic. In particular I can see it being problematic when:

* The key is encrypted (which right now is always the case), and
* The topic is compacted

I.e.

1. Client sends message with key `mk` to compacted topic `t`
2. Proxy encrypts with encryption key `k_1` and forwards to broker
3. Encryption key is changed to `k_2`.
4. Client sends message with key `mk` and null value to compacted topic `t`
5. Proxy encrypts with encryption key `k_2` and forwards to broker, which persists
6. Messages is never deleted/garbage collected because `enc(mk, k1) != enc(mk, k2)`

There also also interesting race conditions if the producer and consumer were accessing the cluster via different proxies (which might well be the case with a sidecar deployment).

To be clear, I think it would be fine to say "We don't support this in the initial version", but it should be explicit what's going to be implemented under this proposal, and whether these problems might be addressed eventually, or not at all.

015-kafka-topic-encryption.md
chris-giblin: I Kafka record is comprised of Key and Value fields. Encryption is applied to the Value field only. The message key therefore is the same across encryptions.  

As an example, the Kafka javadoc for ProducerRecord defines Key and Value as distinct fields:
https://kafka.apache.org/10/javadoc/org/apache/kafka/clients/producer/ProducerRecord.html

chris-giblin: We have considered supporting a regular expression for topic names. Because regular expressions can result in multiple matches and thus can be a bit unpredictable, we initially have taken a conservative stance that either policy is explicitly specified for each topic or a global wildcard (*) policy is applied.  We are definitely open to discussion. We have also considered the idea of a default policy.  Supporting multiple topics does seem to make sense either as a comma-separated list of topic names or the slightly risky regex.

chris-giblin:
Regarding message keys, encryption is applied only to the record value, not the record key. As a result, there is no impact on compaction since message keys are not encrypted.
As an example of the record structure, the javadoc for the Kafka Produce message demonstrates key and value fields being discreet and distinct:
https://kafka.apache.org/10/javadoc/org/apache/kafka/clients/producer/ProducerRecord.html

tombentley:
@chris-giblin thanks for the clarification, I didn't see that spelled out in the propsal, did I miss it?

015-kafka-topic-encryption.md
chris-giblin: We leave open the full functionality of the policy service. The matching expression does not have to be an exact match or a regular expression but rather also some sort of query expression. Same caveat with regular expressions apply as mentioned above apply - a topic may match more than one expression and therefore more than one policy.  The convenience of regular and query expressions must be weighed with potential ambiguities which may arise.

chris-giblin:
You are right @tombentley that is not spelled out in the proposal. Will gladly add that.

chris-giblin:
@tombentley it would be good to elaborate on potential race conditions. Is there an example we could begin with?

tombentley:
@chris-giblin I was thinking of things like:

1. Producer P1 sends message M1 via sidecar proxy using encryption key K1
2. Encryption key changes to K2
3. Producer P1 sends message M2 via sidecar proxy using encryption key K2
4. Producer P2 sends message M3 via its own sidecar proxy using encryption key K1, because for whatever reason it doesn't know about K2 yet.
5. Consumer sees `enc(M1,K1), enc(M2,K2), enc(M3,K1)`

But I guess this is pretty much what you already expected, and consumer can still obtain old keys from the key management system unless they were purposely deleted there. Which makes me wonder how the proxy should handle (at the protocol level) the case where it really doesn't and can't know the key.

chris-giblin:
@tombentley there many subtleties and options with key rotation, in part arising from Kafka being an immutable store. Starting with a simple case first, let's assume there is one proxy instance,  a sidecar sharing a pod with a kafka broker. It retrieves an AES key from a KMS such as Vault. We change keys and are now encrypting with k2. The topic in question now has a mix of messages in storage, encrypted by k1 or k2.  When decrypting, the proxy needs information on which key was used to encrypt it (we use the record headers to pass this metadata). Using this information, it retrieves the appropriate key, decrypts and forwards to the client.

Now onto the scenario you have described with multiple proxies in play. They would need to be linked into the same key service and be informed when a key changes. Agreed, it is possible for a variety of reasons why that not all proxy instances have switched at the exact same time. However, this does not pose a problem because, as you suggest,  all proxies rely on the encryption metadata in the message to retrieve the right key to use. 

In the event the proxy cannot retrieve the key, be it for technical reasons such as network outage or an inconsistency, it should return an error message. This would mean supporting a new set of error conditions.

ppatierno:
> When decrypting, the proxy needs information on which key was used to encrypt it (we use the record headers to pass this metadata).

@chris-giblin If we are using record headers for this info, it means that we cannot encrypt headers of a Kafka message as I raised before.

chris-giblin:
@ppatierno Each record header is a name/value pair. We already address  potential collisions with the header names we use and those used by clients. Because we can discern between the two, we can control that “our” headers are not encrypted.  

A policy could contain an optional list of  names of headers to encrypt. It conceivably could contain a wildcard, '*', as well, indicating encrypt all headers. All except "ours" would be encrypted.

I agree, this should be clarified in the proposal. We will commit an update in the next 24 hours.

chris-giblin:
@ppatierno @tombentley I have committed updates to the proposal which clarify much of what we have discussed so far.

015-kafka-topic-encryption.md
```diff
@@ -0,0 +1,169 @@
...
+
+An example is now described to illuminate the essential details of the message encryption process.
+The encryption module is embedded in a proxy and the proxy, as an intermediary, terminates  connections between clients and the broker, placing the proxy in the position of intercepting all Kafka traffic. The message handling component examines all messages to identify the apikey (i.e., the Kafka message type). All message types besides Produce and Fetch are  immediately forwarded without any processing. With the help of the message handling library, Produce requests are deserialized and inspected more deeply to identify the topic for which records are destined.
```
tombentley, line 100: > All message types besides Produce and Fetch are immediately forwarded without any processing.

You probably want to do something with `ApiVersions` requests. They're used to inform the client about the protocol versions (for each api key) supported by the broker. So it would be possible for a new broker version to add a new version of a `Produce` request which the proxy didn't understand, and by just sending the `ApiVersions` response to the client unchanged a client which also supported that version of `Produce` would use it. So for the `Produce` and `Fetch` APIs the max version in the `ApiVersions` response should the `min(brokerMaxVersion, proxyMaxVersion)`, likewise the min versions should be the `max(brokerMinVersion, proxyMinVersion)`. If you're using Envoy it's possible that it already handles this, but it's worth checking.

015-kafka-topic-encryption.md
```diff
@@ -0,0 +1,169 @@
...
+A proxy can be deployed as a free-standing process or as a sidecar in a Kubernetes pod. 
+In both cases, the proxy needs to intercept and de-serialize Kafka messages, encrypt/decrypt the relevant records, 
+and reserialize them into a new message. The overhead of a proxy potentially has an effect on performance.
+It also means that the proxy version must always be in phase with that of the broker.
+A proxy-based solution however has the advantage that both Kafka client and broker are unaware
+of the proxy and do not require any modification or configuration to support encryption at rest.
+
+Whether the proxy runs under the control of the client or the broker is considered a configuration
+issue, i.e. exactly the same proxy would be used in both cases, but when under the control of the
+client it would be their responsibility to ensure that producers/consumers are passing through a proxy with the same configuration.
```
tombentley, line 122: Client identity might be a problem here. When the proxy is just proxying a single client it can forward the `SaslHandshake` and `SaslAuthenticate` messages, so SASL authentication (configure on the client) should just work. Likewise TLS authentication would work fine, because the proxy can be configured with the TLS keystore and truststore which the client would have used. But it gets more complicated when more than a single client it using the same proxy (and that could even happen in the sidecar deployment, if there were multiple client instances being used in the main container of the pod). You probably don't want to worry about this now, which is fine, but I think we should be explicit about that.

Similarly there can be issues with client quotas if the proxy is serving multiple clients but using a single identity for its connection to the broker.

tombentley (commented):
Thanks for the updates. I left a few more comments which are worth thinking about, even if some of them don't have an easy solution.

015-kafka-topic-encryption.md
```diff
@@ -0,0 +1,169 @@
+<!-- This template is provided as an example with sections you may wish to comment on with respect to your proposal. Add or remove sections as required to best articulate the proposal. -->
+
+# Proxy-Based Kafka Per-Topic Encryption
+
+The goal of this proposal is to provide topic-level encryption-at-rest for Kafka such that distinct encryption keys can be used for different topics. 
+A core library is proposed which is deployed into a proxy. Proxies can flexibly be deployed in a variety of topologies and require no changes to Kafka clients and brokers. All encryption/decryption takes place at the proxy. An unencrypted topic's messages pass through the proxy without being changed and without
+paying any performance penalty.
```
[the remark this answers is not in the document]
chris-giblin: Agree. I will include more precise statement in the next update.

015-kafka-topic-encryption.md
chris-giblin: Agree, good point. Will add to the next update.

015-kafka-topic-encryption.md
chris-giblin: You raise important integration issues which will have to be addressed. Indeed, they may vary depending on how the proxy is deployed.

015-kafka-topic-encryption.md
```diff
@@ -0,0 +1,169 @@
...
+### Embedding Topic Encryption in a Proxy
+A proxy can be deployed as a free-standing process or as a sidecar in a Kubernetes pod. 
+In both cases, the proxy needs to intercept and de-serialize Kafka messages, encrypt/decrypt the relevant records, 
+and reserialize them into a new message. The overhead of a proxy potentially has an effect on performance.
+It also means that the proxy version must always be in phase with that of the broker.
+A proxy-based solution however has the advantage that both Kafka client and broker are unaware
+of the proxy and do not require any modification or configuration to support encryption at rest.
+
+Whether the proxy runs under the control of the client or the broker is considered a configuration
+issue, i.e. exactly the same proxy would be used in both cases, but when under the control of the
```
scholzj, line 121: I think there are some notable difference depending on where it is running which should be covered here:

* When running as a sidecar inside the broker pod, the proxy has to listen to a single port which will only get the connections for this particular broker. This can be also easily orchestrated and this port / address might be used as a advertised host in the broker configuration. And thus the proxy really needs to only deal with the _Produce_ and _Fetch_ requests.

* But on the client side, the advertised hosts metadata make it a way more complicated. They would essentially lead the clients to bypass the proxy and connect to the brokers directly to produce / fetch messages. So I think this would need some special handling, or? I guess there will be two options. One the proxy can just open more different ports (one per broker) and modify the advertised hosts in metadata. Or I guess it can approach it Istio-style and just high-jack all connections into the proxy. But that would need to be able to filter out non-Kafka connections and possibly also connections to other Kafka clusters which should not be encrypted.

Or did I missed some option?

scholzj (commented):
I left some more comments. But this is already pretty good proposal. Thanks for opening it.

015-kafka-topic-encryption.md
scholzj: I agree that things like regexes or prefixes would need to be handled carefully since they open a big space for misconfiguration.

015-kafka-topic-encryption.md
```diff
@@ -0,0 +1,169 @@
...
+]
+```
+
+In this example, two topics will be encrypted with two different keys residing in the same KMS.
+
+### Message Handling
+Message handling is concerned with facilities for intercepting messages, deserializing and inspecting them, invoking encryption as accorded by policy, modifying messages, and finally serializing and forwarding messages onward to Kafka brokers and clients. To accomplish encryption and decryption of data payloads, only Kafka *Produce requests* and *Fetch responses* need to be examined and potentially modified. High quality open source libraries exist already for implementing message introspection and serialization/deseriaization, e.g.,
+https://github.com/Shopify/sarama.
+
+### Message Encryption
```
scholzj, line 96: Maybe a stupid question, but I do not have much experience with encryption other then TLS connections. Is there any impact on this from things like:
* Expiration of the encryption keys?
* Keys needing to be replaced because they were leaked / compromised?

Or is this not something we need to address here?

015-kafka-topic-encryption.md
```diff
@@ -0,0 +1,169 @@
...
+
+The encryption metadata is currently stored in Kafka record headers. Record headers are optional name value/pairs stored in an array associated with the record. Record header names are not required to be unique, thus name "collisions" are possible. Ordering and versioning are used to avoid problems arising from names used for metadata which coincidentally equal names chosen by a Kafka client. Ordering means that the record header is rewritten such that the encryption-related headers
+occur first in the array. The first header always contains a version from which the exact
+metadata field names can be derived. The decryption algorithm therefore can discern between
+headers added by the encryption module and those added by the client. The version field
+is enables future migration and backwards compatibility should changes in semantics occur such as modified message formats or encryption options.
```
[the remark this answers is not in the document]
zrlvep: Yes, added headers will be removed before forwarding them to the consumers.

015-kafka-topic-encryption.md
zrlvep: Partially. The idea is to let implementations of the key service interface deal with such requirements. I.e. implement envelope encryption to be able to expire root keys or support multiple different keys per topic.

I think one thing we need to mention is, that re-encryption of topics is not supported by this proposal. Meaning, if a key is compromised, data encrypted with it is compromised too and can only be mitigated by copying the topic and using a different encryption key.

scholzj: Yeah, I think it would be great to mention this to make it more clear.

015-kafka-topic-encryption.md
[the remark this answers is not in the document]
scholzj: Ok, great. Maybe you could add it to the proposal to make it more clear.

015-kafka-topic-encryption.md
chris-giblin: @scholzj No, you did not miss an option and yes this requires special handling. Kafka clients need a broker list and broker metadata which connects them through one or more proxies. Meanwhile, the brokers  need to interact with each other and answer metadata requests to their clients. This can be approached from different angles. We have viewed this as an integration issue and outside the scope of the core encryption proposal. Although we have deployed an implementation, I am unsure whether we are at liberty to provide details. I will clarify. In any case, this is an area where we very much welcome the expertise of the strimzi community. In the meantime, we will add a section to the proposal to call this out and possibly outline potential solutions.

scholzj: Sounds like a plan. I'm fine if there are not too many details ... but as you said, might be just worth mention either that it is out of scope or just outline the possible solutions and leave the details for implementation.

chris-giblin:
@tombentley @scholzj  We have updated the proposal with information on the following topics:

- Key rotation
- Mention removal of encryption headers from responses
- Correct "no performance penalty"
- Mention ApiVersions
- SASL implications
- Coordinating proxy and broker with advertising property
- AES-GCM for initial version

scholzj (approved):

tombentley (commented):
One big question I have is how the policy is configured. It is passed into the process, or could it be done dynamically (even if that's not implemented to start with)? For example it could be very painful (depending on how the proxy is deployed) if adding a new topic requires restarts. 

I'd also like to know whether/which a KMS interface implementation is part of the proposal. I.e. to what extent are the batteries included with the implementation of this proposal. 

Another question I had was what if you needed to migrate from KMS A to KMS B. This doesn't seem to be possible at all with the current policy. I'm not saying that this needs to be implemented under the proposal, but it might be worth considering how it could be accommodated in the future, or explicitly rule it out. 

The "credential" in the example policy: Is that sensitive data? I'm not completely clear on what `kms_instance` is either, though it's likely just a reflection of my ignorance. But it would be really helpful to readers to describe each of the properties in the policy in a little more detail.

015-kafka-topic-encryption.md
```diff
@@ -0,0 +1,194 @@
...
+explicitly call for application- or database-level encryption to protect sensitive information against such exposures.
+This document proposes a technical solution for providing upper-layer encryption-at-rest for Kafka systems.
+
+## Proposal
+
+### Overview
+An implementation of topic encryption is proposed whereby the message stream between client and broker is intercepted. Incoming data messages (Produce requests) are inspected to determine whether their payload
+should be encrypted according to a policy. If so, the data portions are encrypted and the modified
+message is forwarded to the broker. As a result, the topic data is stored by the broker in encrypted form.
+On the reverse direction, encrypted responses (responses to Fetch requests) are decrypted prior to being sent to clients.
```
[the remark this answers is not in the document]
chris-giblin: @tombentley These are all important points you raise. 

- Agree, policy must be dynamically updated to, at the very least, accommodate topic creation and deletion. This should be done without restarting.  
- The proposal includes the KMS interface and at least one implementation of that interface initially, IBM Key Protect. We have started work with IBM Secrets Manager which is Vault. Very glad to discuss whether Vault would be a preferred initial KMS implementation.
- Migrating a key to another KMS can be supported however could require a restart, at least initially. The new KMS must of course be of a supported type. As long as the key material remains the same, migration is a configuration change.
- Yes, the credential is very sensitive data. It is used to access the KMS. Therefore the entire policy, in its current form,  must be treated as a secret.
- kms_instance is an optional parameter. To access IBM Key Protect, a URL and an instance ID are required. If a KMS does not require this parameter (e.g., Vault) it can be dropped. Perhaps there are other KMS systems out there which require an additional parameter in addition to a URL. kms_instance is a placeholder for such a param. 

I will provide an update with a brief description of the policy parameters.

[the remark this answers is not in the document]
tombentley: Thanks!

I think there's been most user interest (i.e. for TLS-related stuff) around Vault (@scholzj correct me if I'm wrong), so it would be great to have support for that.

I guess the KMS migration via restart should be fine (since we don't expect it to be a common thing). I hadn't realised that the way to do this would be to just export the old keys etc from the one to the other.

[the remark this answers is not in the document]
chris-giblin: KMS semantics should be the same as well when migrating. If AES data encryption keys are stored in the KMS, then migrating should be straightforward. However if an encryption model is employed which involves some sort of computation on the part of the KMS (e.g., envelope encryption with wrap/unwrap primitives) then one KMS is  presumably not a drop-in replacement since vendors may implement functions (e.g., envelope encryption) differently. The topic would then contain cryptographic artifacts from both KMS systems and could not be supported.  So migration needs to be qualified.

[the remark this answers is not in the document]
chris-giblin: @tombentley fyi I have updated the proposal to include clarification on policy updates.

SeanRooooney:
@tombentley @ppatierno is there anything more you need from us to complete the review ?

ppatierno:
@SeanRooooney @chris-giblin I think this PR needs a rebase because it has 16 files changed coming from other proposals (i.e. admin server UI, canary and so on).

scholzj:
I can do rebase while merging ... Do it for most proposals.

chris-giblin:
> I can do rebase while merging ... Do it for most proposals.

Thanks, much appreciated!

015-kafka-topic-encryption.md
```diff
@@ -0,0 +1,216 @@
...
+
+As defined in the encryption policy, each topic can be encrypted by a different key,
+allowing brokers to store a mix of encrypted and unencrypted data, where
+data owners can manage the keys to their topics.
+Keys ideally are stored in a key management system with access policies and logging.
```
tombentley, line 37:
```suggestion
Keys will be stored in an external key management system with access policies and logging. 
```

015-kafka-topic-encryption.md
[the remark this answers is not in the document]
tombentley: @chris-giblin could you fix the typo?

015-kafka-topic-encryption.md
```diff
@@ -0,0 +1,216 @@
...
+
+#### Encrypter-Decrypter
+The Encrypter-Decrypter sub-component performs the actual encryption and decryption of messages, implementing a particular encryption algorithm and making use of the KMS service to obtain topic keys.
+
+#### Key Service
+The Key Service is used in the Encryption Module to handle interactions with a key management system and to perform optional primitives such as those related to envelope encryption. The interface is generic in order to accommodate various key management implementations.
+
+#### Policy Service
+Encrypter-decrypters require a means to lookup encryption requirements and parameters for a topic. As a corresponding standard does not exist, the core component will interact with a generic internal interface for retrieving policy on how a topic is, or not is, to be encrypted. Implementers can develop and configure their own implementation of the policy service, be it a remote microservice, a local pre-configured cache, a cluster registry, etc.
```
tombentley, line 57:
```suggestion
Encrypter-decrypters require a means to lookup encryption requirements and parameters for a topic. As a corresponding standard does not exist, the core component will interact with a generic internal interface for retrieving policy on how a topic is, or is not, to be encrypted. Implementers can develop and configure their own implementation of the policy service, be it a remote microservice, a local pre-configured cache, a cluster registry, etc.
```

tombentley (commented):
At this point my main reservation is the lack of transparent key rotation. On the one hand it's often a compliance requirement, on the other hand what's being proposed requires the topic to either be deleted and recreated with re-encrypted data (incurring downtime for producers and consumers), or for producers and consumers to agree to switch to using a new topic in a coordinated way (which could be difficult if they're managed by different teams, for example), and even then could be non-trivial. OTOH, I can understand that it's not something that should be in scope right now.

ppatierno (approved):
Last nit but LGTM.

chris-giblin:
@tombentley and @ppatierno : thank you for pointing out the corrections and improvements. Done.

SeanRooooney:
@tombentley is there anything else you need from us for approval ?

tombentley (approved):
Thanks!

Frawless (approved):

SeanRooooney:
@scholzj can you rebase or should we ?

scholzj:
I can do the rebase and rename the file etc. But I'm afraid I will need you to fix the DCO signoff :-/: https://github.com/strimzi/proposals/pull/19/checks?check_run_id=1704582165

chris-giblin:
> I can do the rebase and rename the file etc. But I'm afraid I will need you to fix the DCO signoff :-/: https://github.com/strimzi/proposals/pull/19/checks?check_run_id=1704582165

@scholzj I am having difficulty with the sign-off. As per https://github.com/strimzi/proposals/runs/1823797785 I have:
`git commit --amend --no-edit --signoff`
`git push --force-with-lease origin master`

Opening the most recent merges, one sees `Signed-off-by: C. Giblin <christopher.giblin@gmail.com>` yet the PR has a number commits with red crosses. I also tried signing off specific commits with:

`git commit --amend -s -c 8103...`

All this does not seem to result in the required sign-off(s). 

Any suggestions how to resolve this?

scholzj:
@chris-giblin I set it to pass, renamed the file to match the latest proposal and merged it. So I guess next we should create the repository.

SeanRooooney:
@scholzj are you going to create the repo ?

scholzj:
@SeanRooooney Sorry, I was off ... I have now created https://github.com/strimzi/topic-encryption

PR #20, opened 2020-11-20
Cgi encdetails

chris-giblin:


chris-giblin:
This was a mistaken PR. Please ignore

PR #21, opened 2020-11-23
Kafka Connect Build proposal

scholzj:
This PR proposes declarative build of Kafka Connect container images with additional connector plugins and their use in the Kafka Connect clusters. This should make it easier for users to use Connect with additional connector plugins and also make it easier to build things such as connector catalogs on the Strimzi website / in the Strimzi UI.

I have also a working PoC of this proposal: https://youtu.be/sW4GpJ7SU-4


016-kafka-connect-build.md
```diff
@@ -0,0 +1,127 @@
...
+    config.storage.topic: connect-cluster-configs
+    status.storage.topic: connect-cluster-status
+```
+
+When the `build` section is specified in the `KafkaConnect` resource, Strimzi will generate `Dockerfile` which will correspond to the configured connectors and their artifacts.
+It will build the container image from this `Dockerfile` in a separate pod and push it to the configured output (Docker registry / ImageStream).
+It will take the digest of this image and configure the Kafka Connect cluster to use this image.
+
+The build will use two different technologies:
+* [Kaniko](https://github.com/GoogleContainerTools/kaniko) will be used on Kubernetes
```
ppatierno, line 101: just to be clear, it's on user shoulders to run this Kaniko on the same Kubernetes cluster, right? It's not Strimzi going to deploy it.

scholzj: Kaniko is basically just an executable / container image which can build Dockerfiles. Strimzi will create a pod which uses it, pass the Dockerfile to it and it builds the new image and pushes it to the registry. There is nothing to be deployed by the users since it is not some complicated framework which needs to be installed upfront.

016-kafka-connect-build.md
```diff
@@ -0,0 +1,127 @@
...
+This section will allow users to configure:
+* connector plugins
+* build output
+
+The connector plugins section will be a list where one or more connectors can be defined.
+Each connector consists of a name and list of artifacts which should be downloaded.
+The artifacts can be of different types.
+For example:
+* `jar` for a directly download of Java JARs
+* `tgz` or `zip` to _download and unpack_ the artifacts
```
tombentley, line 43: What assumptions/expectations do we make about the contents? We search it and look for jars?

016-kafka-connect-build.md
```diff
@@ -0,0 +1,127 @@
...
+* `maven` to download JARs from Maven based on Maven coordinates
```
tombentley, line 44: Very keen to see this one. It should be doable to have it fetch the pom and pull in all the runtime dependencies, so it should be much simpler to configure in the common case.

016-kafka-connect-build.md
```diff
@@ -0,0 +1,127 @@
...
+This proposal is fully backwards-compatible.
+When users don't specify the `build` section inside the `KafkaConnect` CR, the build will be never activated and used.
+So existing users who do not the new build will not be affected in any way.
+They will be also still able to use the old manually built container images without any change.
+
+## Rejected alternatives
+
+There are several ways how to build container images.
+In addition to Kaniko and OpenShift Builds, I also looked at Buildah.
+But Kaniko and OpenShift Builds worked better (read as: _I didn't got Buildah to work_).
```
tombentley, line 127: I think what's important is that our API isn't bound to any particular tool, so we could always switch to something else at a later date.

tombentley (commented):
A couple of questions/comments, but overall this looks great. 

I thought a bit about whether it made any sense to put the `plugins` in the `KafkaConnector`. It would be more analogous to Kube itself that way (where it's not necessary for someone to configure a node with a list of images it can use). I can see some pros to that approach, but also significant cons:

* KafkaConnect doesn't have enough ability to isolate connectors for it to really work, I think. 
* Updating a single connector would cause the redeployment (and consequent loss of availability) of the connect cluster. That's true of this proposal too, of course, but it's more easily brought under the control of an administrator.

So I think what's in this proposal would work fine for most use cases, and be simpler to implement.

016-kafka-connect-build.md
scholzj: I did not implement this one in my prototype. It will be obviously more complicated then the others. But yes, it could be useful in some cases.

016-kafka-connect-build.md
scholzj: The suggested user API for the `KafkaConnect` is in the proposal. The internal code doing the builds will always depend on the exact tool being used. But it currently just builds a `Dockerfile` generated by the operator. So in theory any tool which can do that should do and it should be possible to implement it.

scholzj:
@ppatierno @tombentley FYI: I tried to add the points from your comments to the proposal to make it more clear there.

016-kafka-connect-build.md
```diff
@@ -0,0 +1,137 @@
...
+This section will allow users to configure:
+* connector plugins
+* build output
+
+The connector plugins section will be a list where one or more connectors can be defined.
+Each connector consists of a name and list of artifacts which should be downloaded.
+The artifacts can be of different types.
+For example:
+* `jar` for a directly download of Java JARs. 
+* `tgz` or `zip` to _download and unpack_ the artifacts.
```
tombentley, line 43: I guess that perhaps tar.gz could include non-regular files/directories, such as symlinks. I guess we'd need a way to detect those and deal with them, since they shouldn't be allowed.

016-kafka-connect-build.md
```diff
@@ -0,0 +1,137 @@
...
+* [Kaniko](https://github.com/GoogleContainerTools/kaniko) will be used on Kubernetes
+* OpenShift Builds will be used on OKD / OpenShift 
+
+Kaniko is a daemon-less container builder which does not require any special privileges and runs in user-space.
+Kaniko is just a container to which we pass the generated `Dockerfile` and let it build the container and push it to the registry.
+It does not require any special installation - neither from the user nor from Strimzi it self. 
+We just use the container.
+
+OpenShift Builds are part of the OKD / OpenShift Kubernetes platform and also do not require any special privileges.
+So it should work for most users across wide variate of environments.
```
tombentley, line 113:
```suggestion
So it should work for most users across wide variety of environments.
```

016-kafka-connect-build.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Affected/not affected projects
+
+The existing `KafkaConnectS2I` resource is not affected by it.
+But it should be deprecated and later removed.
```
tombentley, line 118: Using should here is ambiguous. Unless this proposal is deprecating it then it's best not to mention it. If it is proposing deprecation then it should say so unambiguously: "KafkaConnectS2I will be deprecated."

tombentley (approved):
A few nits, but I'm basically happy enough with this.

016-kafka-connect-build.md
scholzj: Updated.

016-kafka-connect-build.md
scholzj: I see this proposal as superior, so I think we should deprecate the `KafkaconnectS2I` (and remove later of course) it once it is implemented. I updated the proposal accordingly.

016-kafka-connect-build.md
```diff
@@ -0,0 +1,138 @@
...
+This might be also enhanced to support different Maven repositories etc.
+* and possibly others if required
+
+The second part - `output` - will configure how the newly built image will be handled.
+It will support two types:
+* `docker` to push the image to any Docker compatible container repository.
+* `imagestream` to push the image to OpenShift ImageStream (supported only on OpenShift)
+
+Strimzi will not run its own container repository.
+So users of the `docker` type build will need to provide credentials to their own Docker compatible container registry.
```
ppatierno, line 56: I know it's in the following example but I think it's better to make it clear that it happens through a Secret that has to be referenced in the output section.

016-kafka-connect-build.md
```diff
@@ -0,0 +1,138 @@
...
+This does not have to run within the same Kubernetes cluster - it can be also SaaS registry such as Docker Hub or Quay.io.
+
+Following is an example of the Kafka Connect CR:
```
ppatierno, line 59: KafkaConnect

016-kafka-connect-build.md
```diff
@@ -0,0 +1,138 @@
...
+    strimzi.io/use-connector-resources: "true"
+spec:
+  version: 2.6.0
+  replicas: 1
+  bootstrapServers: my-cluster-kafka-bootstrap:9093
+  build:
+    output:
+      type: docker
+      image: docker.io/user/image:tag
+      pushSecret: dockerhub-credentaials
```
ppatierno, line 76: what will be the structure in the case of the imagestream?

ppatierno (approved):
Just a couple of things but overall lgtm.

scholzj:
@Frawless @samuel-hawker Any comments from you?

Frawless (approved):
Cool video example :)

samuel-hawker (approved):
LGTM

PR #23, opened 2020-12-09
Exploratory testing tool

see-quick:
Description in the `015-exploratory-test-module.md`

### DIFF
--- a/015-exploratory-test-module.md
+++ b/015-exploratory-test-module.md
@@ -1,12 +1,13 @@
 # Exploratory test repository
 
-The exploratory test contains dialogue application, which gives you the possibility to create your test scenarios. 
+The exploratory test will contain two dialogue applications, which gives you the possibility to create your test scenarios. 
 
 ## Motivation
 
 Every release, we do exploratory testing and test Strimzi. 
-This tool also provides `transcript` for all scenarios that the user tried. In other words, the application stores success and wrong scenarios. These scenarios end the end of exploration are provided to a user. User then can 
-share with developers or testers reproducible procedure with all YAMLs executed by the user. 
+This tool also provides `transcript` for all scenarios that the user tried. In other words, the application stores success and wrong scenarios. 
+These scenarios end the end of exploration are provided to a user. 
+User then can share with developers or testers reproducible procedure with all YAMLs executed by the user. 
 
 For brevity, here is one of the scenarios:
 
@@ -28,8 +29,9 @@ The module provides two dialogue applications, which has a lot in common:
 1. BaselineExploratory - dialogue application, which executes the predefined scenario 
 2. CreativeExploratory - dialogue application, which executes input YAMLs provided by the user
 
-Moreover, applications remember all your choices, which means that if the user sees the problem and for instance `Kafka` is not working properly then on the question `Is [Kafka] running? [y/n]` user will type `n`, which will store this path to `bug` paths. 
- On the other, it has also the list of `success` paths.
+Moreover, applications remember all your choices, which means that if the user sees the problem and for instance `Kafka` 
+is not working properly then on the question `Is [Kafka] running? [y/n]` user will type `n`, which will store this path to `bug` paths. 
+On the other, it has also the list of `success` paths.
 
 #### BaselineExploratory 
 
@@ -99,8 +101,8 @@ In the last phase of testing user will have a transcript of all his decisions wi
     ...
 ```
 
-As you can see it is the list of commands, which is needed to do to reproduce the same behaviour that the user had. In the plain text
-it will look like this:
+As you can see it is the list of commands, which is needed to do to reproduce the same behaviour that the user had. 
+In the plain text it will look like this:
 
 ```
 ======
--- a/015-exploratory-test-module.md
+++ b/015-exploratory-test-module.md
@@ -1,62 +1,60 @@
 # Exploratory test module
 
-Exploratory test module contains dialog application, which gives you possibility to create your own test scenarios. 
+The exploratory test contains dialogue application, which gives you the possibility to create your test scenarios. 
 
 ## Motivation
 
-Every release, we do exploratory testing. We test Strimzi but with this tool it would be very easy to do such a thing. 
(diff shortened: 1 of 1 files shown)


scholzj (commented):
I personally do not think you can do exploratory testing with a tool like this because it sets too many boundaries for you. But if everyone else agrees I'm not going to block it. So I basically vote 0.

015-exploratory-test-module.md
```diff
@@ -0,0 +1,163 @@
+# Exploratory test repository
+
+The exploratory test will contain two dialogue applications, which gives you the possibility to create your test scenarios. 
+
+## Motivation
+
+Every release, we do exploratory testing and test Strimzi. 
+This tool also provides `transcript` for all scenarios that the user tried. In other words, the application stores success and wrong scenarios.
```
Frawless, line 8: It sounds a little bit weird....the `tool also do something` even when don't know what it's doing in basic function :)

015-exploratory-test-module.md
```diff
@@ -0,0 +1,163 @@
...
+8. Provide YAMLs for installation of Prometheus
+9. Provide YAMLs for installation of Grafana
+
+And many more. This is just one scenario, which anyone can create. 
+Note, that in every step you have to manually check if a specific component is running. 
+For instance, if you choose to 'Provide Kafka CR YAML' then you need to list Kafka CR, Pods and verify that everything works.
+
+### Idea
+
+The module provides two dialogue applications, which has a lot in common:
```
Frawless, line 28: Why two applications? Wouldn't be enough only one with different options?

015-exploratory-test-module.md
```diff
@@ -0,0 +1,163 @@
...
+3. application applies first pre-defined YAML
+4. (interaction with app) yes/no (is component running?)
+5. application applies next pre-defined YAML
+6. (interaction with app) yes/no (is component running?)
+7. if there is another component to apply to go to `5th step` else go to next step 
+8. transcript with wrong and success paths (reporting) = user end the end of testing will have a report about his work
+
+#### CreativeExploratory
+
+Creative exploration consists of several steps:
```
Frawless, line 50: Maybe you can add `quit` option as well

015-exploratory-test-module.md
```diff
@@ -0,0 +1,163 @@
...
+        oc apply -f -
+    ]
+    ...
+```
+
+As you can see it is the list of commands, which is needed to do to reproduce the same behaviour that the user had. 
+In the plain text it will look like this:
+
+```
+======
```
Frawless, line 108: It would be nice it will be an executable bash I think.

015-exploratory-test-module.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Proposal
+
+This proposal suggests to:
+* Create new repository `exploratory-test` in `strimzi`
+
+##### Advantages:
+* reduce time when manually testing some components
+* pre-cursor for marathon testing
+* gives us report (which combination were buggy)
```
Frawless, line 159: Some shorter report about what was wrong would be useful as well.

015-exploratory-test-module.md
```diff
@@ -0,0 +1,163 @@
...
+* everybody can play with it and find bugs
+##### Dis-advantages:
+* It may lead to the fact that the scenarios will not be original and sometimes it will be tested redundantly
+* the user has to input syntactically valid YAMLs
```
Frawless, line 163: YAMLs could be easily verified in the code. However, the overall Kafka yaml have to be correct (I point to use-case when user will put the only spec of Kafka for example)

015-exploratory-test-module.md
see-quick: `quit` option here is implicit when a user doesn't have another component, which he wants to apply then it will quit with the following `transcript`.

015-exploratory-test-module.md
see-quick: Yes of course

015-exploratory-test-module.md
```diff
@@ -0,0 +1,165 @@
+# Exploratory test repository
+
+The exploratory test will contain two dialogue applications, which gives you the possibility to create your test scenarios. 
+
+## Motivation
+
+Every release, we do exploratory testing and test Strimzi. 
+Basic functionality is to find potential bugs by trying various scenarios, which user has to input.
+This tool also provides `transcript` for all scenarios that the user tried. 
+In other words, the application stores success and wrong scenarios.
```
samuel-hawker, line 10:
```suggestion
In other words, the application stores successful and failing scenarios. 
```

samuel-hawker:
I can see what you're trying to achieve with this, but it seems incredibly similar to:
https://cucumber.io/
Which uses the gherkin syntax. What you seem to want to design here is an interactive gherkin based manual testing framework that allows for switchable steps (i.e. switch the kafka CR to exhaustively do similar scenarios but with same assertions)
I wonder then why even bother doing this manually at all - and instead just automate this, the human y/n prompts in theory can be answered by a machine that queries the state of a resource or its sub resources.
This seems to be an acceptance that manual testing has to be necessary and exhaustive, which in my opinion is rarely sustainable.
Wouldn't more exhaustively system testing a release give more assurances?

see-quick:
Thanks for the comment @samuel-hawker .
----
Yes, I know gherkin syntax with behave (python) implementation. First of all, Gherkin uses pre-defined scenarios given, when, then, which in my case can be interpreted as `BaselineExploration` class but that's all. If you take a closer look at`CreativeExploration` there is no similarity because you are letting the user input his ideas to this kind of testing and discover paths, which can invoke bug. Second of all the idea behind doing it manually is that you will never create a test case which has for instance 12 steps and more. (1.deploy co, 2.deploy Kafka, 3. deploy metrics, 4.deploy Grafana... etc.). We design our system tests to be as quick as possible. 

> Wouldn't more exhaustively system testing a release give more assurances?
I think no because you can't achieve with system tests this goal. As I previously mentioned system tests has to be compact and not so long. In this type of testing, you can do whatever you want and create really bizarre scenarios which can find a system error.

PR #24, opened 2021-01-05
feat: add a proposal to use a REST API for the UI

pmuir:


016-rest-admin-api.md
```diff
@@ -0,0 +1,67 @@
...
+* a way for clients to describe the data they want, and get returned only that data
+* a typed schema (with the ability to define new types)
+* built in support for schemas
+
+It also brings a number of challenges:
+
+* support for API versions is not so well defined as REST
+* federating APIs is much harder and introduces a performance hit
+* complexity both in the server code and in the client code
+* whilst querying data is simple, mutating data is not as simple using GraphQL
```
wtrocki, line 28:
```suggestion
* whilst querying data is simple, mutating data is not as simple when using GraphQL
```

scholzj (commented):
I do not have a strong preference between REST API and GraphQL. And I think we are still early enough in this effort to be able to switch.

But I do not feel like we have enough people to maintain both of them. And I think there is a huge risk that one of them will be lacking (at the end this proposal basically directly suggests it) or that one will implement feature X while the other feature Y just because someone implemented one in REST and someone else the other in GraphQL. That brings a big risk of users choosing _the wrong one_ in the beginning and having to change later etc.

As I see it, we should either decide for one or the other. Or we need to commit into strictly supporting both of them (i.e. no PR is merged unless it implements both the REST and GraphQL path etc.)

scholzj:
@matthew-chirgwin @gmcrobert Since you are the authors of the original proposals I would be quite curious what do you think about it.

sjwoodman:
Unless we have specific uses for both REST and GraphQL I would support only having a single API to develop and maintain. If REST is preferred for Strimzi Admin then let's build that out. If GraphQL is required at a later date it can be added in such a way that it supports the desired use case. At that point I agree with @scholzj that no PR should be merged unless feature parity is retained.

gmcrobert:
I don't have a strong preference either but there are a few things I think which are worth considering. Firstly I am not sure that GraphQL is more complicated than REST. I think the problems are more familiarity and maturity as GraphQL is a younger technology and is still developing whereas REST is well known but I think the benefits of GraphQL in the long run outweigh the areas where the GraphQL implementation is possibly still a little immature and as the technology is still progressing, I am sure that the complexities will be ironed out over time. I still question what benefit the admin server brings in Strimzi and I feel that adding a REST admin server gives very little benefit over the HTTP Bridge to be honest. The initial proposals envisaged the admin server providing an all encompassing API, in other words the API would not be limited to the Kafka admin API but could be extended to create higher order APIs built on top of the Kubernetes API for example but, I am not sure that Strimzi would want to do that as it kind of breaks the purity of the idea that Strimzi is a Kafka cluster on Kubernetes and simply "Kubernetes enables" Kafka allowing add-on features to be implemented on downstream projects. The fact that Strimzi is closely tied to Kafka authentication and authorization makes it difficult to figure out how you would use Kafka defined users to call non-Kafka APIs. The other idea for the admin server was that it creates a single API which could support both a CLI and the UI but, given the simplicity of the API that is going to be implemented, I think it is overkill having a separate admin server and again, you are duplicating functionality already available in the bridge. For my money, I would stick with GraphQL and I would drop the admin server and use the Express server that is embedded in the UI to serve requests. GraphQL-java does not support schema stitching or schema federation properly whereas the javascript version is the reference implementation and does support it.

pmuir:
Updated the proposal as per comments from @scholzj and @sjwoodman to support just REST and remove GraphQL support.

pmuir:
> I don't have a strong preference either but there are a few things I think which are worth considering. Firstly I am not sure that GraphQL is more complicated than REST. I think the problems are more familiarity and maturity as GraphQL is a younger technology and is still developing whereas REST is well known but I think the benefits of GraphQL in the long run outweigh the areas where the GraphQL implementation is possibly still a little immature and as the technology is still progressing, I am sure that the complexities will be ironed out over time. 

Yes, without a doubt this is a somewhat subjective area. For me the complexity is in understanding how resolvers work, understanding how mutations work (which for me don't feel natural at all vs post requests), understanding how error handling works (which still feels wrong to me as it breaks HTTP)  and then in the code which is brings in more components than a pure REST endpoint.

My experience on the GraphQL projects that I've worked on to date is that GraphQL is very good where you have large API surface areas or entities with a lot of relations between them (for example the modelling user identity, or the GitHub API) doesn't provide benefits for small surface area APIs or entities without many relationships. It feels to me like the Strimzi API surface area is clearly the latter.

>  I still question what benefit the admin server brings in Strimzi and I feel that adding a REST admin server gives very little benefit over the HTTP Bridge to be honest.

My understanding is that it is primarily to provide support for authentication and authorization, as well as passthrough TLS, but one of the @strimzi/maintainers can probably provide more context.

> The initial proposals envisaged the admin server providing an all encompassing API, in other words the API would not be limited to the Kafka admin API but could be extended to create higher order APIs built on top of the Kubernetes API for example but, I am not sure that Strimzi would want to do that as it kind of breaks the purity of the idea that Strimzi is a Kafka cluster on Kubernetes and simply "Kubernetes enables" Kafka allowing add-on features to be implemented on downstream projects. The fact that Strimzi is closely tied to Kafka authentication and authorization makes it difficult to figure out how you would use Kafka defined users to call non-Kafka APIs. 

This doesn't seem like a good goal to me. Kubernetes already supports creating higher order APIs (via custom resources) which can exposed via REST (using the Kubernetes API server). A standalone project could be created to expose this as a GraphQL if desired.

The point around authn also points to this.

> For my money, I would stick with GraphQL and I would drop the admin server and use the Express server that is embedded in the UI to serve requests. 

I like the three tier archiecture (react frontend, BFF server, HTTP API) as it allows for easy multi-tenancy (multiple Strimzis in the same cluster, all using the same UI), and also means the BFF doesn't have to speak the Kafka binary protocol. Though I think you may have meant to use the bridge api from the BFF?

> GraphQL-java does not support schema stitching or schema federation properly whereas the javascript version is the reference implementation and does support it.

Right - schema stitiching/federation does work in GraphQL, but it isn't mature.

matthew-chirgwin:
As others have mentioned, maintaining multiple APIs for the UI makes no sense. So I absolutely agree that one and only one should be used for the UI. I would also suggest separating a discussion on ‘should we have an admin server at all’ vs ‘which API should be used’ - they are different subjects, and I will talk to the API one here.

One thing I would add (which hasn’t been mentioned/touched upon yet) is that both the UI and Admin proposals/architectures were written/approached with a view for extension/future work. For instance, effort was made to detach and modularise the admin server into areas of responsibility (eg Kafka Admin), so others could be added later (eg k8s access, Kafka Connect, Metrics) as requirements called for them. In addition, those modules should make an effort to separate the ‘business logic’ of getting the data, and the API logic to send that data to the client. That helps here if we were to swap from GraphQL (GQL) to REST the work should hopefully be minimal, but also if the need arises for other API requirements for say an Admin CLI (as an example), the ability to easily add other API support.

I agree that the current admin/UI needs/requirements are simple, and that does make the usage of GQL overkill. GQL also (vs REST) is still pretty new, pretty different in how it would be implemented, and is not as mature. However, as more capability is added to the UI (and a topic listing was always meant to be just the start), more Kafka entities (for instance) will be needed, and these entities do have relationships to each other. GQL can model/represent these. Combined with the modules in admin mentioned previously (and additional entities/their relationships being returned - say for k8s resources, Kafka Connect instances etc), I see GQL resolvers/schemas for each module scaling/being easier to maintain in the long run.

As with most things, what we think the right tool for the job should be used. Long term I think this would be GraphQL myself, but do understand the value/motivations of the REST approach.

dlabaj:
> The initial proposals envisaged the admin server providing an all encompassing API, in other words the API would not be limited to the Kafka admin API but could be extended to create higher order APIs built on top of the Kubernetes API for example ...
I like this idea however I don't think it belongs as part of the strimzi project.  On other projects we ran into some similar issues and something like this would be hugely beneficial.  But for strimzi I think we might want to leave the other APIs out.

scholzj:
> I like this idea however I don't think it belongs as part of the strimzi project. On other projects we ran into some similar issues and something like this would be hugely beneficial. But for strimzi I think we might want to leave the other APIs out.

It does belong. At least long term there is for sure an interest to be able to create for example Connect clusters, users, Connectors or even Kafka clusters using the UI.

dlabaj:
> > I like this idea however I don't think it belongs as part of the strimzi project. On other projects we ran into some similar issues and something like this would be hugely beneficial. But for strimzi I think we might want to leave the other APIs out.
> 
> It does belong. At least long term there is for sure an interest to be able to create for example Connect clusters, users, Connectors or even Kafka clusters using the UI.

Ah that makes sense. Thanks for the clarification.

ppatierno:
@pmuir should we assume that the Security part of it is still in place as defined by https://github.com/strimzi/proposals/blob/master/010-ui-and-admin-server-security.md ?

pmuir:
@ppatierno correct, there would be no change to the authentication and authorization architecture/design. Proposal 10 would still hold in it's entirity

ppatierno:
So I am actually fine with this proposal. @scholzj @tombentley do you think we need to add more? I was thinking about a potential API contract to understand what will be provided initially but tbh we haven't had that for the previous proposal with GraphQL, so I don't think it's worth to add it now here.

tombentley (approved):

ppatierno (approved):

scholzj (approved):
I do not really see the technical value described in the proposal. But if the UI contributors feel like this is a good direction I'm fine with it.

ppatierno:
@samuel-hawker @Frawless do have you time for reviewing this one? We would like to move forward. Thanks!

Frawless (approved):

dlabaj (approved):

samuel-hawker (approved):

PR #27, opened 2021-02-25
Restructure the installation files

scholzj:
This proposal suggests changes to the way we use the examples and installation files in the `master` branch of the operators repository. It aims to make it more friendly to new users.

I was a bit undecided between keeping the existing directories for the release files or moving them to `deploy` subdirectory. If people prefer the rejected alternative I'm fine to change the proposal.

### DIFF
--- a/019-restruture-the-installation-files.md
+++ b/019-restruture-the-installation-files.md
@@ -7,7 +7,7 @@ This proposal suggests restructuring to the folders of the [`strimzi-kafka-opera
 Currently, the Strimzi installation files in our master branch correspond to the master branch.
 That means they are under constant development and often work only with the freshly pulled images.
 That suites well for development work, but not necessarily for the users.
-They often checkout the GitHub repository, see the install or example folders and try them, not realising that they're using `master` rather than a stable release.
+They often checkout the GitHub repository, see the install or example folders and try them, not realizing that they're using `master` rather than a stable release.
 If they just use them for a short time it should not matter.
 But if they keep using them, they will run into issues later.
 
@@ -23,9 +23,11 @@ This proposal suggests following:
     * This subdirectory will contain the _in-development_ versions of `install`, `examples` and `helm-chart` directories.
     * Any changes in regular PRs, generated CRD files etc. will be done in this directory
     * New releases will use these files for packaging
+    * The release artifacts will remain unchanged, they will just be built from this directory and not from the original ones
 * The original `install`, `examples` and `helm-chart` directories will be kept, but will contain the files for the latest stable release
     * Updates to them will be done only during / after release
     * They will not be changed with regular PRs
+    * A README.md file will be aded to them to clarify they contain the latest stable release and should not be changed as part of development but only when new versions are released
 
 ### Risks
 
--- a/019-restruture-the-installation-files.md
+++ b/019-restruture-the-installation-files.md
@@ -7,7 +7,7 @@ This proposal suggests restructuring to the folders of the [`strimzi-kafka-opera
 Currently, the Strimzi installation files in our master branch correspond to the master branch.
 That means they are under constant development and often work only with the freshly pulled images.
 That suites well for development work, but not necessarily for the users.
-They often checkout the GitHub repository, see the install or example folders and try them.
+They often checkout the GitHub repository, see the install or example folders and try them, not realising that they're using `master` rather than a stable release.
 If they just use them for a short time it should not matter.
 But if they keep using them, they will run into issues later.
 
@@ -21,7 +21,7 @@ We should change the way we manage these files.
 This proposal suggests following:
 * Create a new subdirectory `packaging`
     * This subdirectory will contain the _in-development_ versions of `install`, `examples` and `helm-chart` directories.
-    * An changes in regular PRs, generated CRD files etc. will be done in this directory
+    * Any changes in regular PRs, generated CRD files etc. will be done in this directory
     * New releases will use these files for packaging
 * The original `install`, `examples` and `helm-chart` directories will be kept, but will contain the files for the latest stable release
     * Updates to them will be done only during / after release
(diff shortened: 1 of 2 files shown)


019-restruture-the-installation-files.md
```diff
@@ -0,0 +1,65 @@
+# Restructure the installation files
+
+This proposal suggests restructuring to the folders of the [`strimzi-kafka-operator` GitHub repository](https://github.com/strimzi/strimzi-kafka-operator).
+
+## Current situation
+
+Currently, the Strimzi installation files in our master branch correspond to the master branch.
+That means they are under constant development and often work only with the freshly pulled images.
+That suites well for development work, but not necessarily for the users.
+They often checkout the GitHub repository, see the install or example folders and try them.
```
tombentley, line 10:
```suggestion
They often checkout the GitHub repository, see the install or example folders and try them, not realising that they're using `master` rather than a stable release.
```

019-restruture-the-installation-files.md
```diff
@@ -0,0 +1,65 @@
...
+
+Improve the experience of the users who checkout Strimzi and want to install / try it.
+
+## Proposal
+
+We should change the way we manage these files.
+This proposal suggests following:
+* Create a new subdirectory `packaging`
+    * This subdirectory will contain the _in-development_ versions of `install`, `examples` and `helm-chart` directories.
+    * An changes in regular PRs, generated CRD files etc. will be done in this directory
```
tombentley, line 24:
```suggestion
    * Any changes in regular PRs, generated CRD files etc. will be done in this directory
```

019-restruture-the-installation-files.md
```diff
@@ -0,0 +1,65 @@
...
+    * New releases will use these files for packaging
+* The original `install`, `examples` and `helm-chart` directories will be kept, but will contain the files for the latest stable release
+    * Updates to them will be done only during / after release
+    * They will not be changed with regular PRs
+
+### Risks
+
+The proposal keeps existing directories with more or less the same files.
+But changes their purpose.
```
tombentley, line 33:
```suggestion
The proposal keeps existing directories with more or less the same files but changes their purpose.
```

ppatierno:
To be honest I am for the rejected alternative to use the `deploy` folder that is more or less well known and used in other projects (as already mentioned by you)

019-restruture-the-installation-files.md
```diff
@@ -0,0 +1,64 @@
+# Restructure the installation files
+
+This proposal suggests restructuring to the folders of the [`strimzi-kafka-operator` GitHub repository](https://github.com/strimzi/strimzi-kafka-operator).
+
+## Current situation
+
+Currently, the Strimzi installation files in our master branch correspond to the master branch.
```
sknot-rh, line 7: Does this make sense?
```suggestion
Currently, the Strimzi installation files in our master branch correspond to the images built from the master branch.
```

019-restruture-the-installation-files.md
```diff
@@ -0,0 +1,64 @@
...
+    * New releases will use these files for packaging
+* The original `install`, `examples` and `helm-chart` directories will be kept, but will contain the files for the latest stable release
+    * Updates to them will be done only during / after release
+    * They will not be changed with regular PRs
+
+### Risks
+
+The proposal keeps existing directories with more or less the same files but changes their purpose.
+That can be confusing for people using them today for the right purpose.
+It might be also confusing for new contributors who might try to change them in their PRs
```
samuel-hawker, line 34: I think these very good points, I think we can mitigate this somewhat by adding/injecting comments into these files to make clear their purpose, like the install files could have large bannered comments saying 
"Release files - DO NOT EDIT" or something equally scary

samuel-hawker (approved):
Seems like a great proposal to me. I think as long as the HACKING/DEVELOPMENT readme is updated to reflect these new locations of files the disruption should be minimal.

samuel-hawker:
I should mention I am a little against using `deploy`, just because I think the vanilla yaml + helm being located there clashes a little with tthe convention where deploy is used exclusively for packaging csvs etc for tools such operator-sdk/olm.

Do challenge me on this though if you don't agree with this!

Alternative to `packaging` could also simply be `install-dev` which is arguably more clear and easier transition for developers.

Frawless (approved):
STs will be affected by this as well. We should change the links in proper places during implementation.

019-restruture-the-installation-files.md
```diff
@@ -0,0 +1,64 @@
...
+## Motivation
+
+Improve the experience of the users who checkout Strimzi and want to install / try it.
+
+## Proposal
+
+We should change the way we manage these files.
+This proposal suggests following:
+* Create a new subdirectory `packaging`
+    * This subdirectory will contain the _in-development_ versions of `install`, `examples` and `helm-chart` directories.
```
tombentley, line 23: So the release zips will pretty much consist of the contents of this directory, right?

tombentley (approved):
I'm in favour of this.

019-restruture-the-installation-files.md
scholzj: I do not expect the release artifacts to change from how they look today. They will just be built based on the files from `packaging` and not from the project _root_. I can make it more clear in the proposal.

scholzj:
> Alternative to packaging could also simply be install-dev which is arguably more clear and easier transition for developers.

@samuel-hawker I probably prefer to keep the `packaging` folder over `install-dev`. I think the clearer separation from the `install` directory helps to distinguish those. (it is not that I insist on `packaging` it self - I just don't fancy to have there another directory starting with `install`)

019-restruture-the-installation-files.md
scholzj: I was more thinking about having some `README.md` there with explanation. The `DO NOT EDIT` warning is tricky since users are actually expected to edit some of these to change namespaces or the default configuration.

scholzj:
@sknot-rh @ppatierno Any more comments?

@ppatierno It looks like you are the only one strongly inclined to the `deploy` folder.

sknot-rh (approved):

ppatierno (approved):
Quite often it happens btw lgtm ;)

scholzj:
@ppatierno So wanna keep it open and discuss it on the next community call if you are unhappy with it? I do not think there is any hurry about this if you want to do that.

PR #30, opened 2021-04-14
Feature Gates

scholzj:
This PR proposes to start using Feature Gates in the operators to enable or disable functionality.

### DIFF
--- a/022-feature-gates.md
+++ b/022-feature-gates.md
@@ -29,6 +29,8 @@ But there are also limitations:
   So adding fields to the Kafka CR is not a good option to gradually roll out changes or change a default behavior over multiple releases.
 * Sometimes, you also want to just change the behavior of the operator in general.
   But you want to avoid having endless list of `enableFeatureX`, `disableFeatureY`, or `useZ` fields in the custom resources.
+  For such fields, it would be also hard to change their defaults later in backwards-compatible way.
+  So in case such feature would be disabled by default at first, users would need to explicitly enable it even later when the feature is mature.
 * The use of the state of the existing Kubernetes resources works well for one-of operations.
   That fits well for upgrades or for triggering actions such as restarting a pod.
   But not for features which change the operator behavior during the whole run of the operator.
--- a/022-feature-gates.md
+++ b/022-feature-gates.md
@@ -29,7 +29,7 @@ But there are also limitations:
   But you want to avoid having endless list of `enableFeatureX`, `disableFeatureY`, or `useZ` fields in the custom resources.
 * The use of the state of the existing Kubernetes resources works well for one-of operations.
   That fits well for upgrades or for triggering actions such as restarting a pod.
-  But not for configuring an ongoing feature.
+  But not for features which change the operator behavior during the whole run of the operator.
 
 This shows the there might be space for other ways how to enable or disable features.
 
@@ -39,6 +39,21 @@ Strimzi operators should support _feature gates_ to gate selected functionality 
 Each feature gate will have a default state (enabled or disabled by default) and users will be able to manually enable or disable them.
 The feature gates will apply for the whole operator and alter its behavior.
 
+The feature gates should go through 3 different maturity stages:
+* Alpha (typically disabled by default)
+* Beta (typically enabled by default)
+* General Availability / GA (typically enabled by default or always enabled)
+
+Alpha stage indicates that the feature might be unstable (i.e. it might change in the future) or buggy.
+Beta stage indicates a that the feature is already well tested and most probably should not change.
+GA means that the feature is stable and should not change anymore.
+GA features might optionally also change to _always enabled features_.
+In this case the feature gate will be removed and it will not be possible to disable the feature anymore.
+The alpha and beta features might be removed in case they are deemed as not useful or not the solution.
+
+Typically, the gated features should be introduced using a Strimzi Proposal
+The proposal should also suggest the plan for how the feature will move through the different maturity stages.
+
 ### Use cases 
 
 The feature gates should be used for the following use-cases:
@@ -47,11 +62,11 @@ _Note: The examples are just examples. Approving this proposal does not approve 
 
 #### Gradually introducing changes to behavior while minimizing disruption
 
-The feature gate will be introduced and disabled by default.
+The feature gate will be introduced at the alpha stage and disabled by default.
 Users will be able to manually enable it to test or use the new feature.
-After one or more releases, the feature gate default would change to enabled by default.
+After one or more releases, the feature gate will move to beta stage and change to enabled by default.
 Users who are not ready to use it will be still able to disable it manually.
-After another one or more releases, the feature gate might be removed and the feature will be enabled by default without any possibility to disable it.
+After another one or more releases, the feature gate might move to GA and be removed / always enabled.
 
 An example of this use case might be the introduction of the separate control plane listener.
 The listener it self can be added by default.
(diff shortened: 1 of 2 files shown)


022-feature-gates.md
```diff
@@ -0,0 +1,105 @@
...
+An example of this use case might be feature gates to disable automatic creation of Pod Disruption Budgets or Network Policies.
+
+### Implementation
+
+Initially the feature gates support should be implemented in Cluster Operator.
+But the same logic can be also applied to User and Topic Operators.
+The feature gates will be configured in an environment variable `STRIMZI_FEATURE_GATES`.
+Since the feature gates might be enabled or disabled by default, the configuration needs to consist of coma-separated key-value pairs.
+The key will be the name of the feature gate and the value will be a boolean defining the state -  either enabling it (`true`) or disabling it (`false`).
+For example `PodDisruptionBudgets=false,NetworkPolicies=false,ControlPlaneListener=true`.
```
tombentley, line 89: Alternative: `STRIMZI_FEATURE_GATES=-PodDisruptionBudgets,-NetworkPolicies,+ControlPlaneListener`. I don't have a strong opinion though.

I assume it's an error for a name to appear more than once?

022-feature-gates.md
```diff
@@ -0,0 +1,105 @@
...
+## Motivation
+
+In many cases, the mechanisms described above work fine and are well suited.
+But there are also limitations:
+
+* Changes to the custom resources APIs have permanent character.
+  We can remove fields from the CRDs only when changing the API versions.
+  So adding fields to the Kafka CR is not a good option to gradually roll out changes or change a default behavior over multiple releases.
+* Sometimes, you also want to just change the behavior of the operator in general.
+  But you want to avoid having endless list of `enableFeatureX`, `disableFeatureY`, or `useZ` fields in the custom resources.
```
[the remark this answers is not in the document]
scholzj: Sorry, I'm not sure I follow this comment. I mean in theory, you can change the default of the field in the API in the same way you change the you change the default of the feature gate. Or do you mean that that would be against best API practices?

scholzj:
@tombentley I restructured it a bit. But apart from the one comment where I left some follow-up question, the rest should be incorporated.

Frawless (approved):

022-feature-gates.md
```diff
@@ -0,0 +1,125 @@
...
+
+This proposal suggests to add support for Feature Gates in the Strimzi operators.
+Feature Gates should give us additional options how to control and mature different behaviors in the operators.
+
+## Current situation
+
+Currently, any features or behaviors of the operator are controlled by one of these two mechanisms:
+
+* The Strimzi custom resources (as a form of an API)
+* The state of the Kubernetes resources
```
samuel-hawker, line 11:
```suggestion
* The state of the Kubernetes resources
* Environmental variables in the operator deployment
```
I think it's worth calling this out explicitly as if differs slightly from the other kube resources

022-feature-gates.md
```diff
@@ -0,0 +1,125 @@
...
+* The `useServiceDnsDomain` in the listener configuration of the `Kafka` custom resource defines what kind of address should be used for given listener.
+  Whether fully qualified DNS name or not.
+* When the operator during Kafka upgrade sees that the old pods are still using the TLS sidecars, it will automatically loosen the TLS configuration to make sure the upgrade can be completed.
+
+## Motivation
+
+In many cases, the mechanisms described above work fine and are well suited.
+But there are also limitations:
+
+* Changes to the custom resources APIs have permanent character.
```
samuel-hawker, line 25:
```suggestion
* Changes to the custom resources APIs have upgrade and forwards compatibility concerns.
```
Just reworded as I understand what it was getting at, but also you then refute its permanence in the next line, I think being more specific here is better, but your point is still well delivered.

022-feature-gates.md
```diff
@@ -0,0 +1,125 @@
...
+* The use of the state of the existing Kubernetes resources works well for one-of operations.
+  That fits well for upgrades or for triggering actions such as restarting a pod.
+  But not for features which change the operator behavior during the whole run of the operator.
+
+This shows the there might be space for other ways how to enable or disable features.
+
+## Proposal
+
+Strimzi operators should support _feature gates_ to gate selected functionality and switch it on or off.
+Each feature gate will have a default state (enabled or disabled by default) and users will be able to manually enable or disable them.
```
samuel-hawker, line 39: Not to over-engineer here, but is there ever a case where we may want enums instead of booleans?
I think Booleans can cover almost all use cases, but I suppose there might be interesting scenarios where whole different options could be considered?

022-feature-gates.md
```diff
@@ -0,0 +1,125 @@
...
+
+Alpha stage indicates that the feature might be unstable (i.e. it might change in the future) or buggy.
+Beta stage indicates a that the feature is already well tested and most probably should not change.
+GA means that the feature is stable and should not change anymore.
+GA features might optionally also change to _always enabled features_.
+In this case the feature gate will be removed and it will not be possible to disable the feature anymore.
+The alpha and beta features might be removed in case they are deemed as not useful or not the solution.
+
+Typically, the gated features should be introduced using a Strimzi Proposal
+The proposal should also suggest the plan for how the feature will move through the different maturity stages.
```
samuel-hawker, line 55: I suppose that wherever the feature flags are listed, we will comment/annotate the lines with their maturity level?
I fear if there is a split and the maturity is tracked via issues the knowledge may be harder to find/update

022-feature-gates.md
```diff
@@ -0,0 +1,125 @@
...
+If any feature gate appears in the list more then once, the operator will throw an exception and exit.
+The operator will throw an exception and exit also in case a non-existing or removed feature gate appears in the list.
+Feature gates not in the list will have the default value.
+
+The feature gates will be parsed and stored in the `ClusterOperatorConfig` class / object.
+The `ClusterOperatorConfig` object is already used in different places in the operator and makes the feature gates easily accessible without any major changes to the code structure.
+
+## Compatibility
+
+There are no compatibility issues.
```
samuel-hawker, line 117: I know this is just a proposal, but I wonder how we plan to have feature gated CRD API changes, i.e. if we add a new experimental/alpha field, say, `spec.alpha` , how do we ensure this doesnt make it into the GA release?
Are we going to maintain multiple CRDs?
Or simply just allow devs to overwrite the CRD locally? (I think the latter is ill advised, but the former sounds like more generated file checkins!)

samuel-hawker (approved):
I think this sounds wholly reasonable & a good addition, left some comments, thoughts and questions. Some of which I think are important, but overall the goal of this proposal cant be rejected on the grounds of those questions being answered.
Great work!

022-feature-gates.md
scholzj: I added a new sentence to specifically mention that it should be listed int he documentation.

022-feature-gates.md
scholzj: I would probably prefer to avoid this. I think that it would make it more complicated not just to use them in the code but also make the user experience worse and more complicated.

022-feature-gates.md
scholzj: I do not think we can do it as that would be:
* Breaking the compatibility
* Forcing the people to often convert the resources which we want to avoid

The main idea of the feature gates for me is to avoid the CRD changes where not needed to control things. But if you really need to change the API, I'm not sure you need the feature gate. E.g. when you need to add new field to the CRD to configure something, do you really need the feature gate to control it? Or is either having it or not having it in the CR enough? At the end, the maturity level can be carried also in the API reference documentation.

022-feature-gates.md
samuel-hawker: Makes sense

022-feature-gates.md
samuel-hawker: Ah that makes a lot of sense, apologies I was still viewing it from the old way, but indeed, if there is new config sections, I suppose we can include the new fields and just ignore doing anything with them if certain features not enabled.

samuel-hawker (approved):
All questions addressed

sknot-rh (approved):

022-feature-gates.md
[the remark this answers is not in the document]
tombentley: Changing a default value is often not backwards compatible, because it's often difficult to reason that it couldn't possibly break something for users. Because of this it means that not only do early adopters end up with lots of `enableFeatureX`, `disableFeatureY` it means that anyone who wants to use those features (even when they become _recommended_ things which all Strimzi users should use) ends up with those fields.

But I can see this is rather a convoluted argument, so I'm fine with not adding it to the proposal.

tombentley (approved):

022-feature-gates.md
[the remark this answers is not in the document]
scholzj: Got it. I tried to add it to the text.

scholzj:
Approved by 4 maintainers (+ my implicit vote).

PR #35, opened 2021-05-12
Proposal for Network Policy Generation Environment Variable.

davidjlynn:


027-network-policy-generation-feature-gate.md
```diff
@@ -0,0 +1,63 @@
...
+This Feature Gate will be a long lived Feature Gate, and is not meant to phase in a feature. Instead, it should remain so users can turn off this functionality when needed.
+
+When this feature gate is disabled, all generation of Network Policies from the Strimzi Operator is also disabled.
+
+This extends to:
+
+* CruiseControl
+* KafkaCluster
+* KafkaConnectCluster
+* ZookeeperCluster
```
scholzj, line 42: What about Kafka Mirror Maker 2 and Kafka Bridge? I think it should cover all of them.

027-network-policy-generation-feature-gate.md
```diff
@@ -0,0 +1,63 @@
...
+It is possible, in the select case of listeners, to further restrict the Ingress using NetworkPolicyPeers, however this does not extend to all Network Policy generation, for example Metrics.
+It is also not possible to set up Egress Network Policies. 
+
+## Motivation
+
+The current Network Policies that are generated are great for getting Strimzi up and running in a deployment where a deny-all network policy is in place.
+
+However, this does not allow users of Strimzi to reach a fully secured Network Policy set, due to Strimzi not providing the full feature set of Kubernetes Network Policies.
+
+By adding a feature gate, it allows users to disable this generation, and instead write fully secure Network Policies.
```
scholzj, line 20: I guess you should make it clear here that when the user disables the network policies generated by Strimzi, he/she is responsible for setting up the right network policies required by Strimzi to work. I think we will also need to document what these are in the implementation PR (but this proposal can mention that this will be documented).

027-network-policy-generation-feature-gate.md
```diff
@@ -0,0 +1,63 @@
...
+This setup is valid in 2 situations:
+
+* A deployment with no deny-all Network Policy, where all pods can already communicate with each other.
+* A deployment where the consumer wishes to write their own secure Network Policies.
+
+## Proposal
+
+The proposal would be to introduce a Feature Gate which would provide the ability to turn of the generation of Network Policies by Strimzi.
+
+As the functionality already exists to generate Network Policies, a sensible naming of this gate would be `NetworkPolicyGeneration` which by default is set to `true`.
```
scholzj, line 31: I wonder if just `NetworkPolicies` would be better name for the feature gate. Or maybe we can turn it around and call it `DisableNetworkPolicies` and have it disabled by default? Let's see what others think.

027-network-policy-generation-feature-gate.md
```diff
@@ -0,0 +1,63 @@
...
+If users require secure Network Policies, they can disable this using the regular Feature Gate syntax.
```
scholzj, line 32: The proposal in many places suggests the network policies managed by Strimzi are not secure. I disagree with it - I think they are secure. So maybe it would be more clear if you rephrase sentences like this to make it more clear that this is about custom network policies than secure versus insecure.

027-network-policy-generation-feature-gate.md
```diff
@@ -0,0 +1,63 @@
...
+This Feature Gate will be a long lived Feature Gate, and is not meant to phase in a feature. Instead, it should remain so users can turn off this functionality when needed.
```
scholzj, line 33: I think it would be maybe good to add some statement about alpha, beta, GA stages. I agree that this should be long-lived gate which should remain in the same default state. But maybe we want to still go through the alpha and beta stages in case we find out that it causes some unexpected problems and decide we want to implement this differently?

027-network-policy-generation-feature-gate.md
```diff
@@ -0,0 +1,63 @@
...
+There is no way to stop this generation.
+
+It is possible, in the select case of listeners, to further restrict the Ingress using NetworkPolicyPeers, however this does not extend to all Network Policy generation, for example Metrics.
+It is also not possible to set up Egress Network Policies. 
+
+## Motivation
+
+The current Network Policies that are generated are great for getting Strimzi up and running in a deployment where a deny-all network policy is in place.
+
+However, this does not allow users of Strimzi to reach a fully secured Network Policy set, due to Strimzi not providing the full feature set of Kubernetes Network Policies.
```
scholzj, line 18: As mentioned below, I would prefer for this to talk about custom network policies rather than secure / insecure policies.

027-network-policy-generation-feature-gate.md
```diff
@@ -0,0 +1,63 @@
+# Network Policy Generation Feature Gate
+
+The feature would consist of a new Feature Gate which can be used to disable the generation of Network Policies by Strimzi, where the user wishes to set up a more restrictive Network Policy set.
```
scholzj, line 3: As mentioned below, I would prefer for this to talk about custom network policies rather than more strict policies. Based on previous issues, there will be users who will use this to disable network policies what so ever for example.

027-network-policy-generation-feature-gate.md
```diff
@@ -0,0 +1,63 @@
...
+This only impacts the [strimzi-kafka-operator](https://github.com/strimzi/strimzi-kafka-operator) project, which currently generates these network policies.
+
+## Compatibility
+
+By setting this Feature Gate to `true` by default, this maintains existing compatibility.
+
+## Rejected alternatives
+
+The other alternative to adding a feature gate was to attempt to support the full functionality of Network Policies within Strimzi.
+The origin requests for this ticket was to allow more strict rules for the Metrics ports, however further requests also exists for adding Egress support in these Network Policies.
```
scholzj, line 60: Same as above ... I would talk about more customisable network policies.

scholzj (commented):
Thanks for the proposal. I think the idea is good ... I left some comments about the working and formal side of things.

027-network-policy-generation-feature-gate.md
davidjlynn: Good point. KafkaMirrorMaker2 is already included in the PR, so I will add it here.
Kafka Bridge however does not appear to generate any Network Policies?

027-network-policy-generation-feature-gate.md
davidjlynn: Happy to switch to `NetworkPolicies`, the "Generation" part is superfluous here as everything is generation in this context.
I am less of a fan of `DisableNetworkPolicies` as Network Policies are a feature, Disable Network Policies is a setting, but not overly fussed as long as the ability to turn these off is available.
I will switch to `NetworkPolicies` for now, and it can change again based on further feedback.

027-network-policy-generation-feature-gate.md
davidjlynn: I agree, I will change the phrasing to be more along the lines of it becoming their responsibility, rather than they get the ability.

027-network-policy-generation-feature-gate.md
```diff
@@ -0,0 +1,63 @@
...
+By adding a feature gate, it allows users to disable this generation, and instead write fully secure Network Policies.
+
+This setup is valid in 2 situations:
+
+* A deployment with no deny-all Network Policy, where all pods can already communicate with each other.
+* A deployment where the consumer wishes to write their own secure Network Policies.
+
+## Proposal
+
+The proposal would be to introduce a Feature Gate which would provide the ability to turn of the generation of Network Policies by Strimzi.
```
tombentley, line 29:
```suggestion
The proposal would be to introduce a Feature Gate which would provide the ability to turn of the generation off Network Policies by Strimzi.
```

027-network-policy-generation-feature-gate.md
tombentley: Would users always want to enable/disable this feature across all these things simultaneously? I can imagine use cases for KafkaConnect where it might be necessary to do something custom because of a particular connector, but otherwise the user might be happy to have the normal Strimzi behaviour, for example.

Frawless (commented):
Overall LGTM. I agree with comments/proposals from Tom and Jakub

027-network-policy-generation-feature-gate.md
ppatierno: Tbh I am not a fan of "negative" logic so I am more for `NetworkPolicies` defaulted to `true` than a `DisableNetworkPolicies` defaulted to false.

scholzj:
@ppatierno Any comments from you so that @davidjlynn can incorporate all the comments?

ppatierno:
@scholzj I already left one comment about using `NetworkPolicies` as feature gate name and not the negative logic. Other than that I agree with fixing your and Tom's comments. I will review again when they will be all fixed.

scholzj:
@ppatierno Ok, great. Sorry, for some reason it does not show that you had comments on the PR, so that confused me.

scholzj:
@davidjlynn Any chance you can have a look at the comments and incorporate them?

scholzj:
@davidjlynn Any chance you can have a look at the comments and incorporate them? Or should we close the PR?

davidjlynn:
Thank you for all the replies and apologies for the delay, I am away from my computer this week but will try and fit the corrections in next week.

scholzj:
@davidjlynn Great, thanks.

samuel-hawker:
I think this seems like a reasonable proposal. I saw some of the comments sort of skirted around this already, but I am curious why an override for the network policies can't exist the in the CR's, i.e. for `Kafka`
```
spec:
  kafka:
    networkPolicies:
      spec:
        <MY_CUSTOM_SPEC>
```
and allow a user to supply their own specs, I assume this is because these custom netpols can have different api versions, or the specs can be extened by different implementors?
Apologies if I am misunderstanding something here.

scholzj:
> I am curious why an override for the network policies can't exist the in the CR's

TBH, I would not go for this. How would this work? Would it create single policy? Will it have a list of custom policies? How will it affect ZooKeeper network policies or network policies around Cruise Control etc.? I think you can make this work for some folks. But sooner or later you get a request for something where it will not work. So having instead the feature gate to choose whether the user wants to follow Strimzi way for the policies or whether the user wants to do it them self is more future proof and ensures we are not back at the same point in a month or two. 

(just the other day I talked with another user who has network policies use blocked completely by their cluster admins which is another usecase this handles)

samuel-hawker:
I think this makes a lot of sense, thanks Jakub.
it is clear to me then that Network policies are a bit different from the other resources Strimzi manages, because they are so customizable and use-case dependent.

027-network-policy-generation-feature-gate.md
davidjlynn: @scholzj As mentioned, I have added KafkaMirrorMaker2, however Kafka Bridge does not appear to have any network policy generation currently so is not included.

@tombentley This is a good point, however I would say that the most likely situation is that a user is either happy with Strimzi's generated rules, or has their own guidelines for NetworkPolicies in which they will want to write these all themselves anyway. Certainly possible but I think most users would just end up disabling all generation anyway.
I have put it in the rejected solutions, as it is a good idea if it was needed, however we can switch to multiple flags if you feel this is required.

027-network-policy-generation-feature-gate.md
davidjlynn: I switched to `NetworkPolicies`

027-network-policy-generation-feature-gate.md
davidjlynn: @scholzj For all these points, I have avoided the word "secure" and moved to using "custom".

027-network-policy-generation-feature-gate.md
davidjlynn: As above, now using "custom".

027-network-policy-generation-feature-gate.md
davidjlynn: As above, now using "custom".

027-network-policy-generation-feature-gate.md
davidjlynn: As above, now using "custom".

027-network-policy-generation-feature-gate.md
davidjlynn: Added a comment about it moving through the stages, however remaining the same value unlike other Feature Gates.

028-network-policy-generation-feature-gate.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Proposal
+
+The proposal would be to introduce a Feature Gate which would provide the ability to turn off the generation of Network Policies by Strimzi.
+
+As the functionality already exists to generate Network Policies, a sensible naming of this gate would be `NetworkPolicies` which by default is `true`.
+If users require custom Network Policies, they can dirsable this using the regular Feature Gate syntax.
+This Feature Gate will be a long-lived Feature Gate, and is not meant to phase in a featue. Instead, it should remain so users can turn off this functionality when needed.
+
+This Feature Gate will also transition through the Alpha, Beta, GA stages. However, as this is a long-lived feature gate which the behaviour does not become default eventually the value will remain set to `true`.
```
scholzj, line 39: I wonder if we really need to go through Alpha and Beta here. Maybe we can make it GA right away? @strimzi/maintainers WDYT?

Assuming we want to take it through the phases, it would be nice to add some preliminary plan of the versions when it goes Beta and GA as we have in https://github.com/strimzi/proposals/blob/main/026-service-account-patching.md#proposal or https://github.com/strimzi/proposals/blob/main/025-control-plain-listener.md#proposal.

scholzj (approved):
One comment / question - more on other maintainers than you @davidjlynn. But apart from that it looks good to me. Thanks for updating it.

Frawless (approved):

028-network-policy-generation-feature-gate.md
```diff
@@ -0,0 +1,82 @@
...
+* A deployment with no deny-all Network Policy, where all pods can already communicate with each other.
+* A deployment where the user wishes to write their own custom Network Policies.
+  * A user may have strict guidelines on the network policy requirements which differ from Strimzi's generated policies.
+
+## Proposal
+
+The proposal would be to introduce a Feature Gate which would provide the ability to turn off the generation of Network Policies by Strimzi.
+
+As the functionality already exists to generate Network Policies, a sensible naming of this gate would be `NetworkPolicies` which by default is `true`.
+If users require custom Network Policies, they can dirsable this using the regular Feature Gate syntax.
```
sknot-rh, line 36: typo in "dirsable"

sknot-rh (commented):
What are the advantages of using Feature Gates instead of envVar disabling/allowing the generation?

028-network-policy-generation-feature-gate.md
davidjlynn: From my side, I think it would be sensible to skip the Alpha, Beta stages as the behaviour does not change.
In fact, the [proposal for Feature Gates](https://github.com/strimzi/proposals/blob/main/022-feature-gates.md#:~:text=an%20example%20of%20this%20use%20case%20might%20be%20feature%20gates%20to%20disable%20automatic%20creation%20of%20pod%20disruption%20budgets%20or%20network%20policies.) uses this feature as an example of when to skip the stages.
However happy to take your decision on this.

davidjlynn:
@sknot-rh Feature Gates seemed like a nice way to represent the activation and de-activation of features, and this feature was even given as an [example in the documentation on Feature Gates](https://github.com/strimzi/proposals/blob/main/022-feature-gates.md#:~:text=an%20example%20of%20this%20use%20case%20might%20be%20feature%20gates%20to%20disable%20automatic%20creation%20of%20pod%20disruption%20budgets%20or%20network%20policies.).
@scholzj can likely give you a better justification than me on the benefits, however I get the feeling that feature gates may be the prefered way of toggling these behaviours.

scholzj:
@davidjlynn We discussed it today on the Strimzi Community Call and it sounded like other maintainers might prefer to just have a separate environment variable to control this instead of using feature gate. But lets wait for some more to comment on it here before changing anything (technically it should not make much difference 🤷‍♂️ ) ... @tombentley @ppatierno @samuel-hawker

tombentley:
@davidjlynn, sorry for the late-in-the-day thinking here. Essentially I'm wondering whether we really want to use Feature Gates for the "Enabling / disabling a feature" use case, or whether we should seek to limit them only to the "Introducing experimental or unstable APIs" and "Gradually introducing changes to behaviour while minimizing disruption" use cases. I personally think that the value of feature gates is not in having another way of setting a boolean flag, but in the alpha/beta/GA process around them, and the possibility to adding something in alpha or beta and then for it _not_ to go to GA for whatever reason.

So for me this is a option where there will always be the use case for warning off the network policies, and therefore it's better as an option. The other maintainers need to reach a consensus about whether they also see the feature gates in this way.

sknot-rh (approved):

tombentley:
@sknot-rh you approved it, but didn't express an opinion on the question of [whether to use a feature gate for this](https://github.com/strimzi/proposals/pull/35#issuecomment-872208817). Wdyt about it?

ppatierno:
I agree with Tom. The feature gates should work the same way as Kubernetes where AFAIK they are more related to the cycle alpha/beta/GA process than just enabling/disabling a feature. With feature gates, people can try stuff which is not stable but that we want to release before a final and consistent status. I would say it's not the case for network policy.

Frawless:
+1 for controlling it only with env variable set in operator deployment

sknot-rh:
> @sknot-rh you approved it, but didn't express an opinion on the question of [whether to use a feature gate for this](https://github.com/strimzi/proposals/pull/35#issuecomment-872208817). Wdyt about it?

Ah, I forgot to put it to the comment 🤦‍♂️ My +1 to control generation by an envVar.

davidjlynn:
Looks like a majority for setting with an ENV variable.
I will update to reflect this unless there are any strong objections.
It does seem like a good pattern that I am happy with, feature gates are for phasing in features, while environment variables are for the long lived features. Perhaps at some point the proposal for feature gates should [have the section removed/edited](https://github.com/strimzi/proposals/blob/main/022-feature-gates.md#enabling--disabling-a-feature) to describe this preference.

scholzj (approved):

ppatierno (approved):

tombentley (approved):
Thanks!

028-network-policy-generation-feature-gate.md
davidjlynn: As discussed, we will now go with a Environment Variable approach, rather than the Feature Gate approach.

Frawless (approved):

scholzj:
@samuel-hawker Do you want more time to review this? Or can we merge this? Thanks.

PR #39, opened 2021-06-30
Adopt the Strimzi Drain Cleaner project

scholzj:
This proposal suggests to adopt the [Strimzi Drain Cleaner](https://github.com/scholzj/strimzi-cistic-odpadu) project under the Strimzi umbrella. That should help to make it better integrated in out docs and system tests and make users more aware of it.

_Disclaimer: I'm the original author of the project._

### DIFF
--- a/029-adopt-the-drain-cleaner-utility.md
+++ b/029-adopt-the-drain-cleaner-utility.md
@@ -16,11 +16,11 @@ For example:
 
 * Does the broker accept client connections? (this one corresponds roughly to Kubernetes readiness concept)
 * Are all partition replicas in sync?
-* Do all partitions have enough in-sync replicas to be able to receive messages from producers (i.e. do the have more in-sync replicas than what is the minimal in-sync replicas sestting)?
-* Are the replicas leaders and ready to actually receive messages from prodcuers
+* Do all partitions have enough in-sync replicas to be able to receive messages from producers (i.e. do the have more in-sync replicas than what is the minimal in-sync replicas setting)?
+* Are the replicas leaders and ready to actually receive messages from producers
 
 Strimzi currently maps the Kubernetes readiness probe to whether the Kafka broker is able to accept the client connections.
-It monitors te broker state and when the listebners are started, it marks it as ready.
+It monitors the broker state and when the listeners are started, it marks it as ready.
 It does not reflect whether the replicas are in-sync, whether the partitions are under replicated etc.
 This was intentional decisions and users can use other tools Strimzi provides to monitor the other aspects of Kafka readiness.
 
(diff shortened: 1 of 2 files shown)


029-adopt-the-drain-cleaner-utility.md
```diff
@@ -0,0 +1,102 @@
...
+* During voluntary disruptions / evictions
+
+The Kubernetes readiness probe concept does not naturally fit into how Apache Kafka works.
+Apache Kafka brokers have several different levels of readiness which do not always correspond to Kubernetes view of readiness.
+For example:
+
+* Does the broker accept client connections? (this one corresponds roughly to Kubernetes readiness concept)
+* Are all partition replicas in sync?
+* Do all partitions have enough in-sync replicas to be able to receive messages from producers (i.e. do the have more in-sync replicas than what is the minimal in-sync replicas setting)?
+* Are the replicas leaders and ready to actually receive messages from producers
```
ppatierno, line 20: "and" to be removed?

029-adopt-the-drain-cleaner-utility.md
```diff
@@ -0,0 +1,102 @@
...
+
+Kubernetes distinguishes between two types of disruptions: voluntary and involuntary.
+Involuntary disruptions are situations when the pod need to be moved / restarted which are not controlled by Kubernetes.
+This includes for example hardware failures or kernel failures and there is not much we can do about them.
+Voluntary disruptions on the other hand are planned and controlled.
+This includes for example draining worker nodes because of cluster scale-down or because the worker node needs some maintenance.
+
+During the voluntary disruptions, Kubernetes will evict the pods from the node being drained.
+That means that it will shut them down and start them on some other node (if available / possible).
+During the evictions, Kubernetes will follow Pod Disruption Budget rules.
```
ppatierno, line 40: a link to documentation about it could be useful here.

samuel-hawker (commented):
This seems like a great bit of value to users, added some minor grammar suggestions.

I wonder if `drain-cleaner` is the best name for this, or if something like `kafka-safe-disruption-drainer` might be more accurate? More specifically not sure about the 'cleaner' part here, maybe I am missing something though in the implementation or in the context that would make it clean?

ppatierno (approved):

sknot-rh (approved):

scholzj:
> I wonder if drain-cleaner is the best name for this, or if something like kafka-safe-disruption-drainer might be more accurate? More specifically not sure about the 'cleaner' part here, maybe I am missing something though in the implementation or in the context that would make it clean?

@samuel-hawker It is called drain cleaner because when you set the `maxUnavailable` in PDB to `0`, it will block the draining of the worker nodes. And the drain cleaner toll will unblock it => clean it. I'm open to other name suggestions, u TBH `kafka-safe-disruption-drainer` is equally cryptic and just much longer. I do not think you can give it a name from which the user will understand what it really does. and why to use it.

samuel-hawker:
You're right that my name isn't better! :P I think leaving as is works- just curious to see if yourself or the other maintainers had an inspirational name alterative, bu I doubt we will find one

samuel-hawker (approved):

scholzj:
> You're right that my name isn't better! :P I think leaving as is works- just curious to see if yourself or the other maintainers had an inspirational name alterative, bu I doubt we will find one

If you don't like the `drain-cleaner` or think it is particularly confusing, maybe we can use something simpler ... `drainer` or `pod-drainer` 🤷‍♂️

samuel-hawker:
I'll give it some thought. Initial thoughts are something like `kafka-disruption-drainer`
but i think drain-cleaner is serviceable.

tombentley (approved):

Frawless (approved):

PR #41, opened 2021-09-20
Propose custom authentication

Sam-Kramer:


031-custom_authentication.md
```diff
@@ -0,0 +1,96 @@
...
+          type: custom
+          principal.builder.class: SimplePrincipal.class
+          security.protocol: SASL_SSL
+          sasl.enabled.mechanisms:
+            - OAUTHBEARER
+          listener-config:
+            oauthbearer.sasl.client.callback.handler.class: client.class
+            oauthbearer.sasl.server.callback.handler.class: server.class
+            oauthbearer.sasl.login.callback.handler.class: login.class
+            oauthbearer.connections.max.reauth.ms: 999999999
```
scholzj, line 61: I think this should all be in some `config` section which I guess should be just a `Map<String, String>` because each connector will need different options.

031-custom_authentication.md
```diff
@@ -0,0 +1,96 @@
...
+```
+listener.name.<listener-name>.oauthbearer.sasl.client.callback.handler.class=
+listener.name.<listener-name>.oauthbearer.sasl.server.callback.handler.class=
+listener.name.<listener-name>.oauthbearer.sasl.login.callback.handler.class=
+listener.name.<listener-name>.oauthbearer.connections.max.reauth.ms=
+listener.name.<listener-name>.sasl.enabled.mechanisms=
+listener.name.<listener-name>.oauthbearer.sasl.jaas.config=
+listener.name.<listener-name>.AWS_MSK_IAM.sasl.jaas.config=
+listener.name.<listener-name>.AWS_MSK_IAM.sasl.client.callback.handler.class=
+security.protocol=
```
scholzj, line 30: Why does it need to pass `security.protocol`? That impacts all listeners if it is not prefixed, so that seems wrong.

031-custom_authentication.md
```diff
@@ -0,0 +1,96 @@
...
+principal.builder.class=
```
scholzj, line 31: We need to clarify and figure out how to ensure this does not break the existing functionality.

031-custom_authentication.md
```diff
@@ -0,0 +1,96 @@
...
+listener.name.tls-9093.oauthbearer.sasl.server.callback.handler.class=server.class
+listener.name.tls-9093.oauthbearer.sasl.login.callback.handler.class=login.class
+listener.name.tls-9093.oauthbearer.connections.max.reauth.ms=999999999
+principal.builder.class=SimplePrincipal.class
+listener.security.protocol.map=CONTROLPLANE-9090:SSL,REPLICATION-9091:SSL,TLS-9093:SASL_SSL
+```
+
+## Affected/not affected projects
+
+[strimzi-kafka-operator](https://github.com/strimzi/strimzi-kafka-operator)
```
scholzj, line 86: I assume that this is affected. But you should make it more clear since the title is `Affected/not affected`

scholzj (commented):
Thanks for the PR. I left some comments.

031-custom_authentication.md
Sam-Kramer: Sure, that works. One thing to note: `security.protocol: SASL_SSL` is going to be a bit tricky, as we need to append this to `listener.security.protocol.map`, which is why I was thinking of having it be a separate property. 

Also, should have made this a bit more clear, but listener-config I was intending on being a `Map<String, String>`, however  we would forcefully prefix all config settings with `listener.name.<listener-name>` to prevent users from manually patching in forbidden config prefixes.

031-custom_authentication.md
Sam-Kramer: Sorry, I meant for `listener.security.protocol.map` rather than `security.protocol`.

031-custom_authentication.md
Sam-Kramer: More than happy to. Do you want me to flesh this out a bit more?

031-custom_authentication.md
```diff
@@ -0,0 +1,97 @@
...
+
+However, there is no ability to specify custom authentication, which is what this proposal will focus on.
+
+## Motivation
+
+Supporting custom authentication allows greater flexibility for users of Strimzi who may have bespoke/proprietary authentication requirements. This would greatly broaden user adoption of the operator, as it would also support more common authentication schemes, such as [Amazon MSK](https://docs.aws.amazon.com/msk/latest/developerguide/security_iam_service-with-iam.html), and attract users who have much more stringent requirements (i.e. proprietary auth). 
+
+The “cost” of supporting such feature should be minimal as well, as the expectation for custom authentication passes the complexity onto the user. Strimzi should strictly support the passing of configuration values, and nothing else. It’s up to the user to ensure their authn/z schemes can be integrated with Strimzi. Considering this also requires custom built Strimzi Kafka images in order to load custom classes on to the classpath, it’s also expected that will only really apply to highly technical/pro users.
```
tombentley, line 15: There are cons to this:

* If we add support for this in the CRD schema then users can configure clusters with a customer authenticator but the operator has no way of validating that the broker image actually has those classes. Until now the operator and apiserver schema validation together have done a reasonable job of validating CR instances. 

* It's easy for more technical users to make use of this, but if they keep that to themselves then the community doesn't benefit. There would likely end up being multiple "technical users" each with their own MSK authentication classes (so they don't get the benefit of there being a single well-maintained implementation), and less sophisticated users don't get any benefit. In other words the bar remains high even when someone scratches their own itch. Obviously there are limited options for doing better:
    * Building support for each authenticator into the Strimzi images. This doesn't feel great to me. While it lowers the bar for end users it raises a different bar: Getting an authenticator into Strimzi is now an impediment.
    * Add support for building broker images, using the Strimzi images as a base, but adding plugin jar files. Similar to what we're doing with the ability to build `KafkaConnect` images which include connector plugins. This would open the door to being able to configure other plugins, such as policy plugins, custom quota managers, logging classes and so on. It would make it easier for less technical users to make use of plugins, and allows communities to grow around those plugins. So there could be a single open source MSK plugin which everyone could easily use. It doesn't preclude users writing their own plugins in private, either, but makes them easier to deploy because they don't have to engage with the Strimzi build system. Obviously this is not a trivial piece of work, but I'd be interested to hear how crazy people thought it was.

031-custom_authentication.md
tombentley: @Sam-Kramer yes please, if you don't mind?

031-custom_authentication.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Affected/not affected projects
+
+Affected: [strimzi-kafka-operator](https://github.com/strimzi/strimzi-kafka-operator)
+
+## Compatibility
+
+Backwards compatibility shouldn’t be an issue, as current configuration should not affect custom authentication. 
+
+Forwards compatibility will need to be mindful of the custom authentication setting, and how to handle the case of multiple listeners/multiple auth types.
```
tombentley, line 93: Can you elaborate this point (how would we be 'mindful')?

tombentley (commented):
Thanks for the proposal, I left a few comments.

031-custom_authentication.md
Sam-Kramer: ```If we add support for this in the CRD schema then users can configure clusters with a customer authenticator but the operator has no way of validating that the broker image actually has those classes. Until now the operator and apiserver schema validation together have done a reasonable job of validating CR instances.```

Yeah this is true, and unfortunately, unavoidable. However, this is already the case with the CustomAuthorizer already. 

```It's easy for more technical users to make use of this, but if they keep that to themselves then the community doesn't benefit. There would likely end up being multiple "technical users" each with their own MSK authentication classes (so they don't get the benefit of there being a single well-maintained implementation), and less sophisticated users don't get any benefit. In other words the bar remains high even when someone scratches their own itch. Obviously there are limited options for doing better:```
Agreed that the incentive structure needs to be put-in place to ensure users are contributing back to the community, especially the more technical ones. However, one thing to be mindful is not gatekeeping too hard, as otherwise users will just end up forking, and adding their custom auth anyways. I do think a soft-measure that can be taken here, is that rather the explicitly supporting custom auth in the documentation, simply state Amazon MSK support (+ others which we've images for). Hopefully this lead to the workflow of the user filling an issue asking about auth scheme X, which then a boilerplate response can be given on how-to add support. 

```
Add support for building broker images, using the Strimzi images as a base, but adding plugin jar files. Similar to what we're doing with the ability to build KafkaConnect images which include connector plugins. This would open the door to being able to configure other plugins, such as policy plugins, custom quota managers, logging classes and so on. It would make it easier for less technical users to make use of plugins, and allows communities to grow around those plugins. So there could be a single open source MSK plugin which everyone could easily use. It doesn't preclude users writing their own plugins in private, either, but makes them easier to deploy because they don't have to engage with the Strimzi build system. Obviously this is not a trivial piece of work, but I'd be interested to hear how crazy people thought it was.
```
You could take this a step further, and have the jar just be dynamically downloaded at runtime/pod launch. Although this runs counter to exactly why you've images in the first place, and would make rolls much longer.

File: 031-custom_authentication.md (hunk 3 above)
Comment (Sam-Kramer) [reply 031-custom_authentication.md:31]: Please see: https://github.com/strimzi/proposals/pull/41/files#diff-d084fccbfc36b011113ad5b180e07db06bb1f1b443053d002c8f9620427114f8R67-R72

**Comment by Sam-Kramer** [comment]
Any additional concerns @tombentley / @scholzj ? Is the format of the proposal hard to follow? Needs more refinement?

File: 031-custom_authentication.md [hunk 7]
```diff
@@ -0,0 +1,104 @@
...
+```
+    listeners:
+      - name: tls
+        port: 9093
+        type: internal
+        tls: true
+        authentication:
+          type: custom
+          principal.builder.class: SimplePrincipal.class
+          security.protocol: SASL_SSL
```
scholzj, line 54: The `_SSL` part is driven by the `tls: true` flag on a higher level. So - assuming we support only SASL mechanisms - this has to be configured automatically in the code. If we would support also some _non-SASL_ mechanisms, it would need to be configurable with same `SASL` flag.

031-custom_authentication.md
```diff
@@ -0,0 +1,104 @@
...
+          tlsTrustedCertificates: ...
+```
+
+Then, when constructing the broker config:
+
+* `principal.builder.class` is set cluster-wide for all authentication methods (this is the status quo for OAuth authentication). This can only be set once, and we will fail hard (i.e. reject the config) if there are multiple principals defined. 
+    
+    It is expected that the provided principal builder *must* support Strimzi authentication (i.e. ssl-based auth). This is expected to be identical, or very similar to [Strimzi's OAuth library](https://github.com/strimzi/strimzi-kafka-oauth/blob/main/oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthKafkaPrincipalBuilder.java). In particular, it leverages [Kafka's default principal builder class](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/security/authenticator/DefaultKafkaPrincipalBuilder.java#L73-L79), which allows the building of a principal based upon the name of the of peer certificate. Thus, it is expected that the custom principal builder provides a principal of type "user" with the name being that of the ssl peer certificate.
```
scholzj, line 69: Can you add here some examples of how will the principal class look like to handle something custom and yet make sure it works for other supported mechanisms? This will anyway be needed for the docs etc. later.

031-custom_authentication.md
```diff
@@ -0,0 +1,104 @@
...
+    
+    It is expected that the provided principal builder *must* support Strimzi authentication (i.e. ssl-based auth). This is expected to be identical, or very similar to [Strimzi's OAuth library](https://github.com/strimzi/strimzi-kafka-oauth/blob/main/oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthKafkaPrincipalBuilder.java). In particular, it leverages [Kafka's default principal builder class](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/security/authenticator/DefaultKafkaPrincipalBuilder.java#L73-L79), which allows the building of a principal based upon the name of the of peer certificate. Thus, it is expected that the custom principal builder provides a principal of type "user" with the name being that of the ssl peer certificate. 
+
+    In addition, it also expected that the custom authorizer supports principals of type "user", and loads the `super.users` property to know which user's (i.e. Strimzi services) are authorized.
```
scholzj, line 71: Not sure I understand how does this fit into it this. The authorization is independent on authentication, so this seems a bit random here and unrelated to content around it. Maybe you can clarify it?

031-custom_authentication.md
```diff
@@ -0,0 +1,104 @@
...
+    
+    It is expected that the provided principal builder *must* support Strimzi authentication (i.e. ssl-based auth). This is expected to be identical, or very similar to [Strimzi's OAuth library](https://github.com/strimzi/strimzi-kafka-oauth/blob/main/oauth-server/src/main/java/io/strimzi/kafka/oauth/server/OAuthKafkaPrincipalBuilder.java). In particular, it leverages [Kafka's default principal builder class](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/security/authenticator/DefaultKafkaPrincipalBuilder.java#L73-L79), which allows the building of a principal based upon the name of the of peer certificate. Thus, it is expected that the custom principal builder provides a principal of type "user" with the name being that of the ssl peer certificate. 
+
+    In addition, it also expected that the custom authorizer supports principals of type "user", and loads the `super.users` property to know which user's (i.e. Strimzi services) are authorized. 
+
+
+* We parse out `security.protocol` and pass that value into the `listener.security.protocol.map` ** as we currently do for OAuth authentication types.
```
scholzj, line 74: As mentioned above, the `security.protocol` would need to be constructed also based on the other flags.

031-custom_authentication.md
```diff
@@ -0,0 +1,104 @@
...
+## Compatibility
+
+Backwards compatibility shouldn’t be an issue, as current configuration should not affect custom authentication. However, users who try to have multiple auth types (i.e. OAuth and Custom) could run into issues if the custom auth requires `principal.builder.class` (as OAuth setting requires that this be set).
+
+Forwards compatibility will need to be mindful of the custom authentication setting, and how to handle the case of multiple listeners/multiple auth types. Explicitly, understand that: 
+- `listener.security.protocol.map` is partially built by the custom authentication setting
+- `principal.builder.class` can be set in custom auth, and therefore can only be set once
+- listener-specific overrides occur here, and therefore `listener.name.<listener-name>.*` should not be overriden anywhere by other config settings
+
+## Rejected alternatives
```
scholzj, line 102: Might be worth adding here in more detail why can't you use the regular OAuth type authentication.

scholzj (commented):
Sorry, I was away for some time. I left some more comment. These custom things are also very hard to test. So it would be good to also include some strategy for system tests covering this feature. Since without that it would be very hard to maintain this code.

031-custom_authentication.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Current situation
+
+Currently, numerous authentication methods are [supported](https://github.com/strimzi/strimzi-kafka-operator/tree/0.25.0/api/src/main/java/io/strimzi/api/kafka/model/authentication) in Strimzi, and allows the user to configure which one to use. In addition, in the case a custom *authorizer* needs to be used, this is also currently [supported](https://github.com/strimzi/strimzi-kafka-operator/blob/0.25.0/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java#L541-L549). 
+
+However, there is no ability to specify custom authentication, which is what this proposal will focus on.
+
+## Motivation
+
+Supporting custom authentication allows greater flexibility for users of Strimzi who may have bespoke/proprietary authentication requirements. This would greatly broaden user adoption of the operator, as it would also support more common authentication schemes, such as [Amazon MSK](https://docs.aws.amazon.com/msk/latest/developerguide/security_iam_service-with-iam.html), and attract users who have much more stringent requirements (i.e. proprietary auth).
```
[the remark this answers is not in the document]
Sam-Kramer: Hey @scholzj  sorry for the bad link, I was actually thinking of this: https://github.com/aws/aws-msk-iam-auth#configuring-a-kafka-client-to-use-aws-iam rather than docs/ideas about IAM configuraiton. I.e., how can we enable MSK-based auth on the Kafka-cluster itself.

031-custom_authentication.md
```diff
@@ -0,0 +1,104 @@
...
+          principal.builder.class: SimplePrincipal.class
+          security.protocol: SASL_SSL
+          sasl.enabled.mechanisms:
+            - OAUTHBEARER
+          listener-config:
+            oauthbearer.sasl.client.callback.handler.class: client.class
+            oauthbearer.sasl.server.callback.handler.class: server.class
+            oauthbearer.sasl.login.callback.handler.class: login.class
+            oauthbearer.connections.max.reauth.ms: 999999999
+          tlsTrustedCertificates: ...
```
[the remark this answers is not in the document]
Sam-Kramer: Yup to be clear, I was thinking `tlsTrustedCertificates` to modeled identically to how it's currently implemented for other listeners. So ideally, we can leverage what already pre-exists?

031-custom_authentication.md
```diff
@@ -0,0 +1,104 @@
...
+
+Supporting custom authentication allows greater flexibility for users of Strimzi who may have bespoke/proprietary authentication requirements. This would greatly broaden user adoption of the operator, as it would also support more common authentication schemes, such as [Amazon MSK](https://docs.aws.amazon.com/msk/latest/developerguide/security_iam_service-with-iam.html), and attract users who have much more stringent requirements (i.e. proprietary auth). 
+
+The “cost” of supporting such feature should be minimal as well, as the expectation for custom authentication passes the complexity onto the user. Strimzi should strictly support the passing of configuration values, and nothing else. It’s up to the user to ensure their authn/z schemes can be integrated with Strimzi. Considering this also requires custom built Strimzi Kafka images in order to load custom classes on to the classpath, it’s also expected that will only really apply to highly technical/pro users.
+
+## Proposal
+
+Add a CustomAuthenticationKafkaListener class, which supports an example of the following properties
+
+```
```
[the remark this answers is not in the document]
Sam-Kramer: > Sorry, I was away for some time. I left some more comment. These custom things are also very hard to test. So it would be good to also include some strategy for system tests covering this feature. Since without that it would be very hard to maintain this code.

No worries whatsoever! Agreed, I'll include a testing plan as well.

031-custom_authentication.md
```diff
@@ -0,0 +1,104 @@
...
+listener.name.<listener-name>.AWS_MSK_IAM.sasl.client.callback.handler.class=
+listener.security.protocol.map=
+principal.builder.class=
+```
+
+Really, this can be boiled down to: 
+
+```
+listener.name.<lister-name>.<auth-type>.*
+listener.name.<listener-name>.sasl.enabled.mechanisms
```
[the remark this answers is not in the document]
scholzj: I guess the question is whether `PLAINTEXT` or `SSL` can offer some custom authentication ... it sounds like not to me, but I wasn't 100% sure.

031-custom_authentication.md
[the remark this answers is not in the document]
scholzj: Well, I think the main issue to solve here is what kind and how many certificates will you need. Not every custom authentication will probably need the same.

031-custom_authentication.md
```diff
@@ -0,0 +1,142 @@
...
+
+Add a CustomAuthenticationKafkaListener class, which would support the following properties.
+
+```
+listener.name.<listener-name>.oauthbearer.sasl.client.callback.handler.class=
+listener.name.<listener-name>.oauthbearer.sasl.server.callback.handler.class=
+listener.name.<listener-name>.oauthbearer.sasl.login.callback.handler.class=
+listener.name.<listener-name>.oauthbearer.connections.max.reauth.ms=
+listener.name.<listener-name>.sasl.enabled.mechanisms=
+listener.name.<listener-name>.oauthbearer.sasl.jaas.config=
```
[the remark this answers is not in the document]
Sam-Kramer: Hi @scholzj , I only listed the oauthbearer options for brevity, and also as I'm not entirely sure what the sub-properties are for PLAIN/SCRAM/GSSAPI. Happy to maybe leavea comment here saying this is an example for configuring OAUTHBEARER, but PLAIN/SCRAM/GSSAPI should also be applicable?

031-custom_authentication.md
Sam-Kramer: Good point, my only follow up question is, should we support non-SASL based authentication workflows? To my understanding, there really wouldn't be a need for any, as mTLS should already be handled by a TLS listener? I suppose there is http basic authentication, which potentially wouldn't ever be compatible with Strimzi as the usage of certs for authentication.

031-custom_authentication.md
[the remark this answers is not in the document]
Sam-Kramer: That's a good point, we _probably_ want them to be able to also mount kerberos credentials (keytab) if they're using GSSAPI.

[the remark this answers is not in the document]
Sam-Kramer: I'll push a change to the proposal where we include a field called `secretMounts`.

Sam-Kramer:
> Sorry for the time I needed to get back to this. I left some more comments, but I think we are moving in the right direction.

No worries, thanks for getting back to me and providing feedback! Do you've any suggestions in terms of maybe getting the community to review this on a tighter timescale? My hope was if this proposal is accepted, we could maybe get the feature out by EOY (given that we'd do a 0.27.0 release at all).

031-custom_authentication.md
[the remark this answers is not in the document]
scholzj: Well, that is really a part of the point, right? We do not want a custom authentication which just fits your individual use-case. We want something what fits ideally everyone who might need to extent the supported authentication mechanisms. Understanding all the different requirements is also important to properly design the API.

031-custom_authentication.md
scholzj: TBH, I don't know what other methods there would be. I guess it is only mTLS? Or is there something else? There were one or two requirements for a combination of mTLS and SASL in the past. If it is just mTLS, it might be maybe possible to add some more flag got that. If it is more than just mTLS, it might be easier to focus on SASL only.

scholzj:
> No worries, thanks for getting back to me and providing feedback! Do you've any suggestions in terms of maybe getting the community to review this on a tighter timescale? My hope was if this proposal is accepted, we could maybe get the feature out by EOY (if there are plans for a 0.26.1, or 0.27.0 at all). Obviously from my side, I can focus/deliver 100% of my time to getting the PR up.

So the, easy part to answer => yes, I would expect the 0.27.0 to happen before the end of the year.

The second part whether this might be in it is a bit harder to answer. First of all, I apologise for not being always able to react more quickly on your updates. I know that probably wasn't the best experience. Plus I don't know why there was not more feedback from other @strimzi/maintainers.

However, TBH the topic here is quite hard to deal with because it has many consequences. Especially the API (the changes tot he CRD resources) - if not done the right way - is really hard to get rid of or change while keeping proper levels of backwards compatibility. So any problem which we don't catch in this phase might hinder the development for months or years. So that is why at least I personally try to be careful about it.

scholzj:
PS: I also shared the proposal on Slack. Let's see if we can get more people in here with some more requirements.

031-custom_authentication.md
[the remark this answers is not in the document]
Sam-Kramer: Fair point :) I'll update the proposal.

031-custom_authentication.md
Sam-Kramer: So mTLS should already be handled by `KafkaListenerAuthenticationTls`, as it enabled mTLS, plus from what I can tell, there are no additional properties a user would really need to modify.

Sam-Kramer:
> > No worries, thanks for getting back to me and providing feedback! Do you've any suggestions in terms of maybe getting the community to review this on a tighter timescale? My hope was if this proposal is accepted, we could maybe get the feature out by EOY (if there are plans for a 0.26.1, or 0.27.0 at all). Obviously from my side, I can focus/deliver 100% of my time to getting the PR up.
> 
> So the, easy part to answer => yes, I would expect the 0.27.0 to happen before the end of the year.
> 
> The second part whether this might be in it is a bit harder to answer. First of all, I apologise for not being always able to react more quickly on your updates. I know that probably wasn't the best experience. Plus I don't know why there was not more feedback from other @strimzi/maintainers.
> 
> However, TBH the topic here is quite hard to deal with because it has many consequences. Especially the API (the changes tot he CRD resources) - if not done the right way - is really hard to get rid of or change while keeping proper levels of backwards compatibility. So any problem which we don't catch in this phase might hinder the development for months or years. So that is why at least I personally try to be careful about it.

That's fair, I just wanted to make sure there is nothing on my end that's missing/lacking that could be potentially blocking the proposal. I also am aligned with ensuring our API changes are non-breaking/won't hinder future development, seeing as we'll probably be one of the primary consumers of it!

031-custom_authentication.md
scholzj: Yeah, but what I meant was that there are people interested in having mTLS and SASL in parallel on the same listener => you cannot combine your custom type authentication with `KafkaListenerAuthenticationTls` in a single listener (while that would be useful, it is one of the things where the API was not thought through properly and we missed this option and now it is hard to change).

Sam-Kramer: I'm most likely misunderstanding, so apologies for the back-and-forth, but is it possible to have both SASL + mTLS? Confluent docs mention (https://docs.confluent.io/platform/current/kafka/authentication_ssl.html): 
```
If any SASL authentication mechanisms are enabled for a given listener, then SSL client authentication is disabled–even if you have specified ssl.client.auth=required and the broker authenticates clients only using SASL on that listener.
```

scholzj: I don't know TBH. I was not aware of any such limitation - but that does not mean it is not there.

tombentley: [KIP-684](https://cwiki.apache.org/confluence/display/KAFKA/KIP-684+-+Support+mutual+TLS+authentication+on+SASL_SSL+listeners) will change this.

Sam-Kramer: Ah thanks for the link @tombentley ! Cool, then my proposal should definitely support this.

tmitchell10:
IBM Event Streams also require this capability and we were just about to raise a proposal that was very similar to this one.

Our use case is that we have multiple secured Rest endpoints, which act as a proxy to kafka. The request gets authenticated at the Rest endpoint, but as we don’t have the original credentials, the kafka client within the Rest container that performs the User action, uses system credentials to authenticate with Kafka, and the previously authenticated Username is passed through as the identity to use when generating the Kafka Principal.
An example of this would be using a kafka user that has been configured to use Mutual TLS. Once the SSL handshake is processed, we get the UserName, and then forward this into kafka, where the authorization can process the username.

Our solution requires a custom KafkaPrincipalBuilder and also a custom JAAS login module on the kafka listener. We use this to process the “RunAs” identity and to flow that into Kafka, via the PrincipalBuilder.

The only issue that I wasn’t 100% sure about was whether we would be able to configure custom login modules and corresponding custom sasl protocols e.g.listener.name.<listener-name>.<custom protocol>.sasl.jaas.config, and listener.name.<listener-name>.sasl.enabled.mechanisms=<custom protocol> .
The proposal and corresponding comments implied that we would be able to do this, but I just wanted to check that this would be the case. If we can configure these listener config properties using custom protocols, I think that covers everything that we will need.

Sam-Kramer:
> IBM Event Streams also require this capability and we were just about to raise a proposal that was very similar to this one.
> 
> Our use case is that we have multiple secured Rest endpoints, which act as a proxy to kafka. The request gets authenticated at the Rest endpoint, but as we don’t have the original credentials, the kafka client within the Rest container that performs the User action, uses system credentials to authenticate with Kafka, and the previously authenticated Username is passed through as the identity to use when generating the Kafka Principal. An example of this would be using a kafka user that has been configured to use Mutual TLS. Once the SSL handshake is processed, we get the UserName, and then forward this into kafka, where the authorization can process the username.
> 
> Our solution requires a custom KafkaPrincipalBuilder and also a custom JAAS login module on the kafka listener. We use this to process the “RunAs” identity and to flow that into Kafka, via the PrincipalBuilder.
> 
> The only issue that I wasn’t 100% sure about was whether we would be able to configure custom login modules and corresponding custom sasl protocols e.g.listener.name...sasl.jaas.config, and listener.name..sasl.enabled.mechanisms= . The proposal and corresponding comments implied that we would be able to do this, but I just wanted to check that this would be the case. If we can configure these listener config properties using custom protocols, I think that covers everything that we will need.

Hey @tmitchell10, yes, this should be possible! If you'd like, if you provide me with the exact configuration you need, I'm also happy to incorporate this into the proposal, to ensure it will meet your needs.

tmitchell10:
@Sam-Kramer Thanks for offering to add this to the proposal. This is an example of the listener config for a custom JAAS login module:

```
listener.name.<listener-name>.ssl.keystore.location=/tmp/kafka/cluster.keystore.p12
listener.name.<listener-name>.ssl.keystore.password=<password>
listener.name.<listener-name>.ssl.keystore.type=PKCS12
listener.name.<listener-name>.custom-sasl-protocol.sasl.jaas.config=my.custom.loginmodule.CustomLoginModule required;
listener.name.<listener-name>.sasl.enabled.mechanisms=CUSTOM_SASL_PROTOCOL
```

The only thing I wasn't sure about was whether the enabled.mechanisms could have custom values or whether you had planned to have a list of available SASL mechanisms that would have to choose from. We would need the ability to provide a custom value for this.

One other thing that would be really helpful would be the ability to provide multiple authentication mechanisms on the same listener. An example of this would be something like:

```
listener.name.<listener-name>.ssl.keystore.location=/tmp/kafka/cluster.keystore.p12
listener.name.<listener-name>.ssl.keystore.password=<password>
listener.name.<listener-name>.ssl.keystore.type=PKCS12
listener.name.<listener-name>.custom-sasl-protocol.sasl.jaas.config=my.custom.loginmodule.CustomLoginModule required;
listener.name.<listener-name>.scram-sha-512.sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required;
listener.name.<listener-name>.sasl.enabled.mechanisms=CUSTOM_SASL_PROTOCOL,SCRAM-SHA-512
```

Having this capability isn't essential for us, because we could configure multiple listeners, each with a single authentication mechanism, and point the client at the relevant listener endpoint, depending on the authentication mechanism, but if this was something that could be easily incorporated, that would very useful.

Sam-Kramer:
> @Sam-Kramer Thanks for offering to add this to the proposal. This is an example of the listener config for a custom JAAS login module:
> 
> ```
> listener.name.<listener-name>.ssl.keystore.location=/tmp/kafka/cluster.keystore.p12
> listener.name.<listener-name>.ssl.keystore.password=<password>
> listener.name.<listener-name>.ssl.keystore.type=PKCS12
> listener.name.<listener-name>.custom-sasl-protocol.sasl.jaas.config=my.custom.loginmodule.CustomLoginModule required;
> listener.name.<listener-name>.sasl.enabled.mechanisms=CUSTOM_SASL_PROTOCOL
> ```
> 
> The only thing I wasn't sure about was whether the enabled.mechanisms could have custom values or whether you had planned to have a list of available SASL mechanisms that would have to choose from. We would need the ability to provide a custom value for this.
> 
> One other thing that would be really helpful would be the ability to provide multiple authentication mechanisms on the same listener. An example of this would be something like:
> 
> ```
> listener.name.<listener-name>.ssl.keystore.location=/tmp/kafka/cluster.keystore.p12
> listener.name.<listener-name>.ssl.keystore.password=<password>
> listener.name.<listener-name>.ssl.keystore.type=PKCS12
> listener.name.<listener-name>.custom-sasl-protocol.sasl.jaas.config=my.custom.loginmodule.CustomLoginModule required;
> listener.name.<listener-name>.scram-sha-512.sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required;
> listener.name.<listener-name>.sasl.enabled.mechanisms=CUSTOM_SASL_PROTOCOL,SCRAM-SHA-512
> ```
> 
> Having this capability isn't essential for us, because we could configure multiple listeners, each with a single authentication mechanism, and point the client at the relevant listener endpoint, depending on the authentication mechanism, but if this was something that could be easily incorporated, that would very useful.

Thanks for this @tmitchell10, just pushed with an example of your desired workflow. To my understanding, yes, it should be possible to support multiple SASL mechanisms enabled on a single listener.

031-custom_authentication.md
```diff
@@ -0,0 +1,198 @@
...
+              org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required ;
+          tlsTrustedCertificates: ...
+          secrets:
+            - name: example
+              mountPath: "/etc/example"
+```
+
+Then, when constructing the broker config, we’ll perform the following tasks:
+
+* `Principal`, if set, is set cluster-wide for all authentication methods. This is a limitation of Kafka, which only allows one principal to be specified for the entire cluster. If set, we need to ensure that no other listeners override this property, and if they do and are different, then fail-hard.
```
ppatierno, line 89: As you side it's cluster-wide and the user could specify the `principal.builder.class` in the `spec.kafka.config` section as well ... why just don't leave it there? Otherwise we have to deal with which configuration takes precedence on the other one and as you said different listeners using custom authn could specify different principal builder class but operator should raise error.

Sam-Kramer: That's a fair point @ppatierno! The primary reason why I left `principal.builder.class` to be defined at the listener level rather than `spec.kafka.config` is that currently the OAuth listener defines it at such level (https://github.com/strimzi/strimzi-kafka-operator/blob/ef576db7975dc40b19b559464caf1ff5f1b70071/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java#L212). I also think this makes sense, as otherwise users would have to configure their OAuth configuration in two separate locations, rather than just one.

ppatierno: My only concern is about the user making a mistake and setting it in the Kafka config section. The implementation should have a well known precedence order failing when there are such mistakes.

Sam-Kramer: Gotcha, that makes sense! Thankfully we already forbid users from setting the value in the Kafka config section: https://github.com/strimzi/strimzi-kafka-operator/blob/ef576db7975dc40b19b559464caf1ff5f1b70071/api/src/main/java/io/strimzi/api/kafka/model/KafkaClusterSpec.java#L45

ppatierno: Ok cool! I should have had a look to this before asking ;-) Thanks!

031-custom_authentication.md
```diff
@@ -0,0 +1,198 @@
...
+        tls: true
+        configuration:
+          brokerCertChainAndKey:
+            secretName: my-secret
+            certificate: my-certifciate.crt
+            key: my-key.key
+        authentication:
+          type: custom
+          sasl: true
+          principal.builder.class: SimplePrincipal.class
```
scholzj, line 72: What if the user has multiple listeners with custom authentication? I don't think that would be unexpected - e.g. one internal and one external etc. Would it make more sense to allow setting this in `.spec.kafka.config` instead of here?

031-custom_authentication.md
```diff
@@ -0,0 +1,198 @@
...
+* The protocol for this listener would be derived from the `tls` and `sasl`, which then would be appended to `listener.security.protocol.map`. For example, if `tls: true` and `sasl: true`, then the protocol will be `SASL_SSL`. 
+* Each configuration entry under `listener-config` would be pre-appended with `listener.name.<listener-name>`. 
+* `brokerCertChainAndKey` can be specified if a custom certificate wishes to be served from this listener. This automatically creates/shoves the specified certifciate into a keystore automatically for the user, as it's done for other authentication mechanisms today.
+* `tlsTrustedCertificates` functions identically to OAuth’s setting. This is needed for workflows which will use custom mTLS, or mTLS + SASL workflows, as they may require to add additional certificate authorities. Similar to `brokerCertChainAndKey`, a truststore is generated automatically for the user, with their provided certifciate authorities.
```
scholzj, line 93: I think this part is problematic:
```
a truststore is generated automatically for the user
```

How does the system know what is truststore and what is keystore (important for automatically creating it). I think to make this versatile, this maybe needs to be merged with the `secrets` field just as opaque list of secrets which will be mounted and the user has to deal with them. If the authentication needs truststore or keystore, it might need to be already in the secret.

031-custom_authentication.md
```diff
@@ -0,0 +1,198 @@
...
+            oauthbearer.sasl.server.callback.handler.class: server.class
+            oauthbearer.sasl.login.callback.handler.class: login.class
+            oauthbearer.connections.max.reauth.ms: 999999999
+            sasl.enabled.mechanisms: oauthbearer
+            oauthbearer.sasl.jaas.config: |
+              org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required ;
+          tlsTrustedCertificates: ...
+          secrets:
+            - name: example
+              mountPath: "/etc/example"
```
scholzj, line 84: I don't think the mount path should be free to specify. I think all these secrets should be mounted into a single subdirectory (similarly to for example done in KafkaConnect with external configurations). That will make it isolated and will also simplify validation of what is valid and what is not. Otherwise you might need to validate whether the mount points don't conflict between listeners, whether there are no security issue with the selected mount path etc.

Or is there any specific value in having this as a free configuration?

scholzj (commented):
3 more nits to clarify. But looks good otherwise.

031-custom_authentication.md
Sam-Kramer: I would prefer having `principal.builder.class` in `.spec.kafka.config`, but does not fit well with the current authentication setup. I.e., the current principal is derived from the OAuth listener, and moving the definition to `spec.kafka.config` would be a breaking change. It'd become a breaking change as it would require the end-user to configure their authentication scheme in two separate locations, rather than just the listener config currently. 

Obviously this hurts user experience as well, but that's not to say that this currently workflow doesn't hurt user experience as well.

scholzj: Right now, the user does not configure the principal anywhere => so why would it be breaking change? Also, if you keep this here, how will you deal with it being configured in multiple listeners ... possibly to different values? I think given it is a central value, it makes more sense to configure it centrally.

031-custom_authentication.md
Sam-Kramer: I'm totally fine with just requiring the end user with creating their truststores/keystores. What I was thinking here was  just leveraging what's already in place: https://github.com/strimzi/strimzi-kafka-operator/blob/main/docker-images/kafka/scripts/kafka_tls_prepare_certificates.sh#L79 & https://github.com/strimzi/strimzi-kafka-operator/blob/79e28149b6dd8dd7b61e0ed60aae6d26fb93b1b7/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java#L200-L206 & https://github.com/strimzi/strimzi-kafka-operator/blob/79e28149b6dd8dd7b61e0ed60aae6d26fb93b1b7/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java#L1482

031-custom_authentication.md
Sam-Kramer: Makes sense to me, and reduces the need to configure another thing. Only value I can see is it would be nice for deduplication, as in, I have two secrets of the same type (tls), and I can just mount them into different directories to avoid a clash. That said, I think this case is rare, and the end user could just rename their secrets. Also would be easy to adapt a rename function here, but I think for now we keep it simple, especially as there is a workaround (i.e. just rename your secret key).

031-custom_authentication.md
Sam-Kramer: `Right now, the user does not configure the principal anywhere => so why would it be breaking change?`
It's a breaking change for OAuth users as currently the principal is derived from the listener (https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java#L235). Therefore, if we moved this to be in `spec.kafka.config`, we'd need all users who have an OAuth listener to then update their `.spec.kafka.config` to include the OAuth principal. 

`Also, if you keep this here, how will you deal with it being configured in multiple listeners ... possibly to different values` Yes, we'll throw in this case (maybe we can perform webhook validation?), but agree that `makes more sense to configure it centrally`. I just don't see a good way of getting there, unless we're OK with the above "break".

scholzj: > `Right now, the user does not configure the principal anywhere => so why would it be breaking change?` It's a breaking change for OAuth users as currently the principal is derived from the listener (https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java#L235). Therefore, if we moved this to be in `spec.kafka.config`, we'd need all users who have an OAuth listener to then update their `.spec.kafka.config` to include the OAuth principal.

But OAuth doesn't allow users to configure the principal. It is configured automatically. So the users using OAuth or OAuth with SCRAM or mTLS will not need to configure anything as it is today. Only if they combine OAuth with the custom authentication, then it would be needed => but due to the custom authentication. So I'm still not sure where you see the breaking change.

Sam-Kramer: `But OAuth doesn't allow users to configure the principal. It is configured automatically. ` yeah this is the piece I think needs addressing -- how do we handle a user using OAuth, who also sets the principal in `spec.kafka.config`?

Sam-Kramer: In theory, this shouldn't happen, but I _think_ it's confusing from a user experience perspective. The principal is being set on behalf of the user, at a separate location from where the location exists. That said, if we still think this the best path, I'll go ahead and update the proposal.

scholzj: > But OAuth doesn't allow users to configure the principal. It is configured automatically. yeah this is the piece I think needs addressing -- how do we handle a user using OAuth, who also sets the principal in spec.kafka.config?

> In theory, this shouldn't happen, but I think it's confusing from a user experience perspective. The principal is being set on behalf of the user, at a separate location from where the configuration exists. That said, if we still think this the best path, I'll go ahead and update the proposal.

I'm not sure. You are right to some extent. But:
* I don't think it is confusing to say that when `principal.builder` is specified for user, it will be used and the user is responsible for how it works. But when it is not specified, the operator is managing it. I think that is more or less the common behaviour applied everywhere. In the listener, you can configure the advertised hosts and ports. If you don't set them, the operator will set them on its own. So I think that is ok.
* It does not break backwards compatibility => the existing configurations will work exactly as before. Users will just have a bit more freedom to shoot them self in the foot by being able to configure the principal builder. But the same configuration should still deliver the same result
* If we see this as a problem, what would it really mean? I guess it would mean that this proposal is problem as well because it allows for even more problems.
* What are the better alternatives? Having a value which exists only once configured multiple times per listener is not really an option as that is definitely way more confusing. So, can we do something like have our own Strimzi principal which would delegate to some per-listener principals? I think we discussed it in the past with @tombentley as something what might work. So we can look into that as well ...
  * You would configure the principal builder per-listener
  * Strimzi will always use its own principal builder which would basically do something like this:
    ```
    if (connectionFromListener1) {
        delegateToPrincipalBuilderFromListener1
    } else if (connectionFromListener2) {
        delegateToPrincipalBuilderFromListener2
    } else {
        useTheDefaultPrincipalBuilder
    }
    ```
    I guess this would need some PoC to proof it can work.

Sam-Kramer: `But when it is not specified, the operator is managing it. I think that is more or less the common behaviour applied everywhere` ok then this is fine then! I'll go ahead and update the proposal to allow such behavior.

031-custom_authentication.md
```diff
@@ -0,0 +1,200 @@
+# Custom Authentication
+
+This proposal focuses on supporting custom authentication for SASL/mTLS in the Strimzi operator.
```
scholzj, line 3: Maybe to make things a bit more clear later ... could you make it clear in the title and maybe in the first paragraph that it is the broker / server part? Maybe _Custom Authentication in Kafka brokers_ would be a more fitting name.

031-custom_authentication.md
```diff
@@ -0,0 +1,200 @@
+# Custom Authentication
+
+This proposal focuses on supporting custom authentication for SASL/mTLS in the Strimzi operator. 
+
+## Current situation
+
+Currently, numerous authentication methods are [supported](https://github.com/strimzi/strimzi-kafka-operator/tree/0.25.0/api/src/main/java/io/strimzi/api/kafka/model/authentication) in Strimzi, and allows the user to configure which one to use. In addition, in the case a custom *authorizer* needs to be used, this is also currently [supported](https://github.com/strimzi/strimzi-kafka-operator/blob/0.25.0/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaBrokerConfigurationBuilder.java#L541-L549).
+
+However, there is no ability to specify custom authentication, which is what this proposal will focus on.
```
scholzj, line 9: Again, just for clarity ... maybe you could add that it focuses only on the broker part.

031-custom_authentication.md
```diff
@@ -0,0 +1,200 @@
...
+        
+        // Create your own KafkaPrincipal here
+        ...
+    }
+}
+```
+
+Another thing to be mindful of, is to ensure your **CustomAuthorizer** supports `super.users`, and the default KafkaPrincipal to ensure seamless integration with Strimzi. This is relevant as in the case of using a custom principal builder, you’re most likely using a CustomAuthorizer as well. 
+
+
```
scholzj, line 167: The double empty line is probably not needed.

scholzj (approved):
I left some more comments, but there are more about nits such as making the proposal name more clear etc. So please implement them, but they are not about the technical side of things.

From the technical side, it looks more or less ok. I feel like during the implementation, we will run into some more details which will need to be clarified and I still wonder if we really want both `tlsTrustedCertificates` and `secrets`. But for me personally, this probably sufficiently covers what we want to achieve and how, so IMHO these details can be ironed out on the implementation PR if needed.

Note:
* This requires approval from other maintainers before it can be merged. @tombentley and @ppatierno would be good candidates I guess.
* In any case, you will need to fix the DCO sign-off before we can merge it (details and instructions are under the _Details_ link next to the DCO status)

Sam-Kramer:
> I left some more comments, but there are more about nits such as making the proposal name more clear etc. So please implement them, but they are not about the technical side of things.
> 
> From the technical side, it looks more or less ok. I feel like during the implementation, we will run into some more details which will need to be clarified and I still wonder if we really want both `tlsTrustedCertificates` and `secrets`. But for me personally, this probably sufficiently covers what we want to achieve and how, so IMHO these details can be ironed out on the implementation PR if needed.
> 
> Note:
> 
> * This requires approval from other maintainers before it can be merged. @tombentley and @ppatierno would be good candidates I guess.
> * In any case, you will need to fix the DCO sign-off before we can merge it (details and instructions are under the _Details_ link next to the DCO status)

Thank you @scholzj , appreciate your patience and feedback on the proposal. I've went ahead and made changes to the text as you suggested to be a bit more clear, as well as renaming the proposal.

031-custom_authentication_in_kafka_brokers.md
```diff
@@ -0,0 +1,199 @@
...
+        tls: true
+        configuration:
+          brokerCertChainAndKey:
+            secretName: my-secret
+            certificate: my-certifciate.crt
+            key: my-key.key
+        authentication:
+          type: custom
+          sasl: true
+          principal.builder.class: SimplePrincipal.class
```
ppatierno, line 127: isn't the principal cluster wide so that user has to set it under the `spec.kafka.config` (as the previous example)?

ppatierno (commented):
Just a couple of last nits but I have actually a question ...
How does this copes with `KafkaUser`? I mean, today in the `KafkaUser` we have the `authentication` field which can have values like `tls` and `scram-sha-512` related to the type of authentication you set on the listeners; the user operator will then generate Secret with information for authentication (password or certificates and corresponding stores). 
How to deal with a `custom` authentication in a listener from a `KafkaUser` resource point of view?

scholzj:
> How does this copes with KafkaUser? I mean, today in the KafkaUser we have the authentication field which can have values like tls and scram-sha-512 related to the type of authentication you set on the listeners; the user operator will then generate Secret with information for authentication (password or certificates and corresponding stores).
How to deal with a custom authentication in a listener from a KafkaUser resource point of view?

My assumption is that it doesn't cope with it in any way => when the users use the _custom_ authentication, they will need to handle user management on their own and not using the User Operator. I would also expect that in most cases, this will be used to integrate with some _external_ authentication systems anyway. So it will be similar to the current `type: oauth` authentication which also doesn't integrate into the User Operator in any way.

ppatierno:
@scholzj ok, fair enough. Thanks.

kfox1111:
@scholzj  does it preclude it though? I'd like to set principal.builder.class= enabling mapping something like
spiffe://example.com/kafka/<username> to <username> and then use KafkaUser with type=tls_external as normal.

ppatierno:
I think that using a different type of authentication in a `KafkaUser` would mean having the support in the UO for generating the corresponding Secret containing the information that the custom type of authentication needs (maybe not just username/passowrd, or TLS cert, it would depends on the authN type) that could be different from one authN to another so I think it's not simple to make it more general. I guess this is more or less the point that Jakub was making.

scholzj:
@kfox1111 Yeah, this should allow you to override the principal builder when using the `tls` authentication, so it should allow you to transform SPIFEE name to some format which can be used with the User Operator and either the `tls-external` authentication or no authentication. That should work.

@ppatierno This will not allow the User Operator to create any users. But if the usernames produced by the custom authentication are compatible with user operator, you can still use it to manage ACLs or Quotas for example.

tombentley (approved):

tombentley:
@Sam-Kramer sorry I've been so slow about reviewing this. Thanks for the proposal!

031-custom_authentication_in_kafka_brokers.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Workflow
+
+Add a CustomAuthenticationKafkaListener class, which would support the following properties.
+
+```
+listener.name.<listener-name>.<sasl-mechanism>.sasl.client.callback.handler.class=
+listener.name.<listener-name>.<sasl-mechanism>.sasl.server.callback.handler.class=
+listener.name.<listener-name>.<sasl-mechanism>.sasl.login.callback.handler.class=
+listener.name.<listener-name>.<sasl-mechanism>.connections.max.reauth.ms=
```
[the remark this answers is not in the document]
Sam-Kramer: Added this to `Affected / not affected projects`

Sam-Kramer:
I addressed requested nits, and renamed the file to have `032` instead of `031` prefix, as I see we have committed a different proposal with that number.

jorisaxual:
Hi all, really great to see this addition to Strimzi. Is it planned to be released as part of 0.27.0?

scholzj:
This proposal has been approved by 3 +1 votes (Tom Bentley, Paolo Patierno and Jakub Scholz) and is now approved. I'm going to merge it.

scholzj:
Thanks for the work on this @Sam-Kramer

PR #38, opened 2021-06-04
Proposal for service binding support in Strimzi

scholzj:
This proposal has covers the support for Service Bindings and Service Bindings operator. It was originally opened in strimzi/strimzi-kafka-operator#2753 by @AndrewJSchofield (who wrote the content). This is a copy of the original PR since the in the mean time the proposals moved into a separate repository. The original comments and discussions are in the original PR. Hopefully we will have more time to get back to this soon.

PR #43, opened 2021-11-04
Service binding proposal

katheris:
This proposal covers the support for Service Bindings and Service Bindings operator. It was originally opened in strimzi/strimzi-kafka-operator#2753 by @AndrewJSchofield (who wrote the content) and then copied to https://github.com/strimzi/proposals/pull/38 since in the mean time the proposals moved into a separate repository. The original comments and discussions are in the original PR.

The proposal has been updated to align with the newly released Service Binding specification version 1.0.0.


028-service-binding.md
```diff
@@ -0,0 +1,340 @@
...
+4. Application can bind to a single custom resource to get all information
+
+### 1 - Application binds to KafkaUser only
+
+In this approach the application creates a `ServiceBinding` for a specific `KafkaUser` CR. The credential information is bound to the application and the cluster certificate and bootstrap server address is obtained by the application in another way.
+
+Required changes:
+
+ - Update the `KafkaUser` CR to include a `status.binding` field where the `name` matches the current field `secret` in the status
+ - Update the secret created for the `KafkaUser` to include the username when the `KafkaUser` is using type SASL SCRAM
```
scholzj, line 41: This should not be a problem. I assume that means that the service binding will pass username / password and not the JAAS configuration? That would make it suitable with non-java clients. But make it harder to use with Java clients I guess.

028-service-binding.md
```diff
@@ -0,0 +1,340 @@
...
+
+### 4 - Application can bind to a single custom resource
+
+In this approach the application binds to a new custom resource type, e.g. called `KafkaConnection`.
+
+Required changes:
+
+ - Add a new Custom Resource called `KafkaConnection`
+ - The `KafkaConnection` spec has several fields, cluster, listener and user, that refer to the `Kafka`, the specific listener and the `KafkaUser` that should be used
+ - The Strimzi operator adds a `spec.binding` to the `KafkaConnection` status
```
scholzj, line 91:
```suggestion
 - The Strimzi operator adds a `status.binding` to the `KafkaConnection` status
```

028-service-binding.md
```diff
@@ -0,0 +1,340 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: my-user
+  labels:
+    strimzi.io/kind: KafkaUser
+    strimzi.io/cluster: my-cluster
+type: Opaque
+data:
+  password: Z2VuZXJhdGVkcGFzc3dvcmQ=
```
scholzj, line 267: For the completeness, it has the JAAS string now as well.

028-service-binding.md
```diff
@@ -0,0 +1,340 @@
...
+### 2 - Application uses multiple ServiceBinding resources
+
+In this approach the application creates two `ServiceBinding` resources, one that will bind to the `Kafka` CR and one that will bind to the `KafkaUser` CR.
+
+Required changes:
+
+- Update the `KafkaUser` CR to include a `status.binding` field where the `name` matches the current field `secret` in the status
+- Update the secret created for the `KafkaUser` to include the username when the `KafkaUser` is using type SASL SCRAM
+- Update the `Kafka` CR to include a `status.binding` field, where the `name` refers to a new secret containing both the cluster certificate and bootstrapServers
+- Add a new secret that contains the bootstrap servers for each listener and the cluster certificate information
```
scholzj, line 56: How does the user choose the listener? Will they need to use the `env` section in the `ServiceBinding` resource?

scholzj (commented):
Thanks for updating the proposal. I left some comments.

Is this expected the service binding to be used? Is it only within the single Kubernetes cluster? Also, does it work across namespaces? The reasons why this is important are:
* Being used only inside the same Kubernetes cluster might make it easier to identify the listener automatically. My general expectation would be that for given authentication method, only one internal listener might exist since I'm not sure about the value of having multiple of them. The additional listeners might be external - with different types and networks. But one would not use these from inside the Kube cluster normally.
* One of the main limitations we have is that all the resources usually end up in the same namespace. But users want to run the applications across many namespaces - e.g. KAfka in namespace X, one application in namespace Y, another one in namespace Z etc. And we currently don't handle this well. They need to copy the secrets around, keep them up to date etc. So, can the ServiceBinding do something like this?
  ``` yaml
  apiVersion: servicebinding.io/v1alpha3
  kind: ServiceBinding
  metadata:
    name: barista-kafka
  spec:
    workload:
      apiVersion: apps/v1
      kind: Deployments
      name: barista-kafka
    service:
      kind: Kafka
      name: my-cluster
      namespace: my-kafka-ns
  ```
  and copy the secrets from other namespaces? 
  
  If yes, it is great. If not, we should focus on options 1, 2 or 3. But if it does only in the same namespace, I think those options are useless because they will run into the same issue with the CRs living in one namespace but applications binding them living somewhere else. 
  
  So in that case, I think we should focus on the option 4 which can have its own operator handling the `KafkaConnection` resources in different namespaces and will just extract the information from the Kafka and KafkaUser resources in other namespace and create its own secret with the collected informations all in one. That might be more complicated and complex on the surface. But it would also add more value in dealing with the resources across the namespaces => so it would bring more than the ServiceBinding alone.

028-service-binding.md
katheris: As noted in the considerations, this would rely on the application developer writing custom logic to grab the listener they want from the ones that were mounted into the app using the secret.

I would expect we would end up with something like:
```
data:
  listeners:
    - name: internal
       bootstrapServers: kafka.svc:9092
    - name: external
       bootstrapServers: myhost.com:443
```

And then the app gets that mounted as a file and can find the one they want. Not ideal, but it's more useful than what we have today for service binding

katheris:
> Is this expected the service binding to be used?

I'm not sure what you mean here. The application defines a service binding CR and then any running service binding operator will look at the service referenced in the CR, check for the `status.binding` value and mount the listed secret into the app.

> Is it only within the single Kubernetes cluster? Also, does it work across namespaces?

Yes, the first version of the spec assumes the app and service are in the same Kubernetes cluster and within the same namespace. All the service binding operator does is mount existing secrets onto the app. It doesn't have any capability to copy those secrets to another namespace. The service binding spec community have talked about this capability but based on their previous discussions it is unlikely this will ever be added to the spec, although some specific service binding operators might add it as an extension.

> So in that case, I think we should focus on the option 4 which can have its own operator handling the KafkaConnection resources in different namespaces and will just extract the information from the Kafka and KafkaUser resources in other namespace and create its own secret with the collected informations all in one. That might be more complicated and complex on the surface. But it would also add more value in dealing with the resources across the namespaces => so it would bring more than the ServiceBinding alone.

I think this is a really good point which I hadn't thought much about. Since the service binding spec is focused on one namespace only I think you're right that we have an opportunity to get a better experience for both users of Strimzi and those wanting to combine Strimzi with Service Binding.

This makes me think the right thing to do would be:
 - Make the changes to the KafkaUser secret and CR so it can be used by service binding since these are small changes but immediately make Strimzi more compatible
 - Put in the work to create a different CR that is designed to be put into the namespace with your apps so that the apps can get connection details easily. And we could do some clever logic around automatically picking out a single listener, or letting you choose one if you do have more than one internal listener

028-service-binding.md
tombentley: I think the service binding spec allow the secret to contain other fields. It just requires that field with the names defined in the spec have the value/meaning defined in the spec.

028-service-binding.md
```diff
@@ -0,0 +1,340 @@
...
+In this approach the application creates a `ServiceBinding` for a specific `KafkaUser` CR. The credential information is bound to the application and the cluster certificate and bootstrap server address is obtained by the application in another way.
+
+Required changes:
+
+ - Update the `KafkaUser` CR to include a `status.binding` field where the `name` matches the current field `secret` in the status
+ - Update the secret created for the `KafkaUser` to include the username when the `KafkaUser` is using type SASL SCRAM
+
+Considerations:
+
+ - In this approach a lot of work is still left up to the user to discover the cluster certificate and the bootstrap server address to use.
```
tombentley, line 45: @katheris can you elaborate the additional work?

028-service-binding.md
```diff
@@ -0,0 +1,340 @@
...
+
+- Update the `KafkaUser` CR to include a `status.binding` field where the `name` matches the current field `secret` in the status
+- Update the secret created for the `KafkaUser` to include the username when the `KafkaUser` is using type SASL SCRAM
+- Update the secret created for the `KafkaUser` to include the bootstrap servers for each listener
+- Update the secret created for the `KafkaUser` to include the cluster certificate information
+
+Considerations:
+
+- This approach results in a lot more information being added to each `KafkaUser` secret
+- It does not help the user to decide which listener to use
```
tombentley, line 77: It could do, if the `KafkaUser` CR supported a `spec.binding.listener` field.

028-service-binding.md
```diff
@@ -0,0 +1,340 @@
...
+
+Possible extensions:
+
+ - Could only add bootstrap server addresses for listeners that support the authentication type that the `KafkaUser` supports
```
tombentley, line 81: This kind of validation is good. Using the `spec.binding.listener` idea failure could be a reason in an unready condition.

028-service-binding.md
```diff
@@ -0,0 +1,340 @@
...
+Required changes:
+
+ - Add a new Custom Resource called `KafkaConnection`
+ - The `KafkaConnection` spec has several fields, cluster, listener and user, that refer to the `Kafka`, the specific listener and the `KafkaUser` that should be used
+ - The Strimzi operator adds a `spec.binding` to the `KafkaConnection` status
+ - The Strimzi operator add a new secret containing all the required information
+ 
+Considerations:
+
+- This approach results in a new Custom Resource type that is just for convenience
```
tombentley, line 96: I would question how convenient it really is over the `KafkaUser` CR supporting a `spec.binding.listener` idea. It requires extra CRs, and keeping the references straight. It doesn't really reduce the coupling to the service binding spec (i.e. we'd have CRs which had such a coupling in either case).

tombentley (commented):
Thanks @katheris I left a few comments.

scholzj:
> > Is this expected the service binding to be used?
> 
> I'm not sure what you mean here. The application defines a service binding CR and then any running service binding operator will look at the service referenced in the CR, check for the `status.binding` value and mount the listed secret into the app.

Sorry, that was supposed to say _Is this expected the service binding to be used only within the single Kubernetes cluster?_ :-)

> > Is it only within the single Kubernetes cluster? Also, does it work across namespaces?
> 
> Yes, the first version of the spec assumes the app and service are in the same Kubernetes cluster and within the same namespace. All the service binding operator does is mount existing secrets onto the app. It doesn't have any capability to copy those secrets to another namespace. The service binding spec community have talked about this capability but based on their previous discussions it is unlikely this will ever be added to the spec, although some specific service binding operators might add it as an extension.
> 
> > So in that case, I think we should focus on the option 4 which can have its own operator handling the KafkaConnection resources in different namespaces and will just extract the information from the Kafka and KafkaUser resources in other namespace and create its own secret with the collected informations all in one. That might be more complicated and complex on the surface. But it would also add more value in dealing with the resources across the namespaces => so it would bring more than the ServiceBinding alone.
> 
> I think this is a really good point which I hadn't thought much about. Since the service binding spec is focused on one namespace only I think you're right that we have an opportunity to get a better experience for both users of Strimzi and those wanting to combine Strimzi with Service Binding.
> 
> This makes me think the right thing to do would be:
> 
> * Make the changes to the KafkaUser secret and CR so it can be used by service binding since these are small changes but immediately make Strimzi more compatible
> * Put in the work to create a different CR that is designed to be put into the namespace with your apps so that the apps can get connection details easily. And we could do some clever logic around automatically picking out a single listener, or letting you choose one if you do have more than one internal listener

I don't think enhancing the KafkaUser would be an issue. But seems almost like a wasted effort. It seems to me like we can have some `strimzi-binding` operator which will manage the `KafkaConnection` CR looking something like this:

```yaml
kind: KafkaConnection
metadata:
  name: barista-kafka
spec:
  kafka:
    kind: Kafka
    name: my-cluster
    namespace: my-namespace # optional
    listener: # optional
      name: tls # optional
  user:
    kind: KafkaUser
    name: my-barista
    namespace: my-namespace # optional
status:
  binding:
    name: barista-kafka
```

It could collect the information from the Kafka CR and KafkaUser CR and their secrets and create the secrets with it in the right namespace to be used by the Service Binding / or by mounting the values directly.

028-service-binding.md
katheris: Yes a service binding operator will mount all fields in the secret, so Java clients could use the JAAS configuration and non-Java clients can use the username/password

028-service-binding.md
katheris: If the user has authorization turned off, then they wouldn't normally have to create a KafkaUser, or use the entity operator, but to make this work they would. I think the actual advantage of this is where Jakub was suggesting that this new CR could live in a different namespace to the install of Strimzi. Is it possible to put your KafkaUser CR in a different namespace to the Strimzi instance it refers to?

tombentley: The `KafkaUser` operator can watch a different namespace than the one it's deployed in, but the `Secrets` will currently always be in the same namespace as the `KafkaUser` CR. But there's nothing stopping us from allowing the `KafkaUser` spec to nominate a (set of) namespace(s) to publish service-binding compatible `Secrets` to. Or doing that at the operator level (e.g. add `PUBLISH_NAMESPACE` alongside `WATCHED_NAMESACE`).

I'm not necessarily opposed to a new CR, but I'd like to understand why that's better than using an improved `KafkaUser`.

scholzj: > The `KafkaUser` operator can watch a different namespace than the one it's deployed in, but the `Secrets` will currently always be in the same namespace as the `KafkaUser` CR. 

The typical use-case is that you have application per-namespace. So for 99% of cases, the _watch a different namespace_ feature is useless since it changes the namespace, but still keeps everything in one namespace.

> But there's nothing stopping us from allowing the `KafkaUser` spec to nominate a (set of) namespace(s) to publish service-binding compatible `Secrets` to. Or doing that at the operator level (e.g. add `PUBLISH_NAMESPACE` alongside `WATCHED_NAMESACE`).

I think one of the challenges is that you cannot use garbage collection, so we will be leaving the secrets behind. (dangerous with TLS where the certificate works even after the user is deleted). It is also a bit messy from the RBAC perspective. it means that the CO will need to have the rights this `strimzi-binding` operator might need. Users might be also asking for a weird things => such as "I want to give it access to secrets in these 10 namespaces but not whole cluster. If CO does the RBAC setup for UO, this will be hard to do. 

So I feel like separation of concerns makes better sense. And it solves the namespace issue for us as well to some extent. 

The advantage of using KafkaUser would be especially if we one day really implement multi-namespace support for the UO.

028-service-binding.md
tombentley: Thanks! Could you add a sentence or two to the proposal itself?

ppatierno:
Looking at the Service Binding CRD spec [1], the `service` field doesn't provide the possibility to specify the namespace where the service related resource lives (i.e. `KafkaUser`, `Kafka` or `KafkaConnection` custom resource), so it sounds to me that the service resource (i.e. `KafkaUser`), the corresponding `Secret` and the `ServiceBinding` resource have to leave all together in the same namespace, right? And following the solution of `KafkaUser` doesn't sound great to me due to the issue Jakub mentioned about copying stuff in different namespaces.
I think that putting more effort but centralizing all connection information in one `KafkaConnection` resource is much better and clear for developers as well.

I also have one question which is not clear when the use case is with the Kafka cluster having TLS enabled and the user want to use TLS client authentication, so doing mTLS overall: the service binding spec is clear about the `private-key` field (private key for the client authN) but not about the `certificates` field.
This `certificates` field is defined as "A collection of PEM-encoded X.509 certificates, representing a certificate chain used in mTLS client authentication". In my specific use case I would need a place to put the CA certificate for TLS/encryption but even the client certificate (public key) for TLS client authN and they would not fit in the same chain (the client certificate would be signed by a different CA as it happens in Strimzi). 
The `certificates` field is really not clear to me about how to use in this case, but anyway the spec says that you could use any other field just not using the "private" ones.

https://github.com/servicebinding/spec/blob/main/servicebinding.io_servicebindings.yaml#L67

028-service-binding.md
ppatierno: > If the user has authorization turned off, then they wouldn't normally have to create a KafkaUser, or use the entity operator, but to make this work they would.

I think this is the bigger point about not using the `KafkaUser`. With solution 4, the `KafkaConnection` would be anyway created with no auth data but just the listener ones and it would make more sense to me.

028-service-binding.md
katheris: @ppatierno @scholzj How is this field designed to be used? Most commonly in Java applications I have seen the following configuration:
```
sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required username="<username>" password="<password>";
```

ppatierno: the content of the field is exactly something like you showed above. See the example in the official Strimzi doc [1].
In your Java application you can just get the value of this field and setting it for the "sasl.jaas.config" property (in your Properties map). From that example, if you decode the base64 value you get "org.apache.kafka.common.security.scram.ScramLoginModule required username="my-user" password="generatedpassword";

[1] https://strimzi.io/docs/operators/latest/using.html#scram_sha_512_authentication

028-service-binding.md
katheris: Done

028-service-binding.md
katheris: I've added this as a suggested enhancement

028-service-binding.md
katheris: Ah yes, I forgot to base64 decode the value that I was looking at in my secret!

028-service-binding.md
```diff
@@ -0,0 +1,356 @@
...
+
+Required changes:
+
+ - Update the `KafkaUser` CR to include a `status.binding` field where the `name` matches the current field `secret` in the status
+ - Update the secret created for the `KafkaUser` to include the username when the `KafkaUser` is using type SASL SCRAM
+
+Considerations:
+
+ - In this approach a lot of work is still left up to the application developer to make sure their application has the correct cluster certificate and bootstrap server address.
+   - Either they would have to look up the cluster certificate and bootstrap server address and hard code it into their application or a secret in their cluster, or they would have to write custom logic into their app so that it can query the Kafka CR.
```
pedjak, line 46: +1 and this speaks strongly against the approach.

katheris:
> I also have one question which is not clear when the use case is with the Kafka cluster having TLS enabled and the user want to use TLS client authentication, so doing mTLS overall: the service binding spec is clear about the private-key field (private key for the client authN) but not about the certificates field.
This certificates field is defined as "A collection of PEM-encoded X.509 certificates, representing a certificate chain used in mTLS client authentication". In my specific use case I would need a place to put the CA certificate for TLS/encryption but even the client certificate (public key) for TLS client authN and they would not fit in the same chain (the client certificate would be signed by a different CA as it happens in Strimzi).
The certificates field is really not clear to me about how to use in this case, but anyway the spec says that you could use any other field just not using the "private" ones. So I guess we are just going to use our fields like ca.p12, user.p12, and so on.

I've been asking on the Service Binding slack (https://kubernetes.slack.com/archives/C012F2GPMTQ/p1636626317015600) and my understanding is that the `certificates` is meant to hold a single certificate, but as you say that won't be sufficient for ours, and perhaps many others, usecase. It is an area that the community is aware can be improved. So for this proposal I think we would define our own fields like you suggested, but if we name them generically there is an opportunity for us to try and contribute them back into the spec.

ppatierno:
@katheris it's really strange they used plural, then talking about chain of certificates but it meant to represent just one certificate :-) Btw I agree with you, we can propose our own fields for their spec, taking the `privatekey` into account as well which is so much general as well as the `certificates` one.
Are you going to propose our own fields here and then moving that proposal to their spec as well?

katheris:
>Are you going to propose our own fields here and then moving that proposal to their spec as well? 
I'll update the proposal to include proposed fields and then once we've agreed them here I will go back to the service binding spec and also propose them there

028-service-binding.md
pedjak: for the record, k8s secret data are flat, i.e. key values are just simple strings, and if the values are anyhow structured, the application should need to parse them appropriately. 

BTW, does the app really need to have a few bootstrap servers to chose from? Should not we simply project just enough connection data (e.g. single bootstrap server and certificates if needed) and the app just take it and try to connect? Similar to what we do in jdbc world?

028-service-binding.md
```diff
@@ -0,0 +1,415 @@
...
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaConnection
+metadata:
+  name: barista-kafka
+spec:
+  kafka:
+    apiVersion: kafka.strimzi.io/v1beta2
+    kind: Kafka
```
pedjak, line 120: not sure if these two fields are really needed, because `KafkaConnection` refers implicitly `Kafka` resource.

028-service-binding.md
```diff
@@ -0,0 +1,415 @@
...
+    name: my-cluster
+    namespace: my-namespace # optional
+    listener: # optional
+      name: tls # optional
+  user: # optional
+    apiVersion: kafka.strimzi.io/v1beta2
+    kind: KafkaUser
```
pedjak, line 127: same as above.

028-service-binding.md
katheris: > for the record, k8s secret data are flat

Ah I hadn't realised that the secret data had to be flat, so I guess any structure would have to be encapsulated in a string.

> BTW, does the app really need to have a few bootstrap servers to chose from?

Strimzi allows users to expose multiple different endpoints for applications to connect to. The user can configure endpoints that are only accessible within the cluster, or accessible externally, or have different security mechanisms. There is no way for Strimzi to know which listener the application will want to use. If there is only one internal listener then Strimzi could guess it is that one, but there could be more than one.

As noted in the considerations, approach number 2 is not ideal for this very reason - it still leaves the application work to do to figure out which endpoint to use. Approaches 3 and 4 both include options that would allow Strimzi to give the application only a single bootstrap server to call.

028-service-binding.md
```diff
@@ -0,0 +1,415 @@
...
+Naming of secret keys:
+
+The names for the keys given above match what is currently provided in other secrets created by Strimzi. The service binding specification does included some suggested fields for certificates, but they do not satisfy the requirements of an application connecting to Strimzi. These fields could be proposed back to the Service Binding specification.
+
+An alternative approach for naming is to name the fields to match the use, for example `ca` -> `truststore` (`truststore.crt`, `truststore.p12`, `truststore.password`), `user` -> `keystore` (`keystore.p12`, `keystore.password`, `keystore.crt`, `keystore.key`)
+ 
+Considerations:
+
+- This approach results in a new Custom Resource type that is just for convenience
+- The application developer still has to decide which cluster, user and listener to choose
```
pedjak, line 170: How application developer decide about that? The app needs to use data projected through `KafkaConnection` to connect to the cluster.

028-service-binding.md
pedjak: > I think the actual advantage of this is where Jakub was suggesting that this new CR could live in a different namespace to the install of Strimzi.

I would see that as a big plus, because application needs to get access only to KafkaConnection living in the same namespace as the app. Speaking about garbage collection - by l8s there are no ownership relationship across namespaces, but it can be tracked by Kafka and User operator, so whenever of them gets removed, `KafkaConnection` can be cleaned up as well.

028-service-binding.md
katheris: The application developer has to put the correct Kafka cluster, Kafka user and potentially listener into their KafkaConnection object. You could argue this is the job of the person provisioning the Kafka, but either way someone has to create the KafkaConnection object with those three pieces in.

028-service-binding.md
katheris: I've updated the examples

028-service-binding.md
katheris: We could certainly make these optional, but the Strimzi operator does have multiple apiVersions it supports and this would mean if Strimzi has been deployed with a differently named CRD for some reason the operator can cope with that.

028-service-binding.md
katheris: @pedjak I've edited this line of the proposal, hopefully it is clearer now

028-service-binding.md
```diff
@@ -55,18 +55,23 @@ Required changes:
...
-  - e.g. for listeners the secret data could be formatted like:
+  - e.g. for listeners, the secret data could be formatted to have one string with all listeners:
     ```yaml
     data:
-      listeners:
-        - name: internal
-          bootstrapServers: kafka.svc:9092
-        - name: external
-          bootstrapServers: myhost.com:443
+      listeners: aW50ZXJuYWxfa2Fma2Euc3ZjOjkwOTIsZXh0ZXJuYWxfbXlob3N0LmNvbTo0NDM= # when base64 decoded something like internal_kafka.svc:9092,external_myhost.com:443
```
pedjak, line 61: why not using just json, instead of encoding it?

katheris: From the Kubernetes documentation "the values for all keys in the data field have to be base64-encoded strings.", but we could use a json string and base64 encode that. I think if we did go for this approach we would have to think carefully about the best format to use.

pedjak: > From the Kubernetes documentation "the values for all keys in the data field have to be base64-encoded strings."

The values are encoded automatically at secret creation and decoded automatically at secret projection/mounting. Hence, for communication/documentation purpose, one can define keys under `stringData`, similar to https://kubernetes.io/docs/concepts/configuration/secret/#bootstrap-token-secrets

028-service-binding.md
```diff
@@ -0,0 +1,420 @@
...
+- It does not help the user to decide which listener to use. The application would need to parse the secret to pick out the correct listener from the list
+
+Possible extensions to allow only one bootstrap server to be listed:
+
+ - Could only add bootstrap server addresses for listeners that support the authentication type that the `KafkaUser` supports
+ - Could update the `KafkaUser` CR to include `spec.binding.listener` field which would then determine the listener bootstrap server address that is added to the secret
+
+### 4 - Application can bind to a single custom resource
+
+In this approach the application binds to a new custom resource type, e.g. called `KafkaConnection`.
```
PaulRMellor, line 106: Share Paolo's concern on name being confused with Kafka Connect. Is something like `KafkaServiceBinding` feasible?

scholzj: Well, as far as I understood it:
* This is not a service binding => this is a resource the service binding will bind to
* The Secret this creates is usable outside of Service binding AFAIU => so we should not tie the name to it

I'm not sure I see the confusion => while the names are similar, they will be used in different contexts a possibly by different personas.

ppatierno: I totally agree that the `Secret` related to this binding resource is actually usable outside of the service binding and it's really useful to users; this could even be generated out of the needs for the service binding operator.
The `KafkaConnection` resource is instead a resource that makes sense only in the service binding operator land, or?
As from the proposal:

> The Strimzi operator adds a `status.binding` to the `KafkaConnection` status that points to a Kubernetes `Secret`

So the `KafkaConnection` is really something related to the support for service binding operator. 
Then, on the service binding specification when they mention the `ServiceBinding` resource they have two main info "workload" (the actual user application) and the "service" (in our case Kafka).
Would it make more sense to call use `KafkaService` instead?

028-service-binding.md
scholzj: I don't know ... I wonder about the usefulness of the `apiVersion` and `kind` fields as well. In particular in the relationship to the implementation. Will we use raw resources to be able to grab the information from any resource which has matching status structure? I don't think we can really do that easily. So will it end up as something what supports selected options? e.g. `kind` will always be `Kafka` and `apiVersion` will always be `kafka.strimzi.io/v1beta2` and maybe one day we add `kafka.strimzi.io/v1` as supported as well?

scholzj (commented):
Left some more comments about some specific details. 

I also think we should cover a bit more about the implementation. Not necessarily tech-design level details. For example my view is that this is more or less independent component which should have its own sub-project / GitHub repo, its own release cycle etc. It should not be needed to release this everytime Strimzi releases this unless the API versions change basically. But not sure everyone sees it the same way. So it might be good to clarify. Similarly, maybe the technologies used might be worth clarifying here. E.g. if it is separate project, do we use Strimzi's code base (crd-generator, etc.)? Or do we use operator framework etc.

028-service-binding.md
```diff
@@ -0,0 +1,412 @@
...
+    password: # SCRAM password
+    sasl.jaas.config: # sasl jaas config string for use by Java applications
+  # Provided if selected user is mTLS:
+    user.p12: # client certificate for the consuming client PKCS #12 archive file for storing certificates and keys
+    user.password: # password for protecting the client certificate PKCS #12 archive file
+    user.crt: # certificate for the consuming client signed by the clients' CA
+    user.key: # private key for the consuming client
+```
+
+Naming of secret keys:
```
pedjak, line 86: In addition to the above keys, the service binding specification requires presence of `type` key and an optional `provider` key: https://github.com/servicebinding/spec#provisioned-service 

In this case, `type` can be equal to `kafka`, helping clients to understand what kind of connection should be made, and `provider` could be set to `strimzi`, helping even further clients to understand what other keys might appear and what their semantic is.

028-service-binding.md
katheris: I agree that the Kind seems unlikely to change, but I do wonder about the apiVersion, since as you say we may find that the `KafkaConnection` apiVersion doesn't match the `Kafka` apiVersion at some point, so allowing it to be specified means we wouldn't need to change anything at that point.

028-service-binding.md
```diff
@@ -0,0 +1,412 @@
...
+
+* **host** - hostname or IP address
+* **port** - port number
+* **userName** - username, optional
+* **password** - password or token, optional
+* **certificate** - certificate, optional
+
+Some of this comes from the `Kafka` CR and some from `KafkaUser`. Below is detailed the requirements for different configurations of Strimzi and why the current behaviour does not conform to the Service Binding spec.
+
+## Binding to a Kafka cluster with no TLS or authentication
```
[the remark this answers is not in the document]
katheris: I included it initially so that those who are unfamiliar with Strimzi (primarily those from the Service Binding spec), understood the starting point of what we have at the moment. Since we selected an approach already I can remove these sections if you feel they are no longer required?

028-service-binding.md
scholzj: > The KafkaConnection resource is instead a resource that makes sense only in the service binding operator land, or?

You need the `KafkaConnection` to get the Secret. So they are tied together. So if you wanna use the secret without the Service Binding operator, you still need to use the KafkaConnection resource => you create the KafkaConnection resource and the KafkaConnection operator will create the secret.

> Would it make more sense to call use `KafkaService` instead?

I think that this is more confusing than `KafkaConnection` -> it mixes with Kubernetes `Service`, so I would expect that to create some endpoint and not just a secret. Although I guess this could be individual.

028-service-binding.md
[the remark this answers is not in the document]
scholzj: It seems disconnected they way it is included right now - it does not seem to link to anything particular. So I would say we should do one of:
* Just delete it
* Merge this into the rejected alternatives since this explains to some extent the logic when coming up with them and the current state
* Give it some _introduction_ to explain what it is and why is it here

It is not a big deal ... I can live with it as it is as well if needed. It just seemed to stand out to me as a weird part - almost if it was something what was supposed to be deleted but wasn't.

028-service-binding.md
katheris: Thanks @pedjak I hadn't spotted that, I've updated the proposal to add them.

028-service-binding.md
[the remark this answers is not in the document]
katheris: I've been finding it useful to refer back to to remember the current state of affairs with where all the connection details are while we discuss, so I've condensed it down and written an introduction to explain why it's there.

028-service-binding.md
katheris: I'm going to mark this conversation as resolved since this option has been rejected.

028-service-binding.md
ppatierno: > You need the KafkaConnection to get the Secret. So they are tied together.

They are tied together from a service binding operator point of view. It's the KafkaConnection referring to the Secret via the `service.binding` field not the other way around. So if you want to use the Secret alone you don't need the KafkaConnection.

Maybe it's more related to how we envisage the implementation as you were mentioning. It's a new operator creating both KafkaConnection and Secret, starting from the CO resources, right? Why you said something like "you create the KafkaConnection resource and the KafkaConnection operator will create the secret." ?

Ragarding the name I think that `KafkaService` is really more consistent with the Service Binding, their spec and their examples and it's really just related to it.

028-service-binding.md
```diff
@@ -0,0 +1,374 @@
...
+7. The operator will then create a secret and add the bootstrapServers for the chosen listener to that secret.
+8. If the chosen listener uses TLS, the operator will then perform a GET request on the `cluster-ca-cert` secret and put the certificate and truststore files in the new secret.
+9. If the `KafkaConnection` references a `KafkaUser`, the operator will perform a GET request on that user to find the name of the secret containing the credentials and the username 
+from the status if it is provided. The operator will also perform a GET request on the user secret to get any credentials. 
+10. The operator will add the cluster certificate and any user related credentials to the secret it created earlier.
+11. The operator will add a `status.binding` to the `KafkaConnection` resource with the name of the secret. 
+12. The operator will then create `Watch` requests for the `Kafka` CR and, if used earlier, the `cluster-ca-cert` secret, the `KafkaUser` resource and the secret created by the `KafkaUser` operator. 
+13. If the `KafkaConnection` CR or any of the other watched resources change the operator will make the matching updates to the secret it created.
```
scholzj, line 117: Why is this approach better than completely separate operator in a completely separate repo? Given how long the current build pipeline runs we should try to look more into how to split things rather than add more.

I would also suspect that this has:
* Different release cycle (most Strimzi releases don't really change anything with regards to the `KafkaConnection`. Only new CRD versions might change it)
* Different installation mode (you cn have this for example installed as cluster wide while hacing the cluster operator installed as namespaced)

028-service-binding.md
```diff
@@ -0,0 +1,374 @@
...
+metadata:
+  name: barista-kafka
+spec:
+  kafka:
+    apiVersion: kafka.strimzi.io/v1beta2
+    kind: Kafka
+    name: my-cluster
+    namespace: my-namespace # optional
+    listener: # optional
+      name: tls # optional
```
tombentley, line 49: When not given what does it default to?

028-service-binding.md
```diff
@@ -0,0 +1,374 @@
...
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaConnection
+metadata:
+  name: barista-kafka
+spec:
+  kafka:
+    apiVersion: kafka.strimzi.io/v1beta2
+    kind: Kafka
+    name: my-cluster
+    namespace: my-namespace # optional
```
tombentley, line 47: When not given presumably it defaults to the same ns as the `KafkaConnection` itself? (If so you should say that)

028-service-binding.md
```diff
@@ -0,0 +1,374 @@
...
+
+At a high level these are the current problems that make integrating Strimzi with a Service Binding Operator hard:
+
+1. Bootstrap Server information is not contained in any secret
+2. Information is spread across multiple `Secret` resources
+3. The user has to determine which `KafkaUser` can access which listener
+
+## Proposal
+
+This proposal introduces a new custom resource type called `KafkaConnection` and the application creates a `ServiceBinding` that binds to this service.
```
tombentley, line 27: Sorry, I can't remember if we reached a consensus already, but did we prefer `KafkaConnection` over `KafkaService`?

028-service-binding.md
```diff
@@ -0,0 +1,374 @@
...
+
+`Secret` created by Strimzi:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: barista-kafka
+  namespace: my-namespace
+type: Opaque
```
tombentley, line 68: My reading of the [spec](https://github.com/servicebinding/spec#provisioned-service) is that this should be `servicebinding.io/kafka`.

028-service-binding.md
```diff
@@ -0,0 +1,374 @@
...
+Naming of secret keys:
+
+The names for the keys given above match what is currently provided in other secrets created by Strimzi. The service binding specification does included some suggested fields for certificates, but they do not satisfy the requirements of an application connecting to Strimzi. These fields could be proposed back to the Service Binding specification.
+
+An alternative approach for naming is to name the fields to match the use, for example `ca` -> `truststore` (`truststore.crt`, `truststore.p12`, `truststore.password`), `user` -> `keystore` (`keystore.p12`, `keystore.password`, `keystore.crt`, `keystore.key`)
+ 
+
+### Implementation
+
+This proposal will result in a new `KafkaConnectionAssemblyOperator` class being added to the `cluster-operator` directory in the `strimzi-kafka-operator` repo. The class will be
```
tombentley, line 97: It's a total aside, but I'm fed up with the confusing class naming we've ended up with. `KafkaConnectionController` (and renaming the other `*AssemblyOperator` to be `*Controller` and the various "resource operators" to be `*Operations` would help us when talking about our code, and likely be more understandable for people new to the project). Wdyt?

028-service-binding.md
```diff
@@ -0,0 +1,374 @@
...
+An alternative approach for naming is to name the fields to match the use, for example `ca` -> `truststore` (`truststore.crt`, `truststore.p12`, `truststore.password`), `user` -> `keystore` (`keystore.p12`, `keystore.password`, `keystore.crt`, `keystore.key`)
+ 
+
+### Implementation
+
+This proposal will result in a new `KafkaConnectionAssemblyOperator` class being added to the `cluster-operator` directory in the `strimzi-kafka-operator` repo. The class will be 
+deployed as a separate Verticle and have its own reconcile loop.
+
+The actions taken during the reconcile loop will be:
+1. On start-up the operator will watch all resources of Kind `KafkaConnection`.
```
tombentley, line 101: Just to be clear, this is always in all namespaces, right?

028-service-binding.md
katheris: I went for this approach initially as it matches the approach we have for other operator such as the KafkaConnect operator. Integrating it into the existing operator means we don't have to maintain a separate build and packaging system. However, I do think it's a good point that this will likely be on a different release cycle.

The point I hadn't considered was the different installation mode. Allowing the KafkaConnection operator to be cluster wide and the cluster operator to be namespaced I think is a big enough advantage that perhaps it outweighs the additional overhead.

At the moment the logic to handle reconcile loops and watches is all contained in the operator-common directory. We could look at creating a new project that still pulls in that code somehow. Alternatively we could look at operator-sdk. Operator-sdk doesn't currently support Java but this operator only has to watch Kubernetes resources, it doesn't have to talk to Kafka so there isn't really a reason why it has to be in Java.

028-service-binding.md
```diff
@@ -0,0 +1,374 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: barista-kafka
+  namespace: my-namespace
+type: Opaque
+  data:
+    type: kafka
+    provider: strimzi
+    bootstrapServer: # comma separated list of host:port
```
tombentley, line 72: Maybe worth mentioning that we use `bootstrapServer` rather than the more standard `host` and `port` because most Kafka clients are configured with a list of bootstrap servers, so using `host` and `port` would be more inconvenient.

028-service-binding.md
```diff
@@ -0,0 +1,374 @@
...
+The actions taken during the reconcile loop will be:
+1. On start-up the operator will watch all resources of Kind `KafkaConnection`.
+2. For any `KafkaConnection` resource present it will perform a GET request on the Kafka cluster that is 
+referenced in the `KafkaConnection` and pick out the configured listeners.
+3. If there are multiple listeners listed in the `Kafka` CR with the type `internal`, the operator 
+will filter the list by comparing the `tls` and `authentication` properties in the `Kafka` and `KafkaUser` CRs.
+4. If there are still multiple listeners it will pick one at random.
+5. The operator will add the name of the listener chosen will be added to the `KafkaConnection` status so that in future reconciles 
+the operator can take this into account in step 4 and avoid cycling through listeners.
+6. If there are no listeners marked as type `internal`, the reconcile loop will fail.
```
tombentley, line 109: So we can't bind to an external listener? Is there a good reason for that?

tombentley (commented):
A bunch of questions, but I think they'll be easy to answer.

028-service-binding.md
ppatierno: I am curious about this as well ... why the bootstrap server in the Secret can't be one for an external listener?

028-service-binding.md
ppatierno: If we want to develop a new operator in Java then this is the way to go https://javaoperatorsdk.io/

028-service-binding.md
katheris: I've currently left `KafkaConnection` everywhere, because I don't think it's been decided.

028-service-binding.md
katheris: When the listener field isn't supplied, the operator will pick a listener based on the steps in the implementation. The name field should be required if the listener field is supplied.

tombentley: Yeah, I think I found it in the end, it was just an unresolved question at this point in the text.

```suggestion
      name: tls # optional, see below
```

028-service-binding.md
katheris: I agree that the names are confusing, I don't know what `AssemblyOperator` actually means. I think `Controller` would make it more obvious what the purpose of the class is.

028-service-binding.md
katheris: The reason I restricted it to an internal listener is because from a service binding point of view the application is expected to bind to the secret, so the application will be running in the cluster. Can operators watch resources in other Kubernetes clusters? If they can that would be a reason to open it up to external listeners as well

028-service-binding.md
katheris: I hadn't realised there was a Java operator-sdk, I had heard rumours but it wasn't linked from the operator-sdk site so I didn't spot it

028-service-binding.md
katheris: I have re-read the spec and I agree, I'll update it.

028-service-binding.md
katheris: @ppatierno @scholzj @tombentley any more thoughts on the naming of the new CR?

028-service-binding.md
katheris: I've updated the proposal to make it clearer.

028-service-binding.md
ppatierno: My personal feeling is that `KafkaService` fits better in the service binding picture, but it's just my opinion of course.

tombentley: I'm also inclined to `KafkaService`, since it's not really a "connection" in a meaningful sense (sure, the information necessary to make a connection, but that's not the same thing), and it's for _service_ binding.

scholzj: Its not a service either ;-). I think `KafkaConnection` better represents what do you use it for.

scholzj: And again ... this is not really depending in any way on the Service Binding. The `Kafkaconnection` or whatever the name is super useful on its own without the Service Binding => it gives you the secret and you can just mount it and user it without the Service Binding operator.

katheris: I prefer `KafkaService` over `KafkaConnection` because if I was looking for the address of where my Kafka was in Kubernetes I would look for a `Service`. Going in a different direction though we could go with `KafkaBindingRequest`, as it's not just the Service Binding spec that uses binding to mean tying an application to a service, and you are making a request to get the information you need to bind to the Kafka.

ppatierno: I still prefer `KafkaService` ...

katheris: Throwing some other ideas down.. `KafkaAccess`, `KafkaApplication`, `KafkaTenant`, `KafkaCredentials`, `KafkaPortal`, and the two I like but we probably can't use due to their use elsewhere `KafkaIngress`, `KafkaGateway`

katheris: I still feel like `KafkaService` and `KafkaConnection` are probably better than any of the other ones I've come up with though

028-service-binding.md
```diff
@@ -0,0 +1,379 @@
...
+The actions taken during the reconcile loop will be:
+1. The operator can be deployed at cluster-scope where it will watch resources across all namespaces, or in a single namespace, where it will only watch resources in that namespace.
+2. On start-up the operator will watch all resources of Kind `KafkaConnection`.
+3. For any `KafkaConnection` resource present it will add a status of NotReady to that CR.
+4. For each `KafkaConnection` resource it will perform a GET request on the Kafka cluster that is 
+referenced in the `KafkaConnection` and pick out the configured listeners.
+   1. If this or any subsequent GET requests fail due to a 401 (unauthorised) error, the status reason will be updated to reflect this.
+5. If there are multiple listeners listed in the `Kafka` CR with the type `internal`, the operator 
+will filter the list by comparing the `tls` and `authentication` properties in the `Kafka` and `KafkaUser` CRs.
+6. If there are still multiple listeners it will pick one at random.
```
tombentley, line 110: This non-determinism bothers me because it means users could experience weird behaviour. E.g. if you had two `KafkaConnection` CRs that differed only in the `KafkaUser` the pointed to then the resulting bindings might end up selecting different listeners, which would be inconsistent from the users PoV. I think we need a way to make it deterministic. The best way imho would be to require selection by listener name if there is ambiguity. Alternatives include using the first listener in the filtered list, or sorting the filtered list by name and picking the first, but imho they're inferior because they're too implicit and prone to breakage by seemingly innocuous changes in the `Kafka` CR.

028-service-binding.md
```diff
@@ -0,0 +1,379 @@
...
+7. The operator will add the name of the listener chosen will be added to the `KafkaConnection` status so that in future reconciles 
+the operator can take this into account in step 4 and avoid cycling through listeners.
```
tombentley, line 112: Now step 6?

028-service-binding.md
tombentley: WDYT @scholzj @ppatierno?

028-service-binding.md
```diff
@@ -0,0 +1,379 @@
...
+  name: barista-kafka
+spec:
+  kafka:
+    apiVersion: kafka.strimzi.io/v1beta2
+    kind: Kafka
+    name: my-cluster
+    namespace: my-namespace # optional, defaults to same namespace as the KafkaConnection
+    listener: # optional, when not specified Strimzi will choose an appropriate listener (see implementation below)
+      name: tls
+  user: # optional
```
tombentley, line 50: Are we unnecessarily making the user operator a mandatory requirement if you want to have an authenticated connection? We _could_ support supplying `user` or a secret reference which would supply the cert, key and user name (using a `oneOf` constraint in the CRD so that `user` property was mutually exclusive with such a `userSecret` property). That would allow people not using the UO to use bindings too. If there's value in that, there's no need to implement it immediately. It could be added later.

028-service-binding.md
katheris: I hadn't thought of the inconsistency between two different `KafkaConnection` CRs. Could we not make listener selection a requirement but add a warning to the status of the `KafkaConnection` CR if the operator had to make a choice between multiple options?

028-service-binding.md
katheris: We could make it easy to supply just a secret by keeping the `Kind` option here, and letting you supply a `Kind` of `Secret` instead of `KafkaUser`.

028-service-binding.md
scholzj: What about `KafkaBinding`?

scholzj: > I prefer KafkaService over KafkaConnection because if I was looking for the address of where my Kafka was in Kubernetes I would look for a Service.

But it is not service -> that has consequences on access rights, for network policies and tenant segregation etc. If you have a Kube service, you can connect to it. But this resource has nothing to do with service.

> Throwing some other ideas down.. KafkaAccess, KafkaApplication, KafkaTenant, KafkaCredentials, KafkaPortal, and the two I like but we probably can't use due to their use elsewhere KafkaIngress, KafkaGateway

I think `KafkaAccess` is good -> no conflicts with native Kube resources and it fairly well describes what you can get ... information to access the Kafka cluster.

`KafkaApplication` is probably something I can live with. But I would really expect that you pass some DSL inside the resource and it builds the Kafka application for you ;-). So not sure how understandable it is.

`KafkaTenant`, `KafkaPortal` and `KafkaCredentials` IMHO don't describe what it is and how is it used. `KafkaIngress` and `KafkaGateway` bring the same confusion as `KafkaService` where pre-existing resource create expectations of what it does.

028-service-binding.md
scholzj: > If we want to develop a new operator in Java then this is the way to go https://javaoperatorsdk.io/

Keep in mind that we need to integrate it with our existing resources. So you need it to be able to use our `api` extensively. Not sure how will that goes together or not.

028-service-binding.md
scholzj: I would argus that our existing `*AssemblyOperator` classes are not controllers. They do not follow the Kubernetes controller pattern. So I think renaming them to controller is confusing. TBH, I think the `Assembly` part is much more confusing part of the name than the `Operator` part.

028-service-binding.md
scholzj: I think we should keep our options open and not close the doors on this in the proposal ;-)

028-service-binding.md
scholzj: I think we should have the following logic:
* Find all listeners with matching AuthN
* If more of them are found and one or more are internal, we should prefer the internal
* If only external is found with matching AuthN, we should take it

028-service-binding.md
scholzj: It has to be deterministic not just to not differ between two resources. If it is random, it might change in every reconciliation and just loop in a circle. I think we should just pick a sorting rule ... lowest port number, listener names by alphabet etc.

028-service-binding.md
ppatierno: Sorry I am still with `KafkaService` which matches what service binding is and what I saw in the spec ... but I don't want to force on this, maybe we should start a voting process :P

028-service-binding.md
ppatierno: AFAIK when you have a custom resource to handle you are dealing with an "operator". I agree with Jakub that the `Assembly` part is the most confusing one, I have never seen this as part of operators developed with tools like the Go sdk for example. I think we should stay with `Operator` and trying at some point to remove that `Assembly` part. Maybe we can start with this new operator to do that.

tombentley: Having `KafkaOperator` and `KafkaConnectOperator` etc would be an improvement, but there's still the confusion with the "Cluster Operator" and "Entity Operator" terms, which are used extensively in the docs. Maybe we just can't avoid that confusion.

But what about the "resource operator" classes? Those are not really operators (they have no watch). At best they're "reconcilers", but we use methods like `get`/`getAsync` as much as `reconcile`. Having a better term for them would also be an improvement, simply from having fewer `*Operator` classes floating around.

katheris: For what it's worth it looks like the java-operator-sdk project have currently landed on "Reconciler" as the name for the class that is responsible for updating a resource to look how it should.

028-service-binding.md
katheris: The mechanism described in step 7 will stop it cycling, the intention was only to pick a random one if there isn't one listed in the status already. That way even if the port number changes later the listener chosen will stay as before.

028-service-binding.md
scholzj: > But what about the "resource operator" classes? Those are not really operators (they have no watch). At best they're "reconcilers", but we use methods like get/getAsync as much as reconcile. Having a better term for them would also be an improvement, simply from having fewer *Operator classes floating around.

To be honest, these are really just `*Utils` classes, right? They have a `reconcile` method, but they do not run the reconciliation loop. But TBH, I'm not sure the current naming is really a blocker for anything and being consistent at this point is more useful than changing the names around in a hope they will be more clear to someone else.

028-service-binding.md
scholzj: How does that deal with the changes in the Kafka CR? I don't know, random seems weird -> I think it is easy to come up with some fixed logic which works in most cases.

katheris: Ok, I think sorting based on listener name is the simplest and least likely to change. I will update the proposal to use that

028-service-binding.md
scholzj: Discussed on the community call on 2nd Dec -> we should go with `KafkaAccess`.

028-service-binding.md
tombentley: SGTM, so long as unsupported kinds result in a status condition with suitable message.

tombentley (approved):
LGTM. Thanks Kate!

ppatierno (approved):
Thanks Kate!

028-service-binding.md
```diff
@@ -0,0 +1,276 @@
...
+1. The operator can be deployed at cluster-scope where it will watch resources across all namespaces, or in a single namespace, where it will only watch resources in that namespace.
+2. On start-up the operator will watch all resources of Kind `KafkaAccess`.
+3. For any `KafkaAccess` resource present it will add a status of NotReady to that CR.
+4. For each `KafkaAccess` resource it will perform a GET request on the Kafka cluster that is 
+referenced in the `KafkaAccess` and pick out the configured listeners.
+   1. If this or any subsequent GET requests fail due to a 401 (unauthorised) error, the status reason will be updated to reflect this.
+5. The operator will determine which listener to choose:
+   1. If there are no listeners in the `Kafka` CR the reconcile loop will fail.
+   2. If the `KafkaAccess` specifies a listener, the operator will choose this listener.
+   3. If there are no listeners specified in the `KafkaAccess`, but only one listener in the `Kafka` CR, the operator will choose that listener.
```
scholzj, line 108: I think we should validate it against the authentication type from the KafkaUser resource and fail if they are not compatible.

028-service-binding.md
```diff
@@ -0,0 +1,276 @@
...
+   4. If there are no listeners specified in the `KafkaAccess`, but multiple listeners listed in the `Kafka` CR, the operator
```
scholzj, line 109: Why does the `tls` property matter here? The `KafkaAcess` does not seem to touch it.

028-service-binding.md
```diff
@@ -0,0 +1,276 @@
...
+   4. If there are no listeners specified in the `KafkaAccess`, but multiple listeners listed in the `Kafka` CR, the operator 
+   will filter the list by comparing the `tls` and `authentication` properties in the `Kafka` and `KafkaUser` CRs.
+   5. If there are still multiple listeners after filtering for `tls` and `authentication` the operator will choose the one that is of type `internal`.
+   6. If there are still multiple listeners after filtering for `internal` the operator will sort the listeners alphabetically by name and choose the first one.
+   7. If the operator fails to find a suitable listener after filtering the reconcile loop will fail.
+8. The operator will then create a secret and add the bootstrapServers for the chosen listener to that secret.
+9. If the chosen listener uses TLS, the operator will then perform a GET request on the `cluster-ca-cert` secret and put the certificate and truststore files in the new secret.
+10. If the `KafkaAccess` references a `KafkaUser`, the operator will perform a GET request on that user to find the name of the secret containing the credentials and the username
```
scholzj, line 116: You need to have the user already before the step 5.4 -> so this needs to be done earlier.

028-service-binding.md
```diff
@@ -0,0 +1,276 @@
...
+   7. If the operator fails to find a suitable listener after filtering the reconcile loop will fail.
+8. The operator will then create a secret and add the bootstrapServers for the chosen listener to that secret.
+9. If the chosen listener uses TLS, the operator will then perform a GET request on the `cluster-ca-cert` secret and put the certificate and truststore files in the new secret.
+10. If the `KafkaAccess` references a `KafkaUser`, the operator will perform a GET request on that user to find the name of the secret containing the credentials and the username 
+from the status if it is provided. The operator will also perform a GET request on the user secret to get any credentials. 
+11. The operator will add any user related certificates and credentials to the secret it created earlier.
+12. The operator will add a `status.binding` to the `KafkaAccess` resource with the name of the secret. 
+13. The operator will then use a `Watch` or `Informer` to keep track of changes to the `Kafka` CR and, if used earlier, the `cluster-ca-cert` secret, the `KafkaUser` resource and the secret created by the `KafkaUser` operator,
```
scholzj, line 120: I think this goes beyond the proposal and more into the implementation realm. But I think it will not scale well to have a separate watch/informer for each `Kafka` and `KafkaUser`. Maybe it should have a single watch/infromer for these and cross reference the Kafka, KafkaUser and KafkaAccess resources internally.

scholzj (approved):
Left some nits, looks good to me otherwise.

028-service-binding.md
katheris: Agreed

028-service-binding.md
katheris: The Kafka Access operator needs to know if the listener has `tls` true or not, because if it is true it should go and fetch the cluster ca certificate and put that in the secret.

028-service-binding.md
katheris: Good point, I have it earlier in my POC

028-service-binding.md
katheris: I agree this is more implementation realm, but interestingly this is solved quite nicely I think by java-operator-sdk. It allows you to watch all `Kafka` and `KafkaUser` resources and then trigger the reconcile for the relevant `KafkaAccess` resources when they change.

scholzj:
This has now 3 approvals - we will wait till Monday if anyone else has anything more before closing the vote.

scholzj:
@katheris What some reason, I seem to be unable to check this out and update it before merging. Do you think you could please:
* Update the file name to `33`
* Add the record to the README.md file

Thanks

katheris:
@scholzj Filename and readme have been updated

scholzj:
@katheris Great, thanks.

scholzj:
Approved by 3 +1 votes. Merging.

PR #44, opened 2021-11-12
StatefulSet removal

scholzj:
This proposal covers the removal of the StatefulSets from Strimzi. It explains why it might be useful and what might be the alternative. It is accompanied by the draft PR with a proof-of-concept implementation which removes the StatefulSets from the ZooKeeper cluster: strimzi/strimzi-kafka-operator#5884


katheris:
I've had a read through the proposal and it all makes sense to me and seems a reasonable approach to the problem. 

Has there been any investigation done into how other operators managing stateful services like Kafka manage this problem? Do most others just use stateful sets and deal with the problems that are highlighted in the proposal?

scholzj:
> Has there been any investigation done into how other operators managing stateful services like Kafka manage this problem? Do most others just use stateful sets and deal with the problems that are highlighted in the proposal?

There are other operators managing the pods directly, so it is not a unique idea. But I haven't done any in depth analysis to have some statisctics to share or anything like that.

031-statefulset-removal.md
```diff
@@ -0,0 +1,225 @@
...
+
+Pods managed by a single StatefulSet always exist only on single Kubernetes cluster.
+If you want to have a Kafka cluster which would stretch multiple Kubernetes clusters, you would need to create a StatefulSet on each of them and have it create the Pods on given cluster.
+But you would not be able easily move the brokers between the clusters since they would depend on the StatefulSet index numbers to generate their broker IDs.
+
+## Proposal
+
+The StatefulSets should be replaced with a new Strimzi custom resource called `KafkaNodeSet`.
+KafkaNodeSet will be used only internally.
+Users will not be expected to interact with.
```
tombentley, line 88: I wonder whether we can make this more explicit. For example, I wonder about the `Kafka...` in the name. In this case this CR does not map to a Kafka entity at all. Make it would be better to call it `StrimziNodeSet` or `StrimziCluster` or something.

031-statefulset-removal.md
```diff
@@ -0,0 +1,225 @@
...
+
+In the current solution, Kubernetes is responsible for (re)starting the Pods in case of disruptions.
+After implementing this proposal, Strimzi will be responsible for it.
+Any bugs and issues in Strimzi might result in the Pods not being started and might affect the availability of the Kafka clusters.
+By accepting this proposal, we need to be able to accept this risk and responsibility.
+
+Unlike StatefulSets, KafkaNodeSet also doesn't use template to generate Pods.
+That brings us many advantages such as asymmetric configuration etc.
+But it also means that the custom resource will be bigger than StatefulSets because it contains a full YAML for each Pod.
+It will have a limited capacity how many Pods it will be able to handle.
```
tombentley, line 178: Can we estimate, even if very roughly, what this limit is likely to be? Are we talking about 10s of pods, or 100s?

031-statefulset-removal.md
```diff
@@ -0,0 +1,225 @@
...
+In the first phase, we should implement the 1-to-1 replacement for StatefulSets.
+While that will not bring any immediate advantages, it will allow us to easily move from StatefulSets to KafkaNodeSets.
+It would also avoid big PRs and instead allow us to gradually improve in smaller steps.
+
+### Risks
+
+In the current solution, Kubernetes is responsible for (re)starting the Pods in case of disruptions.
+After implementing this proposal, Strimzi will be responsible for it.
+Any bugs and issues in Strimzi might result in the Pods not being started and might affect the availability of the Kafka clusters.
+By accepting this proposal, we need to be able to accept this risk and responsibility.
```
tombentley, line 173: Sometimes, in various weird corner cases, uses can scale down the CO and delete pods if they want/need to. Are we confident that there are no valid use cases for them to actually _need_ to do this?

031-statefulset-removal.md
```diff
@@ -0,0 +1,225 @@
...
+They will identify which pods need to be rolled and delete them.
+The only difference will be that instead of the Kubernetes StatefulSet controller, it will be the KafkaNodeSet controller which will create a new pod.
+For each Pod managed by a KafkaNodeSet, Strimzi will calculate a revision (really a _hash_ of the Pod configuration).
+This revision will be used to identify changes to the desired Pod configuration and trigger rolling of the pod.
+This is a similar process as used by the StatefulSet itself.
+
+Similarly to the StatefulSets, the KafkaNodeSets will be stable resources as well.
+All the annotations currently used on StatefulSets will be applied to them.
+This applies to the annotations used to store the configurations and state but also to the command annotations such as manual rolling update triggers.
+
```
tombentley, line 143: I realise you're trying to limit the scope of the proposal so that it remains tractable... but I think it would be good to elaborate a little bit about how the stretch cluster use case would work with node sets. Presumably you'd have a nodeset in each kube cluster, and hence each nodeset only ever selects pods in the same kube cluster?

031-statefulset-removal.md
```diff
@@ -0,0 +1,225 @@
...
+* The usual metadata, kind and API identifications
+* A `.spec` section with
+    * Pod selector to be able to find the pods managed by this KafkaNodeSet
+    * List of Pods (full Pod YAML definition including metadata section, but without status section)
+* A status section with
+    * Number of managed, current and ready pods
+    * List of error conditions in case of problems
+    * Observed generation
+
+The node sets will be used very similarly to the StatefulSets.
```
tombentley, line 104: Probably worth mentioning:

* that the Pods will be in the same namespace as the NodeSet (assuming that's true)?
* that the NodeSet would be the owner of the Pods and GC will be used

031-statefulset-removal.md
scholzj: Let's see what others think. I'm fine with `StrimziNodeSet`. All our resources are `Kafka...`, but no problem to rename this.

031-statefulset-removal.md
scholzj: No, I do not have any estimate.

031-statefulset-removal.md
scholzj: If the user scales down the operator for whatever reason, than they loose the ability to have the pods restarted. This is good point for documentation. But I do not think we expect anyone to scale down the operator long-term. So they would need to handle this.

031-statefulset-removal.md
scholzj: Is the explanation in the `Stretch clusters` section sufficient? They core of how it is enabled is really that you can take away the pod from the node set in any place and just create it on another cluster.

tomncooper (approved):
LGTM, this work will be vital for KRaft!

031-statefulset-removal.md
tombentley: Well the bound is on the size of CRs supported by kubernetes (which IIUC is inherited from the etcd limit of 1.5MB by default on key-value pairs). So presumably we could estimate this by taking a Kafka `Pod` spec (for a cluster using all the bells and whistles) and seeing how many bytes it requires.

031-statefulset-removal.md
tombentley: So we never currently tell users who are in some kind of tricky situation to stop the CO and delete pods. All problems requiring manual intervention can currently be solved with the CO running?

031-statefulset-removal.md
tombentley: I get that bit. It's whether the node sets (and their selection) is confined to a single cluster that I'm interested in.

tombentley: (by "and their selection" I mean "do we do the selection just the the same cluster as the nodeset?" I think the answer would be yes)

031-statefulset-removal.md
scholzj: Oh, I first thought you are asking about the performance and not about how many pods can fit. 

IIRC the limit is 1MB. But it depends heavily how you define _cluster using all the bells and whistles_. How much metadata does that mean? How many affinity rules? How many volumes does it have? How many topology spread-rules? My ZooKeeper pods have about 4KB each. It has some metadata, but no affinity, spread constraints etc.

Kafka pods will be probably bigger with more volumes etc. ... so maybe 100 pods?

031-statefulset-removal.md
scholzj: I don't think the docs ever recommended it, but I'm not 100% sure. In general, yes, it used to be recommended as a short term solution to typically do manual changes to storage etc. (which requires deletion of STSs and pods)

031-statefulset-removal.md
scholzj: I don't know, I do not have any design document for that. Given our current capacity it is probably years away. If you ask me right now, I would say the KafkaNodeSet will be local and most probably also its controller should run locally and not rely on remote connection. That also fits on the thinking I had around the ZooKeeper removal (https://hackmd.io/@scholzj/HJIZsU52O).

031-statefulset-removal.md
Frawless: `StrimziNodeSet` makes better sense to me than `KafkaNodeSet`

Frawless (approved):

031-statefulset-removal.md
```diff
@@ -0,0 +1,231 @@
...
+
+It would be also possible to manage the Pods from the `KafkaAssemblyOperator`.
+In the same way as it is currently creating the StatefulSets or deleting the pods, it can also create and update the Pods directly.
+This would make things easier because we would not need any additional controller or custom resource.
+However, the Kafka CR reconciliation is currently very _heavy_ process which does lot of things in a serial fashion.
+It manages both ZooKeeper and Kafka and it uses long-running reconciliation.
+The reconciliation often takes minutes while it waits for the Pods getting ready or for load balancers being created.
+It runs multiple reconciliations in parallel, but for each Kafka custom resource only one reconciliation can be in progress.
+That means that it might often take a lot of time before the reconciliation gets to the point when it can recreate the Pods.
+Imagine situation when several Kafka brokers are gone because of failure in one of the data centers but the reconciliation is stuck waiting on some load balancer creation to finish while the pods are missing.
```
PaulRMellor, line 218:
```suggestion
Imagine a situation in which several Kafka brokers are unavailable because of a failure in one of the data centers, and the reconciliation is stuck waiting for a load balancer creation to finish while the pods are missing.
```

031-statefulset-removal.md
```diff
@@ -0,0 +1,231 @@
...
+Such situation might easily happen and would not be acceptable.
```
PaulRMellor, line 219:
```suggestion
A situation such as this can easily arise and would not be acceptable.
```

PaulRMellor (approved):
Read for familiarisation and left a few comments as I went through the content.

031-statefulset-removal.md
```diff
@@ -0,0 +1,231 @@
...
+We could use multiple StatefulSets to separate the different types of nodes.
+But even then it would be hard to change the cluster topologies and the node types.
+Without the StatefulSet limitations, we should have a lot more flexibility to mix the node types and their roles.
+
+### Stretch clusters
+
+Pods managed by a single StatefulSet always exist only on single Kubernetes cluster.
+If you want to have a Kafka cluster which would stretch multiple Kubernetes clusters, you would need to create a StatefulSet on each of them and have it create the Pods on given cluster.
+But you would not be able easily move the brokers between the clusters since they would depend on the StatefulSet index numbers to generate their broker IDs.
+With the KafkaNodeSets, you can easily have different KafkaNodeSets for each Kubernetes cluster and just delete a pod on one Kubernetes cluster and move it to another cluster.
```
ppatierno, line 83: Reading the proposal, it's the first time you mention a `KafkaNodeSets` maybe pointing to its definition that comes later could help at least explain people that is something new covered in this proposal later.

031-statefulset-removal.md
ppatierno: I am with Jakub, I can understand that it doesn't map to any Kafka entity at all but all the others are `Kafka` prefixed. I can get that we have topic, user, connector so they would make sense but for example we have `KafkaRebalance` and not `StrimziRebalance` and the rebalance is not really a Kafka concept as topic and user. I would stick with current naming using `KafkaNodeSet`.

031-statefulset-removal.md
```diff
@@ -0,0 +1,231 @@
...
+This revision will be used to identify changes to the desired Pod configuration and trigger rolling of the pod.
+This is a similar process as used by the StatefulSet itself.
+
+Similarly to the StatefulSets, the KafkaNodeSets will be stable resources as well.
+All the annotations currently used on StatefulSets will be applied to them.
+This applies to the annotations used to store the configurations and state but also to the command annotations such as manual rolling update triggers.
+
+### Feature Gate
+
+The switch from StatefulSets to KafkaNodeSets should be controller by a new feature gate called `StatefulSetRemoval`.
```
ppatierno, line 150: What do you think to do the other way around `KafkaNodeSetEnable` which is false by default? I see the removal as the "negative" logic I don't like, but anyway I am fine to stick with it, was just a suggestion.

031-statefulset-removal.md
scholzj: The docs are valid. This is done by the operator. But in theory there might be other things => resizing, changing storage types etc.

031-statefulset-removal.md
samuel-hawker: ^^ I personally prefer this approach - good suggestion.

samuel-hawker (commented):
This seems like a great proposal! Thanks for putting this together.

Since a feature-gate is being used, I assume the StatefulSetOperator will be sticking around for another couple of releases, to support both STS & NodeSet installs?

scholzj:
@tombentley @ppatierno @samuel-hawker I think your comments should not be addressed. Could you please have another look? Thanks.

031-statefulset-removal.md
```diff
@@ -0,0 +1,242 @@
...
+When a StrimziPodSet is modified, it will check if any pods need to be added or deleted.
+When a Pod is deleted or modified, it will identify the StrimziPodSet it belongs to and reconcile the whole StrimziPodSet.
+
+![StrimziPodSet hierarchy](images/031-strimzi-with-StrimziPodSet.png)
+
+The StrimziPodSet controller will not be responsible for handling the rolling updates.
+They will continue to be handled in the same way by the `KafkaAssemblyOperator` and in the accompanied classes such as `KafkaRoller` and `ZooKeeperRoller`.
+They will identify which pods need to be rolled and delete them.
+The only difference will be that instead of the Kubernetes StatefulSet controller, it will be the StrimziPodSet controller which will create a new pod.
+For each Pod managed by a StrimziPodSet, Strimzi will calculate a revision (really a _hash_ of the Pod configuration).
```
tombentley, line 137: Rather than "Strimzi will calculate" it would be better to say which component would do this (since we're talking about several in this proposal)

031-statefulset-removal.md
```diff
@@ -0,0 +1,242 @@
...
+* Current storage configuration
+* Kafka versions
+* CA generations
+
+If we manage the Pods directly, we would also need to find a way how to replace them.
+
+## Compatibility
+
+The removal of StatefulSets is obviously a significant change.
+But the migration between StatefulSets and StrimziPodSets should be possible without any service interruption or availability issues.
```
tombentley, line 242: Will it be possible to seamlessly upgrade to, and downgrade from StrimziPodSets?

031-statefulset-removal.md
```diff
@@ -0,0 +1,242 @@
...
+
+Pods managed by a single StatefulSet always exist only on single Kubernetes cluster.
+If you want to have a Kafka cluster which would stretch multiple Kubernetes clusters, you would need to create a StatefulSet on each of them and have it create the Pods on each given cluster.
+But you would not be able easily move the brokers between the Kubernetes clusters since they would depend on the StatefulSet index numbers to generate their broker IDs.
+By introducing _StrimziPodSets_ (for more details and explanation, check the _Proposal_ section below), you can have different StrimziPodSets in each Kubernetes cluster and just delete a pod on one Kubernetes cluster and move it to another cluster by modifying the corresponding StrimziPodSets.
+
+## Proposal
+
+The StatefulSets should be replaced with a new Strimzi custom resource called `StrimziPodSet`.
+StrimziPodSet will be used only internally.
```
tombentley, line 89: When we say this, are we expecting to have the same compatibility guarantees for `StrimziPodSet` as for the other resources (I think we are, but I just want everyone to be clear about this).

031-statefulset-removal.md
scholzj: Improved.

031-statefulset-removal.md
scholzj: Yes, it will be possible. Improved in the proposal.

031-statefulset-removal.md
scholzj: Well, I think anything can happen during the alpha phase. Once it leaves the alpha phase, it should be treated as any other resource with backwards compatibility, versioning etc. I added such explanation to the text as well.

031-statefulset-removal.md
```diff
@@ -0,0 +1,252 @@
...
+
+Instead of using a Pod templates, and generating Pods with fixed names, StrimziPodSet will use full Pod structures including custom names.
+The configuration retains the usefulness of a StatefulSet, but removes the limitations we have identified.
+
+The StrimziPodSet will be created by the `KafkaAssemblyOperator` class in the same place in the reconciliation as the StatefulSet.
+This is also where the StrimziPodSet will be patched when configuration changes.
+
+A new controller will be developed to reconcile the StrimziPodSet resources.
+It will run as part of the Strimzi cluster operator container.
+It will follow a more lightweight controller approach rather than the long-running _heavy_ approach of our current assembly operators.
```
ppatierno, line 118: what does this exactly mean?

ppatierno (commented):
The images name have to be fixed because they still have the old kafkanodset in there. 
I also left one question.

031-statefulset-removal.md
scholzj: That is supposed to mean what is described in the sentences below this one:
* It will be designed to iterate quickly rather than wait
* It will not wait for Pod readiness but instead get a new event when the Pod becomes ready
* It will consume events about the StrimziPodSets and their Pods and react to them
* It will run in its own thread and use a work queue to queue the events and the reconciliations to avoid working on too many reconciliations in parallel

I changed the text a bit to make it more clear.

scholzj:
@ppatierno I'm not gonna fix the images. If you think the old name is a problem I can remove them.

scholzj:
> @ppatierno I'm not gonna fix the images. If you think the old name is a problem I can remove them.

Actually, never mind. I fixed them. Sorry.

031-statefulset-removal.md
tombentley: If we can't rule out the possibility of needing to be able to run the node set operator but not running the kafka operator/CO perhaps we should allow it (it seems like a simple enough thing to support via the CO config). It would at least give us that escape hatch, should it turn out we needed it.

tombentley (approved):
Thanks for driving this Jakub.

sknot-rh (approved):

031-statefulset-removal.md
scholzj: You mean having some flag which would tell the CO to run the StrimziPodSet reconciliation only? That would be possible I guess. I think the pausing of reconciliation should be enough for most cases now => but it is quite hard to say on 100% that it will be never needed. So if you want I can add it, although I personally don't see the need right now. It should not be a big deal to implement it I think.

tombentley: It just feels like one of those things where you don't think you need it, you release it and later find you do, and then you've got to release a fix. But it's behind a feature flag, so I guess there's no need to implement this immediately and give it time to mature first.

scholzj: I added it there. I feel like it might be useful in the future for the stretch clusters as well (to manage the pods locally) for example. I will not add it to the PoC PR for ZooKeeper. But will track it as a task.

ppatierno (approved):
lgtm!

scholzj:
It looks like this has currently 5 binding and one non-binding approvals. Unless there are any new comments, I will merge it tonight.

scholzj:
Approved by 5 binding and 1 non-binding approvals

PR #45, opened 2021-11-23
feat: New secure Connect API proposal

katheris:
Signed-off-by: Katherine Stanley <11195226+katheris@users.noreply.github.com>

### DIFF
--- a/031-tls-encrypt-the-kafka-connect-rest-api.md
+++ b/031-tls-encrypt-the-kafka-connect-rest-api.md
@@ -18,7 +18,7 @@ There is currently no way to TLS encrypt the Kafka Connect REST API, leaving the
 
 This proposal adds REST api configuration options to the `KafkaMirrorMaker2` and `KafkaConnect` CRDs.
 
-A new field is added to the `KafkaConnect` and `KafkaMirrorMaker2` specs named `restListeners`, which allows users to configure an HTTP listener on port 8443, an encrypted HTTPS listener on port 8443, or both in the Kafka Connect configuration.
+A new field is added to the `KafkaConnect` and `KafkaMirrorMaker2` specs named `restListeners`, which allows users to configure an HTTP listener on port 8083, an encrypted HTTPS listener on port 8443, or both in the Kafka Connect configuration.
 This enables users to access the REST API with TLS if required.
 The `KafkaConnector`, `KafkaConnect` and `KafkaMirrorMaker2` operators will use the TLS encrypted listener when it is enabled, even if an unencrypted listener is present.
 
@@ -27,8 +27,7 @@ The `restListeners` field is designed for potential future extension, and is a l
 For this proposal, the `protocol` field only supports two values: `http` and `https`. If the `restListeners` list is empty the `KafkaConnect` and `KafkaMirrorMaker2` 
 runtimes will be created with a single unencrypted listener on 8083 to match the existing behaviour.
 
-The `restCertChainAndKey` field will be required if the `https` protocol is selected. This field will contain the certificate and key the Kafka Connect operator will 
-use to create the self-signed certificate. The Kafka Connect and MirrorMaker2 operators will use the certificate to create a truststore for communicating with the REST API.
+The `restCertChainAndKey` field will be required if the `https` protocol is selected.
 
 For example, the following CR will enable the HTTP listener on port 8083 and the HTTPS listener on port 8443:
 ```
@@ -41,7 +40,7 @@ spec:
   ...
 ```
 
-The schema for `restListeners` will also include properties to provide the certificate and key that should be used to create the self-signed certificate and for the operators to use to communicate with Kafka Connect as follows:
+The schema for `restListeners` will be as follows:
 ```
 openAPIV3Schema:
   type: object
@@ -98,6 +97,41 @@ listeners.https.ssl.keystore.password: ***generated password***
 listeners.https.ssl.keystore.type: PKCS12
 ```
 
+### Certificates
+
+If the user decides to enable an encrypted HTTPS listener, they must have the `restCertChainAndKey` 
+property present in the `KafkaConnect` CR. The property must point to an existing secret that contains the 
+CA certificate and CA private key that they want Kafka Connect to use. This secret can be created by the 
+user, or it can be the [Strimzi cluster CA secret](#using-the-strimzi-cluster-ca).
+
+The CA certificate and CA private key configured by the user in the `KafkaConnect` CR will be used by the Kafka Connect 
+operator to generate a self-signed certificate. This self-signed certificate will be used to configure the keystore and 
+truststore for Kafka Connect. The CA certificate will also be used to generate a truststore 
+that the Strimzi operators, e.g. KafkaConnector, MirrorMaker2 can use when communicating with Kafka Connect.
+
+#### Using the Strimzi cluster CA
+
+If the Kafka Connect cluster is connecting to a Kafka instance that is managed by Strimzi, the user can configure the 
+`restCertChainAndKey` property to point to the Strimzi cluster CA.
+
+A `KafkaConnect` instance with a secure endpoint using the cluster CA will look like:
+
+```
+apiVersion: kafka.strimzi.io/v1beta1
+kind: KafkaConnect
+spec:
+  restListeners:
+  - protocol: https
+    restCertChainAndKey:
+      secretName: <cluster-name>-cluster-ca
+      key: ca.key
+      certificate: ca.crt
+  ...
+```
+
+As part of this proposal the `<cluster-name>-cluster-ca` secret will be updated to include the cluster CA certificate 
+as currently it only includes the private key.
+
 ### Certificate renewals
 
 The self-signed TLS certificate will have a 1-year expiration.
@@ -109,28 +143,6 @@ This would need to be a multi-phase process consisting of the following steps:
 2. Distribute the new certificate to all pods and cluster operator truststores
 3. Replace the key and roll all Connect pods
 4. When the old certificate expires, remove it from the truststores and roll all Connect pods
-
-### Using the Strimzi cluster CA
-
-The updated `KafkaConnect` CR requires a single secret containing the CA private key and certificate. When a `KafkaConnect` resource 
-is deployed that uses a Kafka instance managed by Strimzi the user should select the Strimzi cluster CA. Currently the private key and 
-certificate for the Strimzi cluster CA are spread over two secrets: `<cluster-name>-cluster-ca` and `<cluster-name>-cluster-cert`.
-
-As part of this proposal the `<cluster-name>-cluster-ca` will be updated to include the cluster CA certificate. A `KafkaConnect` instance 
-with a secure endpoint using the cluster CA will look like:
-
-```
-apiVersion: kafka.strimzi.io/v1beta1
-kind: KafkaConnect
-spec:
-  restListeners:
-  - protocol: https
-    restCertChainAndKey:
-      secretName: <cluster-name>-cluser-ca
-      key: ca.key
-      certificate: ca.crt
-  ...
-```
 
 ## Affected/not affected projects
 
--- a/031-tls-encrypt-the-kafka-connect-rest-api.md
+++ b/031-tls-encrypt-the-kafka-connect-rest-api.md
@@ -26,12 +26,13 @@ A new field is added to the `KafkaConnect` and `KafkaMirrorMaker2` specs named `
 This enables users to access the REST API with TLS if required.
 The `KafkaConnector`, `KafkaConnect` and `KafkaMirrorMaker2` operators will use the TLS encrypted listener when it is enabled, even if an unencrypted listener is present.
 
-The `restListeners` field is designed for potential future extension, and is a list of items that contains a `protocol` field and an optional `caCertChainAndKey` field.
+The `restListeners` field is designed for potential future extension, and is a list of items that contains a `protocol` field and fields for providing certificates if required `certChainAndKey`, `strimziTruststoreCert`
 
 For this proposal, the `protocol` field only supports two values: `http` and `https`. If the `restListeners` list is empty the `KafkaConnect` and `KafkaMirrorMaker2` 
 runtimes will be created with a single unencrypted listener on 8083 to match the existing behaviour.
 
-The `caCertChainAndKey` field will be required if the `https` protocol is selected.
+The `certChainAndKey` field will be required if the `https` protocol is selected. The `strimziTruststoreCert` will be required if the `https` protocol is selected and the provided 
+certificate is not signed by a well-known CA.
 
 For example, the following CR will enable the HTTP listener on port 8083 and the HTTPS listener on port 8443:
 ```
@@ -63,15 +64,15 @@ openAPIV3Schema:
                 - http
                 - https
                 description: The protocol of the REST listener.
-              caCertChainAndKey:
+              certChainAndKey:
                 type: object
                 properties:
                   certificate:
                     type: string
-                    description: The name of the ca certificate file in the Secret.
+                    description: The name of the certificate file in the Secret.
                   key:
                     type: string
-                    description: The name of the ca private key file in the Secret.
+                    description: The name of the private key file in the Secret.
                   secretName:
                     type: string
                     description: The name of the Secret containing the certificate.
@@ -79,7 +80,20 @@ openAPIV3Schema:
                     - key
                     - certificate
                     - secretName
-                description: CA certificate to use to generate.
+                description: Certificate for Kafka Connect to use to create a keystore for the rest listener.
+              strimziTruststoreCert:
+                type: object
+                properties:
+                  certificate:
+                    type: string
+                    description: The name of the certificate file in the Secret.
+                  secretName:
+                    type: string
+                    description: The name of the Secret containing the certificate.
+                  required:
+                    - certificate
+                    - secretName
+                description: Certificate for Strimzi to trust when making calls to the Connect rest listener.
             required:
             - protocol
           description: List of additional REST listeners.
@@ -100,50 +114,16 @@ listeners.https.ssl.keystore.type: PKCS12
 
 ### Certificates
 
-If the user decides to enable an encrypted HTTPS listener, they must have the `caCertChainAndKey` 
+If the user decides to enable an encrypted HTTPS listener, they must have the `certChainAndKey` 
 property present in the `KafkaConnect` CR. The property must point to an existing secret that contains the 
-CA certificate and CA private key that they want Kafka Connect to use. This secret can be created by the 
-user, or it can be the [Strimzi cluster CA secret](#using-the-strimzi-cluster-ca).
+certificate and private key that they want Kafka Connect to use. The certificate must contain the correct 
+Subject Alternative Names (SANs) similarly to when a user provides their own certificates for listeners:
+https://strimzi.io/docs/operators/latest/using.html#ref-alternative-subjects-certs-for-listeners-str
 
-The CA certificate and CA private key configured by the user in the `KafkaConnect` CR will be used by the Kafka Connect 
-operator to generate a self-signed certificate. This self-signed certificate will be used to configure the keystore and 
-truststore for Kafka Connect. The CA certificate will also be used to generate a truststore 
-that the Strimzi operators, e.g. KafkaConnector, MirrorMaker2 can use when communicating with Kafka Connect.
-
-#### Using the Strimzi cluster CA
-
-If the Kafka Connect cluster is connecting to a Kafka instance that is managed by Strimzi, the user can configure the 
-`caCertChainAndKey` property to point to the Strimzi cluster CA.
-
-A `KafkaConnect` instance with a secure endpoint using the cluster CA will look like:
-
-```
-apiVersion: kafka.strimzi.io/v1beta1
-kind: KafkaConnect
-spec:
-  restListeners:
-  - protocol: https
-    caCertChainAndKey:
-      secretName: <cluster-name>-cluster-ca
-      key: ca.key
-      certificate: ca.crt
-  ...
-```
-
-As part of this proposal the `<cluster-name>-cluster-ca` secret will be updated to include the cluster CA certificate 
-as currently it only includes the private key.
-
-### Certificate renewals
-
-The self-signed TLS certificate will have a 1-year expiration.
-Thirty days before the self-signed TLS certificate expires, the operator will automatically renew the certificate, replace the old self-signed certificate, and restart the pods in a controlled manner to ensure the new certificates are in use by Kafka Connect without losing availability.
-
-This would need to be a multi-phase process consisting of the following steps:
-
-1. Generate new certificate
-2. Distribute the new certificate to all pods and cluster operator truststores
-3. Replace the key and roll all Connect pods
-4. When the old certificate expires, remove it from the truststores and roll all Connect pods
+If the user has enabled an encrypted HTTPS listener and the certificate they referenced in the `certChainAndKey` field 
+is not signed by a well known CA, they must also configure the `strimziTruststoreCert`. This property must point to an 
+existing secret that contains a certificate that Strimzi can use to generate a truststore. The resulting truststore will 
+be used by the Strimzi operators, e.g. KafkaConnector, MirrorMaker2 when communicating with Kafka Connect.
 
 ## Affected/not affected projects
 
@@ -154,6 +134,18 @@ This proposal affects the `strimzi-kafka-operator` project.
 This proposal does not change the default behaviour of the Kafka Connect or MirrorMaker2 clusters so any existing clusters will not be affected.
 
 ## Rejected alternatives
+
+### User provides the CA for the Kafka Connect REST API
+
+In an earlier version of this proposal the user defined only the CA certificate and Strimzi used that to generate the actual certificate 
+that Connect would host. The disadvantages of this proposal are Strimzi is forced to manage the lifecycle of the CA certificate and 
+it does not allow users to use their own CA.
+
+The advantage is that it would allow users to reuse the Strimzi cluster CA if they didn't want to manage their own CA.
+
+This approach was rejected because the certificate management in Strimzi is not ideal, so picking an approach where the certificates 
+are managed by the user is preferable. If the user does not want to manage their own certificates they can leave the Connect rest API 
+as unencrypted.
 
 ### TLS encrypting the Kafka Connect REST API by default
 
--- a/031-tls-encrypt-the-kafka-connect-rest-api.md
+++ b/031-tls-encrypt-the-kafka-connect-rest-api.md
@@ -1,0 +1,156 @@
+# TLS encrypting the Kafka Connect REST API
+
+This proposal adds the option to reduce access to the Kafka Connect REST API by enabling TLS encryption on the Kafka Connect REST listener.
+
+## Current situation
+
+Currently, instances of Kafka Connect that are deployed by the Strimzi operators are configured with the default REST API endpoint settings.
+This means that the Kafka Connect REST API endpoint uses HTTP on port 8083 and that the Strimzi `KafkaConnect`, `KafkaConnector` and `KafkaMirrorMaker2` operators make unsecured REST client calls based on this default configuration.
+
+The default network policies created by the Strimzi operators restrict incoming REST API calls to only allow access from the operator pod.
+Users can define further network policies to override the default policy and allow wider access.
+
+## Motivation
+
+There is currently no way to TLS encrypt the Kafka Connect REST API, leaving the Connect REST listener as perhaps the last key endpoint that cannot currently be restricted using Strimzi.
+
+## Proposal
+
+This proposal adds REST api configuration options to the `KafkaMirrorMaker2` and `KafkaConnect` CRDs.
+
+A new field is added to the `KafkaConnect` and `KafkaMirrorMaker2` specs named `restListeners`, which allows users to configure an HTTP listener on port 8443, an encrypted HTTPS listener on port 8443, or both in the Kafka Connect configuration.
+This enables users to access the REST API with TLS if required.
+The `KafkaConnector`, `KafkaConnect` and `KafkaMirrorMaker2` operators will use the TLS encrypted listener when it is enabled, even if an unencrypted listener is present.
+
+The `restListeners` field is designed for potential future extension, and is a list of items that contains a `protocol` field and an optional `restCertChainAndKey` field.
+
+For this proposal, the `protocol` field only supports two values: `http` and `https`. If the `restListeners` list is empty the `KafkaConnect` and `KafkaMirrorMaker2` 
+runtimes will be created with a single unencrypted listener on 8083 to match the existing behaviour.
+
+The `restCertChainAndKey` field will be required if the `https` protocol is selected. This field will contain the certificate and key the Kafka Connect operator will 
+use to create the self-signed certificate. The Kafka Connect and MirrorMaker2 operators will use the certificate to create a truststore for communicating with the REST API.
+
+For example, the following CR will enable the HTTP listener on port 8083 and the HTTPS listener on port 8443:
+```
+apiVersion: kafka.strimzi.io/v1beta1
+kind: KafkaConnect
+spec:
+  restListeners:
+  - protocol: http
+  - protocol: https
+  ...
+```
+
+The schema for `restListeners` will also include properties to provide the certificate and key that should be used to create the self-signed certificate and for the operators to use to communicate with Kafka Connect as follows:
+```
+openAPIV3Schema:
+  type: object
+  properties:
+    spec:
+      type: object
+      properties:
+        restListeners:
+          type: array
+          items:
+            type: object
+            properties:
+              protocol:
+                type: string
+                enum:
+                - http
+                - https
+                description: The protocol of the REST listener.
+              restCertChainAndKey:
+                type: object
+                properties:
+                  certificate:
+                    type: string
+                    description: The name of the ca certificate file in the Secret.
+                  key:
+                    type: string
+                    description: The name of the ca private key file in the Secret.
+                  secretName:
+                    type: string
+                    description: The name of the Secret containing the certificate.
+                  required:
+                    - key
+                    - certificate
+                    - secretName
+                description: CA certificate to use to generate.
+            required:
+            - protocol
+          description: List of additional REST listeners.
+        ...
+```
+
+the above CR would result in the following properties in the generated connect configuration:
+
+```
+listeners: https://:8443,http://:8083
+rest.advertised.listener: https
+rest.advertised.port: 8443
+listeners.https.ssl.client.auth: none
+listeners.https.ssl.truststore.location: /tmp/kafka/kafka-connect-rest.truststore.p12
+listeners.https.ssl.truststore.password: ***generated password***
+listeners.https.ssl.truststore.type: PKCS12
+listeners.https.ssl.keystore.location: /tmp/kafka/kafka-connect-rest.keystore.p12
+listeners.https.ssl.keystore.password: ***generated password***
+listeners.https.ssl.keystore.type: PKCS12
+```
+
+### Certificate renewals
+
+The self-signed TLS certificate will have a 1-year expiration.
+Thirty days before the self-signed TLS certificate expires, the operator will automatically renew the certificate, replace the old self-signed certificate, and restart the pods in a controlled manner to ensure the new certificates are in use by Kafka Connect without losing availability.
+
+This would need to be a multi-phase process consisting of the following steps:
+
+1. Generate new certificate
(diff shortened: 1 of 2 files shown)


031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Motivation
+
+There is currently no way to TLS encrypt the Kafka Connect REST API, leaving the Connect REST listener as perhaps the last key endpoint that cannot currently be restricted using Strimzi.
+
+## Proposal
+
+This proposal adds REST api configuration options to the `KafkaMirrorMaker2` and `KafkaConnect` CRDs.
+
+A new field is added to the `KafkaConnect` and `KafkaMirrorMaker2` specs named `restListeners`, which allows users to configure an HTTP listener on port 8443, an encrypted HTTPS listener on port 8443, or both in the Kafka Connect configuration.
```
ppatierno, line 21: which ports are used when both HTTP and HTTPS are enabled? It cannot be the same 8443 for both, or? Should be HTTP using 8083?

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,156 @@
...
+This enables users to access the REST API with TLS if required.
+The `KafkaConnector`, `KafkaConnect` and `KafkaMirrorMaker2` operators will use the TLS encrypted listener when it is enabled, even if an unencrypted listener is present.
+
+The `restListeners` field is designed for potential future extension, and is a list of items that contains a `protocol` field and an optional `restCertChainAndKey` field.
+
+For this proposal, the `protocol` field only supports two values: `http` and `https`. If the `restListeners` list is empty the `KafkaConnect` and `KafkaMirrorMaker2` 
+runtimes will be created with a single unencrypted listener on 8083 to match the existing behaviour.
+
+The `restCertChainAndKey` field will be required if the `https` protocol is selected. This field will contain the certificate and key the Kafka Connect operator will 
+use to create the self-signed certificate. The Kafka Connect and MirrorMaker2 operators will use the certificate to create a truststore for communicating with the REST API.
```
ppatierno, line 31: The above two sentences are not pretty clear to me.
So, to enable HTTPS on the REST API we need to generate a public cert and a private key (and it should be the KafkaConnect operator going to do so); the public cert will be the one provided by the HTTPS server to the client during the TLS handshake and it has to be signed by a "CA" certificate which will be "imported" in the truststore of the clients to trust the public cert got from the HTTPS server.
If the HTTPS server is providing a self-signed certificate to the client, it means that the client has to have exactly that certificate in its truststore to accept the TLS handshake but I don't think it's the way we want, right?
I don't think this process is really clear from the above description. 

> ...  to create the self-signed certificate

What's the self-signed certificate you are mentioning there?

> The Kafka Connect and MirrorMaker2 operators will use the certificate to create a truststore ....

Which one are you referring to?

ppatierno:
@katheris I had a first pass but stopped reading in the first part where you start mentioning certs, self-signed certs and so on because I think that that part is important to understand the rest of the proposal. I don't think it's pretty clear, at least to me :-)

031-tls-encrypt-the-kafka-connect-rest-api.md
katheris: Yes, that's a mis-type, HTTP will use 8083 and HTTPS will use 8443. I'll update the proposal.

katheris:
Thanks @ppatierno I'll have a look at your comments and see if I can make it clearer.

031-tls-encrypt-the-kafka-connect-rest-api.md
katheris: @ppatierno I've updated the document and added a specific section the discusses the certificates, let me know if it is clearer now.

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,168 @@
+# TLS encrypting the Kafka Connect REST API
+
+This proposal adds the option to reduce access to the Kafka Connect REST API by enabling TLS encryption on the Kafka Connect REST listener.
```
scholzj, line 3: Does it really reduce access? Or just encrypt the connection?

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,168 @@
...
+
+Currently, instances of Kafka Connect that are deployed by the Strimzi operators are configured with the default REST API endpoint settings.
+This means that the Kafka Connect REST API endpoint uses HTTP on port 8083 and that the Strimzi `KafkaConnect`, `KafkaConnector` and `KafkaMirrorMaker2` operators make unsecured REST client calls based on this default configuration.
+
+The default network policies created by the Strimzi operators restrict incoming REST API calls to only allow access from the operator pod.
+Users can define further network policies to override the default policy and allow wider access.
+
+## Motivation
+
+There is currently no way to TLS encrypt the Kafka Connect REST API, leaving the Connect REST listener as perhaps the last key endpoint that cannot currently be restricted using Strimzi.
```
scholzj, line 15: Again, does it restrict the access? Or just encrypt the communication?

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,168 @@
...
+A new field is added to the `KafkaConnect` and `KafkaMirrorMaker2` specs named `restListeners`, which allows users to configure an HTTP listener on port 8083, an encrypted HTTPS listener on port 8443, or both in the Kafka Connect configuration.
+This enables users to access the REST API with TLS if required.
+The `KafkaConnector`, `KafkaConnect` and `KafkaMirrorMaker2` operators will use the TLS encrypted listener when it is enabled, even if an unencrypted listener is present.
+
+The `restListeners` field is designed for potential future extension, and is a list of items that contains a `protocol` field and an optional `restCertChainAndKey` field.
+
+For this proposal, the `protocol` field only supports two values: `http` and `https`. If the `restListeners` list is empty the `KafkaConnect` and `KafkaMirrorMaker2` 
+runtimes will be created with a single unencrypted listener on 8083 to match the existing behaviour.
+
+The `restCertChainAndKey` field will be required if the `https` protocol is selected.
```
scholzj, line 30: `restCertChainAndKey` does make me expect a server certificate certificate. Not CA. If it should be a CA, maybe the name should make it more clear.

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,168 @@
...
+
+This proposal adds REST api configuration options to the `KafkaMirrorMaker2` and `KafkaConnect` CRDs.
+
+A new field is added to the `KafkaConnect` and `KafkaMirrorMaker2` specs named `restListeners`, which allows users to configure an HTTP listener on port 8083, an encrypted HTTPS listener on port 8443, or both in the Kafka Connect configuration.
+This enables users to access the REST API with TLS if required.
+The `KafkaConnector`, `KafkaConnect` and `KafkaMirrorMaker2` operators will use the TLS encrypted listener when it is enabled, even if an unencrypted listener is present.
+
+The `restListeners` field is designed for potential future extension, and is a list of items that contains a `protocol` field and an optional `restCertChainAndKey` field.
+
+For this proposal, the `protocol` field only supports two values: `http` and `https`. If the `restListeners` list is empty the `KafkaConnect` and `KafkaMirrorMaker2`
```
scholzj, line 27: I wonder if this should be `type` instead of `protocol` ... but really just 🤔  about it ... not sure what is better.

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,168 @@
...
+the above CR would result in the following properties in the generated connect configuration:
+
+```
+listeners: https://:8443,http://:8083
+rest.advertised.listener: https
+rest.advertised.port: 8443
+listeners.https.ssl.client.auth: none
+listeners.https.ssl.truststore.location: /tmp/kafka/kafka-connect-rest.truststore.p12
+listeners.https.ssl.truststore.password: ***generated password***
+listeners.https.ssl.truststore.type: PKCS12
```
scholzj, line 94: Wait ... what is the truststore based on? How do users generate their user certificates for this?

scholzj (commented):
Some things which will be very soon requested for this and should probably be thought through up-front:
* Adding additional SANs so that clients from outside can connect securely
* Users who cannot provide a CA but want it to use a a specific server certificate (similar to the listener certificates in Kafka CR)
* Integration with Let's Encrypt (that might not be possible really) / Cert-manager, Venafi, Vault, etc.
* Changing expiration / renewal period

I think you did a good job writing the proposal. But I'm not really sold for it TBH. The CA handling is our weakest part for a long time - the renewal does not work properly for a 1+ year. And this proposal builds on top of it. Plus the solution by linking to the Kafka cluster CA is not available to users using only Connect / MM2. 

I think we should use the fact that we do not have any historical baggage here and IMHO think a bit more out-of-the-box and consider how to offload this to cert-manager and other specialised tools rather than build on top of problematic parts. That also seems to be the direction many users seem to be asking about.

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,168 @@
...
+listeners.https.ssl.keystore.password: ***generated password***
+listeners.https.ssl.keystore.type: PKCS12
+```
+
+### Certificates
+
+If the user decides to enable an encrypted HTTPS listener, they must have the `restCertChainAndKey` 
+property present in the `KafkaConnect` CR. The property must point to an existing secret that contains the 
+CA certificate and CA private key that they want Kafka Connect to use. This secret can be created by the 
+user, or it can be the [Strimzi cluster CA secret](#using-the-strimzi-cluster-ca).
```
tombentley, line 105: I realise this is tricky, but do we _really_ want Strimzi to get deeper into the business of issuing its own certificates like this? It's often not what people operating Strimzi want: They often want to use things like vault or public CAs to issue a server certificate, in which case there is no CA certificate that Strimzi can access to issue itself a server certificate.

An alternative model would be just to point to a server certificate and key in the `KafkaConnect` or MM CRs, and assume that this is issued by some unspecified means outside the system. The CO could take care to restart the Pod when the certificate generation changed. Then in the CO configuration (i.e. it's ENV) we could have the option of using the cluster CA to issue these things automatically. 

That would allow people to issue their own endpoint certificates however they chose, but it also wouldn't build in the assumption that Strimzi is doing the issuing. It's especially important to be careful about building these assumptions into the CRDs because those are APIs which are really difficult to evolve. Doing it in the CO config makes it a bit easier to deprecate configs at a later date, if needed.

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,168 @@
...
+## Motivation
+
+There is currently no way to TLS encrypt the Kafka Connect REST API, leaving the Connect REST listener as perhaps the last key endpoint that cannot currently be restricted using Strimzi.
+
+## Proposal
+
+This proposal adds REST api configuration options to the `KafkaMirrorMaker2` and `KafkaConnect` CRDs.
+
+A new field is added to the `KafkaConnect` and `KafkaMirrorMaker2` specs named `restListeners`, which allows users to configure an HTTP listener on port 8083, an encrypted HTTPS listener on port 8443, or both in the Kafka Connect configuration.
+This enables users to access the REST API with TLS if required.
```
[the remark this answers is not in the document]
katheris: Yes, the sentence "which allows users to configure an HTTP listener on port 8083, an encrypted HTTPS listener on port 8443, or both in the Kafka Connect configuration" is meant to imply they can choose to have only an HTTP listener or only an HTTPS listener or both, I'll see if I can update it to make it clearer.

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,168 @@
...
+The `KafkaConnector`, `KafkaConnect` and `KafkaMirrorMaker2` operators will use the TLS encrypted listener when it is enabled, even if an unencrypted listener is present.
+
+The `restListeners` field is designed for potential future extension, and is a list of items that contains a `protocol` field and an optional `restCertChainAndKey` field.
```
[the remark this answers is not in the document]
katheris: I added the `rest` on the beginning to align with the `brokerCertChainAndKey` property, but I did go back and forth so would be happy to drop the `rest`.

katheris:
@tombentley @scholzj On the comments around the certificate renewals. I agree that having the operator handle the certificate adds additional complexity and does build on top of a system that is not ideal in the first place. When I was writing the proposal I did actually consider writing it so that instead of providing a CA the user provides the actual certificate they want to use and the user has to manage it. I think that would integrate really well with things like cert manager because as far as I can tell those sorts of services result in a secret anyway.

The obvious trade-off is that it means every user that is using this system has to create a certificate themselves, whether they are integrating with a Strimzi managed Kafka or not. However, since this proposal does not enable the security by default so it does not stop people trying it out and if they are using Strimzi in production then they will likely want to provide their own certificates anyway.

Overall I agree with you both that offloading the management of the certificates would likely still satisfy the user requirements.

I'd be interested to know what @ppatierno's and @ajborley's thoughts are

031-tls-encrypt-the-kafka-connect-rest-api.md
katheris: The "reduce access" phrase I reused from the previous proposal, but I agree changing to "The proposal adds the option to encrypt requests to the Kafka Connect REST API ..." would be clearer

031-tls-encrypt-the-kafka-connect-rest-api.md
katheris: I think this is a good point, I'll hold off on making the change until we have agreed whether this should be a CA or not though.

031-tls-encrypt-the-kafka-connect-rest-api.md
katheris: These properties were copied from the original proposal, but you're right, if this is only hosting a HTTPS endpoint and not doing mutual TLS would we only need a keystore, not a truststore?

scholzj: Yes, correct. It would need only the keystore. And the operator will only need to trust the CA / public key to be able to connect.

scholzj:
> The obvious trade-off is that it means every user that is using this system has to create a certificate themselves, whether they are integrating with a Strimzi managed Kafka or not.

I guess this is the next question if we decide to integrate with something else. The integration can scale from shared Secret up to having something like a cert-manager as pre-requisite for using Strimzi and having Strimzi orchestrate it directly (e.g. Strimzi would create the cert-manager CRs to get a CA and server cert). I guess something in the middle might be best - but at the end this would define how much of this would be user's manual work and how much it would be automated.

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Certificates
+
+If the user decides to enable an encrypted HTTPS listener, they must have the `restCertChainAndKey` 
+property present in the `KafkaConnect` CR. The property must point to an existing secret that contains the 
+CA certificate and CA private key that they want Kafka Connect to use. This secret can be created by the 
+user, or it can be the [Strimzi cluster CA secret](#using-the-strimzi-cluster-ca).
+
+The CA certificate and CA private key configured by the user in the `KafkaConnect` CR will be used by the Kafka Connect 
+operator to generate a self-signed certificate. This self-signed certificate will be used to configure the keystore and
```
ppatierno, line 108: Still I think that "self-signed" is not right here. If you have a CA certificate and a CA private key and you generate a new "server" certificate signed by the CA certificate, the "server" certificate is not self-signed but it's signed by the CA certificate. It's the CA certificate that should be self-signed if a root certificate or it could be signed by another CA certificate and so on until you reach a certificate up to the chain which is the root one and self-signed.

ppatierno:
I agree, I think that right direction is offloading to user to provide Secrets so that it works with cert-manager and things like that. Isn't it the direction that @tombentley was taking while refactoring the certificates handling in Stimzi anyway?

031-tls-encrypt-the-kafka-connect-rest-api.md
katheris: Yes I think you are correct here @ppatierno. Here I mean the certificate signed by the CA we just generated, not technically a self-signed certificate. If we were to use a true self-signed certificate we would still have the issue of Strimzi needing to managing the lifecycle of the certificate. If it were self-signed I guess we would also need a different mechanism for the operator and users to trust the certificate.

katheris:
@tombentley @ppatierno @scholzj are you happy for me to update this proposal to instead propose the user provides to the CR a secret with an already created certificate? So taking away any responsibility from Strimzi to maintain the certificate?

031-tls-encrypt-the-kafka-connect-rest-api.md
ppatierno: When a TCP server provides to the client a self-signed certificate, the only way you have to trust that certificate is having the certificate itself in your truststore, so no a CA certificate but the actual TCP server self-signed certificate; I don't think it's our use case, or? We always want a CA (or a CA chain) used to sign the TCP server certificate that won't be never self-signed.

katheris: I think the original proposal was to have a self-signed certificate, but every other certificate in Strimzi uses a CA. I guess if I update the proposal so the user supplies the certificate though it doesn't matter, we just need one place to specify the certificate that Kafka Connect should host and one place to provide the certificate that the Kafka Connect operator should trust, the one given for operator to trust could be the same certificate or the CA certificate depending on whether the user has decided to use self-signed?

ajborley:
@katheris - I'm also happy with the plan to provide certs via secrets here - sounds a simpler and more flexible way to TLS enable the connection 👍

031-tls-encrypt-the-kafka-connect-rest-api.md
ppatierno: Yes, if we go this direction then it doesn't make any difference, I would just clarify in the proposal that the two scenarios are supported:

* the user is using a CA certificate (or a CA chains) to sign the server cert so the CA certificate (or chain) has to be added  to the truststore for connection.
* the user is using a self-signed cert as a server cert so it has to be added to the truststore for connection.

It could be implicit but most of time people are confused on TLS related stuff, so clarifying it is much better imho.

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Certificates
+
+If the user decides to enable an encrypted HTTPS listener, they must have the `certChainAndKey` 
+property present in the `KafkaConnect` CR. The property must point to an existing secret that contains the 
+certificate and private key that they want Kafka Connect to use. The certificate must contain the correct 
+Subject Alternative Names (SANs) similarly to when a user provides their own certificates for listeners:
+https://strimzi.io/docs/operators/latest/using.html#ref-alternative-subjects-certs-for-listeners-str
+
+If the user has enabled an encrypted HTTPS listener and the certificate they referenced in the `certChainAndKey` field
```
katheris, line 123: Could the `strimziTruststoreCert` also only be required if you want to use the Connectors CR if using Kafka Connect?  I can't see an obvious other time when the Kafka Connect/or operator needs to call the Kafka Connect endpoint.

I'm guessing for MirrorMaker it would be required for the MirrorMaker operator to start and stop the mirror maker connectors.

katheris:
@ppatierno @tombentley @scholzj @ajborley proposal has been updated so the user provides all certificates

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,161 @@
...
+### Certificates
+
+If the user decides to enable an encrypted HTTPS listener, they must have the `certChainAndKey` 
+property present in the `KafkaConnect` CR. The property must point to an existing secret that contains the 
+certificate and private key that they want Kafka Connect to use. The certificate must contain the correct 
+Subject Alternative Names (SANs) similarly to when a user provides their own certificates for listeners:
+https://strimzi.io/docs/operators/latest/using.html#ref-alternative-subjects-certs-for-listeners-str
+
+If the user has enabled an encrypted HTTPS listener and the certificate they referenced in the `certChainAndKey` field 
+is not signed by a well known CA, they must also configure the `strimziTruststoreCert`. This property must point to an
```
ppatierno, line 124: I would add an example of "well known CA" like Verisign, Let's Encrypt and so on.

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,161 @@
...
+
+If the user decides to enable an encrypted HTTPS listener, they must have the `certChainAndKey` 
+property present in the `KafkaConnect` CR. The property must point to an existing secret that contains the 
+certificate and private key that they want Kafka Connect to use. The certificate must contain the correct 
+Subject Alternative Names (SANs) similarly to when a user provides their own certificates for listeners:
+https://strimzi.io/docs/operators/latest/using.html#ref-alternative-subjects-certs-for-listeners-str
+
+If the user has enabled an encrypted HTTPS listener and the certificate they referenced in the `certChainAndKey` field 
+is not signed by a well known CA, they must also configure the `strimziTruststoreCert`. This property must point to an 
+existing secret that contains a certificate that Strimzi can use to generate a truststore. The resulting truststore will
```
ppatierno, line 125: Can't we use `trustedCertificates` instead of `strimziTruststoreCert`, I am not sure how much it makes sense having "strimzi" in the name and how much it could be confusing for users.
The `trustedCertificates` name is the same we use in other place of our CRs like bridge connecting to Kafka (so trusting the Strimzi CA for example) or even the same Kafka Connect connecting to Kafka (see `tls.trustedCertificates` in the spec).
I would also specified that the contained certificate was used to sign the cert provided in the `certChainAndKey`, not just saying "that Strimzi can use to generate a truststore" which sounds to be too much general.

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,161 @@
...
+                description: The protocol of the REST listener.
+              certChainAndKey:
+                type: object
+                properties:
+                  certificate:
+                    type: string
+                    description: The name of the certificate file in the Secret.
+                  key:
+                    type: string
+                    description: The name of the private key file in the Secret.
```
tombentley, line 75: Do we need to define the expected/supported format for the key?

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,161 @@
...
+  
+This enables users to access the REST API with TLS if required.
+The `KafkaConnector`, `KafkaConnect` and `KafkaMirrorMaker2` operators will use the TLS encrypted listener when it is enabled, even if an unencrypted listener is present.
+
+The `restListeners` field is designed for potential future extension, and is a list of items that contains a `protocol` field and fields for providing certificates if required `certChainAndKey`, `strimziTruststoreCert`
+
+For this proposal, the `protocol` field only supports two values: `http` and `https`. If the `restListeners` list is empty the `KafkaConnect` and `KafkaMirrorMaker2` 
+runtimes will be created with a single unencrypted listener on 8083 to match the existing behaviour.
+
+The `certChainAndKey` field will be required if the `https` protocol is selected. The `strimziTruststoreCert` will be required if the `https` protocol is selected and the provided
```
tombentley, line 34: The HTTP server would only need a trust store if it was going to authenticate clients using their client certificates. IIUC this proposal is just for adding support for TLS encryption, so I don't think we need any trust store (perhaps listing authentication as a non goal would make sense). Instead the `certChainAndKey` will contain: i) The end-entity certificate of the server (with appropriate subject alternative name etc), ii) And needed intermediate CA certificates iii) The root CA certificate that REST clients will need to trust if they're going to be able to authenticate the server certificate.

It's then entirely up to the client to configure _its_ truststore with the CA certificate in the server's certificate chain. 

In the case where _we_ are managing connectors via `KafkaConnector` that client is in the CO, so we will have to add it to the truststore for the HTTP client we use. I suppose it would be fine to just do that automatically, using the `certChainAndKey` in the `KafkaConnect` resource, rather than having an extra place to configure this HTTP client (it doesn't seem to add any value that I can see).

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,161 @@
...
+              protocol:
+                type: string
+                enum:
+                - http
+                - https
+                description: The protocol of the REST listener.
+              certChainAndKey:
+                type: object
+                properties:
+                  certificate:
```
tombentley, line 70: I think this should be `certificateChain`, as mentioned above. We should also define the format.

031-tls-encrypt-the-kafka-connect-rest-api.md
katheris: @tombentley I agree that Kafka Connect should not be creating a truststore, the truststore would only be used by the Strimzi operators. I put a question elsewhere about whether this would only be required if the managing via connectors is true. That was what I thought but wasn't sure if I'd missed something.

In terms of the certificates, the field I based it on is `brokerCertChainAndKey` and my reading of the docs "You can provide a certificate that includes the whole CA chain in the certificate file." is that you don't have to provide the whole chain, but I guess if we switched to requiring the chain for this field that would remove the need for the truststore field?

tombentley: It seemed a bit weird to configure the cert chain, complete with the root CA, and then, in the same resource, force the user to have to configure a trust store which contains that same root CA. But on thinking about it I suppose that (usually?) a cert chain doesn't include the root CA certificate, because it's assumed that the client will have it in its trust store. So maybe having a separate store for the root CA certificate makes sense for that reason alone. If you buy that argument, do you think we should validate that the cert chain _doesn't_ end with a self-signed CA cert (i.e. a root CA cert), to avoid misconfiguration?

> "You can provide a certificate that includes the whole CA chain in the certificate file."

I think this language is confusing. It's a bit ambiguous what a "certificate file" really is. I know what a _certificate_ is, and I know that I can store a sequence of them in PEM format in a file, and that would constitute a CA chain. I can also store certificate chains in other kinds of file (JKS, PKCS#12). But I don't think of any of those as a "certificate file", which sounds like it's only able to store a single certificate.

031-tls-encrypt-the-kafka-connect-rest-api.md
katheris: I think we need restListeners to be an object rather than a list because we need a field to specify the certificates

031-tls-encrypt-the-kafka-connect-rest-api.md
katheris: If I understand correctly there are two possible ways we can go here:
1. The user only has to configure a `certChainAndKey` field, where they provide a complete certificate chain and the private key.
2. The user specifies two things, firstly `certAndKey` field where they provide the certificate that should be hosted by Kafka Connect and the private key. Secondly, only if they are using the KafkaConnector resource and their certificate is not signed by a well-known CA, they provide `truststoreCert` where they give the root CA certificate that the Strimzi Kafka Connect operator can trust when calling Kafka Connect

I think the way I wrote the proposal assumes option 2, but the naming of the cert and key field is then wrong. Which approach do you think makes more sense for the user @tombentley?

ppatierno: As mentioned in a previous comment, we are heavily using `trustedCertificates` instead of `truststoreCert`.

031-tls-encrypt-the-kafka-connect-rest-api.md
katheris: Would `type` be confusing given that the listeners has a `type` field but that is `internal` vs `route` vs `loadbalancer`. Here we aren't specifying the way the listener is exposed in a Kubernetes sense, but just whether it is secure. We could instead of having `protocol` have a boolean field called `tls`. That would match the listeners, but wouldn't be extensible. Although I can't think of what other protocols the Connect rest listeners might support

031-tls-encrypt-the-kafka-connect-rest-api.md
katheris: Based on discussions at the community meeting it sounds like the operator may need to call Kafka Connect even if the Connectors CR is not being used

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,172 @@
...
+spec:
+  restListeners:
+  - tls: true
+    certChainAndKey:
+      secretName: my-secret
+      certificateChain: my-certificate.crt
+      key: my-key.key
+    trustedCertificates:
+      secretName: my-secret
+      certificate: ca.crt
```
scholzj, line 51: Should this be a list to allow multiple trusted certificates to be added?

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+```
+listeners: https://:8443,http://:8083
+rest.advertised.listener: https
+rest.advertised.port: 8443
+listeners.https.ssl.client.auth: none
+listeners.https.ssl.keystore.location: /tmp/kafka/kafka-connect-rest.keystore.p12
+listeners.https.ssl.keystore.password: ***generated password***
+listeners.https.ssl.keystore.type: PKCS12
+```
```
scholzj, line 118: * Does connect support using X509 files directly as KAfka does? If yes, maybe we should prefer that.
* With multi-node connect, the nodes talk with each other as well using the REST API. So this will also need to be configured to trust the other nodes. That should be added here to the config I guess as well.

031-tls-encrypt-the-kafka-connect-rest-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+This proposal adds REST api configuration options to the `KafkaMirrorMaker2` and `KafkaConnect` CRDs.
+
+A new field is added to the `KafkaConnect` and `KafkaMirrorMaker2` specs named `restListeners`, which allows users to configure the Kafka Connect configuration to have:
+  - a single HTTP listener on port 8083
+  - a single encrypted HTTPS listener on port 8443
+  - both an HTTP listener on port 8083 and an encrypted HTTPS listener on port 8443
+  
+This enables users to access the REST API with TLS if required.
+The `KafkaConnector`, `KafkaConnect` and `KafkaMirrorMaker2` operators will use the TLS encrypted listener when it is enabled, even if an unencrypted listener is present.
```
scholzj, line 27: I touched on it below as well -> but it should be probably added here as well. The TLS is also used internally between Connect nodes to forward REST requests etc. So it is not just the user and the operator but also the Connect node themself.

scholzj (commented):
So, what does #46 mean for this? Formally, this talks about Connect and #46 talks about Kafka. And the requirements are probably slightly different. But it seems that we are going in a different direction in each of them which is a bit weird. I wonder if there should be some synergy and re-use between them.

031-tls-encrypt-the-kafka-connect-rest-api.md
katheris: The intention was that the user would specify the root certificate here, but if they need to update the certificates then I think having the option to add multiple trusted certificates would help with that so I'll update this to be an array.

031-tls-encrypt-the-kafka-connect-rest-api.md
katheris: @scholzj do you mean using X509 file as the keystore and truststore files, or specifying X509 certificates directly in the config? The docs are unclear but I tried supplying a PEM file for my keystore and truststore and ran into a NoSuchAlgorithmException, so it seems the answer is no, but I'll keep investigating to make sure.

katheris:
In terms of how this interacts with #46 I think the high level approach to certificates is aligned, in that it encourages the user to use an external system for generating certificates. The key difference is that in this proposal the user is required to manage the certificate requesting on their own, whereas in the other proposal Strimzi does handle the certificate requesting for you. The approach in this proposal matches the way that certificates are currently provided to the connect operator so that the connect instances can trust Kafka. Currently the connect operator doesn't manage any certificates itself, it only uses the ones it is given.

The advantage of the user owning the certificate requesting process is that the connect operator can work with any certificate generation system as long as that system puts the certificates in a secret. Since the connect operator is often used with Kafka instances that aren't managed by Strimzi I think this flexibility is useful as it means users are less likely to need to change their certificate processes to fit in with what Strimzi supports.

Proposal #46 does not mention the connect operator at all, and from my understanding none of the changes suggested require the connect operator to start handling certificate requests.

If we did want to align this more closely with #46 I think it would update the `restListeners` object to contain fields that match the ones in #46 e.g.
```
    type: cert-manager.io
    validityDays: <integer> # used for CertificateRequest.spec.duration
    issuer:
      name: <string>  # used for CertificateRequest.spec.duration
      kind: <string>
      group: <string>
    generateSecretOwnerReference: <boolean>
    renewalDays: <integer>
```

Since the connect operator runs as part of the cluster operator, I assume it wouldn't be too hard to have it make use of the new certificate request process. Then instead of the cluster operator looking in the Connect CR spec for the certificate it needs to trust when managing connectors, the connect operator could add the resulting certificate secret name to the status and the cluster operator would look there instead.

031-tls-encrypt-the-kafka-connect-rest-api.md
scholzj: > @scholzj do you mean using X509 file as the keystore and truststore files, or specifying X509 certificates directly in the config? The docs are unclear but I tried supplying a PEM file for my keystore and truststore and ran into a NoSuchAlgorithmException, so it seems the answer is no, but I'll keep investigating to make sure.

Yeah, with the first point I meant using the PEM/X509 files directly (i.e. the key and crt files). Kafka supports it, I'm not sure about Connect, but it sounds like it does not. Maybe something @tombentley or someone can look into in Kafka for the longterm

katheris:
> The cases are not the same -> the Kafka cluster needs a CA with server certs, here we basically need only a server cert. But it seems to be a very different direction and it concerns me a bit. Also, there will be users who do not want to use Kubernetes Secrets since they are not that secret. #46 deals with that with suggesting possible integration beyond secrets. This does not really open the doors for this.

For connect if we are using the `KafkaConnectors` type then the operator does need access to the CA as well as server certs. We could certainly adapt this proposal to more closely align to #46. I think including the "external" type that is suggested here: https://github.com/tombentley/proposals/blob/ca-abstraction/035-ca-abstraction.md#rejected-alternatives along with other types like cert manager gives a range of options to the user. @ppatierno @tombentley what do you think?

ppatierno:
> operator does need access to the CA as well as server certs

Can you explain why? Maybe I don't remember exactly how the KafkaConnector works but if there is a TLS handshake to secure a connection, a CA is enough to validate the server certificate or the server certificate is enough with no need of the CA (for example if the server certificate is self-signed but even if it's not).

katheris:
> > operator does need access to the CA as well as server certs
> 
> Can you explain why?

If the `KafkaConnectors` type is in use then the Kafka Connect operator has to call the REST API endpoint. If the REST API is hosting a https endpoint, then either the connect operator has to make an insecure connection or it has to trust the certificate that the REST API is presenting. If the certificate being presented is a self-signed cert, then the operator just trusts that, but in most cases we would assume the certificate has been signed by a CA, so the operator should trust the CA instead of the end certificate.

We did discuss in the community call I think whether the operator could just trust the end certificate, but if I remember correctly it was designed that it was better to follow the approach that users would expect, which would be for the operator to trust the CA. That way if the end certificate changes the connect operator doesn't need to be updated, it is only updated if the CA cert changes.

ppatierno:
> We did discuss in the community call I think whether the operator could just trust the end certificate, but if I remember correctly it was decided that it was better to follow the approach that users would expect, which would be for the operator to trust the CA. That way if the end certificate changes the connect operator doesn't need to be updated, it is only updated if the CA cert changes.

Absolutely, I think no IT system trusts an end certificate but always trust the CA. 

> For connect if we are using the KafkaConnectors type then the operator does need access to the CA as well as server certs.

So is this sentence wrong? The operator needs access to the CA not the server certs.

yoniwg:
Hi,
Do you have any update on this proposal? It's been more than an year since the last update.
This feature is really important...
Thanks in advance

scholzj:
Triaged on the Strimzi Community call on 25.1.2024: This is blocked by the Kafka CA abstraction work that is a prerequisite for this. We should close this and re-open it / open a new proposal for this once the CA work is done.

PR #50, opened 2022-02-12
Support the Replication Factor changes for the kafka topics 

yyang48:
Signed-off-by: Yaodong Yang <yyang48@apple.com>

### DIFF
--- a/034-kafka-topic-replication-factor.md
+++ b/034-kafka-topic-replication-factor.md
@@ -1,20 +1,18 @@
 # Support Replication Factor Change for Kafka Topics
 
-Apache Kafka introduced the Replication Factor (RF) as the configuration that determines the number replicas kafka brokers will retain for a single data partition. A bigger number of the RF means both higher reliability for the data partition and higher infrastructure cost for topics.
+Apache Kafka introduced the Replication Factor (RF) as the configuration that determines the number of replicas kafka brokers will retain for a single data partition. A bigger number of the RF means both higher reliability for the data partition and higher infrastructure cost for topics.
 
-Apache Kafka allows users to change the RF for any topic partition through the native scripts during the runtime. However, it's not a lightweight process, because users usually need to decide which brokers to host new replicas and which replicas need to be dropped, or even conduct a cluster rebalance after the change of the RF.  
+Apache Kafka allows users to change the RF for any topic partition through a partition reassignment process in the runtime. However, it's not a simple process, because users usually need to decide which brokers to host new replicas and which replicas need to be dropped, or even conduct a cluster rebalance after the change of the RF.
 
-The goal of this proposal is to offer users a simple approach to change the topic RF through the KafkaTopic Custom Resources and Cruise Control, in the cloud native, strimzi style. 
+The goal of this proposal is to offer users an easy approach to change the topic RF through the KafkaTopic Custom Resources and Cruise Control, in the cloud native, strimzi style.
 
 ## Current situation
 
 Currently, users have two options to change the topic RF:
-* Leverage the Native Kafka script to change the RF. Users need to generate the partition replica assignment and balance the kafka clusters manually if needed. 
-* Delete the topic and recreate the topic with the desired RF.
+* Leverage the Native Kafka script to change the RF. Users need to generate the partition replica reassignment and balance the kafka clusters manually if needed. Clearly it's not as simple as it should be for the users.
+* Delete the topic and recreate the topic with the desired RF. It's not acceptable because of losing data, unless the user accepts that.
 
-Clearly none of the above approaches is user-friendly.
-
-When users attempt to change the RF in the KafkaTopic Resources, the Status of the KafkaTopic will become `NotReady` with `ReplicationFactorChangeException`. The following error message is shared with users.
+When users attempt to change the RF in the strimzi KafkaTopic Resources, the Status of the KafkaTopic will become `NotReady` with `ReplicationFactorChangeException`. The following error message is shared with users.
 
 `Changing 'spec.replicas' is not supported. This KafkaTopic's 'spec.replicas' should be reverted to 4 and then the replication should be changed directly in Kafka.`
 
@@ -22,16 +20,39 @@ When users attempt to change the RF in the KafkaTopic Resources, the Status of t
 
 With this new feature, it will become much easier to achieve the desired tradeoff between the cost and data reliability in the kafka clusters. 
 
-For instance, users can lower the RFs for selective less critical topics to reduce the infrastructure cost (Compute, Storage and Network). Also, for some critical kafka topics, users can increase the data reliability through the bump of the RFs ONLY for these specific topics. 
+For instance, users can lower the RFs for selective less critical topics to reduce the infrastructure cost (Compute, Storage and Network). 
 
-More importantly, users only need to specify the new RF number in the `KafkaTopic` resource and approve the change, in order to change the RF for kafka topics. The heavy lifting work, such as new replica placement, extra replica deletion and possible cluster rebalance, is completely taken care by the strimzi. 
+Also, for some critical kafka topics, users can increase the data reliability through the bump of the RFs ONLY for these specific topics. 
+
+In addition, users can temporally adjust the RF number to deal with Storage Capacity Shortage. For instance, we could avoid Disk Full Exceptions by temporally decrease the RF number for some topics. 
+
+More importantly, users only need to specify the new RF number in the `KafkaTopic` resource, and it's as simple as changing any properties for a normal k8s Custom Resource. The heavy lifting work, such as new replica placement, obsolete replica deletion and possible cluster rebalance, is completely taken care by the strimzi and Cruise Control. 
 
 ## Proposal
 
 ### API
-The Kafka Topic RF can be changed by users in the following steps:
-* Update the `spec.replicas` properties in the target `KafkaTopic` Resource in the kubernetes cluster
-* Watch the `Status` field of the `KafkaTopic Resource`, for the status update of this RF change.
+
+#### Strimzi KafkaTopic CRD
+The `spec.replicas` property in the `KafkaTopic` CRD will be changed from ReadOnly to ReadWrite Mode. Any change to this `spec.replicas` property from users will trigger an RF change for the targeting kafka topic.
+
+Under the hood, the Strimzi Topic Operator will watch the changes of `spec.replicas` from the `KafkaTopic` CR, interact with Cruise Control Service and return the result in the `Status` field of the `KafkaTopic` CR.
+
+#### Strimzi Topic Operator
+Since the Topic Operator needs to send requests to the Cruise Control Service for the RF changes, the Topic Operator should know the details of the Cruise Control Service Endpoint.
+
+We plan to introduce the following Environment Variables in the Topic Operator. These Environment Variables will be set up by the EntityTopicOperator.
+
+```java
+public static final String TC_CRUISECONTROL_ENABLED = "STRIMZI_CRUISECONTROL_ENABLED";
+public static final String TC_CRUISECONTROL_NAMESPACE = "STRIMZI_CRUISECONTROL_NAMESPACE";
+public static final String TC_CRUISECONTROL_HOST = "STRIMZI_CRUISECONTROL_HOST";
+public static final String TC_CRUISECONTROL_API_AUTH_ENABLED = "STRIMZI_CRUISECONTROL_API_AUTH_ENABLED";
+public static final String TC_CRUISECONTROL_API_SSL_ENABLED = "STRIMZI_CRUISECONTROL_API_SSL_ENABLED";
+public static final String TC_CRUISECONTROL_SECRET_NAME = "STRIMZI_CRUISECONTROL_SECRET_NAME";
+public static final String TC_CRUISECONTROL_API_SECRET_NAME = "STRIMZI_CRUISECONTROL_API_SECRET_NAME";
+```
+
+If the Cruise Control Service is not enabled in the Kafka Cluster, this change of RF will be disabled. Users will be notified with an error message.
 
 ### Design
 We propose to change the topic RFs with the support of `Cruise Control` and `Strimzi Topic Operator`. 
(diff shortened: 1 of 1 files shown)


034-kafka-topic-replication-factor.md
```diff
@@ -0,0 +1,76 @@
...
+
+#### Cruise Control
+Cruise Control Service will be leveraged to change the RF for topics in the kafka clusters, for the following reasons: 
+* Strimzi supports the Cruise Control Service, and it's ready for use with simple configurations. 
+* Cruise Control has native support for the RF changes. It automates the whole process, including selecting of new replicas and extra replicas, triggering the replica creation in the kafka brokers and monitoring the cluster during the process.
+
+#### Kafka Topic Operator
+Strimzi Topic Operator will watch the changes of `KafkaTopic` Resource. 
+
+For any new topic RF change event, the `Strimzi Topic Operator` will validate the change. For instance, the new RF should be >= the min insync replica set size for the target topic.
```
tombentley, line 47: Is this the only validation condition, or are there others? I guess we can't have > #brokers. But perhaps the TO should support a separate limit that's less than that, to allow someone administering the Kafka cluster to place an upper bound for people changing the `KafkaTopic` (that might be different people).

034-kafka-topic-replication-factor.md
```diff
@@ -0,0 +1,76 @@
...
+Cruise Control Service will be leveraged to change the RF for topics in the kafka clusters, for the following reasons: 
+* Strimzi supports the Cruise Control Service, and it's ready for use with simple configurations. 
+* Cruise Control has native support for the RF changes. It automates the whole process, including selecting of new replicas and extra replicas, triggering the replica creation in the kafka brokers and monitoring the cluster during the process.
+
+#### Kafka Topic Operator
+Strimzi Topic Operator will watch the changes of `KafkaTopic` Resource. 
+
+For any new topic RF change event, the `Strimzi Topic Operator` will validate the change. For instance, the new RF should be >= the min insync replica set size for the target topic. 
+
+For the valid changes, the `Strimzi Topic Operator` will send requests to `Cruise Control`. The Kafka Topic Operator will periodically poll the status from the Cruise Control, and update the status in the `Status` field of the `KafkaTopic` resource.
```
tombentley, line 49: I don't know the CC API, but does it return back some kind of cookie or identifier for the request that was made to its endpoint and that needs to be passed when polling for completion? If so, I assume this needs to be added to the status section of the CR. An example CR would help.

We should also consider the various erroneous flows:

1. What should the operator do if the topic is deleted (and possibly recreated) while a change in RF is happening?
2. What happens if CC gets restarted while a change in RF is happening (I assume CC has persisted its state somewhere, so this isn't a problem, but if that's the case let's say so).
3. What happens in the user changes the RF again, while a change in RF already is happening? Does the TO have to cancel the existing change, or wait for it to complete?
4. What happens if the TO crashes after make the request to CC to change the RF but before it's able to update the `KafkaTopic.status` (assuming we are storing some kind of request id/cookie there)?

034-kafka-topic-replication-factor.md
```diff
@@ -0,0 +1,76 @@
...
+For instance, users can lower the RFs for selective less critical topics to reduce the infrastructure cost (Compute, Storage and Network). Also, for some critical kafka topics, users can increase the data reliability through the bump of the RFs ONLY for these specific topics. 
+
+More importantly, users only need to specify the new RF number in the `KafkaTopic` resource and approve the change, in order to change the RF for kafka topics. The heavy lifting work, such as new replica placement, extra replica deletion and possible cluster rebalance, is completely taken care by the strimzi. 
+
+## Proposal
+
+### API
+The Kafka Topic RF can be changed by users in the following steps:
+* Update the `spec.replicas` properties in the target `KafkaTopic` Resource in the kubernetes cluster
+* Watch the `Status` field of the `KafkaTopic Resource`, for the status update of this RF change.
```
tombentley, line 34: The configuration env vars consumed by the TO are also an API. You'll need at least one in order to tell the TO the address to connect to CC on, but potentially you'll need more for any authentication or TLS config needed for the TO -> CC communication. Can we spell all those out please?

tombentley (commented):
Thanks for the proposal! I think this is definitely along the right lines, but there are a few details which it would be good to elaborate on.

034-kafka-topic-replication-factor.md
```diff
@@ -0,0 +1,76 @@
...
+
+Apache Kafka allows users to change the RF for any topic partition through the native scripts during the runtime. However, it's not a lightweight process, because users usually need to decide which brokers to host new replicas and which replicas need to be dropped, or even conduct a cluster rebalance after the change of the RF.  
+
+The goal of this proposal is to offer users a simple approach to change the topic RF through the KafkaTopic Custom Resources and Cruise Control, in the cloud native, strimzi style. 
+
+## Current situation
+
+Currently, users have two options to change the topic RF:
+* Leverage the Native Kafka script to change the RF. Users need to generate the partition replica assignment and balance the kafka clusters manually if needed. 
+* Delete the topic and recreate the topic with the desired RF.
```
kyguy, line 13: Is there any danger of allowing users to use the native Kafka script with this newly proposed feature in place? If so, should we disable it's use somehow as part of this work?

kyguy (commented):
Looks awesome @yyang48!

034-kafka-topic-replication-factor.md
```diff
@@ -0,0 +1,76 @@
...
+
+Clearly none of the above approaches is user-friendly.
```
ppatierno, line 15: I would say that the second option is not about to be "user-friendly", it's not acceptable because of losing data, unless the user accepts that.

034-kafka-topic-replication-factor.md
ppatierno, line 34: I haven't tried the interaction with CC in order to change a RF but does it work as a rebalance? If yes, it means it's going to generate a proposal and then the user has to accept or refuse it before seeing the replication factor being increased. If confirmed, I don't think it's going to be just changing `spec.replicas` and watch the `Status` field. It would need a more "complex" state machine as we have for rebalancing.

034-kafka-topic-replication-factor.md
```diff
@@ -0,0 +1,76 @@
...
+* Update the `spec.replicas` properties in the target `KafkaTopic` Resource in the kubernetes cluster
+* Watch the `Status` field of the `KafkaTopic Resource`, for the status update of this RF change.
+
+### Design
+We propose to change the topic RFs with the support of `Cruise Control` and `Strimzi Topic Operator`. 
+
+#### Cruise Control
+Cruise Control Service will be leveraged to change the RF for topics in the kafka clusters, for the following reasons: 
+* Strimzi supports the Cruise Control Service, and it's ready for use with simple configurations. 
+* Cruise Control has native support for the RF changes. It automates the whole process, including selecting of new replicas and extra replicas, triggering the replica creation in the kafka brokers and monitoring the cluster during the process.
```
ppatierno, line 42: what does this mean `new replicas and extra replicas` ? what do you mean by "extra replicas"?

034-kafka-topic-replication-factor.md
ppatierno: As I mentioned in a previous comment, I think we would need a state machine based on the interaction with the CC endpoint. I would try it and report how it works in the proposal, by describing a potential state machine in the TO. I guess it would need some user interaction to approve, unless we want the TO doing it automatically.

034-kafka-topic-replication-factor.md
```diff
@@ -0,0 +1,76 @@
...
+* Users modify the value of the `spec.replicas` properties in the target `KafkaTopic` resource.
+* Kafka Topic Operator watches this resource changes and send the request to the Cruise Control Service.
+* Kafka Topic Operator will periodically poll the result from the Cruise Control Service, and update the status in the `KafkaTopic` resource.
+* Once it's completed by the Cruise Control Service, the Kafka Topic Operator will mark the Status in the `KafkaTopic` resource as `Completed`.
+
+## Affected/not affected projects
+
+The following components will be affected by this proposal:
+* Strimzi Topic Operator
+* Kafka Rebalance Operator
```
ppatierno, line 64: From the proposal is not clear how the Kafka Rebalance Operator is affected. Are we assuming to factor out some common code to use in the TO?

034-kafka-topic-replication-factor.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Compatibility
+
+There is no backward compatibility issue for this proposal, as there is no new persistent states introduced. 
+
+The status of the RF changes will be cached in the memory of the `Strimzi Topic Operator`, and posted in the `Status` of the `KafkaTopic` resources.
+
+This feature will be disabled, if `Cruise Control` is NOT enabled. Users will get the Error Response in the `Status` field of the `KafkaTopic` resource, if they intend to change the topic RFs.
```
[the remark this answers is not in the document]
ppatierno: I would say that as per "operator" behaviour, on next reconcile the TO should be able to trigger the RF increase. I don't expect the user doing more.

034-kafka-topic-replication-factor.md
tomncooper: AFAIK the flow would be almost identical to the process and logic used for a full cluster rebalance, but initiated by the TO. I wonder if it makes sense to have a common way to execute these CC (POST) interactions, otherwise we are likely to be replicating a lot of existing logic from the KRAO in the TO?

034-kafka-topic-replication-factor.md
tomncooper: I think we need to look at what common code can be shared here as a lot of the interactions will be the same.

034-kafka-topic-replication-factor.md
yyang48: I think we should encourage users to use this feature to change the replication factor, instead of the Kafka native script. At the same time, it's also OK for them to use the native Kafka script, if they really want to. 

There should be no risk, unless they use both options in the same Kafka Cluster, at the same time. No matter which option the users pick, there will be one partition reassignment request that is created in the zookeeper path.

Also, is there any way to block users to use the native kafka script? Users can always ssh to kafka or zk pods and call the native scripts directly.

yyang48: In addition, if we have the cruise control running for a Kafka Cluster, it's better NOT to use the native kafka script for partition reassignment. 1) It will affect the goals of the cruise control. 2) Kafka can only execute one reassignment request at a time, so it's harder to submit any partition reassignment requests to kafka through the native Kafka script, when the cruise control is running.

So this is another reason we should encourage users not to use native kafka script for the partition reassignment tasks.

034-kafka-topic-replication-factor.md
yyang48: For the `extra replicas`, I mean the replicas that we need to remove. For the `new replica`, I meant the replicas we need to create.

For instance, if a user wants to change RF from 4 to 3. The current replica is [101, 102,103,104]. The cruise control may pick the replica [104] and delete it from the Kafka Brokers. So the 104 is the `extra replica`. 

I think `obsolete` is a better term than the `extra`. I will change it in this proposal.

034-kafka-topic-replication-factor.md
```diff
@@ -0,0 +1,76 @@
...
+* Watch the `Status` field of the `KafkaTopic Resource`, for the status update of this RF change.
+
+### Design
+We propose to change the topic RFs with the support of `Cruise Control` and `Strimzi Topic Operator`. 
+
+#### Cruise Control
+Cruise Control Service will be leveraged to change the RF for topics in the kafka clusters, for the following reasons: 
+* Strimzi supports the Cruise Control Service, and it's ready for use with simple configurations. 
+* Cruise Control has native support for the RF changes. It automates the whole process, including selecting of new replicas and extra replicas, triggering the replica creation in the kafka brokers and monitoring the cluster during the process.
+
```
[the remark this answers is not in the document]
yyang48: Good point! I will add it in the proposal. It's basically the endpoint Kyle mentioned above. 
[1] https://github.com/linkedin/cruise-control/wiki/REST-APIs#change-kafka-topic-configuration

yyang48:
Thanks a lot for these great questions, comments and suggestions. I will give a try in the code and will update this proposal with more details soon.

ppatierno:
@yyang48 are there any updates on this? Are you still working on it? Thanks!

034-kafka-topic-replication-factor.md
yyang48: Good idea. I added these Environment variables in the updated version.

yyang48: Based on the discussion, we agreed that we don't need to leverage the state machine to control this process, because the change of RF for a single Kafka topic is a very light weight process, compared with the cluster rebalance process.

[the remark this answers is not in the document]
yyang48: This is a good idea. I will think about it and explore it in the Cruise Control and KafkaRebalance Operator. My guess is we probably don't need all of them in the KafkaTopic Resource, as it will make it very complex for users. 

In the current proposal, I will use the default values for most of them, just to make it work earlier. After that, I can improve it by making some parameters configurable in the `KafkaTopic` Resource.

WDYT?

034-kafka-topic-replication-factor.md
yyang48: I added the #brokers configuration in the proposal.

For the upper bound of the RF, what's the reasonable number from your perspective? I feel 5 should be enough, but I could be wrong.

034-kafka-topic-replication-factor.md
yyang48: Yes, it's correct.

fvaleri:
Hello @yyang48, I'm willing to take this task and continue from where you started. If that's ok with you, please close this PR. I will open a new one where we can discuss the design details.

yyang48:
> Hello @yyang48, I'm willing to take this task and continue from where you started. If that's ok with you, please close this PR. I will open a new one where we can discuss the design details.

Hello @fvaleri , no problem, please continue to work on this design and feature. Feel free to let me know if anything I can help.

I'm closing this PR now.

see-quick:
> @fvaleri just wand to ask if there is any progress on this topic?

Hi @Syndlex, to the best of my knowledge, it has already been implemented (here is a correct proposal [1]).

[1] - https://github.com/strimzi/proposals/pull/101

fvaleri:
@Syndlex this feature was released in 0.41.0, with a couple of minor bug fixes in 0.44.0. In any case I would recommend to use the latest release available: 0.46.0. Note that deploying Cruise Control is a requirement.

https://strimzi.io/docs/operators/0.46.0/full/deploying#proc-cruise-control-topic-replication-str

PR #51, opened 2022-03-22
Cluster wide volume usage monitoring proposal

SamBarker:
Signed-off-by: Sam Barker <sbarker@redhat.com>


k-wall (commented):
Looks like a good start, no comments from me. thanks @SamBarker

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -0,0 +1,129 @@
...
+#### The quota plugin consumer should:
+
+- Should apply the largest delay caused by each disk which breaches the soft limit.
+    - The amount of throttling should be proportional to how much of the difference between the soft and hard limit
+      remains for a given Volume. Applying more throttle the closer to the hard limit things are.
+- Should apply a `PAUSE` of publication if *any* Volume in the cluster breaches the hard limit
+- Connect as an internal service using TLS and client certificates to connect to the Control Plane listener.
+- Use the Kafka Admin client to discover the list of currently active brokers
+- On startup and periodically thereafter read the Volume usage stats from the topic
+- Ensure that the quota plugin Volume updates are not throttled
```
scholzj, line 72: This probably deserves more explanation as to how it will be done.

scholzj (commented):
Thanks for the proposal. I left some comments in the text. But I think there are some additional things to consider:
* It might be good to describe how the expected configuration of the quota plugin will look like. That gives people better idea how it will be integrated etc.
* The Topic Operator is a key for configuring the topics. Once the broker is at the hard-limit, you might need to use it to change the retention of some topics etc. to _clean_ the storage. So we need to make sure that it works. (it is based on Kafka Streams API so it will be publishing and consuming messages)
* The approach this proposal seems to take is to fence the cluster rather than fence the broker with the disk running out of space. I wonder why it was chosen over fencing the broker and what kind of consequences were considered (the introduction briefly mentions that KIP-73 was not designed for this, but doesn't go into detail what the issue would be). This does not seem to fully protect the broker against replication for example in cases when the replicas on the disk getting full are behind the leaders and will still get a lot more data through replication even when the outside producers are blocked. So this proposal basically works only with the caveat that all replicas are _in-sync_.
* I agree that the JMX or other metric sources might not be the best source of information. But using a Kafka topic for the integration IMHO creates a very tightly closed loop. It would be hard to add any other mechanisms to control the Quota plugin - for example through outside. I think we should consider whether some better and more pluggable mechanism can be found.

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -0,0 +1,129 @@
...
+- Apply a freshness check to the volume usage it reads and ignore stale state [3]. The threshold for staleness
+  should be a configuration parameter
+
+#### Message schema
+
+##### Message
+
+| Field      | Type                                    | Description                                                                   |
+|------------|-----------------------------------------|-------------------------------------------------------------------------------|
+| SnapshotAt | `ISO 8601 Date Time`                    | Timestamp when the snapshot of volume usage was generated [1]                 |
```
ppatierno, line 86: any specific reasons for using a timestamp as part of the message payload instead the Kafka record timestamp set by the producer? It should match with the timestamp of the snapshot of the volume if you are sending the message right after getting volume info, or?

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -0,0 +1,129 @@
...
+- On startup and periodically thereafter read the Volume usage stats from the topic
+- Ensure that the quota plugin Volume updates are not throttled
+- Should take a configured action when it can not determine the state of all known brokers.
+    - The action should be one of `PAUSE` or `OPEN`.
+    - Defaulting to `PAUSE` as the fail-safe option.
+- Publish a metric showing when it is applying throttling due to a lack of data.
+- Apply a freshness check to the volume usage it reads and ignore stale state [3]. The threshold for staleness
+  should be a configuration parameter
+
+#### Message schema
```
ppatierno, line 80: Is it going to be a binary message or using JSON, XML or something else?

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -0,0 +1,129 @@
...
+   useful for hard limits as an absolute minimum of free space.
+3. Minimum Free Percentage. Triggers if the amount of free space on the volume drops below the configured proportion of
+   the Volumes total capacity.
+
+#### The quota plugin data publisher should:
+
+- Collect local volume usage
+- Publish usage to an internal compacted topic keyed by broker ID.
+- Collect and publish usage on start up and periodically after that.
+- Connect as an internal service using TLS and client certificates to connect to the Control Plane listener.
```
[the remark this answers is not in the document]
SamBarker: No, there is no specific reason to use the Control Plane listener, I was under the impression that's what the other Strimizi components used, but hadn't actually checked.  

I'll update to an appropriate listener once I work out what that is.

035-cluster-wide-volume-usage-quota-managemnet.md
SamBarker: Nothing particularly strong: 
 
 - A personal preference that in general things that are important to the business logic should be explicit in the model/domain. 
 - Preventing any issues with delays between reading the disk and message publication.
 - I have half a thought in my mind about not using a compacted topic and looking back through previous updates to deal with transient IO problems when publishing. So it felt more flexible to include it now.

035-cluster-wide-volume-usage-quota-managemnet.md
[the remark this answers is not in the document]
SamBarker: I've updated the text but will answer here as well for record.

I'd default to JSON as its human readable and easy to debug and I doubt the data volumes are going to be high enough for the extra verbosity to be a problem. 

Enabling compression on the topic might be worth doing.

cc: @ppatierno as you have asked the same question else where.

SamBarker: covered in https://github.com/strimzi/proposals/pull/51#discussion_r833855911

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -0,0 +1,129 @@
...
+- Publish metrics showing its connection status and a counter of the number of publication errors - to allow the
+  detection of connectivity issues.
+- Publish Volume Information with the following [schema](#Message-schema)
+
+#### The quota plugin consumer should:
+
+- Should apply the largest delay caused by each disk which breaches the soft limit.
+    - The amount of throttling should be proportional to how much of the difference between the soft and hard limit
+      remains for a given Volume. Applying more throttle the closer to the hard limit things are.
+- Should apply a `PAUSE` of publication if *any* Volume in the cluster breaches the hard limit
```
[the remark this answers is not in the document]
SamBarker: I've clarified the wording. Yes it should block external producers rather than stop publishing new metrics. 

I think the storage requirement of the metrics is roughly steady state. Given the number of messages published is 1 per broker and the number of volumes per broker is static in 99% of cases.

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -0,0 +1,129 @@
...
+
+Configuring the soft and hard limit thresholds is currently done in terms of bytes consumed, which forces users
+deploying the plugin to calculate a threshold for throttling and would require separate thresholds per volume to avoid
+out of space issues.
+
+## Proposal
+
+### Leverage kafka to distribute volume usage metrics throughout the cluster.
+
+By publishing per volume usage metrics to a compacted topic keyed by the broker ID each instance of the plugin will be able
```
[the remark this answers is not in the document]
SamBarker: It is possible to infer it from the [updateClusterMetadata](https://kafka.apache.org/30/javadoc/org/apache/kafka/server/quota/ClientQuotaCallback.html#updateClusterMetadata(org.apache.kafka.common.Cluster)) method on the ClientQuotaCallback interface.

However it might be simpler and more timely to inject it through configuration.

SamBarker:
> * I agree that the JMX or other metric sources might not be the best source of information. But using a Kafka topic for the integration IMHO creates a very tightly closed loop. It would be hard to add any other mechanisms to control the Quota plugin - for example through outside. I think we should consider whether some better and more pluggable mechanism can be found.

I've added an `Implementation details` section to try and address this thought, but that still needs a bit more thought to make sure it fully addresses the concern. 

I need to spend a bit more time thinking and developing thoughts before addressing the rest of the points above.

035-cluster-wide-volume-usage-quota-managemnet.md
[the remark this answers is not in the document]
scholzj: Strimzi compoenents connect to the replication listener.

SamBarker:
> * The approach this proposal seems to take is to fence the cluster rather than fence the broker with the disk running out of space. I wonder why it was chosen over fencing the broker and what kind of consequences were considered (the introduction briefly mentions that KIP-73 was not designed for this, but doesn't go into detail what the issue would be).

I've added a discussion of KIP-73 to the rejected alternatives section which is largely why I've ended up with a cluster wide fence. My thinking works as follows:

1. The quota API from KIP-257 does not provide a mechanism to identify replication traffic.
2. Reading the [rejected alternatives](https://cwiki.apache.org/confluence/display/KAFKA/KIP-257+-+Configurable+Quota+Management#KIP257ConfigurableQuotaManagement-Enablemanagementofclientquotasandreplicationquotasusingasinglecallbackinterface) section of KIP-257 clearly suggests its is not applied to replication traffic 
3. Provides a consistent latency profile across the cluster. Regardless of which broker is leader for a given partition.

Given the configuration of KIP-73 via `partitionId` and `replica` we would need a mechanism, within the quota plugin? but certainly at runtime, to reverse engineer replica layout from the logical volume. I'm not aware of any Kafka API's which would allow us to do that. It might be possible to via scanning the file system but that seems relatively more complicated and error-prone compared to the current proposal. 

> This does not seem to fully protect the broker against replication for example in cases when the replicas on the disk getting full are behind the leaders and will still get a lot more data through replication even when the outside producers are blocked. So this proposal basically works only with the caveat that all replicas are in-sync.

Yes, this proposal doesn't have a great story to tell when the cluster is out of sync. By giving all brokers within the cluster a view of the usage of all volumes with the soft limits the graceful degradation would hopefully buy enough time for some combination of natural attrition to help (data passing topic retention thresholds), external intervention such as a cluster admin or Kafka Cruise Control to resolve the situation.    

I'm not aware of any other solution to the out of sync problem. KIP-73 would only be partially successful given a way to dynamically step down the replication limit. 

I'd welcome any other suggestions or avenues to investigate.

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -89,14 +89,116 @@ To make all this work the quota plugin will need to introduce some new interface
...
+ */
+public interface QuotaPolicy {
+
+    /**
+     * Does the particular volume breach this policy's defined hard limit.
+     *
+     * @param volumeDetails details of the disk in question.
+     * @return <code>true</code> if this policy considers the volume to breach the limit otherwise <code>false</code>
+     */
+    boolean breachesHardLimit(VolumeDetails volumeDetails);
```
SamBarker, line 122: I'm wondering if this should accept  `Set<VolumeDetails>` instead so that global capacity quotas could be enforced, purely for backwards compatibility. 

However I don't know if that backwards compatibility is actually worth having as it's not really protecting the cluster/broker from running out of disk space.

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -89,14 +89,116 @@ To make all this work the quota plugin will need to introduce some new interface
...
+
+    /**
+     * Returns the fraction of the original quota this policy thinks is appropriate. Represented as percentage value between <code>0</code> and <code>1</code>
+     * <p>
+     * Where a fraction of <code>1.0</code> is un affected <br>
+     * Breaching the hard limit implies a quota factor of <code>0.0</code>
+     * @param volumeDetails details of the disk in question.
+     * @return A value between <code>0</code> and <code>1</code>.
+     */
+    double quotaFactor(VolumeDetails volumeDetails);
```
SamBarker, line 140: While this would appear to duplicate the `QuotaFactorSource` this acts on a single volume rather than the set of volumes within the cluster and provides a clearer expectation at an API level.

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -89,14 +89,116 @@ To make all this work the quota plugin will need to introduce some new interface
...
+
+import java.util.function.Supplier;
+
+/**
+ * Determines the current restriction factor to be applied to the client quota.
+ * Values returned are required to be in the range <code>0.0..1.0</code> inclusive.
+ * Where a value of `1.0` implies no additional restriction over and above the defined quota.
+ * A value of `0.0` implies that there is no quota available regardless of the defined quota.
+ */
+public interface QuotaFactorSource extends Supplier<Double> {
```
SamBarker, line 178: I've kept this deliberately generic to allow it support the Kafka Consumer of this proposal or an external service as a possible future extension. 

It would make little sense for each plugin instance to query an external service by supplying some set of disk usage statistics.

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -89,14 +89,116 @@ To make all this work the quota plugin will need to introduce some new interface
...
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+package io.strimzi.kafka.quotas;
+
+import java.util.concurrent.TimeUnit;
+
+public interface DataSourceTask extends Runnable {
```
SamBarker, line 192: This is the interface I have least clarity around. 

My basic thinking is it is a simple marker to be submitted to an external `ScheduledExecutorService` however I think there is a case for it being as simple as the `QuotaFactorSource` and entirely encapsulating the periodic nature of data publication.

PaulRMellor (commented):
Looks good. I made a few suggestions for clarity and consistency

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -0,0 +1,264 @@
...
+
+#### Limit Types
+
+1. Consumed space limit: Triggers if the consumed space of a volume breaches the value. Candidate for deprecation.
+2. Minimum Free Bytes: Triggers if the amount of free space on a volume drops below the configured level. Particularly
+   useful for hard limits as an absolute minimum of free space.
+3. Minimum Free Percentage. Triggers if the amount of free space on the volume drops below the configured proportion of
+   the Volumes total capacity.
+
+#### The quota plugin data publisher should:
```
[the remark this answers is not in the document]
SamBarker: The terminology around this certainly needs revisited. I'll try and clarify

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -0,0 +1,264 @@
...
+- Should apply a `PAUSE` level throttle to client produce requests if *any* Volume in the cluster breaches the hard
+  limit.
+- Connect as an internal service using TLS and client certificates to connect to the Replication listener.
+- Use the Kafka Admin client to discover the list of currently active brokers
+- On startup and periodically thereafter read the Volume usage stats from the topic
+- Ensure that the quota plugin Volume updates are not throttled
+- Should take a configured action when it can not determine the state of all known brokers.
+    - The action should be one of `PAUSE` or `OPEN`.
+    - Defaulting to `PAUSE` as the fail-safe option.
+- Publish a metric showing when it is applying throttling due to a lack of data.
```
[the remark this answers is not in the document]
SamBarker: Hmm. Interesting thought. 

It is probably a different level of requirement to others. I guess metrics probably are closer to implementation details than high level requirements. 

I'll move it

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -0,0 +1,277 @@
...
+
+##### Metrics
+- Throttling applied due lack of data. - Gauge with the value `0` not applied or `1` applied
+- Connection status - Gauge with the value `0` disconnected or `1` Connected
+- publication errors - Counter to allow the detection of connectivity issues.
+- Metrics showing usage per local volume.
+
+##### Configuring the plugin
+###### Existing properties
+Preserved for backwards compatability.
```
SamBarker, line 237: Could potentially call them deprecated. 

In the single volume case these are easy to support. For the multi volume case they can't be auto migrated sensibly and applying them as global limits while consistent has all of the issues mentioned in the introduction section.

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -0,0 +1,277 @@
...
+Notes:
+
+1. Explicitly mark when the volume usage was captured rather than depending on message publication timestamps.
+2. One could view the limits defined at the message level as being broker wide defaults, they could then optionally be
+   overridden at the VolumeDetails level with a volume specific limit.
+3. Separating freshness from topic eviction as freshness is a "business" concern of the plugin.
+
+##### Metrics
+- Throttling applied due lack of data. - Gauge with the value `0` not applied or `1` applied
+- Connection status - Gauge with the value `0` disconnected or `1` Connected
```
SamBarker, line 231: It could also be worth including `2` Error in the enum, but I'm not convinced. 

The other option is to include a message field on the guage and include any errors there on the disconnected state.

Its not clear to me how helpful either of those options would actually be.

LiamClarkeNZ:
> * I agree that the JMX or other metric sources might not be the best source of information. But using a Kafka topic for the integration IMHO creates a very tightly closed loop. It would be hard to add any other mechanisms to control the Quota plugin - for example through outside. I think we should consider whether some better and more pluggable mechanism can be found.

@scholzj does this need to be pluggable currently though? Or could it be deferred until later when there's a subsequent use case to help shape design of that pluggable mechanism?

Alternatively, if we're looking to go pluggable, I'd suggest we should extract a set of base interfaces that the core functionality works on as a new artifact with common configuration applied at that level, and then make the existing static quota and @SamBarker's new one both pluggable implementations...

scholzj:
> @scholzj does this need to be pluggable currently though? Or could it be deferred until later when there's a subsequent use case to help shape design of that pluggable mechanism?

@LiamClarkeNZ Well, I think that the current design is not something what can be easily adapted to be pluggable. So doing it later might mean yet another redesign. I don't think this proposal necessarily needs to handle all the details and implementation of the pluggability. But I think it is something what would be good to take into account.

LiamClarkeNZ:
> > @scholzj does this need to be pluggable currently though? Or could it be deferred until later when there's a subsequent use case to help shape design of that pluggable mechanism?
> 
> @LiamClarkeNZ Well, I think that the current design is not something what can be easily adapted to be pluggable. So doing it later might mean yet another redesign. I don't think this proposal necessarily needs to handle all the details and implementation of the pluggability. But I think it is something what would be good to take into account.

Hi @scholzj 

When you say current design, do you mean what's described in this SIP, or the related PR in the quota plugin repo?

scholzj:
I mean this proposal -> with syncing through Kafka topic which seemes like fairly closed loop.

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -0,0 +1,280 @@
...
+#### The data source should:
+- Collect local volume usage
+- Publish usage to an internal compacted topic keyed by broker ID.
+- Collect and publish volume usage metrics (using this [message schema](#Message-schema)) on start up and periodically after that.
+- Connect as an internal service using TLS and client certificates to connect to the Replication listener.
+
+#### The data sink should:
+- Connect as an internal service using TLS and client certificates to connect to the Replication listener.
+- On startup and periodically thereafter read the volume usage stats from the topic
+- Use the Kafka Admin client to discover the list of currently active brokers
```
mimaison, line 63: We'll also need to use `describeLogDirs()` to find the logdir where each replica lives.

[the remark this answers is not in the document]
mimaison: How does the data sink know the size of each volume?

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -0,0 +1,280 @@
...
+1. Consumed space limit: Triggers if the consumed space of a volume breaches the value. Candidate for deprecation.
+2. Minimum Free Bytes: Triggers if the amount of free space on a volume drops below the configured level. Particularly
+   useful for hard limits as an absolute minimum of free space.
+3. Minimum Free Percentage. Triggers if the amount of free space on the volume drops below the configured proportion of
+   the volumes total capacity.
+
+#### The data source should:
+- Collect local volume usage
+- Publish usage to an internal compacted topic keyed by broker ID.
+- Collect and publish volume usage metrics (using this [message schema](#Message-schema)) on start up and periodically after that.
```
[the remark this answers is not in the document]
SamBarker: > Assuming there's nothing else on the volume, we can get the size of each replica via the `describeLogDirs()` Admin API. If we sum all sizes per logdir, we get the disk usage. So do we really need to emit this usage metric?

I'm not sure how valid that assumption is, probably fine for strimzi but not in general. While it would give us the usage per logDir we still wouldn't know how logDir's map to volumes and the capacities of each volume. So I think we still need to emit volume details. At which point I think its easier to just emit a single message with both consumed bytes and capacity in bytes.



> Also Cruise Control has a metrics reporter that write metrics in a topic to be used by other components. Have we considered reusing it?

The initial thought process was to do something which could stand on its own without rather than relying on external services. I.e. cruise control or a metrics service.

035-cluster-wide-volume-usage-quota-managemnet.md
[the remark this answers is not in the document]
SamBarker: > How does the data sink know the size of each volume?

The data source publishes the usage metrics: Bytes consumed and capacity. 



> We'll also need to use `describeLogDirs()` to find the logdir where each replica lives.

As the source is publishing the complete set of metrics for each broker my current thinking is the sink only needs to determine the active set of brokers, so it knows which keys to read from the topic, and doesn't need to understand the relationship between logDirs and volumes.

PaulRMellor (approved):

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -0,0 +1,280 @@
...
+
+[KIP-257](https://cwiki.apache.org/confluence/display/KAFKA/KIP-257+-+Configurable+Quota+Management) describes configurable quota management, where quotas are defined in terms of how much of a delay to apply to a given produce request. This leads to defining the
+possible states of the plugin as:
+`OPEN`, `THROTTLE` and `PAUSE` to accurately reflect the effects of transitioning between states.
+
+This implies there are three roles.
+1. **Data source** - Publish volume usage metrics about the volumes underlying the brokers log dirs.
+2. **Data Sink** - Consume volume usage metrics and make those available for making Quota Policy decisions.
+3. **Quota Policy** - Convert volume usage metrics for the entire cluster into throttling decisions.
+
```
tombentley, line 44: I think a [communication diagram](https://en.wikipedia.org/wiki/Communication_diagram) showing the interaction between components running over two brokers would be good.

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -0,0 +1,280 @@
...
+To make all this work the quota plugin will need to introduce some new interfaces:
+
+1. A Quota Policy - To encapsulate the logic for evaluating if a volume `complies` or `breaches` a policy
+2. A Quota Factor Source - To calculate the factor in the range `0..100`% to apply to the entries in the quota map
+3. A Data Source - To publish the current volume usage for the broker in question.
+
+One can envisage a model where the Quota policy was externalised to a separate service which calculated the quota factor
+and thus the Quota Factor Source would just provide a wrapper around that external service. For the purposes of this
+proposal all three interfaces would be implemented with the current quota plugin.
```
tombentley, line 88: It's not clear to me whether each instance of the plugin has an active instance of the policy, making local decisions, or whether there's a single active instance of the policy for the whole cluster (for example, the consumer group leader)

035-cluster-wide-volume-usage-quota-managemnet.md
```diff
@@ -0,0 +1,280 @@
...
+    void run();
+
+    long getPeriod();
+
+    TimeUnit getPeriodUnit();
+}
+```
+
+##### Message schema
+For simplicity and debug ability the message should be encoded as JSON, but could be converted to a more space efficient
```
tombentley, line 198: Clarify that this is the record _value_, and mention again that the key is the id of the producing broker?

tombentley: TBH it might be easier to just define the json schema here.

035-cluster-wide-volume-usage-quota-managemnet.md
SamBarker: I've added a diagram to the proposal. Its not a diagram format I have any experience with so keen to hear any feedback on it.

035-cluster-wide-volume-usage-quota-managemnet.md
SamBarker: I've updated the text to clarify I envisage each broker operating independently.

035-cluster-wide-volume-usage-quota-managemnet.md
SamBarker: I've added the `brokerId` to the message and included a JSON schema as well, though I have kept the tables describing the format as I find them much easier to read than the schema.

scholzj:
@SamBarker I saw you did some changes -> does that already cover the things discussed last week on the community call? Is it ready for another review round?

SamBarker:
No, I haven't finished following up on the conversation yet.

scholzj:
Ok, no problem ... I just wanted to make sure if I should review it again already or not.

SamBarker:
Now that [KIP-827](https://cwiki.apache.org/confluence/display/KAFKA/KIP-827%3A+Expose+logdirs+total+and+usable+space+via+Kafka+API) has appeared on the horizon  all the details in the proposal around Kafka topics etc become moot and thus I don't think there is value in keeping this proposal open.

I think the high level changes discussed are worth preserving and will revisit them once KIP-827 becomes a practical option

PR #52, opened 2022-04-21
Extending the `KafkaRebalance` resource with rebalance types to help scaling the Apache Kafka cluster

ppatierno:
This proposal is about adding the Cruise Control support for running rebalancing when the Apache Kafka cluster is scaled up or down.
It leverages the usage of a couple of different Cruise Control endpoints to avoid a full rebalance across all the already existing brokers but just moving replicas to the newly added ones (after scaling up) or move replicas out from the removed ones (before scaling down) in order to have less data movement, CPU usage and so on.
It's doesn't provide automation but it's user driven by using an enhanced `KafkaRebalance` resource with new types of rebalancing.

### DIFF
--- a/035-cruise-control-scaling-brokers.md
+++ b/035-cruise-control-scaling-brokers.md
@@ -1,0 +1,98 @@
+# Cruise Control rebalancing on scaling the Apache Kafka cluster
+
+This proposal is about leveraging the Cruise Control integration in the Strimzi cluster operator in order to run an Apache Kafka cluster rebalancing when it's scaled up or down.
+
+## Current situation
+
+Currently, when Cruise Control is deployed alongside the Apache Kafka cluster, no automatic rebalancing happens when the cluster itself is scaled up or down.
+
+It means that, if the cluster is scaled up, no replicas will be moved to the newly added brokers.
+These brokers will only get partition replicas of newly created topics.
+Furthermore, by using the `KafkaRebalance` custom resource to run a rebalancing operation after the scaling up, it will move replicas across all the already existing brokers and not only to the new ones.
+It actually runs a full-rebalance with the drawback of more data movement, CPU usage and so on.
+
+In the same way, if the cluster is scaled down, the replicas hosted on the target brokers to remove are not moved out before the scaling down happens.
+Right now the scaling down is not blocked but doing it is not part of this proposal.
+This could result in some partitions having less ISR (In-Sync Replicas) than before the scaling down, with an impact on availability of the cluster.
+The current Cruise Control integration has no support, via the `KafkaRebalance` custom resource, to allow moving these replicas out of the target brokers before the scaling down happens.
+
+## Motivation
+
+When the Apache Kafka cluster is scaled up, it should be possible to use the Cruise Control integration to run a rebalancing by moving some replicas to the newly added brokers without involving the already existing ones.
+This would provide a faster rebalancing, not involving the entire cluster.
+
+Not stopping the scaling down operation when replicas are hosted on the target brokers to remove is not part of this proposal but it should be possible to move replicas out of the target brokers before the scaling down happens.
+It would allow to remove "empty" brokers which are not hosting any replicas, avoiding less ISR problem.
+
+## Proposal
+
+### Cruise Control add and remove brokers endpoints
+
+Cruise Control provides two specific HTTP REST endpoints to rebalance a cluster when scaling up or down:
+
+* [`/add_broker`](https://github.com/linkedin/cruise-control/wiki/REST-APIs#add-a-list-of-new-brokers-to-kafka-cluster): this endpoint allows to trigger a rebalancing operation by moving replicas from the already existing brokers to the new ones but not moving them among existing brokers. It can be used right after the scaling up, for getting a proposal (as dryrun) and running the actual rebalancing.
+* [`/remove_broker`](https://github.com/linkedin/cruise-control/wiki/REST-APIs#decommission-a-list-of-brokers-from-the-kafka-cluster): this endpoint allows to trigger a rebalancing operation by moving replicas out from the target brokers to remove, in order to leave them "empty" and not hosting any replicas. It can be used just before the scaling down, for getting a proposal (as dryrun) and running the actual rebalancing.
+
+In order to use the above two endpoints in the Strimzi cluster operator, they would be added to the [`CruiseControlApi`](https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApi.java) interface and developing the corresponding implementation.
+
+### KafkaRebalance modes
+
+The current usage of the `KafkaRebalance` custom resource allows to get an optimization proposal and start a full-rebalance with data movement across all the brokers in the cluster.
+At Cruise Control API level, it's reflected by the `KafkaRebalanceAssemblyOperator` using the `/proposals` endpoint for getting an optimization proposal first and then the `/rebalance` endpoint for starting the full-rebalance.
+
+In order to start a rebalancing on a cluster being scaled up or down, the proposal is about adding a `spec.mode` field in the `KafkaRebalance` custom resource with the following possible values:
+
+* `full-rebalance`: it starts a full-rebalance across all the brokers in the cluster as it works today. It is the default behavior if not specified (for backward compatibility).
+* `add-broker`: it starts a rebalancing by moving replicas out from the already existing brokers to the new added ones after a scale up.
+* `remove-broker`: it starts a rebalancing by moving replicas out from the target brokers to remove before a scale down.
+
+When the rebalancing is related to cluster being scaled up or down, the newly added brokers or the ones that are going to be removed have to be specified through a new `spec.brokers` field as an array.
+This needs come from how the Cruise Control API endpoints works. There is a specific query parameter which needs that list.
+
+A `KafkaRebalance` custom resource would look like this.
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaRebalance
+metadata:
+  name: my-rebalance
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  mode: <value> # full-rebalance, add-broker or remove-broker. Assuming full-rebalance by default if the field is not specified.
+  brokers:
+    - 3
+    - 4
+  goals:
+    - RackAwareGoal
+    - ReplicaCapacityGoal
+```
+
+### Flow
+
+The proposal doesn't assume any automation to run the rebalancing when the cluster is scaled up or down.
+Using the `StatefulSet` for Kafka brokers, the corresponding broker IDs are implied in one direction (added from the higher one up, removed from the higher one down) and the cluster operator could get them for the rebalancing.
+But on the other side, Strimzi is moving to use the `StrimziPodSet` and towards a ZooKeeper-less cluster support so that the broker IDs are handled differently and the user has the flexibility to scale down by removing specific brokers or scale up by specifying new broker IDs.
+For this reason the proposal is about having the user to specify the added or removed brokers for the rebalancing.
+If needed, it also provides the flexibility to scale up the cluster but running the rebalancing taking into account only a subset of the newly added brokers and leaving others just for new topics.
+The same could happen by running a rebalance for removing brokers to make them "empty" but then not doing the actual scale down.
+
+In order to scale up the cluster, the user increases the number of replicas through the `spec.kafka.replicas` of the `Kafka` custom resource.
+When the scaling is done, a new `KafkaRebalance` custom resource can be created by setting the `spec.mode: add-broker` and the list of the new broker IDs in the `spec.brokers` field.
+The `KafkaRebalanceAssemblyOperator` will start the interaction with Cruise Control via the `/add_broker` for getting an optimization proposal first and then, after being accepted by the user applying the `strimzi.io/rebalance=approve` annotation on it, to run the actual rebalancing.
+
+Before scaling down the cluster, the user can create a `KafkaRebalance` custom resource with `spec.mode: remove-broker` and the list of the target broker to remove in the `spec.brokers` field.
+The `KafkaRebalanceAssemblyOperator` will start the interaction with Cruise Control via the `/remove_broker` for getting an optimization proposal first and then, after being accepted by the user applying the `strimzi.io/rebalance=approve` annotation on it, to run the actual rebalancing.
+When the rebalancing is done, the user can finally scale down the cluster by decreasing the number of replicas through the `spec.kafka.replicas` of the `Kafka` custom resource.
+
+## Affected/not affected projects
+
+This proposal impacts the Strimzi Cluster Operator only related to the Cruise Control API and the `KafkaRebalanceAssemblyOperator` more specifically.
+
+## Compatibility
+
+The proposed changes are backward compatible because the default mode for the `KafkaRebalance` custom resource is the current full-rebalance when it's not specified.
+
+## Rejected alternatives
+
+No rejected alternatives.
--- a/035-cruise-control-scaling-brokers.md
+++ b/035-cruise-control-scaling-brokers.md
@@ -77,13 +77,28 @@ For this reason the proposal is about having the user to specify the added or re
 If needed, it also provides the flexibility to scale up the cluster but running the rebalancing taking into account only a subset of the newly added brokers and leaving others just for new topics.
 The same could happen by running a rebalance for removing brokers to make them "empty" but then not doing the actual scale down.
 
-In order to scale up the cluster, the user increases the number of replicas through the `spec.kafka.replicas` of the `Kafka` custom resource.
-When the scaling is done, a new `KafkaRebalance` custom resource can be created by setting the `spec.mode: add-broker` and the list of the new broker IDs in the `spec.brokers` field.
-The `KafkaRebalanceAssemblyOperator` will start the interaction with Cruise Control via the `/add_broker` for getting an optimization proposal first and then, after being accepted by the user applying the `strimzi.io/rebalance=approve` annotation on it, to run the actual rebalancing.
+#### Rebalance on scale up
 
-Before scaling down the cluster, the user can create a `KafkaRebalance` custom resource with `spec.mode: remove-broker` and the list of the target broker to remove in the `spec.brokers` field.
-The `KafkaRebalanceAssemblyOperator` will start the interaction with Cruise Control via the `/remove_broker` for getting an optimization proposal first and then, after being accepted by the user applying the `strimzi.io/rebalance=approve` annotation on it, to run the actual rebalancing.
-When the rebalancing is done, the user can finally scale down the cluster by decreasing the number of replicas through the `spec.kafka.replicas` of the `Kafka` custom resource.
+When scaling up the cluster, the user can go through the following procedure:
+
+1. The user increases the number of replicas through the `spec.kafka.replicas` of the `Kafka` custom resource.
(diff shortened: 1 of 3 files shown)


kyguy (commented):
Looking good Paolo! Just one comment for now :)

035-cruise-control-scaling-brokers.md
```diff
@@ -0,0 +1,113 @@
...
+kind: KafkaRebalance
+metadata:
+  name: my-rebalance
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  mode: <value> # full-rebalance, add-broker or remove-broker. Assuming full-rebalance by default if the field is not specified.
+  brokers:
+    - 3
+    - 4
```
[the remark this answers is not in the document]
ppatierno: Good point @kyguy !
I don't think it's so simple. Taking a look at the Cruise Control API, the `destination_broker_ids` parameter is available on the `/proposal` and `/rebalance` endpoints (that are involved during a `full-rebalance`) but even on the `/remove_broker` endpoint (involved during a `remove-broker`) where it's used to specify the destination brkoers where to move replicas out of the brokers going to be removed. So in the specific `remove-broker` case you would need the mandatory `brokers` list (brokers to remove) and the optional "destination brokers". 
This means that using the `brokers` list as `destination_broker_ids` parameter for `full-rebalance` would make things not clear if we want to add the same `destination_broker_ids` parameter for the `remove-broker` where we also need the `brokers` list.
My feeling is leaving the `destination_broker_ids` out for now leaving the door opened for the future with a different field.
Regarding the `brokers` field being specified when `full-rebalance` we can: ignore it as you suggested and print a warning in the operator log, or doing a proper validation and updating the condition status in the `KafkaRebalance` resource to make the user aware that the field has to be removed. Or maybe a mix of them, ignoring and a warning condition in the status (other than the log).

035-cruise-control-scaling-brokers.md
```diff
@@ -0,0 +1,113 @@
...
+  goals:
+    - RackAwareGoal
+    - ReplicaCapacityGoal
+```
+
+### Flow
+
+The proposal doesn't assume any automation to run the rebalancing when the cluster is scaled up or down.
+Using the `StatefulSet` for Kafka brokers, the corresponding broker IDs are implied in one direction (added from the higher one up, removed from the higher one down) and the cluster operator could get them for the rebalancing.
+But on the other side, Strimzi is moving to use the `StrimziPodSet` and towards a ZooKeeper-less cluster support so that the broker IDs are handled differently and the user has the flexibility to scale down by removing specific brokers or scale up by specifying new broker IDs.
```
tombentley, line 75: Can you link to the proposal for `StrimziPodSet`?

035-cruise-control-scaling-brokers.md
```diff
@@ -0,0 +1,113 @@
...
+For this reason the proposal is about having the user to specify the added or removed brokers for the rebalancing.
+If needed, it also provides the flexibility to scale up the cluster but running the rebalancing taking into account only a subset of the newly added brokers and leaving others just for new topics.
+The same could happen by running a rebalance for removing brokers to make them "empty" but then not doing the actual scale down.
+
+#### Rebalance on scale up
+
+When scaling up the cluster, the user can go through the following procedure:
+
+1. The user increases the number of replicas through the `spec.kafka.replicas` of the `Kafka` custom resource.
+2. When the scaling up is done, a new `KafkaRebalance` custom resource can be created by setting the `spec.mode` field as `add-broker` and the list of the new brokers in the `spec.brokers` field.
```
tombentley, line 85: Clarify that it's the user who creates the `KafkaRebalance` (not the CO).

035-cruise-control-scaling-brokers.md
```diff
@@ -0,0 +1,113 @@
...
+When scaling down the cluster, the user can go through the following procedure:
+
+1. The user creates a `KafkaRebalance` custom resource with the `spec.mode` field as `remove-broker` and the list of the target brokers to remove in the `spec.brokers` field.
+2. The `KafkaRebalanceAssemblyOperator` starts the interaction with Cruise Control via the `/remove_broker` endpoint for getting an optimization proposal (by using the dryrun feature).
+3. The user accepts the proposal by applying the `strimzi.io/rebalance=approve` annotation on it.
+4. The `KafkaRebalanceAssemblyOperator` starts the interaction with Cruise Control via the `/remove_broker` endpoint for running the actual rebalancing.
+5. When the rebalancing is done, the user can finally scale down the cluster by decreasing the number of replicas through the `spec.kafka.replicas` of the `Kafka` custom resource.
+
+It could happen that before the scaling down but after the rebalancing a new topic is created and replicas are added on the target brokers to remove.
+It is not part of this proposal, but the Strimzi cluster operator should avoid the scale down in this case.
```
tombentley, line 101: What does 'avoid' mean? Prevent?

tombentley (commented):
Thanks for the proposal!

I think this would be a great step in the right direction. I'd love to see a short section on the bigger picture, for how we might eventually remove the user from the loop entirely. I can see there being a number of parts to that:

1. Support for auto-approval, where the user can specify some criteria to use which the rebalance operator would evaluate so as to skip the approval step if the balance was acceptable, and otherwise use a status condition to denote being stuck.
2. Having two additional `spec.mode`: `add-broker-template` and `remove-broker-mode`. These would be ignored by the rebalance operator. They would instead be referred to from the `Kafka` CR, and the CO would use them for _operator manged_ cluster scaling, where it (rather than the user) would create corresponding `add-broker` or `remove-broker` CRs. The CO's scaling reconciliation would then block on the CC operation. 

I don't think we necessarily have to spec this out entirely in this proposal (but we could, I don't thin it's that much work), but sketching it would be useful.

035-cruise-control-scaling-brokers.md
ppatierno: Yeah I meant prevent, I will change it.

ppatierno:
@tombentley ...

> I'd love to see a short section on the bigger picture, for how we might eventually remove the user from the loop entirely.

My only concern of adding more stuff is that they are not strictly related to the implementation that this proposal will drive. I see a proposal as something that is followed by the development implementing the proposal content itself. Adding more stuff would derail it or making different expectations.

> Support for auto-approval, where the user can specify some criteria to use which the rebalance operator would evaluate so as to skip the approval step if the balance was acceptable, and otherwise use a status condition to denote being stuck.

If you really want a paragraph about this future work, I would be really at an higher level mostly rephrasing what you are already saying here but not adding more details (i.e. how criteria are defined, how the operator evaluates them, ....)

> Having two additional spec.mode: add-broker-template and remove-broker-mode. These would be ignored by the rebalance operator. They would instead be referred to from the Kafka CR, and the CO would use them for operator manged cluster scaling, where it (rather than the user) would create corresponding add-broker or remove-broker CRs. The CO's scaling reconciliation would then block on the CC operation.

Does it make sense mentioning something like this when we don't have much details. For example, based on what the CO is going to create `KafkaRebalance` resource, with which content, ...? Do we really want CO watching `KafkaRebalance` resources as well? Is your idea a replacement for the first one? It's not clear what you have in mind.

035-cruise-control-scaling-brokers.md
```diff
@@ -0,0 +1,113 @@
...
+This could result in some partitions having less ISR (In-Sync Replicas) than before the scaling down, with an impact on availability of the cluster.
+The current Cruise Control integration has no support, via the `KafkaRebalance` custom resource, to allow moving these replicas out of the target brokers before the scaling down happens.
+
+## Motivation
+
+When the Apache Kafka cluster is scaled up, it should be possible to use the Cruise Control integration to run a rebalancing by moving some replicas to the newly added brokers without involving the already existing ones.
+This would provide a faster rebalancing, not involving the entire cluster.
+
+Not stopping the scaling down operation when replicas are hosted on the target brokers to remove is not part of this proposal but it should be possible to move replicas out of the target brokers before the scaling down happens.
+It would allow to remove "empty" brokers which are not hosting any replicas, avoiding less ISR problem.
```
scholzj, line 25: I find this point a bit weird. Not stoping the scaledown when the brokers are not empty can be seen as a _missing feature_ today. But how does it fit with this proposal? Is this something to consider as a feature when the operator can move the replicas out of the broker automatically?

035-cruise-control-scaling-brokers.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Proposal
+
+### Cruise Control add and remove brokers endpoints
+
+Cruise Control provides two specific HTTP REST endpoints to rebalance a cluster when scaling up or down:
+
+* [`/add_broker`](https://github.com/linkedin/cruise-control/wiki/REST-APIs#add-a-list-of-new-brokers-to-kafka-cluster): this endpoint allows to trigger a rebalancing operation by moving replicas from the already existing brokers to the new ones but not moving them among existing brokers. It can be used right after the scaling up, for getting a proposal (as dryrun) and running the actual rebalancing.
+* [`/remove_broker`](https://github.com/linkedin/cruise-control/wiki/REST-APIs#decommission-a-list-of-brokers-from-the-kafka-cluster): this endpoint allows to trigger a rebalancing operation by moving replicas out from the target brokers to remove, in order to leave them "empty" and not hosting any replicas. It can be used just before the scaling down, for getting a proposal (as dryrun) and running the actual rebalancing.
```
scholzj, line 34: Do these endpoints handle multiple brokers added / removed? Or do they handle only one as suggested by the endpoint name?

035-cruise-control-scaling-brokers.md
```diff
@@ -0,0 +1,113 @@
...
+    - 3
+    - 4
+  goals:
+    - RackAwareGoal
+    - ReplicaCapacityGoal
+```
+
+### Flow
+
+The proposal doesn't assume any automation to run the rebalancing when the cluster is scaled up or down.
```
scholzj, line 73: I think this is quite important point which probably should not come up only on line 73. I guess it answers at least one of my previous comments. TBH at least for me, this is a bit misleading since from the title and intro I assumed it takes the full automation into account.

scholzj (commented):
I think this is fine as a _phase 1_ and I do not think I have any major comment to what it proposes. 

What I'm missing a bit is some more _forward looking aspect_. I assume that one day, we would want to have this fully automated and integrated as well (maybe as an optional feature which can be turned on or off, but I think it would be good to have some option in the long term).

I don't think I will be the only one thinking about this. So I wonder what the view of others is. I think it would be good to cover this more clearly. Either by explaining that this is just first phase and that later this will be automated (plus outlining how - for example by the operator creating the KafkaRebalance resources) without too many technical details pointing to future proposal. Or by making it more explicit that this is all what is planned.

035-cruise-control-scaling-brokers.md
ppatierno: > Not stopping the scaling down operation when replicas are hosted on the target brokers to remove is not part of this proposal ....

I was trying to make it clear that this proposal is about moving out replicas from brokers and then scaling them down. There is no operator feature that stops a user to scale down when a broker is hosting replicas. AFAIU it's something that @ShubhamRwt is looking into by using Admin API. If you feel that this sentence just creates confusion, we can remove it.

035-cruise-control-scaling-brokers.md
ppatierno: Yeah the endpoint name is "singular", even the `brokerid` query parameter they handle is singular but in the end it gets a list of brokers.

035-cruise-control-scaling-brokers.md
ppatierno: Are you suggesting to mention it in the introduction as well?

ppatierno:
@scholzj about making it clear that it's "only" phase 1 I would agree (@tombentley mentioned the same).
What I am not sure is about mentioning any kind of idea we have in mind to do that because it would go anyway through a different proposal and even after implementing this one, better ideas could come (so what I can write here would be stale).
I would just raise the point that this is a "phase 1" where everything is manual with user interaction but that at some point it will be somehow automated with the operator doing rebalancing during scale up/down with no user intervention but just some criteria to met that the user has to specified upfront somehow.

035-cruise-control-scaling-brokers.md
scholzj: Either in the introduction or make the title more clear. The title was what really set my expectation. The title suggests it happens on scaling the cluster. I read that as _the operator does it automatically when scaling the cluster_. If it would have been more something like _Add new rebalance types to Kafka Rebalance resource to help scling the cluster up or down_, it would have been more clear that this is actually not about automatic operations but just extending KafkaRebalance.

scholzj:
I think the proposal serves in a way as a roadmap as well - it lists the things which might come (unless they are already implemented). So it is useful to mention the future plans. I do not think it needs too many details. But saying _this is a first phase where we add support for doing this manually and later it will be automated (will be covered in a separate proposal)_ would be good.

ppatierno:
@scholzj I added more information in the introduction. Also changed title and renamed the file as well.

scholzj (approved):
Nice proposal. Thanks.

tombentley (approved):

035-rebalance-types-scaling-brokers.md
```diff
@@ -0,0 +1,117 @@
...
+
+Currently, when Cruise Control is deployed alongside the Apache Kafka cluster, no automatic rebalancing happens when the cluster itself is scaled up or down.
+
+It means that, if the cluster is scaled up, no replicas will be moved to the newly added brokers.
+These brokers will only get partition replicas of newly created topics.
+Furthermore, by using the `KafkaRebalance` custom resource to run a rebalancing operation after the scaling up, it will move replicas across all the already existing brokers and not only to the new ones.
+It actually runs a full-rebalance with the drawback of more data movement, CPU usage and so on.
+
+In the same way, if the cluster is scaled down, the replicas hosted on the target brokers to remove are not moved out before the scaling down happens.
+Right now the scaling down is not blocked but doing it is not part of this proposal.
```
tomncooper, line 19: Have we got an issue for this? I think @ShubhamRwt said he was planning to add this check?

ppatierno: Yeah @ShubhamRwt is working on it but we didn't have an actual issue :-(
I created one https://github.com/strimzi/strimzi-kafka-operator/issues/6736 and now adding the link to the proposal.

Frawless (approved):

035-rebalance-types-scaling-brokers.md
```diff
@@ -0,0 +1,118 @@
...
+For this reason the proposal is about having the user to specify the added or removed brokers for the rebalancing.
+If needed, it also provides the flexibility to scale up the cluster but running the rebalancing taking into account only a subset of the newly added brokers and leaving others just for new topics.
+The same could happen by running a rebalance for removing brokers to make them "empty" but then not doing the actual scale down.
+
+#### Rebalance on scale up
+
+When scaling up the cluster, the user can go through the following procedure:
+
+1. The user increases the number of replicas through the `spec.kafka.replicas` of the `Kafka` custom resource.
+2. When the scaling up is done, the user creates a `KafkaRebalance` custom resource with the `spec.mode` field as `add-broker` and the list of the new brokers in the `spec.brokers` field.
```
tomncooper, line 90: I don't think the user should have to do this. If CC is configured, then the user clearly wants its functionality. I think that Strimzi should create a KafkaRebalance CR with `spec.mode.add_broker` with all the fields filled in. Maybe even run the dry-run proposal (with the default CC settings) and have the results in the status, so that all the user has to do it approve it? The user could then, if they want, customize this auto-created KRCR with the other tuning options / goals and refresh it to get a new proposal.

This could be the standard behaviour and then we could have some flag in the Kafka cc config to turn that auto-KRCR creation functionality on/off.

I understand why we are still, at this stage, sticking with the user approval flow, but we should make things as easy as possible for users.

035-rebalance-types-scaling-brokers.md
```diff
@@ -0,0 +1,118 @@
...
+3. The `KafkaRebalanceAssemblyOperator` starts the interaction with Cruise Control via the `/add_broker` endpoint for getting an optimization proposal (by using the dryrun feature).
+4. The user accepts the proposal by applying the `strimzi.io/rebalance=approve` annotation on it.
+5. The `KafkaRebalanceAssemblyOperator` starts the interaction with Cruise Control via the `/add_broker` endpoint for running the actual rebalancing.
+
+#### Rebalance on scale down
+
+When scaling down the cluster, the user can go through the following procedure:
+
+1. The user creates a `KafkaRebalance` custom resource with the `spec.mode` field as `remove-broker` and the list of the target brokers to remove in the `spec.brokers` field.
```
tomncooper, line 99: Again, as with the scale up, I think Strimzi should create (if CC is configured) a KRCR with the default settings and the brokers field filled in. With StatefulSets we know which brokers are to be removed. I am not sure how the PodSet scale down works but presumable the user has to say which brokers to remove (or strimzi decides) so we should be able to create a KRCR for them.

tomncooper: We can block the scale down until the remove broker KRCR is in `Ready` state.

035-rebalance-types-scaling-brokers.md
```diff
@@ -0,0 +1,118 @@
...
+2. The `KafkaRebalanceAssemblyOperator` starts the interaction with Cruise Control via the `/remove_broker` endpoint for getting an optimization proposal (by using the dryrun feature).
+3. The user accepts the proposal by applying the `strimzi.io/rebalance=approve` annotation on it.
+4. The `KafkaRebalanceAssemblyOperator` starts the interaction with Cruise Control via the `/remove_broker` endpoint for running the actual rebalancing.
+5. When the rebalancing is done, the user can finally scale down the cluster by decreasing the number of replicas through the `spec.kafka.replicas` of the `Kafka` custom resource.
+
+It could happen that before the scaling down but after the rebalancing a new topic is created and replicas are added on the target brokers to remove.
```
tomncooper, line 105: We are gonna need docs on that. Admins should ensure new topics/partitions are not created during a scaling operations. We should also block other Strimzi CC operations during scale down.

ppatierno:
@tomncooper I read you comments but I think we are changing the scope of the proposal or anyway we are on a different page. As I mentioned this is about the phase 1 before having automation in place.
Having the manual approach first can help us to validate the approach itself and it also leaves flexibility of usage.
Also, because we could want the user to disable the automation (when it will be in place), it means that the manual approach should be in place first. I see it as the normal flow.

The automation needs more thoughts, not just about the operator creating a `KafkaRebalance` but even allowing the user to specify some criteria of applying the proposal when no manual approval is needed.
Also, because we are moving to `StrimziPodSet` maybe it would make more sense doing the automation with them in place instead of adding more logic in the operator to check automatically which pods indexes have to be added/removed.

I would start this way then adding the automation with a flag to enable/disable it.

tomncooper:
Ok, I see building it up in stages makes sense. Maybe something like:

1) Add the add/remove broker mode and endpoint functionality.
2) Add the auto-create KRCR on scale operations functionality, still with final user approval.
3) Add the auto-approve within user defined limits functionality. For example "approve if <10% total replicas are moved".

I wonder if we need meta-proposal for this (like KIP-500)?

tomncooper (approved):

kyguy (approved):
Looks great Paolo!

ppatierno:
@tomncooper I didn't think about intermediate step 2 but it could make really sense. I think approaching in steps is the best way for this. Let's start with step 1 for now! :-)

ppatierno:
Leaving the PR still opened for additional comments. I will merge it on Friday Aprile 28th EOD CEST if no further comments to address.

ppatierno:
No further comments so I am going to merge this one. Thanks everyone for reviewing!

PR #54, opened 2022-05-10
Prevent broker scale down

ShubhamRwt:


prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,60 @@
...
+So this check will target the broker which are going to be removed ie. the pods with the highest number.
+
+Methods to be created in the process:
+
+- `getTopicNameAndDescription` (Boolean Type) - This method will be placed inside the `scaleDown()` method in `KafkaReconciler` class and will provide us with the Collection of `TopicDescription`.
+- `describeTopics()` and `topicNames()`  - These two methods will provide us with the description and names of all the topics ie first we will get the names of the topics using the `topicNames` method and using these names we will extract the description of the topics.
+- `brokerHasAnyTopics()` - This method will take as input a podId and a `Collection<TopicDescription>`.
+  It will go through all the broker and see if any partitions are assigned to it.
+
+### Flow :
```
scholzj, line 35: No space before `:`.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,60 @@
...
+Then we will call the `checkPartitions()` method which will give us the `Collection<TopicDescription>`. 
+
+The `checkPartition()` method will call the `topicNames()` method to get the set of topic names and then these topic names will be used by the `describeTopics()` methods to get the `Collection<TopicDescription>`.
+Once the `Collection<TopicDescription>` is computed we will move back to the `scaledown()` method and there we will run a loop from `(previous replicas - 1)` to `current replicas` and inside the loop we will call the `brokersHasAnyTopics()` method which will take the `Collection<TopicDescription>` and `podId(index of the loop)` to check if the broker contains any partition and provide us with a boolean result.
+
+
+### What to do if a broker contains partitions?
+If a partition is detected over a broker then we should trigger some log warnings and let the partitions remain the same as they were before.
+For example, if we try to scale down the broker from 5 to 3, but the check detected that the broker to be scaled down contains a partition then we should keep the replicas as 5 only display the appropriate warnings.
```
scholzj, line 48: This needs more details on how do you achieve keeping 5 replicas. It sounds easier than it might be.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,60 @@
...
+
+### Flow :
+
+When the replicas are changed inside the Kafka resource, reconciliation will happen.
+During this reconciliation, we move to the `scaleDown()` method present in the `KafkaReconciler` class.
+The `scaleDown()` method contains the complete mechanism on how the broker is scaled down. So there, we will first create our `adminClient` and provide it with the Secrets.
+Then we will call the `checkPartitions()` method which will give us the `Collection<TopicDescription>`. 
+
+The `checkPartition()` method will call the `topicNames()` method to get the set of topic names and then these topic names will be used by the `describeTopics()` methods to get the `Collection<TopicDescription>`.
+Once the `Collection<TopicDescription>` is computed we will move back to the `scaledown()` method and there we will run a loop from `(previous replicas - 1)` to `current replicas` and inside the loop we will call the `brokersHasAnyTopics()` method which will take the `Collection<TopicDescription>` and `podId(index of the loop)` to check if the broker contains any partition and provide us with a boolean result.
```
scholzj, line 43: This sounds like a hard to handle situation where you do the scale down _partially_. It might be easier to do none or all.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,60 @@
...
+  It will go through all the broker and see if any partitions are assigned to it.
+
+### Flow :
+
+When the replicas are changed inside the Kafka resource, reconciliation will happen.
+During this reconciliation, we move to the `scaleDown()` method present in the `KafkaReconciler` class.
+The `scaleDown()` method contains the complete mechanism on how the broker is scaled down. So there, we will first create our `adminClient` and provide it with the Secrets.
+Then we will call the `checkPartitions()` method which will give us the `Collection<TopicDescription>`. 
+
+The `checkPartition()` method will call the `topicNames()` method to get the set of topic names and then these topic names will be used by the `describeTopics()` methods to get the `Collection<TopicDescription>`.
```
scholzj, line 42: `checkPartitions()`?

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,60 @@
...
+- `describeTopics()` and `topicNames()`  - These two methods will provide us with the description and names of all the topics ie first we will get the names of the topics using the `topicNames` method and using these names we will extract the description of the topics.
+- `brokerHasAnyTopics()` - This method will take as input a podId and a `Collection<TopicDescription>`.
+  It will go through all the broker and see if any partitions are assigned to it.
+
+### Flow :
+
+When the replicas are changed inside the Kafka resource, reconciliation will happen.
+During this reconciliation, we move to the `scaleDown()` method present in the `KafkaReconciler` class.
+The `scaleDown()` method contains the complete mechanism on how the broker is scaled down. So there, we will first create our `adminClient` and provide it with the Secrets.
+Then we will call the `checkPartitions()` method which will give us the `Collection<TopicDescription>`.
```
scholzj, line 40: You should elaborate on how the error handling will look like here when it fails to connect to the Kafka cluster.

scholzj (commented):
Some more comments:
* Heading levels look a bit weird. The level 3 heading `Flow :` being twice there in the same section is weird in particular.

prevent-broker-scale-down-if-contains-partition .md
ShubhamRwt: @scholzj Hi, do you mean what if any `RuntimeException` pops up and then how will I handle the process to not stop the reconciliation process?

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,57 @@
...
+## Motivation
+
+We should definitely introduce some sort of logic which can detect if the broker which is going to be removed still contains the partitions or not.
+If the partitions are still available on the broker which is going to be scaled, then we should get some warning in the logs which will prompt users to do the reassignment and hopefully reduce the chance of data loss.
+
+## Proposal
+
+This proposal suggest how we can add the check to detect if the broker still contains any partitions and what to do if the broker scale down is not possible.
+
+Checking if the broker contains partition
```
tomncooper, line 21: Is this a title?

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,57 @@
...
+The best way to detect if the broker contains any partitions is to query the topics.
```
tomncooper, line 22: "is to query the cluster metadata"

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,57 @@
...
+To do this, we will have to make use of the `AdminClient` class.
+We should probably put the logic for this check inside the `scaleDown` method present in the `KafkaReconciler` class.
+The `scaleDown` method gets triggered when we scale down the no. of replicas in the `Kafka CR`.
+So this check will target the broker which are going to be removed i.e. the pods with the highest number.
+
+Methods to be created in the process:
+
+- `getTopicNamesAndDescriptions` (Boolean Type) - This method will be placed inside the `scaleDown()` method in `KafkaReconciler` class and will provide us with the Collection of `TopicDescription`.
+- `describeTopics()` and `topicNames()`  - These two methods will provide us with the description and names of all the topics ie first we will get the names of the topics using the `topicNames` method and using these names we will extract the description of the topics.
+- `brokerHasAnyTopics()` - This method will take as input a podId and a `Collection<TopicDescription>`.
```
tomncooper, line 32: Would it not make more sense to do one loop over the TopicDescription collection and produce a simple map from node ID to a boolean. That way if (in future) you want to scale down more than one broker at a time you won't have to re-loop over the collection you can just use the `hasAssignedTopicPartitions` map?

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,57 @@
...
+
+### Flow :
+
+When the replicas are changed inside the Kafka resource, reconciliation will happen.
+During this reconciliation, we move to the `scaleDown()` method present in the `KafkaReconciler` class.
+The `scaleDown()` method contains the complete mechanism on how the broker is scaled down. So there, we will first create our `adminClient` and provide it with the Secrets.
+Then we will call the `getTopicNamesAndDescriptions()` method which will give us the `Collection<TopicDescription>`. 
+
+The `getTopicNamesAndDescriptions()` method will call the `topicNames()` method to get the set of topic names and then these topic names will be used by the `describeTopics()` methods to get the `Collection<TopicDescription>`.
+Once the `Collection<TopicDescription>` is computed we will move back to the `scaledown()` method and there we will run a loop from `(previous replicas - 1)` to `current replicas` and inside the loop we will call the `brokersHasAnyTopics()` method which will take the `Collection<TopicDescription>` and `podId(index of the loop)` to check if the broker contains any partition and provide us with a boolean result.
```
tomncooper, line 43: See comment above, I think its better to just loop over the collection once and produce a map from nodeid to boolean.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,57 @@
...
+Then we will call the `getTopicNamesAndDescriptions()` method which will give us the `Collection<TopicDescription>`. 
+
+The `getTopicNamesAndDescriptions()` method will call the `topicNames()` method to get the set of topic names and then these topic names will be used by the `describeTopics()` methods to get the `Collection<TopicDescription>`.
+Once the `Collection<TopicDescription>` is computed we will move back to the `scaledown()` method and there we will run a loop from `(previous replicas - 1)` to `current replicas` and inside the loop we will call the `brokersHasAnyTopics()` method which will take the `Collection<TopicDescription>` and `podId(index of the loop)` to check if the broker contains any partition and provide us with a boolean result. 
+
+### What to do if a broker contains partitions?
+If a partition is detected over a broker then we should trigger some log warnings and let the partitions remain the same as they were before.
+For example, if we try to scale down the broker from 5 to 3, but the check detected that the broker to be scaled down contains a partition then we should keep the replicas as 5 only display the appropriate warnings.
```
tomncooper, line 47: "the the brokers to be scaled down still contain partitions"

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,57 @@
...
+
+To handle the case, if broker contain some partitions, we can make use of an `if` statement. If we see the current code in the `scaleDown()` method we can see that we are checking whether we are using `StatefulSets` or `StrimziPodSets` and then doing the scale down accordingly, so we can introduce an `if` just before this logic. This `if` condition will check if the boolean result was `true` or `false`. If the result is `false` then we will return a succeeded future i.e. no need to scale down but if the result is `true` then we can run the logic of scale down as per `StatefulSets` or `StrimziPodSets`.
```
tomncooper, line 49: I don't understand this paragraph. If what result is false? What is this if statement checking? Weather we are using STSs or SPSs? Or if the broker has partitions?

tomncooper (changes requested):
A good start, just a few clarifications and efficiency changes needed.

prevent-broker-scale-down-if-contains-partition .md
ShubhamRwt: The `if` condition will check the boolean result stored in a variable w.r.t to whether the broker had partitions or not. If the result comes out to be `True` we will just return a succeeded future and skip the other logic `STS` and `SPS` since we don't need to scale down at all

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,63 @@
+
+# Prevent scale down of the brokers if they contain partitions
```
ppatierno, line 2: I would talk in general about replicas not partition. They could be the same if the topic has 1 partition with 1 replica but not when it has more replicas. We should not scale down in any case, if the broker is hosting a replica leader or a follower (to avoid under-replicated partition or less than min ISR).

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,63 @@
...
+
+This proposal is about adding the mechanism for preventing the scale down of the brokers in the cluster if a partition is assigned to it.
+In the future, when we introduce the automatic scale up/down of the Kafka Cluster, this mechanism would help to prevent any data loss when we are scaling down the brokers when they have some partitions assigned to them.
```
ppatierno, line 5: Which "automatic scale up/down" are you referring to?

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,63 @@
...
+
+## Current situation
+
+Currently, when the number of Kafka replicas is decreased in the Kafka CR the StatefulSet will shut down the highest numbered brokers and there is no check to see if these brokers contain any topic-partition replicas.
```
ppatierno, line 9: you are doing better here mentioning "topic-partition replicas" :-)

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,63 @@
...
+The [documentation](https://strimzi.io/docs/operators/latest/configuring.html#scaling-clusters-str) recommends that all topic-partition replicas are reassigned before doing this scaling action, which means that if someone tries to scale down without re-assigning the brokers, it can cause data loss.
+
+## Motivation
+
+We should definitely introduce some sort of logic which can detect if the broker which is going to be removed still contains the partitions or not.
+If the partitions are still available on the broker which is going to be scaled, then we should get some warning in the logs which will prompt users to do the reassignment and hopefully reduce the chance of data loss.
```
ppatierno, line 15: People not having Cruise Control enabled will have only manual reassignment as the way to go.
I would mention that for people having Cruise Control, they could run a rebalancing in `remove-brokers` mode.
Refer to the corresponding proposal, and maybe documentation if it will be merged before this proposal got accepted.

ppatierno (commented):
I left some comments.
I think that the proposal should be more higher level not mentioning programming related stuff more (`if` statements, `false` or `true`, ...) ... those are all part of the implementation.
It was like reading a block of code but as a description :-)
I would try to avoid this way.

prevent-broker-scale-down-if-contains-partition .md
ppatierno: I think this comment is related to my more general comment where I am saying that the proposal sounds to describe too many programming constructs than describing the overall solution. I found this paragraph hard to understand too.

tomncooper: Yeah the proposal should stick to high level processes and discriptions and not talk about implementation.

prevent-broker-scale-down-if-contains-partition .md
tombentley: > we should get some warning in the logs

You cannot assume that anyone is reading the logs, so a warning achieves very little. I think we should fail the reconciliation in this case.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,63 @@
...
+
+## Proposal
+
+This proposal suggest how we can add the check to detect if the broker still contains any partitions and what to do if the broker scale down is not possible.
+
+### Checking if the broker contains partition
+The best way to detect if the broker contains any partitions is to query the topic metadata.
+To do this, we will have to make use of the `AdminClient` class.
+We should probably put the logic for this check inside the `scaleDown` method present in the `KafkaReconciler` class.
```
tombentley, line 24: "Probably"? In a proposal you should be clear about exactly what you're proposing. 
```suggestion
We will put the logic for this check inside the `scaleDown` method present in the `KafkaReconciler` class.
```

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,63 @@
...
+The `scaleDown` method gets triggered when we scale down the no. of replicas in the `Kafka CR`.
+So this check will target the broker which are going to be removed i.e. the pods with the highest number.
+
+Methods to be created in the process:
+
+- `getTopicNamesAndDescriptions` (Boolean Type) - This method will be placed inside the `scaleDown()` method in `KafkaReconciler` class and will provide us with the Collection of `TopicDescription`.
+- `describeTopics()` and `topicNames()`  - These two methods will provide us with the description and names of all the topics ie first we will get the names of the topics using the `topicNames` method and using these names we will extract the description of the topics.
+- `brokerContainsPartitions` - This method will take as input a `Collection<TopicDescription>` and then generate a map of type `HashMap<Integer podId, Boolean containsPartitions>`.
```
tombentley, line 32: Personally I don't think this level of detail is needed in a proposal. It's an implementation detail.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,63 @@
...
+
+### Flow :
+
+When the replicas are changed inside the Kafka resource, reconciliation will happen.
```
tombentley, line 36:
```suggestion
When the `spec.kafka.replicas` are changed inside the Kafka resource, reconciliation will happen.
```

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,63 @@
...
+The `scaleDown()` method contains the complete mechanism on how the broker is scaled down. So there, we will first create our `adminClient` and provide it with the Secrets.
+Then we will call the `getTopicNamesAndDescriptions()` method which will give us the `Collection<TopicDescription>`. 
+
+The `getTopicNamesAndDescriptions()` method will call the `topicNames()` method to get the set of topic names and then these topic names will be used by the `describeTopics()` methods to get the `Collection<TopicDescription>`. 
+Once the `Collection<TopicDescription>` is computed we will move back to the `scaledown()` method and there we can call `brokerContainsPartition()` method which will take the `Collection<TopicDescription>` and then create a `hasAssignedTopicPartitions` map which contains key as `podId` and value as `boolean` type.
+Now we can use this map to check if the broker contains any partitions or not.
+
+### What to do if a broker contains partitions?
+If a partition is detected over a broker then we should trigger some log warnings and let the partitions remain the same as they were before.
```
tombentley, line 46: Again, fail the reconciliation, since logging won't actually prevent data loss.

prevent-broker-scale-down-if-contains-partition .md
ShubhamRwt: @tombentley Hi, I thought we don't have to fail a reconciliation if somehow there were replicas present on broker. As per the discussion with Jakub, I thought we had to let the reconciliation take place without any scale down of brokers but the reconciliation should still happen

ppatierno: Failing the reconciliation would mean anyway updating the Kafka CR with a warning condition about the error, so the user knows why it is not scaling.

prevent-broker-scale-down-if-contains-partition .md
ppatierno: But updating the Kafka CR with a warning condition.

prevent-broker-scale-down-if-contains-partition .md
scholzj: You cannot fail the reconciliation. The cluster might be in a bad state, but because the scaledown is fairly early in the reconciliation, it would cause the operator to never fix it and in turn the scale-down will never succeed. So it needs to pass:
* When there is any problem with connecting to the broker(s)
* When the partitions don't allow scale-down (because other brokers might have issues and not allow the partitions to move)

prevent-broker-scale-down-if-contains-partition .md
ShubhamRwt: Oh my bad, Let me address this

ShubhamRwt: I was actually trying to point towards the next steps of the Automatic add/ remove of the brokers from cluster. Like cureently we have introduced modes in the rebalance but in future we were going to do all the process without user interaction

ppatierno: so more than "automatic scale up/down of the Kafka Cluster" is "automatic rebalancing when scaling up/down the Kafka cluster". I would also link to the [proposal](https://github.com/strimzi/proposals/blob/main/035-rebalance-types-scaling-brokers.md) where we mention this future plan, so that people know what is in place.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,47 @@
...
+
+### Checking if the broker contains partition
+
+#### Flow :
+
+- The best way to detect if the broker contains any partition replicas is to query the topic metadata.
+- We can create an Admin client instance for this purpose which can be used to get connect with the cluster and get us the topic details(topic name and topic description)
+- We can later on query the topic's description to get the information regarding the `topic partition replicas`.
+- Then we can use this information to check if the broker contains any partition replicas or not.
+- We should make sure that there is no partial scale down by checking the replicas on the brokers(to be scaled down) first and only do the whole scale down after we make sure that the brokers that are going to be removed aren't containing any partition replicas.
```
scholzj, line 29: Again, don't describe what we can do or should do. Just say what we will do.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,47 @@
...
+
+#### What to do if a broker contains partitions?
+
+- If a partition is detected over a broker then we should update the Kafka CR with the desired warnings
```
scholzj, line 33: Make it clear that the warnings will go to the status.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,47 @@
...
+- In case, if the admin client is not able to connect to the cluster, or we are not able to somehow get the topic details then in that case we should not do any scale down and wait for the next reconciliation to happen, but we will update the Kafka CR with the corresponding error message. 
+  Once the user decides to either change the `spec.replicas` back to the previous count or reassign the partition replicas to other brokers, we can move the process accordingly.
+
+### Error Handling
+
+If we somehow fail to connect to the Kafka Cluster or there are some other error which happens during the reconciliation, in that case also we will set the `scaleDown` variable to `false`, and provide a warning log on reason why the `scaleDown` was not done.
```
scholzj, line 39: I think this deserves more details. When will the check be done? Which steps are affected by not scaling down? You refer to some `scaleDown` variable, but it is not clear what it is, where and how is it used.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,49 @@
...
+#### Flow
+
+- A check will be placed just before the scale down steps happens. 
+- If a partition is detected over a broker then the Kafka CR will be updated with the desired warnings.
+  The warnings will be displayed in the status.
+- When the scale down is happening, if the admin client is not able to connect to the cluster, or not able to get the topic details, scale down will not be proceeded, and we will wait for the next reconciliation to happen.
+Once the user decides to either change the `spec.replicas` back to the previous count or reassign the partition replicas to other brokers, we can move the process accordingly.
+  
+
+
```
scholzj, line 42: No need for 3 empty lines I guess.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,49 @@
...
+# Prevent scale down of the brokers if they contain replicas
+
+This proposal is about adding the mechanism for preventing the scale down of the brokers in the cluster if one or more replicas are hosted on it.
+In the future, when we introduce the automatic re-balancing when scaling up/down the Kafka Cluster ([proposal](https://github.com/strimzi/proposals/blob/main/035-rebalance-types-scaling-brokers.md)), this mechanism would help to prevent any data loss when we are scaling down the brokers when they have some partitions assigned to them.
+
+## Current situation
+
+Currently, when the number of Kafka replicas is decreased in the Kafka CR the StatefulSet will shut down the highest numbered brokers and there is no check to see if these brokers contain any topic-partition replicas.
```
scholzj, line 9: You can probably write this more generic:

```suggestion
Currently, when removing brokers from the Kafka cluster, there is no check to see if these brokers contain any partition replicas.
```

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,49 @@
...
+The [documentation](https://strimzi.io/docs/operators/latest/configuring.html#scaling-clusters-str) recommends that all topic-partition replicas are reassigned before doing this scaling action, which means that if someone tries to scale down without re-assigning the brokers, it can cause data loss.
```
scholzj, line 10:
```suggestion
The [documentation](https://strimzi.io/docs/operators/latest/configuring.html#scaling-clusters-str) recommends that all topic-partition replicas are reassigned before doing this scaling action, which means that if someone tries to scale down without re-assigning the brokers, it can cause availability issues or data loss.
```

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,49 @@
...
+
+## Motivation
+
+We should introduce the logic which can detect if the broker which is going to be removed still contains the partition replicas or not.
+If the partition replicas are still assigned to the broker which is going to be scaled up/down, then we should get some warning in the status of the Kafka resource which will prompt users to do the reassignment and hopefully reduce the chance of data loss.
```
scholzj, line 15:
```suggestion
If any partition replicas are still assigned to the broker which is going to be removed from the cluster, then we should get some warning in the status of the Kafka resource which will prompt users to do the reassignment and prevent the broker from being removed until the partition replicas are reassigned.
```

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,49 @@
...
+
+## Proposal
+
+This proposal suggest how we can add the check to detect if the broker still contains any partition replicas and what to do if the broker scale down is not possible.
+
+### Checking if the broker contains partition
+
+#### Flow :
+
+- The best way to detect if the broker contains any partition replicas is to query the topic metadata.
```
scholzj, line 25: Again, write it as a fact:

```suggestion
- The topic metadata will be queried to detect if the broker contains any partition replicas.
```

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,49 @@
...
+- We can create an Admin client instance for this purpose which can be used to get connect with the cluster and get us the topic details(topic name and topic description)
```
scholzj, line 26:
```suggestion
- An Admin client instance will be used to get connect with the cluster and get us the topic details(topic name and topic description)
```

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,49 @@
...
+- We can later on query the topic's description to get the information regarding the `topic partition replicas`.
```
scholzj, line 27: How does this differ from the previous point?

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,49 @@
...
+- Then we can use this information to check if the broker contains any partition replicas or not.
+- The scale down is done after we make sure that the brokers that are going to be removed aren't containing any partition replicas. By doing this we avoid any partial scale down.
+
+#### What to do if a broker contains partitions?
+
+#### Flow
+
+- A check will be placed just before the scale down steps happens.
```
scholzj, line 35: It probably needs to be placed earlier to make sure that everything is proerly configured if the scaledown needs to be aborted.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,49 @@
...
+- We can later on query the topic's description to get the information regarding the `topic partition replicas`.
+- Then we can use this information to check if the broker contains any partition replicas or not.
+- The scale down is done after we make sure that the brokers that are going to be removed aren't containing any partition replicas. By doing this we avoid any partial scale down.
+
+#### What to do if a broker contains partitions?
+
+#### Flow
+
+- A check will be placed just before the scale down steps happens. 
+- If a partition is detected over a broker then the Kafka CR will be updated with the desired warnings.
```
scholzj, line 36: This is a bit confusing ... you should make it clear in this sentence if it is just the sattus of the Kafka CR which will be updated.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,49 @@
...
+  The warnings will be displayed in the status.
+- When the scale down is happening, if the admin client is not able to connect to the cluster, or not able to get the topic details, scale down will not be proceeded, and we will wait for the next reconciliation to happen.
```
scholzj, line 38: This should have more details about how it will be done. You also need to explain what will you do when it connects but the partitions are still assigned to the removed brokers.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,49 @@
+
+# Prevent scale down of the brokers if they contain replicas
+
+This proposal is about adding the mechanism for preventing the scale down of the brokers in the cluster if one or more replicas are hosted on it.
+In the future, when we introduce the automatic re-balancing when scaling up/down the Kafka Cluster ([proposal](https://github.com/strimzi/proposals/blob/main/035-rebalance-types-scaling-brokers.md)), this mechanism would help to prevent any data loss when we are scaling down the brokers when they have some partitions assigned to them.
```
tomncooper, line 5: It think this is a good feature to have weather we have auto-rebalancing or not. Warning a user of dataloss would be a good for robustness.

tomncooper (changes requested):
As @scholzj has highlighted, I think there needs to be more detail on how some of the bullet points in the "Flow" sections will be achieved.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,45 @@
+
+# Prevent scale down of the brokers if they contain replicas
```
ppatierno, line 2: as already done for the auto-rebalancing proposal, let's try to use `partition replicas` instead of just `replicas` everywhere.

prevent-broker-scale-down-if-contains-partition .md
```diff
@@ -0,0 +1,45 @@
...
+
+This proposal is about adding the mechanism for preventing the scale down of the brokers in the cluster if one or more replicas are hosted on it.
+In the future also, when we introduce the automatic re-balancing when scaling up/down the Kafka Cluster ([proposal](https://github.com/strimzi/proposals/blob/main/035-rebalance-types-scaling-brokers.md)), this mechanism would helpful to prevent any data loss when we are scaling down the brokers when they have some partitions assigned to them.
```
ppatierno, line 5: add the reference to the PR #57 and not the 035 you have here.

prevent-broker-scale-down-if-contains-partition .md
ShubhamRwt: Sure on it

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,45 @@
...
+- The topic metadata will be queried to detect if the broker contains any partition replicas.
+- An Admin client instance will be used to get connect with the cluster and get us the topic details(topic name and topic description)
+- Then we can use this information to check if the broker contains any partition replicas or not.
+- The scale down is done after we make sure that the brokers that are going to be removed aren't containing any partition replicas. By doing this we avoid any partial scale down.
+
+#### What to do if a broker contains partitions?
+
+#### Flow:
+
+- A check will be placed before the reconciliation for brokers take place.
```
scholzj, line 34: What exactly does this mean?

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,45 @@
...
+  This check will make sure that the brokers that are going to be removed doesn't contain any partition replicas on them.   
+- During the check, if the admin client is not able to connect to the cluster (not able to get the topic details) or if the partition replicas are assigned to the brokers that are going to be removed, we will update the status of the Kafka CR with the respective warning and fail the Kafka reconciliation.
```
scholzj, line 36: This is greatm but what do you do with the ongoing reconciliation?

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,45 @@
...
+- Later on we can let the user check the status and decide to either change the `spec.replicas` back to the previous count or reassign the partition replicas to other brokers, we can move the process accordingly.
+  
+## Affected/not affected projects
+
+This change will affect the `KafkaAssembly operator` class, and will be place a check before we run the reconciliations for the brokers.
```
scholzj, line 41: The name is `KafkaAssemblyOperator` class. But the logic probably now fits into `KafkaReconciler` instead?

```suggestion
This change will affect the `KafkaAssemblyOperator` class, and will be place a check before we run the reconciliations for the brokers. 
```

scholzj (approved):

kyguy (approved):
Looks solid @ShubhamRwt

fvaleri:
@ShubhamRwt reassigning partitions can be costly in terms of time and data you have to more around. What if I know that a cluster scale down would not cause any service disruption or data loss. Currently, I can skip the reassignment process. Is this logic smart enough to detect this state and let me scale down the cluster?

prevent-broker-scale-down-if-contains-partition-replicas.md
ShubhamRwt: By check , I meant we will be introducing the logic i.e. a condition to check whether the Broker contains any partition replicas or not before the scale down happens

scholzj: I think the check is fairly clear ... but the palcement not so much.

scholzj: I think it should definitely be in the KafkaReconciler. But I don't know where exactly. That is why I think this needs to be part of the proposal because it is not trivial and not completely clear.

ShubhamRwt:
Hi @fvaleri . This mechanism is only to detect whether our broker contains partition replicas or not, If during the scale down process there seems to be some partition replicas present on the broker then it will prompt the user about that in the status of Kafka CR. So the user will decide what is the next step they want to do as per the warning they get in the status of the Kafka CR. The main motive of this logic is just  to make sure that no scale down is happening if partition replicas are present on the broker. I

scholzj:
@ShubhamRwt Well, your proposal says that it will basically revert the scale-down as long as there are partitions on the brokers ot be removed. So I do not think the user can do what Federico suggests.

@fvaleri That said, I'm not sure I see the use-cases where this would make sense to me. To you have any usecases where this would add some value? Even if the scale-down doesn't necessarily break the topics right away, it will certainly reduce the availability, it can block further rolling updates etc. So I'm not entirely convinced there should be a way to override this.

fvaleri:
@scholzj I see your point, so I'm not -1 on this. I just feel that without having a way to override this, we are losing some flexibility that may be handy, especially in dev/test environments, where a K8s admin may need some resources back in a short period of time.

ppatierno:
>  your proposal says that it will basically revert the scale-down as long as there are partitions on the brokers ot be removed. So I do not think the user can do what Federico suggests.

@scholzj where you read it's reverting the scale-down? And what do you mean by that? AFAIU, the scale down is just blocked, so it doesn't start at all if replicas are placed on the brokers that should be shut down.

Regarding @fvaleri point ... I am also in doubt about the use case but does it really hurt to provide a way to "force" the scale down? It's risky for the user of course. It's something that should be disabled by default and well documented.

scholzj:
> @scholzj where you read it's reverting the scale-down? And what do you mean by that? AFAIU, the scale down is just blocked, so it doesn't start at all if replicas are placed on the brokers that should be shut down.

I think that is just wording. I call it revert because during the reconciliation it will need to be _reverted_ for any reconciliation steps which come afterward. I guess that is the same what you call block. In any case, I did not saw any way of overriding this in the proposal which AFAIU is what @fvaleri is asking about. The truth is that you never know when it comes handy. So while I don't see any use-case right now, maybe we should think about it. E.g. some annotation which can be set to force the scale down.

fvaleri:
It turns out that moving all replicas out of a broker may be very difficult on busy clusters, because new topics are constantly created and replicas may be assigned to that broker. This would block the cluster scale down indefinitely, even if brokers are almost empty.

AFAIK, in Kafka there is no broker "maintenance mode" that would avoid new replicas being assigned to it (LinkedIn branch has this feature). Probably the operator may help this, but I think it is something that should be part of Kafka.

For example, in a cluster where you are enforcing a minimum RF and after reassigning the bulk of the data, you may need to use the override flag in order to complete the scale down.

scholzj:
@fvaleri I'm not sure such clusters are typical. But it is a good point, such cases will exist.

I think we can have for example an annotation which overrides this. The same should probably apply also to @ppatierno's automatic scale-down rebalance.

ppatierno:
> The same should probably apply also to @ppatierno's automatic scale-down rebalance.

@scholzj wdym? The automatic rebalance after scale down starts only when the scale down is done (forced or not, by an annotation we are discussing for this proposal). Anyway, when the rebalance starts, as the auto-rebalance proposal stands today, you could stop by applying the `stop` annotation on the auto-created `KafkaRebalance` (even if it should not be touched by the user). I have to find time to come back to the proposal because there is another possible approach on the table.

scholzj:
> The automatic rebalance after scale down starts only when the scale down is done

Really? That does not make sense to me. But I guess let's leave it to the other proposal then to not mix the discussions.

ppatierno:
> Really? That does not make sense to me. But I guess let's leave it to the other proposal then to not mix the discussions.

No sorry! It's of course the other way around :-D The scale down starts only when the rebalance ends :-P

Frawless (approved):

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,58 @@
...
+### Class
+
+A new utility class named `PreventBrokerScaleDownUtils` will be added to the assembly package of cluster operator module
+
+The reason for having this class is to keep all the utilities for checking the broker partition replicas separate such that we don't create a mess in the `KafkaReconciler` class by adding multiple large methods.
+Also, it will help us to test the changes easily.
+
+This class will have utility methods like `canScaleDownBrokers`, `canScaleDownBrokerCheck` and `brokerHasAnyReplicas` methods to check the if broker contains any partition replicas or not.
+
+This class will also contain methods like `topicNames`, `describeTopics` to query the topic metadata.
```
tombentley, line 32: As a rule of thumb, it's typically not necessary in a Proposal to go into great detail about the implementation (which will have its own PR). It's not universally true (e.g. adding or removing dependencies), but in this case I think you could drop this section, personally.

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,58 @@
...
+- For doing so, the topic metadata will be queried to detect if the broker contains any partition replicas.
+- An Admin client instance will be used to connect with the cluster and get us the topic details(topic name and topic description)
+- Then we can use this information to check if the broker contains any partition replicas or not.
+- The scale down is done after we make sure that the brokers that are going to be removed doesn't contain any partition replicas. By doing this we avoid any partial scale down.
+
+#### What to do if a broker contains partitions?
+
+#### Flow:
+
+- If partition replicas are found out on the broker we will revert back the kafka replicas to the previous count by changing the replica count directly in the kafka object and update the status with a warning stating that the scale down is not done and the replica count is changed back to previous one in the memory.
```
tombentley, line 48: 1. What exactly are we changing it back _to_? The current value of the `spec.replicas` in the STS/SPS?
2. Be explicit whether or not the rest of the reconciliation carries on as normal (I think it does, right?)
3. Be explicit about the revised schema for the `status`. Is this situation modelled as a new condition in the `status.conditions`, or some other way?  In particular how does the user know what the current number of brokers in their cluster is? For example if you're saying the `Ready` condition will be present (because the rest of the reconciliation completed OK), and some other new condition (e.g. `CannotScaleDown`) is also present (to indicate that the scale down specifically can't be done right now), then the `spec.kafka.replicas` != the actual number of replicas, and the user can't know from the `Kafka` resource what the actual number of 
brokers in their cluster is.

prevent-broker-scale-down-if-contains-partition-replicas.md
ShubhamRwt: Sorry for not resolving this earlier but this is now updated

prevent-broker-scale-down-if-contains-partition-replicas.md
ShubhamRwt: 1. Yes. I am changing the `spec.replicas` in STS/SPS
2. Yes, it will run normally
3. I was planning to print the `spec.replicas` in the condition's message

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -41,11 +30,12 @@ This class will also contain methods like `topicNames`, `describeTopics` to quer
...
 - The scale down is done after we make sure that the brokers that are going to be removed doesn't contain any partition replicas. By doing this we avoid any partial scale down.
 
-#### What to do if a broker contains partitions?
+### What to do if a broker contains partitions?
 
 #### Flow:
 
-- If partition replicas are found out on the broker we will revert back the kafka replicas to the previous count by changing the replica count directly in the kafka object and update the status with a warning stating that the scale down is not done and the replica count is changed back to previous one in the memory.
+- If partition replicas are found out on the broker we will revert back the kafka replicas to the previous count by changing the `spec.replicas` in the STS/SPS and the rest of the reconciliation will be done normally.
+- We will also generate a new condition which will be added to Kafka resource status depicting that the scale down is not done. It will also contain the `spec.replicas` count(which is being currently being used) in the condition message.
```
tombentley, line 38: Can you give an example YAML?

Would the fact that this resource wasn't completely reconciled be apparent from any metrics? Off the top of my head this would be the first time a resource with a `Ready` condition wasn't completely reconciled. I'm not sure how well that aligns with user expectations, TBH. We might need to add a metric so that users can alert on resources like this.

ppatierno (approved):

prevent-broker-scale-down-if-contains-partition-replicas.md
ShubhamRwt: @tombentley Hi, do you want the yaml with the warning condition?

ShubhamRwt: This is the Kafka resource yaml when the scaled down fails --> https://gist.github.com/ShubhamRwt/1d26531f5866529eba85820b93825110. I have added the `status` section to the proposal also. Iam not sure on if we need metrics or not for this since we are not failing the reconciliation. @ppatierno @scholzj wdyt?

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,61 @@
...
+- For doing so, the topic metadata will be queried to detect if the broker contains any partition replicas.
+- An Admin client instance will be used to connect with the cluster and get us the topic details(topic name and topic description)
+- Then we can use this information to check if the broker contains any partition replicas or not.
+- The scale down is done after we make sure that the brokers that are going to be removed doesn't contain any partition replicas. By doing this we avoid any partial scale down.
+
+### What to do if a broker contains partitions?
+
+#### Flow:
+
+- If partition replicas are found out on the broker we will revert back the kafka replicas to the previous count by changing the `spec.replicas` in the STS/SPS and the rest of the reconciliation will be done normally.
```
scholzj, line 37: There is no `.spec.replicas` in StrimziPodSets. This seems to also ignore all the other parts - e.g. generated broker certificates, generated services, ingresses, routes etc.

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,61 @@
...
+- We will also generate a new condition which will be added to Kafka resource status depicting that the scale down is not done. It will also contain the `spec.replicas` count(which is being currently being used) in the condition message.
+```yaml
+Status:
+  Cluster Id:  WQvEjYUMS1aiIKtMTvkMIw
+  Conditions:
+    Last Transition Time:  2023-01-13T13:30:18.880555172Z
+    Message:               Can't Scale down since broker contains partition replicas. Ignoring `replicas` setting in Kafka custom resource: my-cluster-kafka. Current `spec.replicas` value is 3
```
scholzj, line 44: I'm not sure if these are just typos in the message or misunderstanding how it works. But normally, the custom resource would be `my-cluster`. `my-cluster-kafka` would be the workload resource (e.g. StatefulSet). Also, in the custom resource, the path would be in `.spec.kafka.replicas` and not `.spec.replicas`.

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,61 @@
...
+    Reason:                ScaleDownException
+    Status:                True
+    Type:                  Warning
+    Last Transition Time:  2023-01-13T13:30:20.844245196Z
+    Status:                True
+    Type:                  Ready
+```
+- During the check, if the admin client is not able to connect to the cluster (not able to get the topic details) or if the partition replicas are assigned to the brokers that are going to be removed, we will update the status of the Kafka CR with the respective warning and revert back the replica count.
```
scholzj, line 52: If would be good to describe how do you plan to deal with less usual situations:
* Initial deployment
* Scale-down, but no pods running etc.

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Proposal
+
+This proposal suggest how we can add the check to detect if the broker still contains any partition replicas and what to do if the broker scale down is not possible.
+
+## Implementation
+
+### Process:
+
+- When the broker count is changed in the Kafka resource, the `reconcile` method of the `KafkaReconciler` will be triggered to reconcile the Kafka brokers.
```
scholzj, line 25: You should describe how do you actually detect if scale-down is happening or not as presumably you will run this only in scale-down situation?

ShubhamRwt:
I have left the comments as unresolved for the moment to make sure that all of them are closed only when they are addressed properly

prevent-broker-scale-down-if-contains-partition-replicas.md
scholzj: Well, until we have some auto-scaling solution, they will probably notice it:
* They will take a manual action and expect to see the result
* It will not be just the status warning they will see, they will also see the pods which did not scale down

So I think this will be more of a problem if we ever have some form of auto-scaling.

scholzj (commented):
I left some comments, but in addition to that, please ...
* Look at the formatting - Kafka with capital `K`, no double spaces, when using multiline bullet listes use indentation to make it more readable, when reffering to class format it as code etc. (e.g. you say _Kafka Cluster_ and it is not really clear if that is just a general _Kafka cluster_ reference or reference to `KafkaCluster` class etc.)
* On the beginning, we discussed the possibility to have a switch to disable this. I do not see it implemented here. If that is the concensus, fine. But since it was discussed, might be worth to be explicit that if this for some reason doesn't work, there is no way around than to not scale-down. (I personally would vote for a switch - e.g. an annotation on the Kafka CR to bypass the validation ... but as I said, if nto planned, at least mention it)

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,70 @@
...
+  Cluster Id:  WQvEjYUMS1aiIKtMTvkMIw
+  Conditions:
+    Last Transition Time:  2023-01-13T13:30:18.880555172Z
+    Message:               Can't Scale down since broker contains partition replicas. Ignoring `spec.kafka.replicas` setting in Kafka custom resource: my-cluster. Current kafka replicas count is 3
+    Reason:                ScaleDownException
+    Status:                True
+    Type:                  Warning
+    Last Transition Time:  2023-01-13T13:30:20.844245196Z
+    Status:                True
+    Type:                  Ready
```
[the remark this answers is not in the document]
ShubhamRwt: Yeah, I used `kubectl describe`. I will fix this

prevent-broker-scale-down-if-contains-partition-replicas.md
ppatierno: Good point from Jakub. User has ways to see that what they asked for didn't happen. Anyway, I am not against having a metric. It's just not sure how much it's useful today.

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,80 @@
...
+        status: "True"
+        type: Warning
+      - lastTransitionTime: "2023-02-22T10:18:57.664668863Z"
+        status: "True"
+        type: Ready
+  ```
+
+### How to bypass the broker scale down mechanism
+
+- To bypass the broker scale down mechanism you can use the annotation `strimzi.io/bypass-brokerScaleDownCheck: "true"` on the Kafka custom resource. For e.g.
```
fvaleri, line 63: It's good to have a bypass option, but I found the annotation name a bit weird. For consistency with the other annotations we provide, I would suggest `strimzi.io/bypass-broker-scaledown-check: "true"`.

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,80 @@
...
+- The broker certificates, services, ingresses, routes etc. will be treated with the original number of nodes and the rest of the reconciliation will be done normally.
+- There is a small risk at this point since the KafkaCluster instance will already exist and the storage validation will be already done. This can cause forbidden storage changes which can lead to some problems since the validation was done with the new count of Kafka replicas but at the same time the likelihood of both illegal storage changes and scale-down happening at the same time is low. 
+- We also generate a new condition which will be added to Kafka resource status depicting that the scale down is not done. It will also contain the `spec.replicas` count(which is being currently being used) in the condition message.
+  ```yaml
+  status:
+    clusterId: DoRj5f84Sruq_7TJ31y7Zw
+    conditions:
+      - lastTransitionTime: "2023-02-22T10:18:56.578009768Z"
+        message: 'Can''t Scale down since broker contains partition replicas. Ignoring
```
fvaleri, line 50: Here I prefer the error message style that we use for `ReplicationFactorChangeException`, where we suggest to revert back and we use the exact configuration path (i.e. `spec.kafka.replicas`).

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,80 @@
...
+By doing this we avoid any partial scale down.
+
+### What to do if a broker contains partitions?
+
+#### Flow:
+
+- If partition replicas are found out on the broker we will revert back the Kafka replicas to the previous count by setting replicas directly in the `KafkaCluster` class using `setReplicas()` method. 
+  Changing the Kafka replica count directly in the Kafka Cluster would help us ensure that we keep same replicas everywhere like while generating certificates, services, ingresses, routes etc.
+- The broker certificates, services, ingresses, routes etc. will be treated with the original number of nodes and the rest of the reconciliation will be done normally.
+- There is a small risk at this point since the KafkaCluster instance will already exist and the storage validation will be already done. This can cause forbidden storage changes which can lead to some problems since the validation was done with the new count of Kafka replicas but at the same time the likelihood of both illegal storage changes and scale-down happening at the same time is low.
```
fvaleri, line 43: This sounds scary. Can we elaborate more? I think a concrete example would help here.

fvaleri (commented):
Thanks @ShubhamRwt. Left a couple of comments.

prevent-broker-scale-down-if-contains-partition-replicas.md
ShubhamRwt: Does this look good --> `'Cannot Scale down since broker contains partition replicas. The `spec.kafka.replicas` should be reverted to 4 directly in the Kafka resource'.`

prevent-broker-scale-down-if-contains-partition-replicas.md
ShubhamRwt: @fvaleri @scholzj Does this sound less scary and better -->   Note : - By the time the replicas are reverted back, the storage validation will be already complete based on the replica count present in Kafka custom resource. This can can cause some issues if someone tries to make some forbidden changes (changes that might not be supported) to the storage during this time frame. To fix this problem, the user should revert back the `spec.kafka.replicas` in the Kafka custom resource back to the replica count currently being used by the `KafkaCluster` class and the next reconciliation will pick up those changes.

ShubhamRwt: Fixed now as per suggestion by @scholzj

prevent-broker-scale-down-if-contains-partition-replicas.md
fvaleri: Yep, much better. Thanks.

prevent-broker-scale-down-if-contains-partition-replicas.md
fvaleri: > Note: By the time the replicas are reverted back, the storage validation will be already complete based on the replica count present in Kafka custom resource. This can can cause some issues if someone tries to make some forbidden changes (changes that might not be supported) to the storage during this time frame. To fix this problem, the user should revert back the `spec.kafka.replicas` in the Kafka custom resource back to the replica count currently being used by the `KafkaCluster` class and the next reconciliation will pick up those changes. 

That's better, but you have "back" two times. The final phrase can simply be: 

To fix this problem, the user should revert back the `spec.kafka.replicas` in the Kafka custom resource and the next reconciliation will pick up those changes.

fvaleri (approved):
LGTM. Thanks!

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,82 @@
...
+## Proposal
+
+This proposal suggest how we can add the check to detect if the broker still contains any partition replicas and what to do if the broker scale down is not possible.
+
+## Implementation
+
+### Process:
+
+- When the broker count is changed in the Kafka resource, the `reconcile` method of the `KafkaReconciler` will be triggered to reconcile the Kafka brokers.
+- The `canScaleDownBrokers()` utility method will be present at the top of the compose chain in the `reconcile()` method of the `KafkaReconciler` to make sure that every other method  which requires the replica count use the correct replica count based on the outcome of the check.
```
scholzj, line 26: Double space.

```suggestion
- The `canScaleDownBrokers()` utility method will be present at the top of the compose chain in the `reconcile()` method of the `KafkaReconciler` to make sure that every other method which requires the replica count use the correct replica count based on the outcome of the check.
```

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,82 @@
...
+      - lastTransitionTime: "2023-02-22T10:18:56.578009768Z"
+        message: 'Cannot Scale down since broker contains partition replicas. The `spec.kafka.replicas` should be reverted back to 4 directly in the Kafka resource'.
+        reason: ScaleDownException
+        status: "True"
+        type: Warning
+      - lastTransitionTime: "2023-02-22T10:18:57.664668863Z"
+        status: "True"
+        type: Ready
+  ```
+  Note :  By the time the replicas are reverted back, the storage validation will be already complete based on the replica count present in Kafka custom resource.
```
scholzj, line 57:
```suggestion
  Note: By the time the replicas are reverted back, the storage validation will be already complete based on the replica count present in Kafka custom resource.
```

scholzj (approved):
Left some nits. But LGTM otherwise.

see-quick (approved):

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,82 @@
...
+## Implementation
+
+### Process:
+
+- When the broker count is changed in the Kafka resource, the `reconcile` method of the `KafkaReconciler` will be triggered to reconcile the Kafka brokers.
+- The `canScaleDownBrokers()` utility method will be present at the top of the compose chain in the `reconcile()` method of the `KafkaReconciler` to make sure that every other method which requires the replica count use the correct replica count based on the outcome of the check.
+- The `canScaleDownBrokers()` method will only run if we see the current Kafka replicas (replicas before the Kafka custom resource is modified) count gets greater than the Kafka replicas present in the Kafka custom resource.
+  We can get the desired Kafka replica count by using `kafka.getReplicas()` where `kafka` is an object of `KafkaCluster` class .
+- This method will check if the broker contains any partition replicas or not and will continue the process based on the outcome.
+- For doing so, the topic metadata will be queried to detect if the broker contains any partition replicas.
```
PaulRMellor, line 30:
```suggestion
- To do so, the topic metadata will be queried to detect if the broker contains any partition replicas.
```

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,82 @@
...
+- An Admin client instance will be used to connect with the cluster and get us the topic details(topic name and topic description)
+- Then we can use this information to check if the broker contains any partition replicas or not.
```
PaulRMellor, line 32:
```suggestion
- We can then use this information to check if the broker contains any partition replicas or not.
```

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,82 @@
...
+- The scale down is done after we make sure that the brokers that are going to be removed doesn't contain any partition replicas.
```
PaulRMellor, line 33:
```suggestion
- The scale-down is made after we make sure that the brokers that are going to be removed don't contain any partition replicas.
```

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,82 @@
...
+  By doing this we avoid any partial scale down.
```
PaulRMellor, line 34:
```suggestion
  By doing this we avoid any partial scale-down.
```

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,82 @@
...
+
+### What to do if a broker contains partitions?
+
+#### Flow:
+
+- If partition replicas are found out on the broker we will revert back the Kafka replicas to the previous count by setting replicas directly in the `KafkaCluster` class using `setReplicas()` method.
```
PaulRMellor, line 40:
```suggestion
- If partition replicas are found on the broker, we will revert back the Kafka replicas to the previous count by setting replicas directly in the `KafkaCluster` class using `setReplicas()` method. 
```

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,82 @@
...
+- Then we can use this information to check if the broker contains any partition replicas or not.
+- The scale down is done after we make sure that the brokers that are going to be removed doesn't contain any partition replicas.
+  By doing this we avoid any partial scale down.
+
+### What to do if a broker contains partitions?
+
+#### Flow:
+
+- If partition replicas are found out on the broker we will revert back the Kafka replicas to the previous count by setting replicas directly in the `KafkaCluster` class using `setReplicas()` method. 
+  Changing the Kafka replica count directly in the Kafka Cluster would help us ensure that we keep same replicas everywhere like while generating certificates, services, ingresses, routes etc.
```
PaulRMellor, line 41:
```suggestion
  Changing the Kafka replica count directly in the Kafka Cluster helps to ensure that we keep the same replicas everywhere, such as when generating certificates, services, ingresses, routes etc.
```

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,82 @@
...
+- The broker certificates, services, ingresses, routes etc. will be treated with the original number of nodes and the rest of the reconciliation will be done normally.
+- We also generate a new condition which will be added to Kafka resource status depicting that the scale down is not done. It will also contain the `spec.replicas` count(which is being currently being used) in the condition message.
```
PaulRMellor, line 43:
```suggestion
- We also generate a new condition which will be added to Kafka resource status depicting that the scale down is not done. It will also contain the `spec.replicas` count (that is being currently being used) in the condition message.
```

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,82 @@
...
+  This can can cause some issues if someone tries to make some forbidden changes (changes that might not be supported) to the storage during this time frame.
+  This is hard to prevent in the current code.
+  But the likelihood of this happening at the same time should be relatively small.
+  So this proposal suggests to ignore this risk.
+
+### How to bypass the broker scale down mechanism
+
+- To bypass the broker scale down mechanism you can use the annotation `strimzi.io/bypass-broker-scaledown-check: "true"` on the Kafka custom resource. For e.g.
+  ```sh
+  kubectl annotate Kafka my-cluster strimzi.io/bypass-broker-scaledown-check: "true"
```
PaulRMellor, line 67:
```suggestion
- To bypass the broker scale down mechanism you can use the annotation `strimzi.io/bypass-broker-scaledown-check: "true"` on the Kafka custom resource:
  ```sh
  kubectl annotate Kafka my-cluster strimzi.io/bypass-broker-scaledown-check: "true"
```

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,82 @@
...
+  ```
+
+### Other Scenarios
+
+- During the check, if the admin client is not able to connect to the cluster (not able to get the topic details), we will update the status of the Kafka CR with the respective warning and revert back the replica count in the `KafkaCluster` class.
```
PaulRMellor, line 72:
```suggestion
- During the check, if the Admin client is not able to connect to the cluster (not able to get the topic details), we will update the status of the Kafka CR with the respective warning and revert back the replica count in the `KafkaCluster` class.
```

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,82 @@
...
+- If the Kafka cluster is just initialized and the pods are not ready, the `canScaleDownBrokers()` utility method will not work because the current Kafka replicas (replicas before the Kafka custom resource is modified) count will be equal to 0 and the Kafka replicas present in the Kafka custom resource will also be 0. Hence, the mechanism will not run since the condition requires the current replica count to be greater than the Kafka replicas present in the Kafka custom resource
+- If the current Kafka replicas/pods are 0 the mechanism will not work since if there are zero brokers it will not be considered a scaledown.
```
PaulRMellor, line 74:
```suggestion
- If the current Kafka replicas/pods are 0, the mechanism will not work since if there are zero brokers it will not be considered a scale-down.
```

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,82 @@
...
+ 
+## Affected/not affected projects
+
+This change will affect the Strimzi Cluster Operator Module and mostly the `KafkaReconciler` class.
```
PaulRMellor, line 78:
```suggestion
This change will affect the Strimzi Cluster Operator module and mostly the `KafkaReconciler` class.
```

PaulRMellor (commented):
Sounds like a useful mechanism to introduce.
Approved with some minor suggestions :+1:

PaulRMellor (approved):

prevent-broker-scale-down-if-contains-partition-replicas.md
```diff
@@ -0,0 +1,82 @@
...
+        message: 'Cannot Scale down since broker contains partition replicas. The `spec.kafka.replicas` should be reverted back to 4 directly in the Kafka resource'.
+        reason: ScaleDownException
+        status: "True"
+        type: Warning
+      - lastTransitionTime: "2023-02-22T10:18:57.664668863Z"
+        status: "True"
+        type: Ready
+  ```
+  Note: By the time the replicas are reverted back, the storage validation will be already complete based on the replica count present in Kafka custom resource.
+  This can can cause some issues if someone tries to make some forbidden changes (changes that might not be supported) to the storage during this time frame.
```
im-konge, line 58:
```suggestion
  This can cause some issues if someone tries to make some forbidden changes (changes that might not be supported) to the storage during this time frame.
```

im-konge (approved):

tombentley (approved):

scholzj:
This proposal now have sufficient number of votes. If there are no new comments tomorrow, we will proceed and merge it as approved.

scholzj:
@ShubhamRwt Can you please rename the file and add it to the index so that it can be merged? Thanks.

ShubhamRwt:
Done @scholzj

README.md
```diff
@@ -9,6 +9,7 @@ This repository list of proposals for the Strimzi project. A template for new pr
 
 |  #  | Title                                                                 |
 | :-: |:----------------------------------------------------------------------|
+| 49  | [Prevent broker scale down if it contains partition replicas](./049-prevent-broker-scale-down-if-it-contains-partition-replicas]) |
```
scholzj, line 12: Is it missing the `.md` at the end in the link?

ShubhamRwt: Sorry my bad, Done now

scholzj:
Approved with 7 binding +1 votes and 2 non-binding +1 votes (+ Shubham as the author in addition).

PR #56, opened 2022-06-08
Add proposal about auto-approval mechanism for optimization proposals

ppatierno:
This proposal is about adding a mechanism for auto-approval of rebalancing optimization proposals.
It is related to https://github.com/strimzi/strimzi-kafka-operator/issues/6910.

### DIFF
--- a/038-optimization-proposal-autoapproval.md
+++ b/038-optimization-proposal-autoapproval.md
@@ -7,37 +7,21 @@ It means that the user can just create the `KafkaRebalance` custom resource and 
 
 Currently, when the user wants to run a cluster rebalance, he has to create a `KafkaRebalance` custom resource for getting an optimization proposal first.
 After getting the proposal, the only way to start the rebalancing operation is by approving it by annotating the custom resource with the `strimzi.io/rebalance=approve` annotation.
-There is the need for a manual intervention of the user and this is a two step process.
+There is the need for a manual intervention of the user and this is a two steps process.
 
 ## Motivation
 
-A user would have the desire to reduce the manual intervention on a cluster for rebalancing operations.
+In some cases, the manual approval process of the rebalance proposal is not necessary.
 Automatically approving an optimization proposal can save time from an operational point of view.
 It enables more automation where just creating a `KafkaRebalance` custom resource can go straight to the cluster rebalance without any additional manual interaction.
 
 ## Proposal
 
 The `KafkaRebalance` custom resource can be extended with a new `spec.autoApproval` field for this purpose.
-The auto-approval can then be enabled or disable through a boolean `spec.autoApproval.enabled` field.
 
 ### Auto-approval
 
-The auto-approval can be enabled through the boolean `spec.autoApproval.enabled` field set to `true`.
-
-```yaml
-apiVersion: kafka.strimzi.io/v1beta2
-kind: KafkaRebalance
-metadata:
-  name: my-rebalance
-  labels:
-    strimzi.io/cluster: my-cluster
-spec:
-  mode: # any mode
-  autoApproval:
-    enabled: true
-```
-
-The same result can be easily achieved if the `spec.autoApproval` is specified as empty.
+The auto-approval can be enabled just adding an empty `spec.autoApproval` field.
 
 ```yaml
 apiVersion: kafka.strimzi.io/v1beta2
@@ -57,31 +41,15 @@ The reason for not having a simple boolean `spec.autoApproval` field is for leav
 
 If the `spec.autoApproval` field is not specified at all, the default behavior will be the current one, so the need for manual approval by the user annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation.
 
-Of course, setting the `spec.autoApproval.enabled` to `false` drives to the same result.
-
-```yaml
-apiVersion: kafka.strimzi.io/v1beta2
-kind: KafkaRebalance
-metadata:
-  name: my-rebalance
-  labels:
-    strimzi.io/cluster: my-cluster
-spec:
-  mode: # any mode
-  autoApproval:
-    enabled: false
-```
-
 ### Flow
 
 As described before, the user interaction flow assumes that the `spec.autoApproval` field is specified in the `KafkaRebalance` custom resource.
 
 1. The user creates a `KafkaRebalance` custom resource.
 2. The cluster operator asks Cruise Control to generate an optimization proposal via the REST API.
-3. When the optimization proposal is ready, the cluster operator checks if the user has specified the `spec.autoApproval` and the `spec.autoApproval.enabled` is set or not:
+3. When the optimization proposal is ready, the cluster operator checks if the user has specified the `spec.autoApproval`.
     * If the `spec.autoApproval` is not specified at all, the user has to approve the proposal manually as usual.
-    * If the `spec.autoApproval` is specified but it's not enabled with `spec.autoApproval.enabled` set to `false`, the user has to approve the proposal manually as usual.
-    * If the `spec.autoApproval` is specified and it's enabled with `spec.autoApproval.enabled` set to `true`, the cluster operator approves the proposal automatically by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation.
+    * If the `spec.autoApproval` is specified, the cluster operator approves the proposal automatically by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation.
 
 ## Affected/not affected projects
 
@@ -91,9 +59,8 @@ This proposal impacts the Strimzi Cluster Operator only and mostly the `KafkaReb
 
 The manual approval will be still in place as it is today.
 As described before, if the `spec.autoApproval` field is not specified in the `KafkaRebalance` custom resource, the default behavior will be the current one, so the need for manual approval by the user.
-The same happens if the `spec.autoApproval.enabled` is set to `false`.
 
 ## Rejected alternatives
 
-Having a simple boolean `spec.autoApproval` field was rejected because the current proposal allows more extensibility for the future if it's needed to add more configuration like for example criteria or rules to met for auto-approving the proposal.
+Having a simple boolean `spec.autoApproval` field was rejected because the current proposal allows more extensibility for the future if it's needed to add more configuration like for example criteria or rules to be met for auto-approving the proposal.
 Using a boolean, would have need an additional field for that like `spec.autoApprovalRules`.
--- a/038-optimization-proposal-autoapproval.md
+++ b/038-optimization-proposal-autoapproval.md
@@ -5,7 +5,7 @@ It means that the user can just create the `KafkaRebalance` custom resource and 
 
 ## Current situation
 
-Currently, when the user wants to run a cluster rebalance, he has to create a `KafkaRebalance` custom resource for getting an optimization proposal first.
+Currently, when the users want to run a cluster rebalance, they have to create a `KafkaRebalance` custom resource for getting an optimization proposal first.
 After getting the proposal, the only way to start the rebalancing operation is by approving it by annotating the custom resource with the `strimzi.io/rebalance=approve` annotation.
 There is the need for a manual intervention of the user and this is a two steps process.
 
--- a/038-optimization-proposal-autoapproval.md
+++ b/038-optimization-proposal-autoapproval.md
@@ -17,11 +17,11 @@ It enables more automation where just creating a `KafkaRebalance` custom resourc
 
 ## Proposal
 
-The `KafkaRebalance` custom resource can be extended with a new `spec.autoApproval` field for this purpose.
+The `KafkaRebalance` custom resource can be annotated with a new `strimzi.io/rebalance-auto-approval` annotation for this purpose.
 
 ### Auto-approval
 
-The auto-approval can be enabled just adding an empty `spec.autoApproval` field.
+The auto-approval can be enabled just annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance-auto-approval=true` annotation.
 
 ```yaml
 apiVersion: kafka.strimzi.io/v1beta2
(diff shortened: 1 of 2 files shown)


ppatierno:
> You will have issues supporting enough different the criteria to make everyone happy

This is why this PR is still in draft, because thinking about how to describe criteria and which ones to allow is not a simple task.

> You create a weird semi-manual process which isn't either automated nor manual

I would not say it's weird. It simplifies the user work if he has clear what are the criteria for auto-approving an optimization proposal. Your approach absolutely works, not sure all users would like it anyway.

> Continuously (or periodically if one wants) create proposals and if some criteria are full-filled (e.g. at least XX move operations etc.), it will auto-approve it.
This would still have some of the limitations mentioned above (e.g. not enough flexibility in the auto-approval criteria), but it will be fully automated.

This is something that is "partially" addressed (or at least as starting point) with this https://github.com/strimzi/proposals/pull/57 that is about scaling up and down but defines a way for the future "continuos" full rebalancing.

Anyway, the above auto-rebalancing proposal relies on the auto-approval mechanism described here and the additions to the `KafkaRebalance` custom resource for that.

Finally, @tombentley when you were referring to auto-approval on this comment, what exactly you had in mind? https://github.com/strimzi/proposals/pull/52#pullrequestreview-949993355

katheris:
I think anything that can be added to make it simpler to approve a rebalance proposal would be useful. At the moment the user is presented with a full proposal and expected to somehow decide whether to apply it without any guidance.

I guess the key missing piece here is what kind of criteria might the user specify. I know this proposal is also being used for the scale up and down, but for that case it seems like there wouldn't be much of a criteria, Strimzi would just apply the proposal once it was created. An obvious choice from the proposal would be to cap the amount of data being moved, or the number of replica or leader movements, but I don't know if in reality that would be important or not when approving a proposal.

I wonder if Jakub's suggestion of some sort of tooling to analyse the proposal and indicate whether it seems appropriate to apply might be a good starting point. That might be an easier thing to iterate on and get feedback to understand what kind of criteria users might actually use for determining whether to auto-approve a proposal.

ppatierno:
I can totally understand that having some criteria could be a bigger problem in order to cover all possible scenarios and new requests coming from users. The simpler way would just enabling the auto-approval with no criteria. This is anyway needed for the auto-rebalancing proposal https://github.com/strimzi/proposals/pull/57 but could be even used by the user for a manual rebalancing when they don't want to bother to take a look at the optimization proposal but just running the rebalancing straight away. If we agree on this I would remove the part mentioning criteria to move forward with this proposal as well (not draft anymore).

ppatierno:
@scholzj @katheris @tombentley @tomncooper because this one is strictly related to the https://github.com/strimzi/proposals/pull/57 would you mind to reach an agreement on criteria usage based on the discussion above? Nothing has changed from a content pov.

scholzj:
@ppatierno As explained in my initial comment -> I feel like the criteria just over complicates it. If needed, it can be probably also added in a later phase if we see the need and demand. So my vote is just for a boolean flag. It also helps you to unblock #57 as you do not need to work through the criteria design ;-). But as I said before, it is not a blocker for me if others want them.

ppatierno:
@scholzj I see your point (already before ;-)) ... I would update the PR taking the criteria idea out but if we make just a boolean field, it would mean an additional field to specify criteria later.
The idea could be something like this:

```yaml
autoApproval:
   enable: <value> # true or false
```

where `autoApproval: { }` could mean enable by default?
It could leave space for adding more in the `autoApproval` section later but I can understand it's not so "beautiful" as a simple `autoApproval: true` :-(

scholzj:
> @scholzj I see your point (already before ;-)) ... I would update the PR taking the criteria idea out but if we make just a boolean field, it would mean an additional field to specify criteria later. The idea could be something like this:
> 
> ```yaml
> autoApproval:
>    enable: <value> # true or false
> ```
> 
> where `autoApproval: { }` could mean enable by default? It could leave space for adding more in the `autoApproval` section later but I can understand it's not so "beautiful" as a simple `autoApproval: true` :-(

I think this is fine yes. I originally thought more about something like this:
```yaml
autoApproval: true
autoApprovalRules: # Future place for the criteria
  - ...
```

But yours is IMHO fine and has probably better space for future extensions. Only thing I wonder about is `enable: true` or `enabled: true`

ppatierno:
This should be now ready to be reviewed again, thanks!

038-optimization-proposal-autoapproval.md
```diff
@@ -0,0 +1,99 @@
...
+kind: KafkaRebalance
+metadata:
+  name: my-rebalance
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  mode: # any mode
+  autoApproval:
+    enabled: false
+```
```
scholzj, line 73: I wonder what others think about this. It might be a bit confusing to the users:
* This defaults to disabled:
  ```yaml
   spec:
     mode: ...
  ```
  => In other words, `enabled` seems to default to false.
* This defaults to enabled:
  ```yaml
   spec:
     mode: ...
     autoApproval: {}
  ```
  => Now suddenly, `enabled` seems to default to true.
* This seems to make not much sense (why would you do this if you can just leave this section out?)
  ```yaml
   spec:
     mode: ...
     autoApproval:
       enabled: false
  ```
* It would be especially weird in the future if you have something like this
  ```yaml
   spec:
     mode: ...
     autoApproval:
       enabled: false
       criteria:
         - ...
         - ...
  ```

I understand the motivation and it can be implemented in the code reasonably. I just wonder how intuitive this will be for a new user who will not read the proposal and know the background.

Would it be more natural to remove the `enabled` field completely and have it just as an empty object? I.e.
* This would disable it
  ```yaml
   spec:
     mode: ...
  ```
* This will enable it:
  ```yaml
   spec:
     mode: ...
     autoApproval: {}
  ```

I know the empty object might seem a bit weird, but if it is an future entry point, why not use it that way? And at the end, most Kafka CRs already use this in a way `userOperator: {}` => enabled with any flag.

038-optimization-proposal-autoapproval.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Flow
+
+As described before, the user interaction flow assumes that the `spec.autoApproval` field is specified in the `KafkaRebalance` custom resource.
+
+1. The user creates a `KafkaRebalance` custom resource.
+2. The cluster operator asks Cruise Control to generate an optimization proposal via the REST API.
+3. When the optimization proposal is ready, the cluster operator checks if the user has specified the `spec.autoApproval` and the `spec.autoApproval.enabled` is set or not:
+    * If the `spec.autoApproval` is not specified at all, the user has to approve the proposal manually as usual.
+    * If the `spec.autoApproval` is specified but it's not enabled with `spec.autoApproval.enabled` set to `false`, the user has to approve the proposal manually as usual.
```
scholzj, line 83: This does not fit with the previous chapter where you say the `enabled` will default to `true`. Maybe you wanted to say this to make it more clear?

```suggestion
    * If the `spec.autoApproval` is specified but is disabled with `spec.autoApproval.enabled` set to `false`, the user has to approve the proposal manually as usual.
```

038-optimization-proposal-autoapproval.md
```diff
@@ -0,0 +1,99 @@
...
+    * If the `spec.autoApproval` is specified and it's enabled with `spec.autoApproval.enabled` set to `true`, the cluster operator approves the proposal automatically by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation.
```
scholzj, line 84:
```suggestion
    * If the `spec.autoApproval` is specified and it's either enabled with `spec.autoApproval.enabled` set to `true` or when the `spec.autoApproval.enabled` field is not set at all, the cluster operator approves the proposal automatically by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation.
```

scholzj, line 84: Also ... you say `the cluster operator approves the proposal automatically by annotating the KafkaRebalance custom resource with the strimzi.io/rebalance=approve annotation`. Will it really set the annotation? Or just approve it inside the code and move it into _rebalancing_ state in the code only?

038-optimization-proposal-autoapproval.md
ppatierno: That's a really good point @scholzj, the behaviour of `enabled` is not unique in this way.
Thinking about having `autoApproval` section with `enabled: false` and maybe some criteria could be a way to "temporary" disable auto-approval? But does it really make any sense? I don't think so.
I will go with your suggestion getting rid of the `enabled` flag even because it could simplify the logic a little bit doesn't taking into account weird situations like the one just described.

038-optimization-proposal-autoapproval.md
ppatierno: By removing the `enabled` flag, this sentence doesn't make sense anymore. I will remove it.

038-optimization-proposal-autoapproval.md
ppatierno: It will change by removing the `enabled` flag.

ppatierno: In general, I would always rely on the current state machine (by applying the annotation) instead of having a specific check to run the same code for this scenario.

038-optimization-proposal-autoapproval.md
```diff
@@ -0,0 +1,66 @@
+# Auto-approval mechanism for optimization proposals
+
+This proposal is about allowing the user to create a `KafkaRebalance` custom resource for getting an optimization proposal and running a rebalance, but without the need of the manual approval via the `strimzi.io/rebalance=approve` annotation.
+It means that the user can just create the `KafkaRebalance` custom resource and the corresponding generated optimization proposal will be automatically approved.
+
+## Current situation
+
+Currently, when the user wants to run a cluster rebalance, he has to create a `KafkaRebalance` custom resource for getting an optimization proposal first.
```
scholzj, line 8: Maybe this way it is better to avoid any he/she discussions ;-)

```suggestion
Currently, when the users want to run a cluster rebalance, they have to create a `KafkaRebalance` custom resource for getting an optimization proposal first.
```

scholzj (approved):
One nit about wording. But LGTM in general.

038-optimization-proposal-autoapproval.md
scholzj: Ok, just wanted to make sure this is what you really wanted to do.

038-optimization-proposal-autoapproval.md
```diff
@@ -0,0 +1,66 @@
+# Auto-approval mechanism for optimization proposals
+
+This proposal is about allowing the user to create a `KafkaRebalance` custom resource for getting an optimization proposal and running a rebalance, but without the need of the manual approval via the `strimzi.io/rebalance=approve` annotation.
```
tomncooper, line 3:
```suggestion
This proposal is about allowing the user to create a `KafkaRebalance` custom resource for getting an optimization proposal and running a rebalance, but without the need for the manual approval via the `strimzi.io/rebalance=approve` annotation.
```

038-optimization-proposal-autoapproval.md
```diff
@@ -0,0 +1,66 @@
...
+It means that the user can just create the `KafkaRebalance` custom resource and the corresponding generated optimization proposal will be automatically approved.
```
tomncooper, line 4:
```suggestion
This means that the user can simply create the `KafkaRebalance` custom resource and the corresponding generated optimization proposal will be approved automatically .
```

038-optimization-proposal-autoapproval.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Current situation
+
+Currently, when the users want to run a cluster rebalance, they have to create a `KafkaRebalance` custom resource for getting an optimization proposal first.
```
tomncooper, line 8:
```suggestion
Currently, when the users want to run a cluster rebalance, they have to create a `KafkaRebalance` custom resource in order to generate an optimization proposal first.
```

038-optimization-proposal-autoapproval.md
```diff
@@ -0,0 +1,66 @@
...
+metadata:
+  name: my-rebalance
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  mode: # any mode
+  autoApproval: {}
+```
+
+The reason for not having a simple boolean `spec.autoApproval` field is for leaving space for a future extension, maybe adding some criteria or rules that have to be matched for auto-approving the optimization proposal.
```
tomncooper, line 38: Ooo I like this, I can foresee some kind of pluggable `RebalanceApprover` interface.

tomncooper (approved):

kyguy (approved):

038-optimization-proposal-autoapproval.md
```diff
@@ -0,0 +1,66 @@
...
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaRebalance
+metadata:
+  name: my-rebalance
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  mode: # any mode
+  autoApproval: {}
```
tombentley, line 35: Do we know that (in the future when this is acting as more than an unconditional "approve it automatically" flag) it would actually be an _object_, rather than a _list_ of criteria to be met? That is, I could imagine that in that future we'd actually want something like

```yaml
spec:
  mode: ...
  approvalCriteria:
    - criterionType: Foo
      threshold: 100
    - criterionType: Bar
      someCriterionParameter: true
```

I _guess_ having in intermediate object might allow some flexibility (e.g. you could have an `anyOf` property and an `allOf` property to provide different ways of combining multiple criteria), but it's equally not clear to me that we'd need such flexibility (and hence no need for an intermediate object either).

038-optimization-proposal-autoapproval.md
```diff
@@ -0,0 +1,66 @@
...
+### Flow
+
+As described before, the user interaction flow assumes that the `spec.autoApproval` field is specified in the `KafkaRebalance` custom resource.
+
+1. The user creates a `KafkaRebalance` custom resource.
+2. The cluster operator asks Cruise Control to generate an optimization proposal via the REST API.
+3. When the optimization proposal is ready, the cluster operator checks if the user has specified the `spec.autoApproval`.
+    * If the `spec.autoApproval` is not specified at all, the user has to approve the proposal manually as usual.
+    * If the `spec.autoApproval` is specified, the cluster operator approves the proposal automatically by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation.
+
```
tombentley, line 53: What should be the current behaviour if the `autoApproval`  object (or `approvalCriteria` list if that's what we implemented) is not empty.

What should happen if it changes during the lifetime of the CR?

038-optimization-proposal-autoapproval.md
tombentley: Note that `approvalCriteria` could allow use to handle the existing manual approval in a more consistent way. I.e. rather than using an annotation we could just say the API was to update the `spec.approvalCriteria` with a criterion that always evaluated as true. For example:

```yaml
spec:
  mode: ...
  # approvalCriteria: abence of property means not approved
```

And this CR

```yaml
spec:
  mode: ...
  approvalCriteria: 
    - criterionType: Always 
```

could be used for either:
* auto approval (if it was created like this) 
* manually approval (if edited to include this after the proposal was generated)

tombentley (commented):
Hi @ppatierno thanks for the proposal. The basic functionality and mechanics makes sense. I had a few questions about the proposed API.

038-optimization-proposal-autoapproval.md
ppatierno: I think that having a field named `approvalCriteria` right now when we don't have support for criteria is totally misleading for users. We have the desire to have criteria and rules but we don't have a real plan now, how much it's really feasible and if it will be really useful for the community so overcommitting could not help.
With the current `autoApproval` we could have a `criteria` or `rules` subfield to match your expectation. Maybe something like this?

```yaml
autoApproval:
  criteria:
    - type: Foo
      threshold: 100
    - type: Bar
      someCriterionParameter: true
```

which anyway leaves space in the `autoApproval` field to not having just a list.

038-optimization-proposal-autoapproval.md
ppatierno: If `autoApproval` is not empty (but it brings criteria, in the future), the KRAO should check that those criteria are met in order to approve or not the optimization proposal. Or is it anything different you were asking?

038-optimization-proposal-autoapproval.md
scholzj: So, what about using annotation instead of changing the CRDs until we know more about the future CRD structure? E.g. if `strimzi.io/rebalance-auto-approval` is set to `true`, it is auto approved?

ppatierno: We can go with annotation but I still don't get what makes `approvalCriteria` as list better than a generic `autoApproval` containing a field for the criteria list so allowing space for improvements. My feeling is that it's possible that we won't come back to have criteria anymore and we'll leave the annotation forever, so having an API (I talk about the `KafkaRebalance` CR) which mixes annotations and fields to provide features.

scholzj: > My feeling is that it's possible that we won't come back to have criteria anymore and we'll leave the annotation forever, so having an API (I talk about the KafkaRebalance CR) which mixes annotations and fields to provide features.

Well, if we see that we don't get back to it and don't need the criteria, we can convert the annotation into a boolean flag inside the KafkaRebalance CR the future. The end result in that case will be better than the empty object.

I approved the original design, so I was fine with it. Just trying to find some compromise to move forward with this. ;-)

ppatierno: I will do the change on the proposal from field to annotation (I see @tombentley thumbs up on this idea) and moving the current one as rejected. Should we just track somewhere (an issue? a card on the project?) that we have to come back to this in the future to evaluate if promoting to field is feasible?

tombentley: > so having an API which (...) mixes annotations and fields to provide features

Well we already have that because we use an annotation for the approval anyway. If we really didn't come back to add support for criteria then leaving people with an API where the have to specify an empty object where they logically mean a boolean is pretty unintuitive. So on balance I think I prefer using an annotation for the time being until we know enough to design the API to support criteria properly.

ppatierno: > Well we already have that because we use an annotation for the approval anyway

I see the use case different tbh. We are using annotation to take actions (approving, refreshing, ...) not for enabling or describing features. I see that the annotation usage fits pretty well when it's about "asking to the operator" to take an action on the CR. But of course, it could be just my view. I see the usage of annotation for configurability just as a temporary solution.

ppatierno: The proposal is now updated to use an annotation.

tombentley (approved):
Thanks Paolo!

PR #57, opened 2022-06-08
Added proposal for automatic rebalance on cluster scaling

ppatierno:
This proposal is about adding a mechaning to automate the rebalance on cluster scaling (up or down).
It is related to https://github.com/strimzi/strimzi-kafka-operator/issues/4112.

### DIFF
--- a/039-auto-rebalance-cluster-scaling.md
+++ b/039-auto-rebalance-cluster-scaling.md
@@ -55,12 +55,12 @@ spec:
 
 The user has the flexibility to specify the [goals and the other rebalancing related configurations](https://strimzi.io/docs/operators/latest/configuring.html#type-KafkaRebalance-reference) the same way they already do today on a `KafkaRebalance` custom resource.
 This allows the user to not rely on the default Cruise Control goals configuration which would be common for any rebalance without overriding goals.
-They could define different goals based on the status of the Kafka cluster or skip hard goals in some scenarios for examples.
+They could define different goals based on the status of the Kafka cluster or skip hard goals in some scenarios for example.
 As already mentioned, specifying goals but more in general customizing the optimization proposal request is already possible in the current usage of a `KafkaRebalance` custom resource so it should be allowed for the auto-rebalancing as well.
 In the end, the cluster operator will use that configuration from the `KafkaRebalanceTemplate` to fill the corresponding `KafkaRebalance` but adding the `spec.mode` and `spec.brokers` accordingly.
-It will also add the `spec.autoApproval` field, as described by this [proposal](https://github.com/strimzi/proposals/pull/56), for enabling auto-approval.
+It will also add the `strimzi.io/rebalance-auto-approval=true` annotation, as described by proposal [038](https://github.com/strimzi/proposals/blob/main/038-optimization-proposal-autoapproval.md), for enabling auto-approval.
 
-We would also need to extend the `Kafka` custom resource to allow the user specifying that he wants auto-rebalancing on scaling by using a template.
+We would also need to extend the `Kafka` custom resource to allow the user specifying that they want auto-rebalancing on scaling by using a template.
 A new `spec.cruiseControl.autoRebalance` field could be added for this purpose.
 Because we are talking about auto-rebalancing, the corresponding field could also be not limited to the rebalancing on scaling but a more general full "continuous” auto-rebalancing to enable on the cluster.
 
@@ -135,9 +135,9 @@ Following the operator flow when auto-rebalancing on scaling up.
 2. The `KafkaReconciler` takes care of it by scaling up the cluster as usual, by adding additional brokers.
 3. If the `spec.cruiseControl.autoRebalance` contains an entry with `add-brokers` mode, the `KafkaReconciler` creates a `KafkaRebalance` custom resource named as `<cluster-name>-auto-rebalance-add-brokers` by using the corresponding template and sets:
    * the `spec.mode` as `add-brokers` and the `spec.brokers` with the list of brokers already added to the cluster.
-   * the `spec.autoApproval` as described by this [proposal](https://github.com/strimzi/proposals/pull/56).
+   * the `strimzi.io/rebalance-auto-approval=true` annotation as described by proposal [038](https://github.com/strimzi/proposals/blob/main/038-optimization-proposal-autoapproval.md).
 4. The `KafkaRebalanceAssemblyOperator` detects the new rebalancing resource and acts as usual by requesting Cruise Control to generate an optimization proposal via the REST API on the `/add_broker` endpoint.
-5. When the optimization proposal is ready, the `KafkaRebalanceAssemblyOperator` approves it automatically, because of the `spec.autoApproval` field, by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation as defined by this [proposal](https://github.com/strimzi/proposals/pull/56).
+5. When the optimization proposal is ready, the `KafkaRebalanceAssemblyOperator` approves it automatically, because of the `strimzi.io/rebalance-auto-approval=true` annotation, by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation as defined by proposal [038](https://github.com/strimzi/proposals/blob/main/038-optimization-proposal-autoapproval.md).
 6. If/When the rebalancing ends successfully, the corresponding `KafkaRebalance` custom resource is deleted by the `KafkaReconciler`.
 
 #### Error handling
@@ -159,9 +159,9 @@ It's the user who should deal with this scenario:
 1. The user scales down the cluster by decreasing the `spec.kafka.replicas` field.
 2. If the `spec.cruiseControl.autoRebalance` contains an entry with `remove-brokers` mode, the `KafkaReconciler` creates a `KafkaRebalance` custom resource named as `<cluster-name>-auto-rebalance-remove-brokers` by using the corresponding template and sets:
    * the `spec.mode` as `remove-brokers` and the `spec.brokers` with the list of brokers that will be removed.
-   * the `spec.autoApproval` as described by this [proposal](https://github.com/strimzi/proposals/pull/56).
+   * the `strimzi.io/rebalance-auto-approval=true` annotation as described by proposal [038](https://github.com/strimzi/proposals/blob/main/038-optimization-proposal-autoapproval.md).
 3. The `KafkaRebalanceAssemblyOperator` detects the new rebalancing resource and acts as usual by requesting Cruise Control to generate an optimization proposal via the REST API on the `/remove_broker` endpoint.
-4. When the optimization proposal is ready, the `KafkaRebalanceAssemblyOperator` approves it automatically, because of the `spec.autoApproval` field, by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation as defined by this [proposal](https://github.com/strimzi/proposals/pull/56).
+4. When the optimization proposal is ready, the `KafkaRebalanceAssemblyOperator` approves it automatically, because of the `strimzi.io/rebalance-auto-approval=true` annotation, by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation as defined by proposal [038](https://github.com/strimzi/proposals/blob/main/038-optimization-proposal-autoapproval.md).
 5. If/When the rebalancing ends successfully, the corresponding `KafkaRebalance` custom resource is deleted by the `KafkaReconciler`.
 6. Finally, the `KafkaReconciler` can continue to scale down the cluster by deleting the brokers.
 
@@ -187,7 +187,7 @@ Of course, the user could change his mind by reverting back the `spec.kafka.repl
 
 This proposal impacts the Strimzi Cluster Operator only.
 The `KafkaReconciler` is impacted on handling the `KafkaRebalanceTemplate` custom resource and generating/deleting the corresponding `KafkaRebalance`.
-The `KafkaRebalanceAssemblyOperator` is impacted mostly on the auto-approval side but it is already defined by this [proposal](https://github.com/strimzi/proposals/pull/56).
+The `KafkaRebalanceAssemblyOperator` is impacted mostly on the auto-approval side but it is already defined by proposal [038](https://github.com/strimzi/proposals/blob/main/038-optimization-proposal-autoapproval.md).
 It handles the `KafkaRebalance` custom resource the same way as it was created by the user but it has to annotate the custom resource automatically.
 
 ## Compatibility
@@ -204,7 +204,10 @@ The only fields that the cluster operator has to specify, when creating the `Kaf
 
 * `spec.mode` : should be `add-brokers` or `remove-brokers` if scaling up or down.
 * `spec.brokers` : with the list of brokers to add/remove.
-* `spec.autoApproval` : for allowing the auto-approval as defined by this [proposal](https://github.com/strimzi/proposals/pull/56).
+
+The cluster operator has also to add the following annotation on the `KafkaRebalance` custom resource:
+
+* `strimzi.io/rebalance-auto-approval` : set to `true`, for allowing the auto-approval as defined by proposal [038](https://github.com/strimzi/proposals/blob/main/038-optimization-proposal-autoapproval.md).
 
 The drawback of this solution is that if the user wants to customize the rebalancing, he has to use the `spec.cruiseControl.config` field in the Kafka custom resource. It’s important to mention that it affects the overall Cruise Control configuration so that any other “empty” rebalancing request, not related to scaling, will use the same options.
 Users won't also be able to skip hard goals to move forward with a rebalancing if they are not met anyway.
@@ -256,6 +259,6 @@ spec:
         # ... other rebalancing related configuration  
 ```
 
-For handling the auto-approval, a `spec.autoApproval` field, as described by this [proposal](https://github.com/strimzi/proposals/pull/56), would be needed as well.
+For handling the auto-approval, a `strimzi.io/rebalance-auto-approval=true` annotation, as described by proposal [038](https://github.com/strimzi/proposals/blob/main/038-optimization-proposal-autoapproval.md), would be needed as well.
 This proposal was rejected because it pollutes the `Kafka` custom resource with too many details and configuration related to a rebalancing operation when the user wants to specify specific desires compared to the default ones in the Cruise Control configuration.
 To understand how big the configuration for each rebalancing mode could be, [here](https://strimzi.io/docs/operators/latest/configuring.html#type-KafkaRebalance-reference) the options that we support today which could anyway increase in the future.
--- a/039-auto-rebalance-cluster-scaling.md
+++ b/039-auto-rebalance-cluster-scaling.md
@@ -1,0 +1,261 @@
+# Automatic rebalance on cluster scaling
+
+This feature is about enabling the cluster operator to get an optimization proposal, which is auto-approved, in order to run the corresponding rebalancing when the user scales the cluster up or down by changing the `spec.kafka.replicas` field on the `Kafka` custom resource.
+Such a rebalancing can happen right after the scale up to move some partition replicas to the newly added brokers or right before the scale down to move partition replicas off of the brokers to be removed.
+
+## Current situation
+
+Currently, there is no automatic rebalancing when the user scales the cluster up or down.
+It means that:
+
+* after scaling up the cluster, in order to rebalance the cluster and move some partition replicas to the newly added brokers, the user has to create a `KafkaRebalance` custom resource using the `spec.mode: add-brokers` mode and add the new broker IDs to the the `spec.brokers` field.
+* before scaling down the cluster, in order to rebalance the cluster and move partition replicas off of the brokers to be removed, the user has to create a `KafkaRebalance` custom resource using the `spec.mode: remove-brokers` mode and add the broker IDs of the brokers to be removed to the `spec.brokers` field.
+
+For more details, the proposal [035](https://github.com/strimzi/proposals/blob/main/035-rebalance-types-scaling-brokers.md) introduced the rebalancing modes mentioned above.
+
+## Motivation
+
+The rebalancing operation during scaling up and down is a manual two steps process.
+
+* scaling up the cluster, then running a rebalance.
+* running a rebalance, then scaling down the cluster.
+
+In order to simplify the process, it would be great for the user to deal with just scaling the cluster by changing the `spec.kafka.replicas` field and then having the operator taking care of running a rebalancing operation automatically if requested.
+
+Taking into account the `StrimziPodSets` usage (instead of the `StatefulSets`), which is the default since the [0.30.0](https://github.com/strimzi/strimzi-kafka-operator/releases/tag/0.30.0) release, a more general use case will be about adding new brokers and/or removing existing ones, not talking just about scaling up and down anymore.
+
+This proposal takes into account this future plan by referring to more general operations like add and remove brokers instead of scaling up and down.
+
+## Proposal
+
+A new `KafkaRebalanceTemplate` custom resource could be a solution allowing the user to define the configuration for the auto-rebalancing but leaving to the operator the creation of the corresponding `KafkaRebalance` custom resource.
+This new CRD should be quite similar to the `KafkaRebalance` one.
+The cluster operator has to copy over the configuration from the template to the actual rebalancing resource.
+Compared to the rejected alternative, of adding the configuration to the `Kafka` custom resource, this will avoid polluting the `Kafka` custom resource and increasing its size.
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaRebalanceTemplate
+metadata:
+  name: my-rebalance-template
+spec:
+  # NOTE: mode and brokers fields are not available as in the KafkaRebalance,
+  #       they are automatically set on the corresponding KafkaRebalance by the operator
+  goals:
+    - CpuCapacityGoal
+    - NetworkInboundCapacityGoal
+    - DiskCapacityGoal
+    - RackAwareGoal
+    - MinTopicLeadersPerBrokerGoal
+    - NetworkOutboundCapacityGoal
+    - ReplicaCapacityGoal
+  skipHardGoalCheck: true
+  # ... other rebalancing related configuration
+```
+
+The user has the flexibility to specify the [goals and the other rebalancing related configurations](https://strimzi.io/docs/operators/latest/configuring.html#type-KafkaRebalance-reference) the same way they already do today on a `KafkaRebalance` custom resource.
+This allows the user to not rely on the default Cruise Control goals configuration which would be common for any rebalance without overriding goals.
+They could define different goals based on the status of the Kafka cluster or skip hard goals in some scenarios for examples.
+As already mentioned, specifying goals but more in general customizing the optimization proposal request is already possible in the current usage of a `KafkaRebalance` custom resource so it should be allowed for the auto-rebalancing as well.
+In the end, the cluster operator will use that configuration from the `KafkaRebalanceTemplate` to fill the corresponding `KafkaRebalance` but adding the `spec.mode` and `spec.brokers` accordingly.
+It will also add the `spec.autoApproval` field, as described by this [proposal](https://github.com/strimzi/proposals/pull/56), for enabling auto-approval.
+
+We would also need to extend the `Kafka` custom resource to allow the user specifying that he wants auto-rebalancing on scaling by using a template.
+A new `spec.cruiseControl.autoRebalance` field could be added for this purpose.
+Because we are talking about auto-rebalancing, the corresponding field could also be not limited to the rebalancing on scaling but a more general full "continuous” auto-rebalancing to enable on the cluster.
+
+The `spec.cruiseControl.autoRebalance` field allows to specify a list of auto-rebalancing modes with the corresponding templates to be used.
+To specify when running the auto-rebalancing, a `spec.cruiseControl.autoRebalance.mode` field could be used to specify if running auto-rebalancing only on adding or removing brokers, or a full rebalance is requested.
+Having a full auto-rebalancing is out of scope for this proposal, the field is added for enabling a future implementation.
+
+The template to be used is referenced by a new mandatory `spec.cruiseControl.autoRebalance.template` field, as a [LocalObjectReference](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/local-object-reference/), and the cluster operator uses that template to create a corresponding `KafkaRebalance` custom resource.
+If the referenced `KafkaRebalanceTemplate` custom resource doesn't exist, the cluster operator will fail the reconcile.
+
+In the following example, a user can decide to have different rebalancing templates when adding or removing brokers and when they want a full rebalancing during the normal cluster execution.
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    replicas: 3
+    # ...
+  cruiseControl:
+    # ...
+    autoRebalance:
+      # using the same rebalancing template when adding or removing brokers
+      - mode: add-brokers
+        template: 
+          name: my-add-remove-brokers-rebalancing-template
+      - mode: remove-brokers
+        template: 
+          name: my-add-remove-brokers-rebalancing-template  
+      # using a different template for a full rebalancing  
+      - mode: full
+        template: 
+          name: my-full-rebalancing-template
+```
+
+The user can even decide to not have full auto-rebalancing and, for example, using different templates when adding or removing brokers.
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    replicas: 3
+    # ...
+  cruiseControl:
+    # ...
+    autoRebalance:
+      # using a specific template when adding brokers
+      - mode: add-brokers
+        template: 
+          name: my-add-brokers-rebalancing-template
+      # using a different template when removing brokers  
+      - mode: remove-brokers
+        template: 
+          name: my-remove-brokers-rebalancing-template
+      # NOTE: no full auto-rebalancing   
+```
+
+The user could even decide to have auto-rebalancing only when adding or removing brokers but not for both.
+
+This way provides the greatest flexibility to describe in which case to run the auto-rebalancing and with which configuration.
+
+### Auto-rebalancing on scaling up
+
+Following the operator flow when auto-rebalancing on scaling up.
+
+1. The user scales up the cluster by increasing the `spec.kafka.replicas` field.
+2. The `KafkaReconciler` takes care of it by scaling up the cluster as usual, by adding additional brokers.
+3. If the `spec.cruiseControl.autoRebalance` contains an entry with `add-brokers` mode, the `KafkaReconciler` creates a `KafkaRebalance` custom resource named as `<cluster-name>-auto-rebalance-add-brokers` by using the corresponding template and sets:
+   * the `spec.mode` as `add-brokers` and the `spec.brokers` with the list of brokers already added to the cluster.
+   * the `spec.autoApproval` as described by this [proposal](https://github.com/strimzi/proposals/pull/56).
+4. The `KafkaRebalanceAssemblyOperator` detects the new rebalancing resource and acts as usual by requesting Cruise Control to generate an optimization proposal via the REST API on the `/add_broker` endpoint.
+5. When the optimization proposal is ready, the `KafkaRebalanceAssemblyOperator` approves it automatically, because of the `spec.autoApproval` field, by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation as defined by this [proposal](https://github.com/strimzi/proposals/pull/56).
+6. If/When the rebalancing ends successfully, the corresponding `KafkaRebalance` custom resource is deleted by the `KafkaReconciler`.
+
+#### Error handling
+
+During the above flow, errors can happen after the scaling up:
+
+* There is an error when getting the proposal.
+* After auto-approving the proposal, there is an error when running the rebalance.
+
+In both scenarios, the reconciliation ends with the `KafkaRebalance` custom resource in the `NotReady` state as per the `KafkaRebalanceAssemblyOperator` states machine.
+The cluster operator logs a warning about the rebalancing not happened.
+It's the user who should deal with this scenario:
+
+* Checking the reason why the `KafkaRebalance` is in the `NotReady` state and deleting the custom resource.
+* If the user still wants rebalancing, they have to create a new `KafkaRebalance` custom resource manually.
+
+### Auto-rebalancing on scaling down
+
+1. The user scales down the cluster by decreasing the `spec.kafka.replicas` field.
+2. If the `spec.cruiseControl.autoRebalance` contains an entry with `remove-brokers` mode, the `KafkaReconciler` creates a `KafkaRebalance` custom resource named as `<cluster-name>-auto-rebalance-remove-brokers` by using the corresponding template and sets:
+   * the `spec.mode` as `remove-brokers` and the `spec.brokers` with the list of brokers that will be removed.
+   * the `spec.autoApproval` as described by this [proposal](https://github.com/strimzi/proposals/pull/56).
+3. The `KafkaRebalanceAssemblyOperator` detects the new rebalancing resource and acts as usual by requesting Cruise Control to generate an optimization proposal via the REST API on the `/remove_broker` endpoint.
+4. When the optimization proposal is ready, the `KafkaRebalanceAssemblyOperator` approves it automatically, because of the `spec.autoApproval` field, by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation as defined by this [proposal](https://github.com/strimzi/proposals/pull/56).
+5. If/When the rebalancing ends successfully, the corresponding `KafkaRebalance` custom resource is deleted by the `KafkaReconciler`.
+6. Finally, the `KafkaReconciler` can continue to scale down the cluster by deleting the brokers.
+
+#### Error handling
+
+During the above flow, errors can happen before the scaling down:
+
+* There is an error when getting the proposal.
+* After auto-approving the proposal, there is an error when running the rebalance.
+
+In both scenarios, the reconciliation ends with the `KafkaRebalance` custom resource in the `NotReady` state as per the `KafkaRebalanceAssemblyOperator` states machine.
+The cluster operator logs a warning about the rebalancing not happened.
+
+The reconcile ends but the scale down cannot proceed because of the brokers to be removed are still hosting some partition replicas.
+This behavior is related to this [proposal](https://github.com/strimzi/proposals/pull/54).
+
+On the next reconciliation, the cluster operator will try to do a scale down again, so running the flow as before.
+The previous `KafkaRebalance` custom resource in the `NotReady` state is overridden to generate a new optimization proposal.
+
+Of course, the user could change his mind by reverting back the `spec.kafka.replicas` field to avoid the scale down.
+
+## Affected/not affected projects
+
+This proposal impacts the Strimzi Cluster Operator only.
+The `KafkaReconciler` is impacted on handling the `KafkaRebalanceTemplate` custom resource and generating/deleting the corresponding `KafkaRebalance`.
+The `KafkaRebalanceAssemblyOperator` is impacted mostly on the auto-approval side but it is already defined by this [proposal](https://github.com/strimzi/proposals/pull/56).
+It handles the `KafkaRebalance` custom resource the same way as it was created by the user but it has to annotate the custom resource automatically.
+
+## Compatibility
+
+The auto-rebalancing on scaling up and down is optional.
+The user has to specify the `spec.cruiseControl.autoRebalance` on the `Kafka` custom resource for the desire of having the operator doing the rebalancing automatically.
+Without that addition, the rebalancing can be done with the manual two step process as today.
+
+## Rejected alternatives
+
+When the `spec.cruiseControl.autoRebalance` is specified, the cluster operator can create an “empty” `KafkaRebalance` custom resource automatically.
+By “empty” custom resource we mean that the default Cruise Control configuration will be used for goals and all other options.
+The only fields that the cluster operator has to specify, when creating the `KafkaRebalance` custom resource, based on scaling up or down, are the:
+
+* `spec.mode` : should be `add-brokers` or `remove-brokers` if scaling up or down.
+* `spec.brokers` : with the list of brokers to add/remove.
+* `spec.autoApproval` : for allowing the auto-approval as defined by this [proposal](https://github.com/strimzi/proposals/pull/56).
+
+The drawback of this solution is that if the user wants to customize the rebalancing, he has to use the `spec.cruiseControl.config` field in the Kafka custom resource. It’s important to mention that it affects the overall Cruise Control configuration so that any other “empty” rebalancing request, not related to scaling, will use the same options.
+Users won't also be able to skip hard goals to move forward with a rebalancing if they are not met anyway.
+
+To allow customization, a potential solution would be to add rebalancing related configuration (goals, and so on) under the `spec.cruiseControl.autoRebalance` directly.
+It would be a kind of copy of the fields from the `KafkaRebalance` custom resource excluding the mode and brokers.
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    replicas: 3
+    # ...
+  cruiseControl:
+    # ...  
+    autoRebalance:
+      # using the same rebalancing configuration when adding or removing brokers ... a lot of duplication
+      - mode: add-brokers
+        goals:
+          - CpuCapacityGoal
+          - NetworkInboundCapacityGoal
+          - DiskCapacityGoal
+          - RackAwareGoal
+          - MinTopicLeadersPerBrokerGoal
+          - NetworkOutboundCapacityGoal
+          - ReplicaCapacityGoal
+        skipHardGoalCheck: true
+        # ... other rebalancing related configuration
+      - mode: remove-brokers
+        goals:
+          - CpuCapacityGoal
+          - NetworkInboundCapacityGoal
+          - DiskCapacityGoal
+          - RackAwareGoal
+          - MinTopicLeadersPerBrokerGoal
+          - NetworkOutboundCapacityGoal
+          - ReplicaCapacityGoal
+        skipHardGoalCheck: true
+        # ... other rebalancing related configuration  
+      - mode: full
+        goals:
+          - CpuCapacityGoal
+          - NetworkInboundCapacityGoal
+          - DiskCapacityGoal
+        skipHardGoalCheck: false
+        # ... other rebalancing related configuration  
+```
+
+For handling the auto-approval, a `spec.autoApproval` field, as described by this [proposal](https://github.com/strimzi/proposals/pull/56), would be needed as well.
+This proposal was rejected because it pollutes the `Kafka` custom resource with too many details and configuration related to a rebalancing operation when the user wants to specify specific desires compared to the default ones in the Cruise Control configuration.
+To understand how big the configuration for each rebalancing mode could be, [here](https://strimzi.io/docs/operators/latest/configuring.html#type-KafkaRebalance-reference) the options that we support today which could anyway increase in the future.
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ This repository list of proposals for the Strimzi project. A template for new pr
 
 |  #  | Title                                                                 |
 | :-: |:----------------------------------------------------------------------|
+| 39  | [Rebalancing on cluster scaling](./039-auto-rebalance-cluster-scaling.md) |
 | 38  | [Auto-approval mechanism for optimization proposals](./038-optimization-proposal-autoapproval.md) |
 | 37  | [Pluggable Pod Security Profiles](./037-pluggable-pod-security-profiles.md) |
 | 36  | [KRaft support: ZooKeeper-less Kafka](./036-kraft-mode.md) |


037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+    autoRebalance:
+      mode: [running, scaling] # or just one of the two
+```
+
+Finally, the user should be able to define what are the [goals and the other rebalancing related configurations](https://strimzi.io/docs/operators/latest/configuring.html#type-KafkaRebalance-reference) which are normally described via a `KafkaRebalance` custom resource.
+In any case, for rebalancing the cluster, an actual `KafkaRebalance` custom resource has to be automatically created by the cluster operator.
+
+### New `KafkaRebalanceTemplate` custom resource
+
+A new `KafkaRebalanceTemplate` custom resource could be a solution allowing the user to define the rebalancing configuration he wants but leaving to the operator the creation of the corresponding `KafkaRebalance` custom resource.
```
tombentley, line 50:
```suggestion
A new `KafkaRebalanceTemplate` custom resource could be a solution allowing the user to define the rebalancing configuration they want but leaving to the operator the creation of the corresponding `KafkaRebalance` custom resource.
```

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+    - DiskCapacityGoal
+    - RackAwareGoal
+    - MinTopicLeadersPerBrokerGoal
+    - NetworkOutboundCapacityGoal
+    - ReplicaCapacityGoal
+  skipHardGoalCheck: true
+  # ... other rebalancing related configuration
+```
+
+The user has the flexibility to specify goals as well as all the other options like he already does today on a `KafkaRebalance` custom resource.
```
tombentley, line 76:
```suggestion
The user has the flexibility to specify goals as well as all the other options like they already do today on a `KafkaRebalance` custom resource.
```

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+The cluster operator will use that configuration for the `KafkaRebalance` but adding the `spec.mode` and `spec.brokers` accordingly.
+
+It would also contain a `spec.autoApproval` field, as described by this [proposal](https://github.com/strimzi/proposals/pull/56), to specify if auto-approval is desired or not when auto-rebalancing is enabled.
+The template to be used is referenced by a new `spec.kafka.autoRebalance.template` field and the cluster operator uses that template to create a corresponding `KafkaRebalance` custom resource.
+If the reference is missing, the cluster operator can create an “empty” `KafkaRebalance` custom resource (which means the optimization proposal will be based on Cruise Control default configuration). It will behave as the rejected alternative.
```
tombentley, line 81: Do we really want to default to CC's default configuration, or would it be better to fail the reconciliation? Consider that it might be deleted by mistake, so you'd pick up the default behaviour by accidentally some future time.

tombentley: Also IDK what "It will behave as the rejected alternative." means.

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    replicas: 3
+    # ...
+    autoRebalance:
+      mode: [running, scaling] # or just one of the two
+      template: # reference to a KafkaRebalanceTemplate custom resource
```
tombentley, line 94: Meta point: When reading this the first time it would have been easier to either present this `template` at the same time you talk about the `mode`, or introduce the `KafkaRebalanceTemplate` resource first, and then explain how it relates to the `Kafka` resource.

Proposal points:
* Is it necessarily the case that users would always want to use the same template in both cases?
* For `running` what would that mean? would there some kind of timer? So would `periodically` be a better term? Also, any way to schedule this using the existing `maintenanceTimes` facility?

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+2. The cluster operator takes care of it by scaling up the cluster as usual, by adding additional brokers.
+3. If `spec.kafka.autoRebalance.mode` contains the `scaling` value, the cluster operator creates a `KafkaRebalance` custom resource.
+   * It will specify `spec.mode` as `add-brokers` and `spec.brokers` with the list of brokers already added to the cluster.
+4. The cluster operator asks Cruise Control to generate an optimization proposal via the REST API.
+5. When the optimization proposal is ready, the cluster operator checks if the user has specified auto-approval or not:
+   * If auto-approval is not specified, the user has to approve the proposal manually as usual.
+   * If auto-approval and criteria are met or they are not specified at all, the cluster operator approves the proposal automatically by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation.
+   * If auto-approval and criteria are not met, the `KafkaRebalance` custom resource status will reflect, by using a warning condition, that the proposal was not approved automatically and the reason for that.
```
tombentley, line 109: Can you explain what happens to the `Kafka` resource status? I assume it's still in some scaleup pending state?

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+2. If `spec.kafka.autoRebalance.mode` contains the `scaling` value, the cluster operator creates a `KafkaRebalance` custom resource.
+   * It will specify `spec.mode` as `remove-brokers` and `spec.brokers` with the list of brokers that will be removed.
+3. The cluster operator asks Cruise Control to generate an optimization proposal via the REST API.
+4. When the optimization proposal is ready, the cluster operator checks if the user has specified auto-approval or not:
+   * If auto-approval is not specified, the user has to approve the proposal manually as usual.
+   * If auto-approval and criteria are met or they are not specified at all, the cluster operator approves the proposal automatically by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation.
+   * If auto-approval and criteria are not met, the `KafkaRebalance` custom resource status will reflect, by using a warning condition, that the proposal was not approved automatically and the reason for that.
+5. Only when the proposal is approved and rebalancing is complete, the cluster operator can continue to scale down the cluster by deleting the brokers.
```
tombentley, line 124: Similar points to the scale up case

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    replicas: 3
+    # ...
+    autoRebalance:
+      mode: [running, scaling] # or just one of the two
```
tombentley, line 42: * Does `running` imply `scaling`, or is this really more of an enum set?

tombentley (commented):
Thanks @ppatierno left a few comments, and questions, but that's a great start.

037-auto-rebalance-cluster-scaling.md
scholzj, line 109: How will you maintain the warning without some new state store?

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+   * If auto-approval and criteria are met or they are not specified at all, the cluster operator approves the proposal automatically by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation.
+   * If auto-approval and criteria are not met, the `KafkaRebalance` custom resource status will reflect, by using a warning condition, that the proposal was not approved automatically and the reason for that.
+5. Only when the proposal is approved and rebalancing is complete, the cluster operator can continue to scale down the cluster by deleting the brokers.
+
+In this use case, the actual scaling down (deleting the brokers) is blocked on the proposal to be approved and the rebalancing done.
+To fix this kind of issue, the user should change the rebalancing configuration and allow the cluster operator to reconcile again for getting a new optimization proposal, running a rebalance and finally unlocking the scaling down.
+
+## Affected/not affected projects
+
+This proposal impacts the Strimzi Cluster Operator only and mostly the `KafkaRebalanceAssemblyOperator`.
```
scholzj, line 131: Can you elaborate? I would say it mostly impacts KafkaReconciler -> I would expect KafkaRebalanceAssemblyOperator based on this proposal to just crunch the `KafkaRebalance` resources without any knowledge about the actor which created them -> user manually or CO during an automated operation.

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+      goals:
+        - CpuCapacityGoal
+        - NetworkInboundCapacityGoal
+        - DiskCapacityGoal
+        - RackAwareGoal
+        - MinTopicLeadersPerBrokerGoal
+        - NetworkOutboundCapacityGoal
+        - ReplicaCapacityGoal
+      skipHardGoalCheck: true
+      # ... other rebalancing related configuration
```
scholzj, line 173: What else would be here? I think it is good to clarify this into more detail to understand how big it would possibly be. It is important part to decide if `KAfkaRebalanceTemplate` is better than this or not (i.e. to balance CRD size with usability and readability).

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+This feature is about enabling the cluster operator to get an optimization proposal (to be auto or manually approved) and run the corresponding rebalancing when the user scales the cluster up or down by changing the `spec.kafka.replicas` field on the `Kafka` custom resource.
+The scaling operation is then subordinated to the rebalancing operation and if it is approved and runs successfully.
+
+## Current situation
+
+Currently, there is no automatic rebalancing when the user scales the cluster up or down.
+It means that:
+
+* after scaling up the cluster, the user has to create a `KafkaRebalance` custom resource with the `spec.mode: add-brokers` and the `spec.brokers` with the added brokers, in order to rebalance the cluster and moving some replicas to the newly added brokers.
+* before scaling down the cluster, the user has to create a `KafkaRebalance` custom resource with the `spec.mode: remove-brokers` and the `spec.brokers` with the brokers to remove, in order to rebalance the cluster and moving replicas out of the brokers to be removed.
```
tomncooper, line 12:
```suggestion
* before scaling down the cluster, in order to rebalance the cluster and move replicas off of the brokers to be removed, the user has to create a `KafkaRebalance` custom resource using the `spec.mode: remove-brokers` mode and add the broker IDs of the brokers to be removed to the `spec.brokers` field.
```

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Motivation
+
+The rebalancing operation during scaling up and down is a manual two step process.
+
+* scaling up the cluster, then running a rebalance.
+* running a rebalance, then scaling down the cluster.
```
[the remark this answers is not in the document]
tomncooper: Good point, we _should_ be able to handle that.

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+
+In order to simplify, it would be great for the user to deal with just scaling the cluster by changing the `spec.kafka.replicas` field and then having the operator taking care to run a rebalancing operation automatically if requested.
+
+## Proposal
+
+We would need to extend the `Kafka` custom resource to allow the user specifying that he wants auto-rebalancing on scaling.
```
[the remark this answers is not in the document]
tomncooper: I agree, I think an annotation would be more flexible at this stage.

037-auto-rebalance-cluster-scaling.md
ppatierno: I wrote "... the user ... " so I used "... he wants ..." or do you want me to change using plural "users"?

037-auto-rebalance-cluster-scaling.md
tomncooper: See my comment above, something like `addBroker, removeBroker, clusterRebalance, all` might be better. Maybe we don't want an `all`? It might be better to have the user explicitly state each mode?

037-auto-rebalance-cluster-scaling.md
ppatierno: Got it offline.

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+    - NetworkOutboundCapacityGoal
+    - ReplicaCapacityGoal
+  skipHardGoalCheck: true
+  # ... other rebalancing related configuration
+```
+
+The user has the flexibility to specify goals as well as all the other options like he already does today on a `KafkaRebalance` custom resource.
+The cluster operator will use that configuration for the `KafkaRebalance` but adding the `spec.mode` and `spec.brokers` accordingly.
+
+It would also contain a `spec.autoApproval` field, as described by this [proposal](https://github.com/strimzi/proposals/pull/56), to specify if auto-approval is desired or not when auto-rebalancing is enabled.
```
tomncooper, line 79: So really, when you say "auto-rebalancing" you are actually saying "auto-proposal-generation + auto-approval"? If a user doesn't set `autoapproval` to true but has auto-rebalacing enabled then the KafkaRebalance CRs will be created and sit there waiting for approval?

I wonder if we should call that behaviour out explicitly by not calling the behaviour in this propsoal "auto-rebalancing" and instead calling it `autoProposalGeneration` or something similar? This would better match today's UX where everything is approved by the user. 

Making this proposal about auto-proposal-generation would be cleaner in my opinion. This would also fit better with the auto-approval proposal. They can be implemented separately but when both are implemented we will then have "auto-rebalancing".

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
+# Automatic rebalance on cluster scaling
+
+This feature is about enabling the cluster operator to get an optimization proposal (to be auto or manually approved) and run the corresponding rebalancing when the user scales the cluster up or down by changing the `spec.kafka.replicas` field on the `Kafka` custom resource.
+The scaling operation is then subordinated to the rebalancing operation and if it is approved and runs successfully.
```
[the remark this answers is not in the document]
ppatierno: By "scaling operation" I mean the entire process that happens across more reconciliations and pods rolling.

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+
+In order to simplify, it would be great for the user to deal with just scaling the cluster by changing the `spec.kafka.replicas` field and then having the operator taking care to run a rebalancing operation automatically if requested.
+
+## Proposal
+
+We would need to extend the `Kafka` custom resource to allow the user specifying that he wants auto-rebalancing on scaling.
+
+A new `spec.kafka.autoRebalance` field could be added for this purpose.
+Because we are talking about auto-rebalancing, the corresponding field could also be not limited to the rebalancing on scaling but a more general “continuous” auto-rebalancing to enable on the cluster. 
+To specify when running the auto-rebalancing, a `spec.kafka.autoRebalance.mode` field could be used by listing if running auto-rebalancing only on scaling, during the normal execution or both.
```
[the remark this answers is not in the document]
tomncooper: I think we don't know how users would want to use this so we should go for more flexibility not less.

037-auto-rebalance-cluster-scaling.md
ppatierno: > Do we really want to default to CC's default configuration, or would it be better to fail the reconciliation? Consider that it might be deleted by mistake, so you'd pick up the default behaviour by accidentally some future time.

I thought that the user could have some "good to go" configuration as defaults in Cruise Control to avoid filling the template each time. Of course, making the template mandatory put some guardrails on how to use this feature. I have no strong opinion against it. The `template` field would be mandatory and the operator has to check that the corresponding resource exists or fails the reconcile.

> Also IDK what "It will behave as the rejected alternative." means.

The rejected alternative works the same when the `autoRebalance` doesn't contain any rebalancing options so it creates an empty `KafkaRebalance` (which is the same when the template is not specified in the proposed solution).

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
scholzj: Yes, correct. This is something what will be possible in the future (where scaling-up / down stops really being the right term ... it is more about adding new or removing existing brokers.)

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+
+Auto-rebalancing on scaling up.
+
+1. The user scales up the cluster by increasing the `spec.kafka.replicas` field.
+2. The cluster operator takes care of it by scaling up the cluster as usual, by adding additional brokers.
+3. If `spec.kafka.autoRebalance.mode` contains the `scaling` value, the cluster operator creates a `KafkaRebalance` custom resource.
+   * It will specify `spec.mode` as `add-brokers` and `spec.brokers` with the list of brokers already added to the cluster.
+4. The cluster operator asks Cruise Control to generate an optimization proposal via the REST API.
+5. When the optimization proposal is ready, the cluster operator checks if the user has specified auto-approval or not:
+   * If auto-approval is not specified, the user has to approve the proposal manually as usual.
```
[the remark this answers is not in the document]
scholzj: I think that means a lot of effort for development and maintenance and a lot of complications in the code and very little added value you get out of it. The manual part is available today as it is. Why not focus on actual improvements?

037-auto-rebalance-cluster-scaling.md
ppatierno: >  or introduce the KafkaRebalanceTemplate resource first, and then explain how it relates to the Kafka resource.

isn't this the flow that we have now? I introduced the `KafkaRebalanceTemplate` first and then I talked about the `template` field in the `Kafka` CR to reference the actual template.

> Is it necessarily the case that users would always want to use the same template in both cases?

Yeah, I thought about this but was not sure if the user had really this need. Maybe yes, like skipping some hard goal when scaling but not during normal operations, or having a different replication throttle (see [here](https://strimzi.io/docs/operators/latest/configuring.html#type-KafkaRebalanceSpec-reference) what we can set today as options for a rebalance).
At this point, `autoRebalance` would be more an array of modes having each one its own template.

```yaml
autoRebalance:
      - mode: scaling
        template: # reference to a KafkaRebalanceTemplate custom resource
      - mode: running
        template: # reference to a KafkaRebalanceTemplate custom resource
```

> For running what would that mean? would there some kind of timer? So would periodically be a better term?

Yeah it's about running a rebalancing overtime. I think there are other suggestions about this field from @tomncooper in another comment.

> Also, any way to schedule this using the existing maintenanceTimes facility?

IIRC the maintenance time feature is not about scheduling but it's about checking that you can do an operation at a specific time. For example, I want to do a Kafka upgrade but only if the window matches. Anyway you have to run the operation there is no scheduler for that.

037-auto-rebalance-cluster-scaling.md
ppatierno: @tomncooper see the other comment where I tagged you about having more templates (one for each mode). Not sure we want different templates for scaling up and down. Maybe between scaling and running. Btw it would mean changing the snippet for mode in the following way.

```yaml
autoRebalance:
      - mode: [addBroker, removeBroker]
        template: # reference to a KafkaRebalanceTemplate custom resource
      - mode: [clusterRebalance]
        template: # reference to a KafkaRebalanceTemplate custom resource
```

Making the `mode` field as an array.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
ppatierno: I could agree but what about sticking with the modes we already have today like `add-brokers`, `remove-brokers`, `full` ?

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+### Flow
+
+Auto-rebalancing on scaling up.
+
+1. The user scales up the cluster by increasing the `spec.kafka.replicas` field.
+2. The cluster operator takes care of it by scaling up the cluster as usual, by adding additional brokers.
+3. If `spec.kafka.autoRebalance.mode` contains the `scaling` value, the cluster operator creates a `KafkaRebalance` custom resource.
+   * It will specify `spec.mode` as `add-brokers` and `spec.brokers` with the list of brokers already added to the cluster.
+4. The cluster operator asks Cruise Control to generate an optimization proposal via the REST API.
+5. When the optimization proposal is ready, the cluster operator checks if the user has specified auto-approval or not:
```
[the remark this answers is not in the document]
ppatierno: The `KafkaRebalanceAssemblyOperator` has its own internal polling (to the CC REST API) for checking that a proposal is ready. It doesn't do that on each reconcile because it would mean to know that a proposal is ready after looooong time (by default 2 minutes).

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
ppatierno: "by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation." as the cluster operator does automatically if needed. I added that sentence.

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+In order to simplify, it would be great for the user to deal with just scaling the cluster by changing the `spec.kafka.replicas` field and then having the operator taking care to run a rebalancing operation automatically if requested.
+
+## Proposal
+
+We would need to extend the `Kafka` custom resource to allow the user specifying that he wants auto-rebalancing on scaling.
+
+A new `spec.kafka.autoRebalance` field could be added for this purpose.
+Because we are talking about auto-rebalancing, the corresponding field could also be not limited to the rebalancing on scaling but a more general “continuous” auto-rebalancing to enable on the cluster. 
+To specify when running the auto-rebalancing, a `spec.kafka.autoRebalance.mode` field could be used by listing if running auto-rebalancing only on scaling, during the normal execution or both.
+Having “continuous” auto-rebalancing is out of scope for this proposal, the field is added for allowing a future implementation.
```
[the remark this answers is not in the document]
ppatierno: Yeah this align with suggestions from @tomncooper and @tombentley in other comments. I think it's a good idea at this point. I will re-arrange the `mode` to reflect the possibility for such a separation.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
ppatierno: This aligns with the idea of having more "modes" and corresponding templates because at a CC REST API level it will mean calling different endpoints (`/add_broker` and/or `/remove_broker`) with a different list of brokers (to add and/or remove).

037-auto-rebalance-cluster-scaling.md
ppatierno: > or introduce the KafkaRebalanceTemplate resource first, and then explain how it relates to the Kafka resource.

I have got what you meant, just moving paragraphs around ;-)

ppatierno: As already mentioned in a different comment, I would use `mode` as an list and `add-brokers`, `remove-brokers` and `full` as possible values (the same that we use today in the `KafkaRebalance` resource).

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+spec:
+  kafka:
+    replicas: 3
+    # ...
+    autoRebalance:
+      mode: [running, scaling] # or just one of the two
+```
+
+Finally, the user should be able to define what are the [goals and the other rebalancing related configurations](https://strimzi.io/docs/operators/latest/configuring.html#type-KafkaRebalance-reference) which are normally described via a `KafkaRebalance` custom resource.
+In any case, for rebalancing the cluster, an actual `KafkaRebalance` custom resource has to be automatically created by the cluster operator.
```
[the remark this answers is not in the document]
ppatierno: I have updated the proposal in this direction by using different modes of auto-rebalancing (with different corresponding templates) and talking more in general about add and remove brokers instead of scaling up and down.

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+
+### New `KafkaRebalanceTemplate` custom resource
```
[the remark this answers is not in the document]
ppatierno: Any different solution you have in mind?

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
ppatierno: Isn't it a more technical implementation detail?

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
scholzj: As I commented on other place, I think we need to understand what the configuration will be to judge whether it is better to have it in the `Kafka` CR or not. If we don't want to have it there then I'm not sure I have a better solution. I think it also still deserves more explanation why the default CC goals should not be used.

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+4. The cluster operator asks Cruise Control to generate an optimization proposal via the REST API.
+5. When the optimization proposal is ready, the cluster operator checks if the user has specified auto-approval or not:
+   * If auto-approval is not specified, the user has to approve the proposal manually as usual.
+   * If auto-approval and criteria are met or they are not specified at all, the cluster operator approves the proposal automatically by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation.
+   * If auto-approval and criteria are not met, the `KafkaRebalance` custom resource status will reflect, by using a warning condition, that the proposal was not approved automatically and the reason for that.
+
+It could happen that the cluster was scaling up but rebalancing didn’t happen because the optimization proposal was not approved.
+To fix this kind of issue, the user should change the rebalancing configuration and allow the cluster operator to reconcile again (but not scaling up because it was already done) for getting a new optimization proposal and running a rebalance.
```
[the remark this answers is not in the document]
ppatierno: After scaling up but rebalancing in error, the corresponding `KafkaRebalance` would be in `NotReady` state.
On each (next) reconcile, the operator could go through the `spec.kafka.autoRebalance` entry and check if there is any "pending" corresponding `KafkaRebalance` to refresh (when in `NotReady`, you can make changes and add `refresh` annotation for getting a new proposal). Of course, it would mean user changing the rebalancing template to fix the issue.

More problematic could be if the scaling up goes ok and operator crashes before creating the `KafkaRebalance` from the template. On next reconcile, the operator doesn't know that before it ran a scaling up and now it needs to run a related rebalancing. We would need some `autoRebalance` related section in the `status`?

> but you don't want to keep track of it forever in case the user just doesn't fix it.

At this point should the user just remove the `autoRebalance` related rebalancing entry to avoid the operator to reconcile?

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+
+Auto-rebalancing on scaling down.
+
+1. The user scales down the cluster by decreasing the `spec.kafka.replicas` field.
+2. If `spec.kafka.autoRebalance.mode` contains the `scaling` value, the cluster operator creates a `KafkaRebalance` custom resource.
+   * It will specify `spec.mode` as `remove-brokers` and `spec.brokers` with the list of brokers that will be removed.
+3. The cluster operator asks Cruise Control to generate an optimization proposal via the REST API.
+4. When the optimization proposal is ready, the cluster operator checks if the user has specified auto-approval or not:
+   * If auto-approval is not specified, the user has to approve the proposal manually as usual.
```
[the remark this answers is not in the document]
ppatierno: > What happens with the scale-down in such case?

I didn't get the question, sorry :-( What do you mean?

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
scholzj: > After scaling up but rebalancing in error, the corresponding KafkaRebalance would be in NotReady state.
On each (next) reconcile, the operator could go through the spec.kafka.autoRebalance entry and check if there is any "pending" corresponding KafkaRebalance to refresh (when in NotReady, you can make changes and add refresh annotation for getting a new proposal). Of course, it would mean user changing the rebalancing template to fix the issue.

That sounds like a possibly never ending story. I think you want to avoid it and have a clear end of the process for all possible paths.

037-auto-rebalance-cluster-scaling.md
ppatierno: The `KafkaRebalance` moves in a `NotReady` state when there is a problem and a warning condition in the status reflects the problem. This is how it already works today.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
scholzj: What happens with the scale-down and the reconciliation while you are waiting for the user approval.

037-auto-rebalance-cluster-scaling.md
ppatierno: I added a link to the Strimzi documentation with the current options we allow to configure today but with possibility they will increase in the future.

037-auto-rebalance-cluster-scaling.md
scholzj: Today, this has nothing to do with KafkaAssemblyOperator and KafkaReconciler. So again, you need to explain it in the context of that. What happens to the reconciliation? How is it handled in further reconciliations etc.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
ppatierno: I put in a couple of places (on the proposal and rejected alternative) the link to the current Strimzi documentation about the configuration parameters we allow for optimization proposal. In any case we are not supporting all of them so the list could even increase in the future. See https://strimzi.io/docs/operators/latest/configuring.html#type-KafkaRebalance-reference

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
ppatierno: If I got it correctly and IIRC ... the `KafkaConnect` used to have a `strimzi.io/use-connector-resources: "true"` annotation to enable the usage of `KafkaConnector` CRs. In this case the API is more complex, it's not just about enabling the auto-rebalancing but specifying modes and corresponding templates.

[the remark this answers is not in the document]
scholzj: Well, you can stick everything into an annotation as a JSON for example. It might help to see what is the best way going forward ... a separate resource, fields inside the Kafka resource etc.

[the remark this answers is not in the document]
scholzj: But in general, I might have not added here if I read the whole proposal first ... this part talks only about the mode basically, which would be much easier. But if you get back to the discussions about node pools, there might be a lot of other changes to the Kafka CRD and maybe long term it might fit also somewhere else. E.g. should the whole CC staff move to a separate CRD which would give it more space for this kind of configuration?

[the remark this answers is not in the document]
scholzj: I know. Lot of questions and not too many answers :-D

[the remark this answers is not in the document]
ppatierno: That was my fear, a big single JSON annotation.
Based on my latest refactoring of the auto-rebalancing configuration, it should be an `autoRebalance` annotation containing the array modes with corresponding templates. 

```JSON
autoApproval: "[
   {
      "mode":[
         "add-brokers",
         "remove-brokers"
      ],
      "template":"my-add-brokers-rebalancing-template"
   },
   {
      "mode":[
         "full"
      ],
      "template":"my-full-rebalancing-template"
   }
]"
```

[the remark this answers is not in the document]
scholzj: Do you really want to have the mode as an array? Validation and error handling will be harder IMHO.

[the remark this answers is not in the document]
ppatierno: Don't cross discussions please. This is about putting in a potential annotation, can you make the same comment on the proposal directly?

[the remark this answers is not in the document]
scholzj: No, I commented here, now you know about it. You can incorporate it or not depending on what you think about it.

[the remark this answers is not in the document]
ppatierno: Regardless of where to comment, if was a good idea so that `mode` could be an enum and we can have validation at Kubernetes API level. I made the change in the proposal.
Let's see here if we reach an overall agreement to use an annotation instead of CRD changes since the beginning.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
ppatierno: After more thinking, having the user involved in the auto-rebalancing does really make the process more complex than expected without taking any additional value. I will assume that the proposal is always auto approved.

ppatierno:
After reviewing the comments, I updated the proposal with the following main changes:

* I removed the possibility to have the user being involved with the approval of the auto-generated rebalancing resource. I would agree with Jakub that for any kind of manual intervention we already have the flow. With this proposal we want to have some automation, so always autoa-approval. Leaving the user the possibility to approve would make the flow pretty complex.
* I added more details on the cluster operator components that are involved in the flow and even the error handling.

@scholzj @tombentley @tomncooper could you have another pass please?

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,251 @@
...
+
+In order to simplify the process, it would be great for the user to deal with just scaling the cluster by changing the `spec.kafka.replicas` field and then having the operator taking care of running a rebalancing operation automatically if requested.
+
+Taking into account the `StrimziPodSets` usage (instead of the `StatefulSets`), which will be the default in the future, a more general use case will be about adding new brokers and/or removing existing ones, not talking just about scaling up and down anymore.
+
+This proposal takes into account this future plan by referring to more general operations like add and remove brokers instead of scaling up and down.
+
+## Proposal
+
+A new `KafkaRebalanceTemplate` custom resource could be a solution allowing the user to define the configuration for the auto-rebalancing but leaving to the operator the creation of the corresponding `KafkaRebalance` custom resource.
```
scholzj, line 31: Have you thought about using a regular `KafkaRebalance` as the template? I'm still a bit concerned by the efforts related to yet another CRD.

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,251 @@
...
+    - MinTopicLeadersPerBrokerGoal
+    - NetworkOutboundCapacityGoal
+    - ReplicaCapacityGoal
+  skipHardGoalCheck: true
+  # ... other rebalancing related configuration
+```
+
+The user has the flexibility to specify the [goals and the other rebalancing related configurations](https://strimzi.io/docs/operators/latest/configuring.html#type-KafkaRebalance-reference) the same way they already do today on a `KafkaRebalance` custom resource.
+The cluster operator will use that configuration to fill the corresponding `KafkaRebalance` but adding the `spec.mode` and `spec.brokers` accordingly.
+It would also contain a `spec.autoApproval` field, as described by this [proposal](https://github.com/strimzi/proposals/pull/56).
```
scholzj, line 60: Is this still needed if the manual approval option is gone now? I would assume that for scale-up / scale-down whatever proposal you get based on the goals, it should be just taken.

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,251 @@
...
+
+We would also need to extend the `Kafka` custom resource to allow the user specifying that he wants auto-rebalancing on scaling by using a template.
+A new `spec.kafka.autoRebalance` field could be added for this purpose.
```
scholzj, line 63: I still wonder if this should be part of `.spec.cruiseControl`. At the end you need to deal with that section if you want to use it.

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,251 @@
...
+    # ...
+    autoRebalance:
+      # using the same rebalancing template when adding or removing brokers
+      - mode: add-brokers
+        template: my-add-remove-brokers-rebalancing-template
+      - mode: remove-brokers
+        template: my-add-remove-brokers-rebalancing-template  
+      # using a different template for a full rebalancing  
+      - mode: full
+        template: my-full-rebalancing-template
```
scholzj, line 92: Should the `template` fields follow ObjectReference, TypedLocalObjectReference or LocalObjectReference APIs from Kubernetes?

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,251 @@
...
+The user could even decide to have auto-rebalancing only when adding or removing brokers but not for both.
+
+This way provides the greatest flexibility to describe in which case to run the auto-rebalancing and with which configuration.
+
+### Auto-rebalancing on scaling up
+
+Following the operator flow when auto-rebalancing on scaling up.
+
+1. The user scales up the cluster by increasing the `spec.kafka.replicas` field.
+2. The `KafkaReconciler` takes care of it by rolling the existing brokers first and then scaling up the cluster as usual, by adding additional brokers.
```
scholzj, line 125: The existing brokers should not be rolled anymore in most cases when StrimziPodSets are used. So maybe you can make it a bit more generic and avoid mentioning it explciitely.

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,251 @@
...
+3. If the `spec.kafka.autoRebalance` contains an entry with `add-brokers` mode, the `KafkaReconciler` creates a `KafkaRebalance` custom resource named as `auto-<cluster-name>-rebalance` by using the corresponding template and sets:
+   * the `spec.mode` as `add-brokers` and the `spec.brokers` with the list of brokers already added to the cluster.
+   * the `spec.autoApproval` from the template as described by this [proposal](https://github.com/strimzi/proposals/pull/56).
```
scholzj, line 128: I left other comment about the`autoApproval` ... but what if it doesn't have the field?

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,251 @@
...
+2. The `KafkaReconciler` takes care of it and, if the `spec.kafka.autoRebalance` contains an entry with `remove-brokers` mode, it creates a `KafkaRebalance` custom resource named as `auto-<cluster-name>-rebalance` by using the corresponding template and sets:
+   * the `spec.mode` as `remove-brokers` and the `spec.brokers` with the list of brokers that will be removed.
+   * the `spec.autoApproval` from the template as described by this [proposal](https://github.com/strimzi/proposals/pull/56).
+3. The `KafkaRebalanceAssemblyOperator` detects the new rebalancing resource and acts as usual by requesting Cruise Control to generate an optimization proposal via the REST API on the `/remove_broker` endpoint.
+4. When the optimization proposal is ready, the `KafkaRebalanceAssemblyOperator` checks the auto-approval:
+   * If criteria are met or they are not specified at all, it approves the proposal automatically by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation.
+   * If criteria are not met, the `KafkaRebalance` custom resource status will move to `NotReady`, using a warning condition to describe why it was not approved automatically.
```
scholzj, line 158: As mentioned above, I do not think this should have any criteria. The only failure should be when the rebalance cannot be done because of e.g. lack of resources etc.

037-auto-rebalance-cluster-scaling.md
tombentley: We might be able to use the existing `mode` for this, e.g. `mode: template`, (or possibly including the existing mode in the template using `mode: add-brokers-template`, but this feels a bit ugly; I think it's cleaner when the add-brokers vs remove-brokers is not part of the template itself, that we you could use the same temple for both the add and remove cases).

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,251 @@
...
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    replicas: 3
+    # ...
+    autoRebalance:
+      # using the same rebalancing template when adding or removing brokers
+      - mode: add-brokers
```
tombentley, line 86: Is `mode` really the right word. It could be confused with the `mode` in the `KafkaRebalance` CR, and it seems to behave more like a trigger or condition for when the rebalance should be used.

037-auto-rebalance-cluster-scaling.md
tombentley: I'm inclined to this too, that way we're not implicitly coupling the `Kafka` API to the `KafkaRebalance` API. I.e. we'd leave the door open to a `BetterKafkaRebalance` kind in the future.

037-auto-rebalance-cluster-scaling.md
ppatierno: > Have you thought about using a regular KafkaRebalance as the template?

Absolutely. It was my first thought when I started to think about this. I had in mind a similar model as Tom but using a new additional field in the `KafkaRebalance`, something like a boolean `spec.template` (which was `false` by default).
It was in my mind the way to avoid the rebalance operator starting to process the `KafkaRebalance` resource when it was created (as opposed to what happens today) because just a reference to use when the auto-rebalancing was needed. But as Tom pointed it was not possible to share the same "template" across two different modes for example which is something that can be useful imho.

037-auto-rebalance-cluster-scaling.md
ppatierno: Taking out this proposal and the #56, the current behaviour of the rebalance operator is that it needs a manual approval (annotation) for running the rebalance. `it should be just taken` would mean having a differentiator on this behaviour and this is where the `spec.autoApproval` comes into account. The `KafkakRebalance` should have this field to allow the auto-approval with no user intervention as we agreeded.

037-auto-rebalance-cluster-scaling.md
scholzj: I still think new CRD is a huge overhead for what it does - not just initially but also maintenance wise. But if you think you need it, fine with me I guess.

037-auto-rebalance-cluster-scaling.md
scholzj: I guess you need to finish the #56 first then?

katheris (commented):
I definitely think this would be a useful feature, so that the user can take a single action to scale up or down and just let Strimzi do the work.

The only area that I think we need to be careful is the KafkaRebalance CRs that are visible to the user. It might be confusing to have some KafkaRebalance CRs that are owned by the user and some that are owned by Strimzi. What happens if a rebalance fails and the user then tries to add refresh and approve annotations, will that conflict with what Strimzi is doing?

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+Finally, the user should be able to define what are the [goals and the other rebalancing related configurations](https://strimzi.io/docs/operators/latest/configuring.html#type-KafkaRebalance-reference) which are normally described via a `KafkaRebalance` custom resource.
+In any case, for rebalancing the cluster, an actual `KafkaRebalance` custom resource has to be automatically created by the cluster operator.
+
+### New `KafkaRebalanceTemplate` custom resource
+
+A new `KafkaRebalanceTemplate` custom resource could be a solution allowing the user to define the rebalancing configuration he wants but leaving to the operator the creation of the corresponding `KafkaRebalance` custom resource.
+This new CRD should be quite similar to the `KafkaRebalance`.
+The cluster operator has to copy over the configuration from the template to the actual rebalancing resource.
```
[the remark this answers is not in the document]
ppatierno: My point is that the user wants anyway to define specific configuration for rebalance operations which happen on scale up/down. This is where the `KafkaRebalanceTemplate` comes into the picture and is used to allow the operator to generate the corresponding `KafkaRebalance`. If not leveraging the current custom resource how the user would be able to configure them? Again, back to one of the rejected alternatives putting everything in the `Kafka` CR? What could be a "different input"?
Anyway the auto generated `KafkaRebalance` is named with something like `auto-rebalance` suffix and of course documentation will help as well.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
ppatierno: The manual approval was removed from the proposal.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
ppatierno: Added in the corresponding "Error handling" section.

037-auto-rebalance-cluster-scaling.md
ppatierno: The manual approval was removed from the proposal, so now it's an "auto-proposal-generation" + "auto-approval" as you said and "auto-rebalancing" sounds to be fine now.

037-auto-rebalance-cluster-scaling.md
ppatierno: Yeah I would agree, in the end without CC (so the `cruiseControl` section) you cannot run a rebalance :-). I will make the change.

037-auto-rebalance-cluster-scaling.md
ppatierno: While I agree with it, on the other side this `mode` is really going to be the one used on the `KafkaRebalance` resource. In the end, we are using the same values. The question could be if the 1:1 mapping from this `mode` and the KR `mode` is actually right. Taking a look at CC REST API, it seems to be in the end. I don't see any other way to rebalance a cluster that could be different from add/remove brokers and full. So does it really make sense to change the name field here? 
I have no strong opinion in the end but if you feel strongly I can make the change. Let me know.

037-auto-rebalance-cluster-scaling.md
ppatierno: Agree, I think `LocalObjectReference` would be enough specifying just the name. I will change it.

037-auto-rebalance-cluster-scaling.md
ppatierno: This sentence is stale now, because we decided to not have "manual" approval anymore, so the  `autoApproval` will be set on the `KafkaRebalance` automatically and it doesn't need to be in the template first. I will fix this.

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,251 @@
...
+Compared to the rejected alternative, of adding the configuration to the `Kafka` custom resource, this will avoid polluting the `Kafka` custom resource and increasing its size.
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaRebalanceTemplate
+metadata:
+  name: my-rebalance-template
+  labels:
+    strimzi.io/cluster: my-cluster
```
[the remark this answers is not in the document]
ppatierno: It would be ignored as it happens for a `KafkaRebalance` custom resource and logging a warning.
In order to provide more info to the user, a warning condition could be added to the `Kafka` status. Wdyt?

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,251 @@
...
+  #       they are automatically set on the corresponding KafkaRebalance by the operator
+  goals:
+    - CpuCapacityGoal
+    - NetworkInboundCapacityGoal
+    - DiskCapacityGoal
+    - RackAwareGoal
+    - MinTopicLeadersPerBrokerGoal
+    - NetworkOutboundCapacityGoal
+    - ReplicaCapacityGoal
+  skipHardGoalCheck: true
```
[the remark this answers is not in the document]
ppatierno: `false` is the default in Cruise Control. In this case the rebalance will fail. So, it means that for the scale up use case we'll have a scaled up but not rebalanced cluster, as described in the "error handling" section, the `KafkaRebalance` will be in a `NotReady` state and the user has to deal with it. For the scale down use case, the failed rebalance will stop the scaling down. As above, the user has to deal with a `KafkaRebalance` in `NotReady` state.

037-auto-rebalance-cluster-scaling.md
ppatierno: Having criteria is anyway related to how the https://github.com/strimzi/proposals/pull/56 will move forward but in any case we can decide to not having criteria for the auto-rebalancing. But why do you think so?
If we think about some feasible criteria in https://github.com/strimzi/proposals/pull/56 (i.e. max MB to move, max replicas to move, ...) why do you think criteria should not be applied for auto rebalancing?

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,251 @@
...
+* After auto-approving the proposal, there is an error when running the rebalance.
+
+In both scenarios, the reconciliation ends with the `KafkaRebalance` custom resource in the `NotReady` state.
+The cluster operator logs a warning about the rebalancing not happened.
+
+The scale down cannot proceed because of the brokers to be removed are still hosting some replicas.
+This behavior is related to this [proposal](https://github.com/strimzi/proposals/pull/54).
+
+On the next reconciliation, the cluster operator will try to do a scale down again, so running the flow as before.
+The previous `KafkaRebalance` custom resource in the `NotReady` state is overridden to generate a new optimization proposal.
```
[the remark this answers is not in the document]
ppatierno: > you wait for 5 minutes and timeout in reconciliation no 1.

When do you mean it times out? When doing the rebalancing or the scale down?
The rebalancing cannot time out because it happens across more reconciliation of course. The KRAO has a different internal periodic pool to check the status through the CC REST API. If at some point it fails and the `KafkaRebalance` moves to `NotReady`, the scale down doesn't start.
If it times out during scale down, it means rebalancing went ok and the `KafkaRebalance` was deleted. Anyway what I see is that on next reconcile due to the auto-rebalancing enabled, the operator could try to restart a rebalance again (creating the `KafkaRebalance`) so we would need to keep track of the rebalancing already happened? Is this what you mean?

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,251 @@
...
+   * the `spec.mode` as `remove-brokers` and the `spec.brokers` with the list of brokers that will be removed.
+   * the `spec.autoApproval` from the template as described by this [proposal](https://github.com/strimzi/proposals/pull/56).
+3. The `KafkaRebalanceAssemblyOperator` detects the new rebalancing resource and acts as usual by requesting Cruise Control to generate an optimization proposal via the REST API on the `/remove_broker` endpoint.
+4. When the optimization proposal is ready, the `KafkaRebalanceAssemblyOperator` checks the auto-approval:
+   * If criteria are met or they are not specified at all, it approves the proposal automatically by annotating the `KafkaRebalance` custom resource with the `strimzi.io/rebalance=approve` annotation.
+   * If criteria are not met, the `KafkaRebalance` custom resource status will move to `NotReady`, using a warning condition to describe why it was not approved automatically.
+5. If/When the rebalancing ends successfully, the corresponding `KafkaRebalance` custom resource is deleted by the `KafkaReconciler`.
```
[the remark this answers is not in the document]
ppatierno: As described in another comment, when there is a pending proposal state or a rebalancing happening, the KRAO as a periodic timer to check so the reconciliation just ends immediately, there is no synchronous wait for the rebalancing done or proposal available in a single reconcile. Is this the waiting or timeout you are referring to?
When the periodic timer detect that the proposal is ready or the rebalancing is ended, it moves the `KafkaRebalance` to a final stable state.

037-auto-rebalance-cluster-scaling.md
ppatierno: I think this comment is stale now, please review the new content.

ppatierno:
@scholzj @tombentley @tomncooper @katheris I have addressed most of the comments and answered questions, I think we need another pass on this proposal to move forward making clearer some parts and addressing others. Thanks!

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
katheris: I agree that the user will need a mechanism to configure the rebalance operations they want to use during a rebalance, so the purpose of the `KafkaRebalanceTemplate` is clear to me. My concern is about the `KafkaRebalance` CRs that are then created as a result. Will it be clear to the user how to interact with the `KafkaRebalance` CRs created by them vs the ones created by the Strimzi operator. For example the `KafkaRebalance` resources that are suffixed with `auto-rebalance` shouldn't be edited by them, the user shouldn't change the config there or add annotations, they should change the config in the `KafkaRebalanceTemplate` instead and create their own `KafkaRebalance` if the automated one fails. This flow to me feels like it is open to confusion from users.

Having said that I haven't been able to think of a flow that would be as simple to implement from a Strimzi perspective. Without the intermediate `KafkaRebalance` you would need to track the status of the rebalancing in a different way.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
scholzj: I don't think it is an implementation detail. The KafkaReconciler/KafkaAssemblyOperator and KafkaRebalanceAssemblyOperator are today completely independent. I think it is important to understand the interaction between them and how they fit together.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
scholzj: KafkaRebalance is acted upon. But this is just a template. From my point of view:
* You make this more complicated then it has to be (you need to add the logic to check the label, have error handling for it etc.)
* You do not allow to share the the template between different clusters

Unless we allow to reference them across namespaces, the sharing of the templates is probably not major use-case. But I wonder more what do you get in exchange for the first point. It seems to me that without the label on the template resource you:
* Have less code complexity and make your life easier
* Do not loose any actual fucntionality or added value
* Possibly in some edge cases enable a new feature (sharing the template between clusters)

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
scholzj: That is the part I don't fully get. You want to have a separate resource for configuring the rebalance for scale-up / scale-down. But then you anyway mix it with the settings from the Kafka CR. My naive expectation would be that if I have to create this template resource, it will be the single source of truth here and not mix with anything else. Which I guess would mean that if this conflicts with the hard goals, they will be always skipped anyway and this field is not needed.

I wonder if this is intuitive for the user. I would either say that it just uses always the default goals from the Kafka CR. Or that the template is the single source of truth here and the pre-configured goals are always ignored.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
scholzj: One thing to keep in mind is GitOps tooling. Ideally we want to avoid it deleting these automatically created KafkaRebalance resources :-o.

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,265 @@
...
+
+During the above flow, errors can happen before the scaling down:
+
+* There is an error when getting the proposal.
+* After auto-approving the proposal, there is an error when running the rebalance.
+
+In both scenarios, the reconciliation ends with the `KafkaRebalance` custom resource in the `NotReady` state.
+The cluster operator logs a warning about the rebalancing not happened.
+
+The reconcile ends but the scale down cannot proceed because of the brokers to be removed are still hosting some replicas.
```
fvaleri, line 184:
```suggestion
The reconcile ends but the scale down cannot proceed because of the brokers to be removed are still hosting some partitions.
```

fvaleri (commented):
This is something users really want, so thanks for putting this together. I also like the idea of keeping the full automatic rebalance out of scope, as it brings additional complexity. Leaved some comments, hope it's helpful.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
scholzj: Creating the KafkaRebalance resource manually can be done today. That does not need any changes or any proposal. The point of this proposal is to do it automatically, without any manual action from the user.

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,265 @@
...
+
+This feature is about enabling the cluster operator to get an optimization proposal, which is auto-approved, in order to run the corresponding rebalancing when the user scales the cluster up or down by changing the `spec.kafka.replicas` field on the `Kafka` custom resource.
+Such a rebalancing can happen right after the scale up to move some replicas to the newly added brokers or right before the scale down to move replicas off of the brokers to be removed.
+
+## Current situation
+
+Currently, there is no automatic rebalancing when the user scales the cluster up or down.
+It means that:
+
+* after scaling up the cluster, in order to rebalance the cluster and move some replicas to the newly added brokers, the user has to create a `KafkaRebalance` custom resource using the `spec.mode: add-brokers` mode and add the new broker IDs to the the `spec.brokers` field.
```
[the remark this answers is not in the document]
scholzj: WDYT it should be a _partition_? You are not moving partitions. You moving replicas or partition-replicas if you want. Partition is a virtual thing which consists of one or more replicas. It is the replicas what is assigned to some broker and is moved.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
fvaleri: If I get this right, the user manually creates the template, so why not directly create the KafkaRebalance? Then, when replicas change, the operator will enrich it with whatever is needed for the specific mode.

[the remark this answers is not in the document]
scholzj: Yeah, but the idea is that you for example create you template once when creating your Kafka cluster. And this is used few months later when you scale it up or down. Not that you create the template manually few seconds before the scaling.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
fvaleri: I know that CC uses replicas, which is a better term, but it is somewhat confusing in this context because we also have broker replicas. Maybe we could use partition replicas, as you suggests.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
fvaleri: Yeah, that's what I meant. The operator should be smart enough to filter them out when it's not a scale up/down situation.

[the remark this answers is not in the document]
scholzj: I think that would be quite confusing for the user and for managing the Rebalance resources when it mixes some used as templates and some used as actual rebalances.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
ppatierno: This is actually how the current `KafkaRebalance` custom resoure and in general Cruise Control works.
From Strimzi documentation about `skipHardGoalCheck` this is what you get:

> Whether to allow the hard goals specified in the Kafka CR to be skipped in optimization proposal generation. This can be useful when some of those hard goals are preventing a balance solution being found. Default is false.

So the user is able to provide goals in his own proposal. There could be some of them as part of the `hard.goals` list but not all of them. It seems that anyway CC takes into account ALL the `hard.goals` (provided by user through the list or the ones in the CC configuration). So it's already mixing things. @kyguy @tomncooper anything I am missing here?

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
ppatierno: Yeah maybe I was adding useless complications on this. Let's remove the cluster label (at least for now).

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
ppatierno: > If I get this right, the user manually creates the KafkaRebalanceTemplate, so why not directly create the KafkaRebalance? Then, when replicas change, the operator will enrich it with whatever is needed for the specific mode.

So `the operator will enrich it` is one of the things that an operator should avoid to do imho. In general, an operator touches/changes only annotations on a resource it's handling (because of the management itself) but it doesn't add/update anything about the `spec` section which is defined by the user. The operator just updates the `status` one.
We could have an already filled `KafkaRebalance` with mode as well but we should think a way to allow the operator to skip it on reconcile if it's not the time to use it (because no scaling up/down). Maybe a `trigger` field could help but anyway it will come to the point by Jakub which I agree (users confused by mixing KafkaRebalance resources).
The other thing is that you can share a template for scaling up and down. Using `KafkaRebalance` directly would mean creating two identical custom resources with a different mode (which is in the end what the operator does for you behind the scenes).

> I think that would be quite confusing for the user and for managing the Rebalance resources when it mixes some used as templates and some used as actual rebalances.

I agree

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
tomncooper: I admit it would be pretty silly to provide custom goals that don't include all the hard goals and then set `skipHardGoalsCheck` to `false`. But this matches what we do with the `KafkaRebalance` CR currently. If anything having the default set to `false` is a useful safety net. 

`skipHardGoalsCheck` is one of the many per-rebalance configs that can be set. I think the template should have all the supported configs that are available in the standard `KafkaRebalance` CR. Not least because in future we may want to use the template for some kind of self-healing rebalance proposals so we should try to make it as flexible as possible.

037-auto-rebalance-cluster-scaling.md
tomncooper: I don't think we should remove the ability for an auto-generated KafkaRebalance to be manually approved. I can definitely see a use case where you specify your template with auto-approval=false and then have Strimzi generate the KafkaRebalance for you using your pre-approved settings and a human gives the final ok.

037-auto-rebalance-cluster-scaling.md
tomncooper: I agree with Paolo, we want to use conistent language when referring to the same things.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
ppatierno: Agree going to use "partition replicas" everywhere.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
ppatierno: Criteria were removed from the auto-approval proposal, I reviewed the steps and now it seems there are actors mentioned to handle what for each step.

037-auto-rebalance-cluster-scaling.md
ppatierno: Criteria were removed by the auto-approval proposal.

037-auto-rebalance-cluster-scaling.md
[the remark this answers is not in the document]
kyguy: +1 @tomncooper

037-auto-rebalance-cluster-scaling.md
kyguy: +1 @ppatierno

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,261 @@
...
+        goals:
+          - CpuCapacityGoal
+          - NetworkInboundCapacityGoal
+          - DiskCapacityGoal
+        skipHardGoalCheck: false
+        # ... other rebalancing related configuration  
+```
+
+For handling the auto-approval, a `spec.autoApproval` field, as described by this [proposal](https://github.com/strimzi/proposals/pull/56), would be needed as well.
+This proposal was rejected because it pollutes the `Kafka` custom resource with too many details and configuration related to a rebalancing operation when the user wants to specify specific desires compared to the default ones in the Cruise Control configuration.
```
kyguy, line 260: I really like this proposal of a `spec.cruiseControl.autoRebalance` field! Is it already too late to discuss this proposal further as an option, was already rejected by everyone?? 

I understand it would add a lot of new fields to the `Kafka` resource but I think it would still be easier for to develop and use!

037-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,261 @@
...
+The cluster operator logs a warning about the rebalancing not happened.
+It's the user who should deal with this scenario:
+
+* Checking the reason why the `KafkaRebalance` is in the `NotReady` state and deleting the custom resource.
+* If the user still wants rebalancing, they have to create a new `KafkaRebalance` custom resource manually.
+
+### Auto-rebalancing on scaling down
+
+1. The user scales down the cluster by decreasing the `spec.kafka.replicas` field.
+2. The `KafkaReconciler` takes care of it and, if the `spec.cruiseControl.autoRebalance` contains an entry with `remove-brokers` mode, it creates a `KafkaRebalance` custom resource named as `<cluster-name>-auto-rebalance-remove-brokers` by using the corresponding template and sets:
```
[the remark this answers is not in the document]
ppatierno: You are right it's misleading. The `KafkaReconciler` will do the scale down later on step 6.

037-auto-rebalance-cluster-scaling.md
ppatierno: It's not late for anything but what I see with this proposal is that the `Kafka` custom resource can really be poisoned with a lot of rebalancing related stuff becoming bigger and bigger even with a lot of duplication.

kyguy: > Kafka custom resource can really be poisoned with a lot of rebalancing related stuff becoming bigger and bigger even with a lot of duplication.

It would definitely bloat the Kafka resource but I feel it would be more intuitive of an interface and less complex to implement than using another custom resource especially one so similar to the existing `KafkaRebalance` resource. 

@katheris @scholzj @tomncooper I assume since this rejected alternative was not commented on that you agree that this route is not worth bloating the Kafka resource, is that correct?

037-auto-rebalance-cluster-scaling.md
katheris: I hadn't realised this was linked to a `mode` elsewhere, I think it is useful to keep consistent language

037-auto-rebalance-cluster-scaling.md
katheris: I am still unsure about the use of both a `KafkaRebalance` and `KafkaRebalanceTemplate` where the `KafkaRebalance` could be owned by the user or could be owned by Strimzi, however I don't think adding it to the `Kafka` resource makes it simpler. I think it is nice to be able to handle rebalance configs in a similar way between the auto-approved ones and the manual approval ones. My worry is more with the different types of `KafkaRebalance` than adding a new CR.

039-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,264 @@
...
+    - ReplicaCapacityGoal
+  skipHardGoalCheck: true
+  # ... other rebalancing related configuration
+```
+
+The user has the flexibility to specify the [goals and the other rebalancing related configurations](https://strimzi.io/docs/operators/latest/configuring.html#type-KafkaRebalance-reference) the same way they already do today on a `KafkaRebalance` custom resource.
+This allows the user to not rely on the default Cruise Control goals configuration which would be common for any rebalance without overriding goals.
+They could define different goals based on the status of the Kafka cluster or skip hard goals in some scenarios for example.
+As already mentioned, specifying goals but more in general customizing the optimization proposal request is already possible in the current usage of a `KafkaRebalance` custom resource so it should be allowed for the auto-rebalancing as well.
+In the end, the cluster operator will use that configuration from the `KafkaRebalanceTemplate` to fill the corresponding `KafkaRebalance` but adding the `spec.mode` and `spec.brokers` accordingly.
```
katheris, line 60: What will happen to the auto-generated `KafkaRebalance` once the rebalance is complete? Will it be left so the user can see what happened, or deleted so it isn't cluttering up the cluster?

039-auto-rebalance-cluster-scaling.md
```diff
@@ -0,0 +1,264 @@
...
+
+* There is an error when getting the proposal.
+* After auto-approving the proposal, there is an error when running the rebalance.
+
+In both scenarios, the reconciliation ends with the `KafkaRebalance` custom resource in the `NotReady` state as per the `KafkaRebalanceAssemblyOperator` states machine.
+The cluster operator logs a warning about the rebalancing not happened.
+It's the user who should deal with this scenario:
+
+* Checking the reason why the `KafkaRebalance` is in the `NotReady` state and deleting the custom resource.
+* If the user still wants rebalancing, they have to create a new `KafkaRebalance` custom resource manually.
```
katheris, line 155: Presumably the user could try to reuse the `KafkaRebalance` by adding a refresh annotation? What would happen in that case?

037-auto-rebalance-cluster-scaling.md
ppatierno: I think we are moving around in circle on this ... 
Having a new `KafkaRebalanceTemplate` CR seems to be much but allows the flexibility to specify rebalancing configuration, but then it comes to user created `KafkaRebalance` vs operator created `KafkaRebalance`.
Bloating the `Kafka` CR doesn't sound great to me unless we decide to move the entire `cruiseControl` related configuration in a new `KafkaCruiseControl` custom resource which will have the current CC configuration and it could host the auto-rebalancing as well. In that case, I would agree that a new `KafkaRebalanceTemplate` wouldn't be necessary (by "bloating the new `KafkaCruiseControl` resource this time). Anyway, it would not "fix" the "issue" to have both user created `KafkaRebalance` vs operator created `KafkaRebalance` because trying to leverage the state machine we already have in place.
Of course, it would be something t   o do before moving forward with this PR. It would be a pre-requisite somehow.
@katheris @kyguy @scholzj @tombentley @tomncooper should we have a thought about this first? Wdyt?

039-auto-rebalance-cluster-scaling.md
ppatierno: It should just work. There is nothing preventing the user to put his hands on a `KafkaRebalance` created by the operator unless we could set up some RBAC for it?

037-auto-rebalance-cluster-scaling.md
scholzj: @ppatierno `KafkaCruiseControl` might not be a bad idea, but if you want to do that, it would probably be for a separate proposal which should be dealt with before this. So not sure how eager you are to go there. My main confusion here is why do we need so many different goals definitions. But that might very well be because of my lack of CC knowledge. My main `KafkaRebalanceTemplate` concern is that it means a lot of maintenance effort going forward. But if you think it is needed despite the effort, then you should go for it -> I do not have problem to approve the PR with it.

ppatierno: > KafkaCruiseControl might not be a bad idea, but if you want to do that, it would probably be for a separate proposal which should be dealt with before this. So not sure how eager you are to go there.

Well let's say that apart from the user goals customization, moving to a `KafkaCruiseControl` would leave us more freedom to approach to the problem without taking the `Kafka` CR anymore and not being scared about polluting it if needed. I am fine to work on a different proposal if we all think that will be helpful anyway. Of course, we'll park this one for later.

scholzj: It would be a big change, so I think wider concensus is needed (i.e. we need more people to comment apart from the two of us ;-))

kyguy: For the `spec.cruiseControl.autoRebalance` argument but also applies to the proposed `KafkaCruiseControl` as well:

Would we have to include the `goal` and other rebalance options [1] as configurable through the Kafka resource? Since the Cluster Operator is the client executing the rebalance call couldn't the Cluster Operator be responsible for determining/setting the optimal rebalance options. [1] I If the Cluster Operator is responsible for those rebalance options [1] then we don't need to expose those rebalance options to users through the Kafka custom resource and we would save the Kafka custom resource from the bloat! I think it makes sense to have the autoscale up and down to rely on the default goals in the `spec.cruiseControl.config` anyway since the overall cluster balancedness score will be based on those `spec.cruiseControl.config` configured goals  before and after the scaling, right? What's the use of the `spec.cruiseControl.config` goals settings if we are overriding them all the time anyway?

[1] https://github.com/linkedin/cruise-control/wiki/REST-APIs#add-a-list-of-new-brokers-to-kafka-cluster

ppatierno:
As discussing during the latest community meeting, I am going to close this one because I want to explore a different idea of doing this after the feedback I have got. It's not planned to write a proposal right now and I will back to this later.

PR #59, opened 2022-07-03
Pluggable security configuration providers

scholzj:
This proposals replaces #48. It suggests pluggable mechanism for PodSecurityProviders and implementation which matches the Kubernetes _restricted_ Pod Security Standards profile. This should make it easier for users to run under this profile.

In addition to the proposal, there is also a draft implementation which can be seen [here](https://github.com/strimzi/strimzi-kafka-operator/compare/main...scholzj:plugable-pod-security-profiles?expand=1). It currently doesn't have the docs, Javadoc and full test coverage, but it is fully functional under the restricted profile.

### DIFF
--- a/037-pluggable-pod-security-profiles.md
+++ b/037-pluggable-pod-security-profiles.md
@@ -197,7 +197,7 @@ In the YAML format, the generated security context will look like this:
 * All containers (apart from Kaniko) will have the following security context:
   ```yaml
   securityContext:
-    allowPriviledgeEscalation: false
+    allowPrivilegeEscalation: false
     runAsNonRoot: true
     secCompProfile:
       type: RuntimeDefault
--- a/037-pluggable-pod-security-profiles.md
+++ b/037-pluggable-pod-security-profiles.md
@@ -93,28 +93,65 @@ The reason for having this interface in the `api` module is that this module is 
 Users will be able to write their implementations by using the `api` module as dependency.
 
 The interface will have methods for providing security contexts for all Pods created by Strimzi, and all containers.
-The methods will take as a parameter the user-configured security context from the `.template` section of the custom resources.
-The methods for the Pods using storage configuration (Kafka and ZooKeeper) will also have the storage configuration as parameter.
-These methods will have the default implementation which will always just return the user-specified context.
-For example:
+The methods providing the security context will get as a parameter an object of type `PodSecurityProviderContext`.
+This will be an object created by Strimzi which will encapsulate the different infromation which strimzi will provide to the Provider plugin to make its decisions when generating the Kubernetes security context.
+The `PodSecurityProviderContext` will be defines as another interface in the `api` module:
 
 ```java
-    default PodSecurityContext kafkaPodSecurityContext(PodSecurityContext userSecurityContext, Storage storage) {
+interface PodSecurityProviderContext {
+    Storage storage();
+    PodSecurityContext userSuppliedContext();
+}
+```
+
+This encapsulation should make it easier to add additional information in the future without changing the signature of the provider methods but only by adding new methods to the Strimzi provided context.
+
+The provider methods will have the default implementation which will always just return the user-specified context.
+For example (just few methods are listed in this example):
+
+```java
+public interface PodSecurityProvider {
+    // ...
+
+    default PodSecurityContext kafkaPodSecurityContext(PodSecurityProviderContext context) {
         return userSecurityContext;
     }
 
-    default SecurityContext kafkaContainerSecurityContext(SecurityContext userSecurityContext, Storage storage) {
+    default SecurityContext kafkaContainerSecurityContext(PodSecurityProviderContext context) {
         return userSecurityContext;
     }
 
-    default SecurityContext kafkaInitContainerSecurityContext(SecurityContext userSecurityContext) {
+    default SecurityContext kafkaInitContainerSecurityContext(PodSecurityProviderContext context) {
         return userSecurityContext;
     }
+
+    // ...
+}
 ```
 
 The interface will also have a `configure` method for configuring the provider.
 The `configure` method will get as a parameter an object describing the platform (Kubernetes version, features / APIs) which can be taken into account when creating the security context.
 This method will not have any default implementation.
+
+```java
+public interface PodSecurityProvider {
+    // ...
+    void configure(PlatformFeatures platformFeatures);
+    // ...
+}
+```
+
+`PlatformFeatures` in this method is another new interface which provides some of the methods provided by the `PlatformFeaturesAvailability` class which is part of the `operator-common` module`
+It uses the features which might be relevant for generating the securitny context:
+
+```java
+public interface PlatformFeatures {
+    KubernetesVersion getKubernetesVersion();
+    boolean isOpenshift();
+}
+```
+
+The `configure(...)` method can in addition to the `PlatformFeatures` also for example use environment variables to configure the provider.
 
 #### Implementations
 
--- a/037-pluggable-pod-security-profiles.md
+++ b/037-pluggable-pod-security-profiles.md
@@ -1,4 +1,4 @@
-# Pluggable Pod Security Profile
+# Pluggable Pod Security Profiles
 
 Kubernetes supports configuration of a Security Context (SC) which can limit what the applications running on top of it are or are not allowed to do.
 The SC can be configured on the Pod level (Pod Security Context) or on the container level.
--- a/037-pluggable-pod-security-profiles.md
+++ b/037-pluggable-pod-security-profiles.md
@@ -114,15 +114,15 @@ public interface PodSecurityProvider {
     // ...
 
     default PodSecurityContext kafkaPodSecurityContext(PodSecurityProviderContext context) {
-        return userSecurityContext;
+        return context.userSuppliedContext();
     }
 
     default SecurityContext kafkaContainerSecurityContext(PodSecurityProviderContext context) {
-        return userSecurityContext;
+        return context.userSuppliedContext();
     }
 
     default SecurityContext kafkaInitContainerSecurityContext(PodSecurityProviderContext context) {
-        return userSecurityContext;
+        return context.userSuppliedContext();
     }
 
     // ...
@@ -162,17 +162,17 @@ We will provide two implementations to support the baseline (default) and restri
 These implementations will be also part of the `api` module so that users can just extend them instead of implementing the provider from scratch.
 
 The `BaselinePodSecurityProvider` will implement the current behaviour:
-* Return the user-configured security context when specified
-* When the user-configured context is `null` and we are on OpenShift:
+* Return the user-supplied security context when specified
+* When the user-supplied context is `null` and we are on OpenShift:
     * A `null` will be returned (OpenShift injects its own Security Context in that case) for all Pods and containers
-* When the user-configured context is `null` and we are not on OpenShift:
+* When the user-supplied context is `null` and we are not on OpenShift:
     * Pods with storage will get a Security Context specifying the file system group as `0`
     * All other pods will get `null` Security Context
     * All containers will get `null` Security Context
 
 The `RestrictedPodSecurityProvider` will implement the following behaviour:
-* Return the user-configured security context when specified
-* When the user-configured context is `null`:
+* Return the user-supplied security context when specified
+* When the user-supplied context is `null`:
     * Pods with storage will get a Security Context specifying the file system group as `0`
     * All other pods will get `null` Security Context
     * For the Kaniko container, an exception will be thrown since Kaniko cannot run under the Restricted profile
--- a/037-pluggable-pod-security-profiles.md
+++ b/037-pluggable-pod-security-profiles.md
@@ -1,0 +1,197 @@
+# Pluggable Pod Security Profile
+
+Kubernetes supports configuration of a Security Context (SC) which can limit what the applications running on top of it are or are not allowed to do.
+The SC can be configured on the Pod level (Pod Security Context) or on the container level.
+When configured on the Pod level, it applies to all containers in given Pod.
+The security context allows you to configure different security related aspects of the containers.
+For example:
+* Users and group used for running the container
+* Capabilities available to the applications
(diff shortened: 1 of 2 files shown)


037-pluggable-pod-security-profiles.md
```diff
@@ -0,0 +1,196 @@
...
+The puggable mechanism should also make it easy for the Strimzi itself to add any additional implementations in the future.
+
+### Plugin implementation
+
+#### Interface
+
+A new interface `PodSecurityProvider` will be added to the `api` module.
+All provider plugins will implement this interface.
+The reason for having this interface in the `api` module is that this module is distributed though Maven repositories.
+Users will be able to write their implementations by using the `api` module as depednency.
```
tombentley, line 93:
```suggestion
Users will be able to write their implementations by using the `api` module as dependency.
```

037-pluggable-pod-security-profiles.md
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    default SecurityContext kafkaInitContainerSecurityContext(SecurityContext userSecurityContext) {
+        return userSecurityContext;
+    }
+```
+
+The interface will also have a `configure` method for configuring the provider.
+The `configure` method will get as a parameter an object describing the platform (Kubernetes version, features / APIs) which can be taken into account when creating the security context.
+This method will not havy any default implementtion.
```
tombentley, line 117:
```suggestion
This method will not have any default implementation.
```

037-pluggable-pod-security-profiles.md
```diff
@@ -0,0 +1,196 @@
...
+The methods for the pods using storage configuration (Kafka and ZooKeeper) will also have the storage configuration as parameter.
+These methods will have the default implementation which will always just return the user-specified context.
+For example:
+
+```java
+    default PodSecurityContext kafkaPodSecurityContext(PodSecurityContext userSecurityContext, Storage storage) {
+        return userSecurityContext;
+    }
+
+    default SecurityContext kafkaContainerSecurityContext(SecurityContext userSecurityContext, Storage storage) {
```
tombentley, line 106: Using the pattern of putting the storage in the signature like this would make for an awkward compatibility story should some pod which today don't use storage change in the future to require storage. 

Using an interface to represent the context for each method might be a more flexible pattern for coping with such changes. 

E.g.

```java
default PodSecurityContext kafkaPodSecurityContext(KafkaPodSecurityContext context) {
    return context.userSuppliedContext();
}

// where

interface KafkaPodSecurityContext {
    Storage storage();
    PodSecurityContext userSuppliedContext();
}

// and similar for all the other methods
```

037-pluggable-pod-security-profiles.md
```diff
@@ -0,0 +1,196 @@
...
+This might be useful for users who use some custom security tooling and might have their own requirement for the security configuration.
+But regular users will not need to write any custom plugins.
+
+The puggable mechanism should also make it easy for the Strimzi itself to add any additional implementations in the future.
+
+### Plugin implementation
+
+#### Interface
+
+A new interface `PodSecurityProvider` will be added to the `api` module.
```
tombentley, line 90: Having a single interface might cause difficulties for implementors when we want to add or remove pods. Using `default` methods helps with adding. But removing a method would break source compatibility when implementers had use `@Override`. Maybe having a separate interface for each kind of pod would make for a better compatibility story.

037-pluggable-pod-security-profiles.md
```diff
@@ -0,0 +1,196 @@
...
+        * Disable privilage escalation
+        * Run as non-root
+        * Specify the Seccomp profile as `RuntimeDefault`
+        * Drop all capabilities
+
+#### Loading the plugins
+
+In the Cluster Operator, the plugins will be loaded using the Java Service Provider Interface which will be used to find the class configured by the user.
+A `PodSecurityProviderFactory` class will be part of the `cluster-oprator` module.
+It will have a static initialize method which will load the class based on the configuration, call its `configure(...)` method and store it in a static field for further use.
```
tombentley, line 151: Can you should the `configure` method in the example above?

tombentley (commented):
Thanks Jakub, I can see the potential benefit of this for users. I left a few comments.

037-pluggable-pod-security-profiles.md
```diff
@@ -0,0 +1,196 @@
...
+When the configured class is not found, an error will be thrown during the intialization and the operator will exit.
+
+### Installation files
+
+There will be no changes to the installation files.
+If user wants to use Strimzi under a _restricted_ security profile, they can edit the installation files to specify the security context.
+That way, the installation files will keep working on the existing environment where they work today.
+This includes OpenShift, which's _restricted_ SCC policy is not compatible with the Kubernetes _restricted_ profile.
+
+In the futere, depending on how many users use the _restricted_ security profile, we can consider adding it to the installtion files and using `RestrictedPodSecurityProvider`  by default.
```
PaulRMellor, line 174:
```suggestion
In the future, depending on how many users use the restricted security profile, we can consider adding it to the installation files and using `RestrictedPodSecurityProvider`  by default.
```

037-pluggable-pod-security-profiles.md
```diff
@@ -0,0 +1,196 @@
...
+
+This proposal will make it easier for the users to run under the different Kubernetes security profiles.
+It will also make it much easier to customize the security context configuration if needed by providing custom plugins.
+
+## Rejected alternatives
+
+### Using a separate module for the interface and its implementations
+
+I considered creating a separate Java module for the plugin interface and its implementations instead of having them in the `api` module.
+But it seemed unnecessary without giving some additional advantage and it would make our builds to run longer.
```
PaulRMellor, line 186:
```suggestion
But it seemed unnecessary without giving some additional advantage and it would take the builds longer to run.
```

037-pluggable-pod-security-profiles.md
scholzj: This is good idea. I added this to the proposal. However, the methods are in general the same. So I propose to use the same context for all of them.

037-pluggable-pod-security-profiles.md
scholzj: Done.

037-pluggable-pod-security-profiles.md
scholzj: TBH, I don't like this approach. The interface has many methods, but they in general do the same. While we want to provide the granularity _just in case it is needed_, I expect that in most cases they will just delegate to a private method. This is also what the implementations of the baseline and restricted profiles which should be part of Strimzi do.

Having a separate interface for each operand would mean a lot of interfaces and lot of implementation classes. They will largely do all the same but just with higher maintenance effort. Also, the _injection_ of the providers into the model classes in the `cluster-operator` model will need to be done multiple times instead of just once.

037-pluggable-pod-security-profiles.md
```diff
@@ -0,0 +1,234 @@
...
+We will provide two implementations to support the baseline (default) and restricted profiles:
+* `BaselinePodSecurityProvider`
+* `RestrictedPodSecurityProvider`
+
+These implementations will be also part of the `api` module so that users can just extend them instead of implementing the provider from scratch.
+
+The `BaselinePodSecurityProvider` will implement the current behaviour:
+* Return the user-configured security context when specified
+* When the user-configured context is `null` and we are on OpenShift:
+    * A `null` will be returned (OpenShift injects its own Security Context in that case) for all Pods and containers
```
fvaleri, line 167: What about returning a `DefaultSecurityContext` instead of `null`? Null makes me think that there is no security context at all.

037-pluggable-pod-security-profiles.md
```diff
@@ -0,0 +1,234 @@
...
+#### Implementations
+
+We will provide two implementations to support the baseline (default) and restricted profiles:
+* `BaselinePodSecurityProvider`
+* `RestrictedPodSecurityProvider`
+
+These implementations will be also part of the `api` module so that users can just extend them instead of implementing the provider from scratch.
+
+The `BaselinePodSecurityProvider` will implement the current behaviour:
+* Return the user-configured security context when specified
```
fvaleri, line 165: I think `userSuppliedContext()` in the above snippets and `user-configured` in this section are the same concept, so we should be consistent with naming.

fvaleri (commented):
LGTM, I just added a couple of comments.

037-pluggable-pod-security-profiles.md
fvaleri: Ok, thanks for the explanation.

037-pluggable-pod-security-profiles.md
```diff
@@ -0,0 +1,234 @@
...
+}
+```
+
+`PlatformFeatures` in this method is another new interface which provides some of the methods provided by the `PlatformFeaturesAvailability` class which is part of the `operator-common` module`
+It uses the features which might be relevant for generating the securitny context:
+
+```java
+public interface PlatformFeatures {
+    KubernetesVersion getKubernetesVersion();
+    boolean isOpenshift();
```
tombentley, line 150: Would something like `platformSuppliesDefaultContext()` actually capture the behaviour we're after? That way the API could support other kube distributions doing similar things to OpenShift, without having to represent every possible distribution in the API.

tombentley (approved):
LGTM.

037-pluggable-pod-security-profiles.md
scholzj: TBH, I don't know. I'm not sure what other platforms are doing -> some other platforms might simply validate the context without injecting it. Even with OpenShift - it currently injects the security context. But not sure how it will do it in the future once it migrates from SCC to Kubernetes Pod Security Profiles. So I'm not entirely sure if something like `platformSuppliesDefaultContext` makes sense as it might be equally insufficient in the future as `isOpenshift` is. It also shifts more of the logic from the plugin into the core operator which might need to identify the platform and decide what it does. With the currently proposed API, you can do more directly in the plugin - for example decide for different behaviour on OpenShift with version A.B.x versus OpenShift with version C.D.x. So I would probably be inclined to keep the `isOpenshift` as proposed right now.

tomncooper (approved):

037-pluggable-pod-security-profiles.md
```diff
@@ -0,0 +1,234 @@
...
+                - ALL
+            runAsNonRoot: true
+            seccompProfile:
+              type: RuntimeDefault
+```
+
+While users can configure the security context, it is not user-friendly.
+The Pod security context does not cover all the different options required by the _restricted_ profile.
+So the security context has to be configured for every container (including init containers, sidecars etc.).
+So a single Kafka cluster might need configuration in 9 different places in the `Kafka` custom resource: ZooKeeper, Kafka broker, Kafka init-container, Topic Operator, User Operator, TLS sidecar, Kafka Exporter, Cruise Control, and another TLS sidecar.
```
ppatierno, line 66: "another TLS sidecar" ? Referring to the CC one, it was removed. Anyway it's not so important here.

037-pluggable-pod-security-profiles.md
```diff
@@ -0,0 +1,234 @@
...
+}
+```
+
+The `configure(...)` method can in addition to the `PlatformFeatures` also for example use environment variables to configure the provider.
+
+#### Implementations
+
+We will provide two implementations to support the baseline (default) and restricted profiles:
+* `BaselinePodSecurityProvider`
+* `RestrictedPodSecurityProvider`
```
ppatierno, line 160: Out of curiosity, aren't we providing a "Privileged" provider for security reasons? I mean, we want the user to be aware of what it means running as "privileged" so he has to create the provider on his own? I can understand that because it would be totally unrestricted.

037-pluggable-pod-security-profiles.md
```diff
@@ -0,0 +1,234 @@
...
+* Return the user-supplied security context when specified
+* When the user-supplied context is `null`:
+    * Pods with storage will get a Security Context specifying the file system group as `0`
+    * All other pods will get `null` Security Context
+    * For the Kaniko container, an exception will be thrown since Kaniko cannot run under the Restricted profile
+    * All other containers will get a Security Context matching the _restricted_ Kubernetes profile:
+        * Disable privilage escalation
+        * Run as non-root
+        * Specify the Seccomp profile as `RuntimeDefault`
+        * Drop all capabilities
```
ppatierno, line 183: Shouldn't we have a couple of examples here to show what is the security context configuration returned when using these two providers?

037-pluggable-pod-security-profiles.md
```diff
@@ -0,0 +1,234 @@
...
+This allows them to customize a security context that matches their own requirements and policies.
+On OpenShift, the security context is automatically injected into the Pods when they are created based on OpenShift Security Context Constraints (SCC) policies.
+This typically involves dropping some capabilities, using some random unprivileged user ID etc.
+In some cases, Strimzi users might also use similar systems to inject the security context using admission controllers and similar tools.
+
+The _baseline_ profile basically means that you just use the default settings without requesting any additional privileges or capabilities.
+So without any of the additional tooling mentioned previously, Strimzi operators and the Pods created by them at present match the _baseline_ profile.
+If you try to run Strimzi in a cluster where the _restricted_ profile is enforced, it does not work out of the box.
+
+Apart from the Kaniko builder used for the Kafka Connect build, all our components are able to run under the _restricted_ profile.
```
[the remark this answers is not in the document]
scholzj: It needs to run as root. I added it to the proposal.

037-pluggable-pod-security-profiles.md
scholzj: I added the YAML examples of what they will return.

037-pluggable-pod-security-profiles.md
ppatierno: Thanks for the clarification!

ppatierno (approved):
LGTM

scholzj:
This proposal now has 5 binding +1 votes which is enough for approving. I will leave it open till tomorrow to see if anyone else has some more comments. Thanks.

PR #60, opened 2022-10-04
Configurable exclusion of labels in User Operator

see-quick:
This PR is proposing a configurable exclusion of labels of User Operator Kubernetes Secrets. There may be some details which I miss, so feel free to comment and we can discuss it :).

Signed-off-by: morsak <maros.orsak159@gmail.com>


039-user-operator-custom-filter-for-labels-.md
```diff
@@ -0,0 +1,139 @@
...
+Moreover, we would need to get the value of the environment variable to such a class.
+In the `UserOperatorConfig` class, we will obtain value, and then in the `KafkaUserOperaor` class, specifically in the following method:
+```java
+protected Future<KafkaUserStatus> createOrUpdate(Reconciliation reconciliation, KafkaUser resource) {
+    KafkaUserModel user;
+    KafkaUserStatus userStatus = new KafkaUserStatus();
+
+    try {
+        user = KafkaUserModel.fromCrd(resource, config.getSecretPrefix(), config.isAclsAdminApiSupported(),config.isKraftEnabled(), 
+            config.getStrimziLabelsExclusionPattern); // <-- this one we will inject into KafkaUserModel class
```
scholzj, line 101: This way, you will ahve to compile the variable again and again in every reocnciliation. That seems like waste of resources.

039-user-operator-custom-filter-for-labels-.md
```diff
@@ -0,0 +1,139 @@
...
+    // filter by value of instance variable `this.getStrimziLabelsExclusionPattern`
+    
+    return secret;
+}
+```
+
+## Compatibility
+
+This proposal does not change any of the existing CRDs or the Kubernetes secrets that are being created.
+The user only needs to modify Kafka's custom resource to use such a feature.
```
scholzj, line 135: Setting it through Kafka CR is only one way. You can use also standalone User Operator. So you should probably not focus too much on it.

scholzj (commented):
Some nits. You should also promote this proposal more widely:
* On Slack
* On the original issue

039-user-operator-custom-filter-for-labels-.md
```diff
@@ -0,0 +1,139 @@
...
+    topicOperator: {}
+    userOperator: {}
+```
+Then, we would obtain a value from the environment variable and parse it in the `UserOperatorConfig` class.
+Based on the environment variable value, we will determine what to do (a) env is not specified, and therefore we do not exclude any label from Secret
+(b) env is specified by a regex, and then we will filter and remove such labels which match the input regex.
+
+### More implementation details
+
+### 1 Parsing part
```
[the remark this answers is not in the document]
see-quick: fixed

039-user-operator-custom-filter-for-labels-.md
```diff
@@ -0,0 +1,139 @@
...
+            .withLabels(Util.mergeLabelsOrAnnotations(labels.toMap(), templateSecretLabels))
+            .withAnnotations(Util.mergeLabelsOrAnnotations(null, templateSecretAnnotations))
+            .withOwnerReferences(createOwnerReference())
+        .endMetadata()
+        .withType("Opaque")
+        .withData(data)
+        .build();
+    
+    // here, we have to do post-processing (i.e., filtering) of labels by exclusion pattern
+    // filter by value of instance variable `this.getStrimziLabelsExclusionPattern`
```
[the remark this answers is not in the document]
see-quick: Well, it could be done here or before the creation of an instance. But I think pre-processing would be better in this case. I have changed the order :)

039-user-operator-custom-filter-for-labels-.md
see-quick: I have changed the String type of strimziLabelsExclusionPattern to the `Pattern` and also compile such regex in the parsing part (i.e., in the UserOperatorConfig), which should eliminate this workload...

039-user-operator-custom-filter-for-labels-.md
```diff
@@ -0,0 +1,139 @@
...
+We can see that User Operator automatically adds the `app.kubernetes.io/instance: my-user` label.
+In some situations, we do not want a such label to be automatically created (i.e., [endless deletion and creation of specific Secret](https://github.com/strimzi/strimzi-kafka-operator/issues/5690))
+and therefore, one way how to get rid of this already assigned label is to filter it.
+
+## Proposal
+
+This proposal suggests adding configurable exclusion of labels.
+We can implement this feature similarly to [issue](https://github.com/strimzi/strimzi-kafka-operator/pull/4791).
+We can create an environment variable and inject such value through the `Kafka` custom resource.
+Specifically, in the `spec.entityOperator.template.topicOperatorContainer.env`.
```
ppatierno, line 48: you meant `userOperatorContainer` here I guess

039-user-operator-custom-filter-for-labels-.md
```diff
@@ -0,0 +1,139 @@
+# User Operator: Configurable exclusion of labels
+
+This proposal is about adding the ability to filter the automatically assigned labels.
+When User Operator creates `KafkaUser`, it also creates an associated Secret, where it automatically adds the label
+`app.kubernetes.io/instance: <username>`.
+Such a label could be used for many uses, and different users have various requirements and expectations.
+Nevertheless, it could also lead to not likeable scenarios (e.g., [repeated deletion and re-creation of specific Secret](https://github.com/strimzi/strimzi-kafka-operator/issues/5690)).
```
PaulRMellor, line 7:
```suggestion
Nevertheless, it could also lead to undesirable scenarios (e.g., [repeated deletion and re-creation of specific Secret](https://github.com/strimzi/strimzi-kafka-operator/issues/5690)).
```

039-user-operator-custom-filter-for-labels-.md
```diff
@@ -0,0 +1,139 @@
...
+Therefore, we propose to implement configurable exclusion of labels.
+
+## Current situation
+
+Presently, when a user attempts to create a KafkaUser defined as follows:
```
PaulRMellor, line 12:
```suggestion
Presently, when a user attempts to create a `KafkaUser` defined as follows:
```

039-user-operator-custom-filter-for-labels-.md
```diff
@@ -0,0 +1,139 @@
...
+kind: KafkaUser
+metadata:
+  name: my-user
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  authentication:
+    type: tls
+```
+The User Operator then creates an associated Kubernetes Secret.
```
PaulRMellor, line 24:
```suggestion
The User Operator then creates an associated Kubernetes `Secret`.
```

039-user-operator-custom-filter-for-labels-.md
```diff
@@ -0,0 +1,139 @@
...
+    app.kubernetes.io/part-of: strimzi-my-user
+    strimzi.io/cluster: my-cluster
+    strimzi.io/kind: KafkaUser
+  name: my-user
+  namespace: myproject
+...
+```
+We can see that User Operator automatically adds the `app.kubernetes.io/instance: my-user` label.
+In some situations, we do not want a such label to be automatically created (i.e., [endless deletion and creation of specific Secret](https://github.com/strimzi/strimzi-kafka-operator/issues/5690))
+and therefore, one way how to get rid of this already assigned label is to filter it.
```
PaulRMellor, line 41:
```suggestion
If we do not want the label that has been assigned, the only way to get rid of it is to filter it.
```

PaulRMellor: I made the above suggestion because we have the same link in the intro.

039-user-operator-custom-filter-for-labels-.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Proposal
+
+This proposal suggests adding configurable exclusion of labels.
+We can implement this feature similarly to [issue](https://github.com/strimzi/strimzi-kafka-operator/pull/4791).
```
PaulRMellor, line 46:
```suggestion
We can implement this feature similarly to [issue 4394](https://github.com/strimzi/strimzi-kafka-operator/pull/4791).
```

039-user-operator-custom-filter-for-labels-.md
```diff
@@ -0,0 +1,139 @@
...
+We can create an environment variable and inject such value through the `Kafka` custom resource.
```
PaulRMellor, line 47:
```suggestion
We can create an environment variable and inject the value through the `Kafka` custom resource.
```

039-user-operator-custom-filter-for-labels-.md
```diff
@@ -0,0 +1,139 @@
...
+      userOperatorContainer:
+        env:
+          name:   STRIMZI_LABELS_EXCLUSION_PATTERN
+          value:  "^app.kubernetes.io/.*$"
+    topicOperator: {}
+    userOperator: {}
+```
+Then, we would obtain a value from the environment variable and parse it in the `UserOperatorConfig` class.
+Based on the environment variable value, we will determine what to do (a) env is not specified, and therefore we do not exclude any label from Secret
+(b) env is specified by a regex, and then we will filter and remove such labels which match the input regex.
```
PaulRMellor, line 68:
```suggestion
Based on the environment variable value, (a) when `env` is not specified, we do not exclude any label from the secret
(b) when `env` is specified by a regex, we filter and remove those labels which match the regex.
```

039-user-operator-custom-filter-for-labels-.md
```diff
@@ -0,0 +1,139 @@
...
+    }
+    ...
+}
+```
+
+#### Exclusion part
+
+The exclusion part will be placed in the `KafkaUserModel` class.
+Moreover, we would need to get the value of the environment variable to such a class.
+In the `UserOperatorConfig` class, we will obtain value, and then in the `KafkaUserOperaor` class, specifically in the following method:
```
PaulRMellor, line 94:
```suggestion
In the `UserOperatorConfig` class, we will obtain the environment variable value, and then in the `KafkaUserOperator` class, specifically in the following method:
```

PaulRMellor: This bit seems unfinished _and then in the `KafkaUserOperaor` class_...?

039-user-operator-custom-filter-for-labels-.md
```diff
@@ -0,0 +1,139 @@
...
+        user = KafkaUserModel.fromCrd(resource, config.getSecretPrefix(), config.isAclsAdminApiSupported(),config.isKraftEnabled(), 
+            config.getStrimziLabelsExclusionPattern); // <-- this one we will inject into KafkaUserModel class
+    } catch (Exception e) {
+        StatusUtils.setStatusConditionAndObservedGeneration(resource, userStatus, Future.failedFuture(e));
+        return Future.failedFuture(new ReconciliationException(userStatus, e));
+    }
+    ...
+}
+```
+Then, we can use such value to store it as an instance attribute of `KafkaUserModel` and implement pre-processing of labels in the
```
PaulRMellor, line 110:
```suggestion
Then, we can use the environment variable value to store it as an instance attribute of `KafkaUserModel` and implement pre-processing of labels in the
```

PaulRMellor (approved):
Really clear proposal. Couple of suggestions.

Frawless (approved):

ppatierno (approved):

im-konge (approved):

tombentley (commented):
I left a few nits. 

Maybe I missed it, but was there an answer to this previous question:

> Do we really need to support regexes, or would a whitelist be enough?

I think it would be good to address that in the proposal. 

We should also be clear about what flavour of regex we would support. The proposal seems to be using `java.util.regex.Pattern`, which would mean we'd be stuck with supporting that. Do we know whether that gives acceptable performance for the kinds of patterns we expect? (See https://swtch.com/~rsc/regexp/regexp1.html if you don't understand why we should at least _think_ about the performance impacts of adding support for arbitrary regex evaluation)

Finally, I think it would be good to be clear about where else in Strimzi we already support label filtering (you link to https://github.com/strimzi/strimzi-kafka-operator/pull/4791 but didn't mention exactly which things that PR affected).  You should also spell out what the consequences are for the access operator.

039-user-operator-custom-filter-for-labels-.md
scholzj: This should still be addressed @see-quick

see-quick:
> Do we really need to support regexes, or would a whitelist be enough?

Well, in https://github.com/strimzi/strimzi-kafka-operator/pull/4791/files, we introduce support for regexes. So I think it would be good to be consistent (from the user's POV). On the other hand, we could possibly have some specific pattern, which would whitelist but I am more inclined to the first approach (i.e., support regexes).

> We should also be clear about what flavour of regex we would support. The proposal seems to be using java.util.regex.Pattern, which would mean we'd be stuck with supporting that. Do we know whether that gives acceptable performance for the kinds of patterns we expect? (See https://swtch.com/~rsc/regexp/regexp1.html if you don't understand why we should at least think about the performance impacts of adding support for arbitrary regex evaluation)

After reading this article, I am convinced that we can give it to the `Thomson implementation, but then we could also change it in the https://github.com/strimzi/strimzi-kafka-operator/pull/4791/files, where we introduce `java.util.regex.Pattern`. But I am not a regex expert so I think it would be great to hear more opinions from others :). 

> Finally, I think it would be good to be clear about where else in Strimzi we already support label filtering (you link to https://github.com/strimzi/strimzi-kafka-operator/pull/4791 but didn't mention exactly which things that PR affected)

Well, this is a kinda widespread question but having a quick look at the code I see that we do only whitelisting whether its basic Kubernetes annotations (i.e., starting with app.kubernetes.io/) or strimzi domain (i.e., io.strimzi.operator.common.model.Label#additionalLabels method where we do only whitelist using `Labels.STRIMZI_DOMAIN` and `Labels.STRIMZI_CLUSTER_LABEL`). 
 
Also in the `Labels.java`, we do label filtering based on input regex. 

```java
/**
     * @param resource The resource to get the labels of.
     * @return A new instance with filtered labels added from the given {@code resource}.
     */
    public static Labels fromResource(HasMetadata resource) {
        Map<String, String> additionalLabels = resource.getMetadata().getLabels();

        if (additionalLabels != null) {
            additionalLabels = additionalLabels
                    .entrySet()
                    .stream()
                    .filter(entryset -> !STRIMZI_LABELS_EXCLUSION_PATTERN.matcher(entryset.getKey()).matches())
                    .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
        }

        return additionalLabels(additionalLabels);
    }
``` 

> You should also spell out what the consequences are for the access operator.

TBH I am not aware of any apart from the labels we would filter by specified regex.

039-user-operator-custom-filter-for-labels-.md
see-quick: So what about something like:
```
The user only needs to modify either User Operator deployment in the case of a standalone approach or Kafka's custom resource to use such a feature.
```

scholzj: For example ... make it clear that this is something available and configurable in both.

scholzj (approved):

tombentley (approved):
Thanks @see-quick!

PR #66, opened 2022-11-21
feat: Add proposal for KRaft liveness and readiness

katheris:
Signed-off-by: Katherine Stanley <11195226+katheris@users.noreply.github.com>


kraft-liveness-readiness.md
```diff
@@ -0,0 +1,112 @@
...
+
+There are three different types of liveness/readiness probe that we need to consider, first for a combined node, 
+then for a broker and then for a controller.
+
+### Combined mode
+
+The proposed probes are:
+
+* Liveness: node is listening on 9090 and 9091
+* Readiness: node is listening on 9090 and 9091
```
scholzj, line 55: I really wonder if there is a situation where 9090 will fail and 9091 will pass for example. Do we really need to check both?

kraft-liveness-readiness.md
```diff
@@ -0,0 +1,112 @@
...
+
+* Liveness: node is listening on 9090
+* Readiness: node is listening on 9090
+
+### Broker only mode
+
+The proposed probes are:
+
+* Liveness: the BrokerState metric < 3 || listening on 9091
+* Readiness: the BrokerState metric >=3 && != 127
```
scholzj, line 69: Why is this not part of the combined node readiness as well? For example in an similar form to how it works today with ZooKeeper when the Readiness probe essentially changes after reaching some point?

scholzj (commented):
Have you considered use of Startup probes as well? I guess now with being Kube 1.19+ we can expect them to be available. Not sure it would add any value ... just wondering.

README.md
```diff
@@ -9,6 +9,7 @@ This repository list of proposals for the Strimzi project. A template for new pr
 
 |  #  | Title                                                                 |
 | :-: |:----------------------------------------------------------------------|
+| 43  | [KRaft liveness and readiness probs](./43-kraft-liveness-readiness.md) |
```
mimaison, line 12: probes?

kraft-liveness-readiness.md
```diff
@@ -0,0 +1,112 @@
+# Liveness and Readiness probes in a KRaft Kafka cluster
```
mimaison, line 1: Should we file be named `43-kraft-liveness-readiness.md` ?

kraft-liveness-readiness.md
katheris: I've explained in the rejected alternatives section why the BrokerState metric check isn't included in the combined node readiness check. Does that answer your question sufficiently or is there something still unclear there?

kraft-liveness-readiness.md
katheris: The ports are used for two different things, so I think for a combined mode to work correctly it would need to be listening on both. That being said, by looking at the code and experimenting I haven't found a scenario where it would get to the point of listening on the replication port (9091) and not be listening on 9090, since the controller starts up first and then starts the broker. So theoretically (as far as I can tell) we could only check 9091, but is it much overhead to just check both?

kraft-liveness-readiness.md
katheris: Good spot, I've aligned the file names properly now

043-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,112 @@
...
+
+There are three different types of liveness/readiness probe that we need to consider, first for a combined node, 
+then for a broker and then for a controller.
+
+### Combined mode
+
+The proposed probes are:
+
+* Liveness: node is listening on 9090 and 9091
+* Readiness: node is listening on 9090 and 9091
```
katheris, line 55: Adding comment from @scholzj that got lost in the file rename:

[scholzj]
I really wonder if there is a situation where 9090 will fail and 9091 will pass for example. Do we really need to check both?

[katheris]
The ports are used for two different things, so I think for a combined mode to work correctly it would need to be listening on both. That being said, by looking at the code and experimenting I haven't found a scenario where it would get to the point of listening on the replication port (9091) and not be listening on 9090, since the controller starts up first and then starts the broker. So theoretically (as far as I can tell) we could only check 9091, but is it much overhead to just check both?

043-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,112 @@
...
+
+* Liveness: node is listening on 9090
+* Readiness: node is listening on 9090
+
+### Broker only mode
+
+The proposed probes are:
+
+* Liveness: the BrokerState metric < 3 || listening on 9091
+* Readiness: the BrokerState metric >=3 && != 127
```
katheris, line 69: Adding comment from @scholzj that got lost in the file rename:

[scholzj]
Why is this not part of the combined node readiness as well? For example in an similar form to how it works today with ZooKeeper when the Readiness probe essentially changes after reaching some point?

[katheris]
I've explained in the rejected alternatives section why the BrokerState metric check isn't included in the combined node readiness check. Does that answer your question sufficiently or is there something still unclear there?

scholzj: I get that it cannot be relied upon exclusively. But can't it be used in the second phase, after the quorum is formed?

I mean, long term, we plan to support all 3 modes so one would expect that the probes should be in the long term good for all modes. So this seems a bit weird:
* If checking for listening on 9091 is fine in combined mode, should it be good enough in broker only mode?

Or the other way around
* If checking the port 9091 is not good enough for broker-only mode, why is it good enough for combined mode?

This makes me a bit nervous TBH as it seems there is some gap. The rejected alternatives explain some of the issue. But I don't think it completely answers my question to about switching what the probe does after the broker starts working.

Maybe it would be also good to see if we have some suggestion for Kafka how to improve this / or get some suggestion from Kafka how to do this, since this is no way unique issue for Strimzi. That might be beyond the scope of this proposal - but it is worth to at least cover that as a prerequisite to do something better.

katheris: I think in an ideal world we would check the BrokerState metric once the quorum is formed. So in combined we would use the current-state metric to determine whether the BrokerState should be checked or not. Then in broker only mode we can just always check BrokerState and the pod won't become ready until the quorum is formed.

However, the problem we have is in combined mode we can't get to the current-state metric via the Kafka Agent since it isn't a Yammer metric. So we have two options:
1. Ignore the BrokerState metric for now in combined mode and see if we hit problems where the brokers get stuck in the STARTING or RECOVERY phase
2. Change the way the Kafka Agent works, or find some other workaround to get hold of the current-state metric

scholzj: I thought that the Yammer metrics issue is about the controller states but not the broker states.

katheris: Yes, but we need a mechanism to know when to check the broker state. We could e.g. have the check be `(brokerState >= 3 && brokerState != 127) || listening on 9091` but is that actually any better than just checking if it's listening on 9091. Because you would never know if the first part ever succeeds.

Thinking about it this way, I feel like the broker **liveness** check should just be listening on 9091 shouldn't it? What's the benefit of having the second part of the check? It doesn't actually enforce that we ever reach that state

scholzj: Yeah,  ut as I said ... if `listening on 9091` is good enough why not use it in the broker. So one way or another, this shows some limitation which needs to be addressed going long-term. I think the operator will anyway need to know if the quorum is formed or not ... so we will anyway need to get this information at some point I guess.

043-kraft-liveness-readiness.md
tombentley: Having looked closely at the `BrokerServer.scala` code (which is used for the `KafkaRaftServer`, i.e. KRaft mode, not ZK model) in `startup()` it:

1. creates the LogManager, but does not start it (i.e. log recovery is deferred)
2. instantiates the `SocketServer` (which results in the `ServerSocket#bind()` call, which results in the `bind` and `listen` syscalls, which result in the kernel putting the socket in the `LISTEN` state).
3. Much later in `startup()` it instantiates the `BrokerMetadataPublisher` which calls `LogManager#startup()` (initiating log recovery) as a side effect of handling Raft commit or snapshots 
4. Only then does `socketServer.enableRequestProcessing()` get called and the broker starts actually accepting queued client requests.

In other words, for a KRaft broker $\texttt{LISTEN} \mathrel{\rlap{\hskip .5em/}}\Longrightarrow brokerState > RECOVERY$. Which is different from a ZK-based broker. The question is how much this is really a problem, given a KRaft server with `process.roles=broker` explicitly binds and listens before starting recovery. So long as the broker can serve `Metadata` requests (which is will be able to do by step 4, I think) the rest of the Kafka protocol should guarantee that a client should be able to bootstrap. It will (eventually) find that the broker isn't the leader for any partitions (while that broker is in log recovery). 

In other words, I'm thinking that perhaps the readiness check we should be using for KRaft (controller, broker or combined) is not on the broker state or server socket LISTEN state, but rather whether we can make a TCP connection from a client (i.e. a client socket gets into the `ESTABLISHED` TCP socket state).

katheris: If I understand what you've said above and the code correctly, for a server with `process.roles=broker` there are three events that happen that we could check readiness on. First it starts listening on the replication port, then (after loading metadata from the controller quorum) it starts accepting TCP connections, then finally it finishes log recovery and moves to a RUNNING state. And it's only at the final step that the broker is actually eligible to be a leader for any topic-partitions. Is that correct?

If so, wouldn't checking for TCP connections mean we mark the broker as ready when it is still doing log recovery? I guess this is at least a better check than just "is it listening", but I still feel unsure about what the risks are of marking a broker as ready if it's still in log recovery. I guess it would mean we wouldn't have the problem where the roller tries to roll the brokers while they are doing recovery?

For the combined mode, from my understanding of the above, if we mark the server as ready only when it starts accepting TCP connections to the replication port, then we would still hit the problem where if all the pods had been pending then a single server by itself can't become ready because there is no controller quorum so the broker endpoint would never start accepting TCP connections right?

tombentley: > For combined mode...

This is an excellent point. I think it's true that a controller will connect itself to participate in Raft, which it can only do if it's ready. In the single node case this will create a cyclic dependency. But it's probably also a problem even when #nodes > 1 (e.g. it might prevent the election of a leader if another voter is down and we a vote for ourselves would win us leadership but we can't connect). Perhaps we should simply not tie readiness to network.

043-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,112 @@
...
+This section contains some background information about the ports and metrics we could use for checks.
+
+### Ports
+In KRaft mode there are two different ports that the node could be listening on.
+If it has a controller process it should be listening on the port defined as the `controller.listener.names` which in Strimzi is currently hard-coded to 9090.
+If it has a broker process it should be listening on the replication port which is hard-coded to 9091.
+If the node is running as a combined mode, it should be listening on both ports
+
+### BrokerState
+The expected [BrokerState](https://github.com/apache/kafka/blob/trunk/metadata/src/main/java/org/apache/kafka/metadata/BrokerState.java) transitions are:
```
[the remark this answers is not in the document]
katheris: I guess what I meant by "expected" was "expected during a normal broker lifecycle, as you say the "happy-path". I can update the wording to make that clearer.

katheris:
Trying to bring together a few of the discussion threads I see that we have this fundamental sticking point: how to handle the fact that in combined mode the pods will need to be marked ready before the broker process is really ready because a quorum needs to be formed for the brokers to become ready.

I see we have three options:
1. In combined mode use the current-state metric to have a conditional readiness check that only checks broker readiness once a controller quorum has been formed
2. Have a more strict readiness check for the broker in broker only mode, than in combined mode
3. Choose a readiness probe for brokers that will pass whether the quorum is formed or not

I'm tempted to say we should bite the bullet and go ahead with option 1 and do the work to get hold of that metric. Mostly because this is the most rigorous check. However, I'm interested to know if anyone thinks options 2 or 3 are good enough.

@scholzj @tombentley

showuon:
I just checked the [`ruok` command source code](https://github.com/apache/zookeeper/blob/master/zookeeper-server/src/main/java/org/apache/zookeeper/server/command/RuokCommand.java#L32) in Zookeeper, because Zookeeper is also a quorum based system. And I found, the handler for `ruok` command in ZK server is actually "no-op". It directly return `imok` when receiving the `ruok` request. That said, the `ruok` command is actually doing a listener check, or request/response check only, instead of doing quorum check. From that point, I'm thinking the (2) or (3) is good enough. If we want to be more strict, maybe we can send a admin command and make sure it is responded as expected. WDYT?

scholzj:
So ...for readiness probes
* Controller node, is ready when it is ready to join the quorum. Not when it joined the quorum.
* Broker, is ready when it is ready to accept the connections from consumers and producers => that might need the quorum to be formed of course. In this case, the quorum does not represent the state of the cluster but the state of an external dependency required for the broker to run.
* Combined mode, combines both => so I would say it should be ready when the one which is _higher_ level is ready. And that would be the broker for me. 

For liveness probes, we need to indicate when the pod should be restarted
* For controller node, it can be likely the same. It really just needs to show that it is still running and alive (the ZooKeeper `ruok` thing is ideal for that).
* For broker node, the broker is alive when it is at least trying to connect to the Quorum even if it cannot succeed as long as it will try to reconnect => I guess we do not mind it crashlooping here, but running not ready might work better I guess?
* For combined node, it has to make sure the liveness of the lowest level component first => so liveness of the controller part (or liveness of the controller and broker parts if they are started in parallel). So in a way:

| Controller alive  | Controller ready | Controller Quorum | Broker alive | Broker Ready | Result                   |
| -------------     | -------------       | -------------          | -----------  | -------------  | ------                    |
| False                   | False                    | False                       | True             | False                | Not Alive / Not Ready |
| True                    | False                    | False                       | True             | False                | Alive / Not Ready |
| True                    | False                    | False                       | False            | False                | Not Alive / Not Ready _(If they are started in parallel)_ |
| True                    | True                     | False                       | True             | False                | Alive / Not Ready |
| True                    | True                     | True                        | True             | False                | Alive / Not Ready |
| True                    | True                     | True                        | True             | True                 | Alive / Ready        |

katheris:
> So The fact that the combined mode is marked as ready only when the broker is ready should not prevent it from establishing the quorum.

The problem I am seeing is if all the pods start in a pending state (e.g. due to node constraints), when you remove the node constraints the KafkaRoller will wait for each pod to be ready before it tries to restart the next one. That is why the system tests in my PR were failing. So we have two options there, either we adjust the readiness probe to account for that or the KafkaRoller should notice that pods are in pending state and restart them all rather than trying one by one.

But I suspect the latter change might be quite complicated as I think the KafkaRoller focuses on one pod at a time right? Maybe if it times out waiting for a pod to be ready and the pod is stuck in pending it could note that the pod is pending and at that point check all pods for a pending state and restart all pending pods in one go.

scholzj:
> The problem I am seeing is if all the pods start in a pending state (e.g. due to node constraints), when you remove the node constraints the KafkaRoller will wait for each pod to be ready before it tries to restart the next one. That is why the system tests in my PR were failing. So we have two options there, either we adjust the readiness probe to account for that or the KafkaRoller should notice that pods are in pending state and restart them all rather than trying one by one.

I don't really follow this ... There is a mechanism in the roller to detect pods in pending state and roll them if something changed in the pod configuration. In your case, some affinity rules should have changed and it should have been rolled. So, maybe this should be fixed there?

As part of the roller improvements, the roller should have better awareness of the broker state. So maybe it can be improved there that it understands that the pod is not ready but is waiting for quorum and should be requeued at the end.

If we need to temporarily call it ready before the broker is ready, I can live with that. But we need to somewhere cover the desired state as well. And I'm a bit confused if this is here or we want to cover only the initial work here ;-)

katheris:
The KafkaRoller does roll the pods that are in pending, but it rolls one first, waits for it to report ready, then rolls the second etc. I think this proposal should describe the final desired state, but I'm just flagging that if we decide the final desired state is what you described above (which I think is a good proposal btw) then to implement it we will likely need different logic in the KafkaRoller or to change our system tests in some way as otherwise the regression system tests will fail. That isn't a problem, just something to note so it isn't a surprise later.

> As part of the roller improvements, the roller should have better awareness of the broker state.

I think this is a good idea, and lines up with the suggestion from Tom that actually the KafkaRoller in KRaft mode needs to not rely on pod readiness alone to know what to do.

scholzj:
> I think this proposal should describe the final desired state

So maybe one of the ways to deal with this is to describe here all the different states, the ways how to check them and how they can be used. And you can establish different phases and how they will use these ... e.g. phase one, we use this, phase two, we improve the roller and use this etc.

And then either here on a separate proposal we can continue with the _agent_ collecting these states and exposing them. That will make it easier to pick it up later. (just an idea how to move forward with this)

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,156 @@
...
+
+As part of this change the readiness probe for a Kafka broker in a ZooKeeper based cluster will be updated to 
+continue being marked as ready when the BrokerState metric > 3 && != 127.
+This means it will tolerate the PENDING_CONTROLLED_SHUTDOWN and SHUTTING_DOWN states.
+This will not cause any compatibility problems because currently the KafkaAgent stops running once the metric reaches 
+the RUNNING state.
+The change will make the code between KRaft and ZooKeeper mode simpler and protect readiness if in future the agent is 
+updated to continue running once the broker is "ready".
+
+## Rejected alternatives
```
scholzj, line 156: You probably don't need the header if you don' have any content.

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,156 @@
...
+    * "alive" if it has a process running.
+    * "ready" when it is ready to start accepting producer/consumer requests.
+
+### Controller only mode
+
+The proposed probes are:
+
+* Liveness: controller is listening on 9090
+* Readiness: it is possible to make a TCP connection to port 9090 on the controller
+(i.e. client socket gets into the `ESTABLISHED` TCP socket state)
```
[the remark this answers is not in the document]
katheris: Since there is a gap between the process starting to listen on that port and actually accepting connections, the intention was that the readiness check does involve actually trying to connect. For the liveness check I guess we would stick with the existing test we've used elsewhere of:
```
netstat -lnt | grep -Eq 'tcp6?[[:space:]]+[0-9]+[[:space:]]+[0-9]+[[:space:]]+[^ ]+:9090.*LISTEN[[:space:]]*'
```

Then for the readiness we could use something like netcat couldn't we? E.g. `nc -w 5 -v localhost 9090 </dev/null; echo $?`

044-kraft-liveness-readiness.md
katheris: Good point, I've removed it

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Current state metric and quorum
+
+Some useful things to be aware of with the current-state metric and how the controller quorum affects the broker nodes:
+
+* The [current-state](https://kafka.apache.org/documentation/#kraft_quorum_monitoring) metric indicates whether a controller is currently a leader or follower.
+If this metric shows one of these two states, this means the leader election has happened and a controller quorum has been successfully formed.
+* The current-state metric is not accessible as a Yammer metric, so cannot be accessed by the KafkaAgent in the same way it currently reads the BrokerState metric.
+* The current-state metric could perhaps be read by the KafkaAgent if it called the MBean server within the JVM directly or read it from `$LOG_DIR/__cluster_metadata-0/quorum-state`.
+* The broker nodes will not move to a RUNNING state until the quorum leader election has happened.
```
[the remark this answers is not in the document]
katheris: I believe it will get stuck in the STARTING state so I've changed it to say that

044-kraft-liveness-readiness.md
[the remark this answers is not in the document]
scholzj: I suspect this would create terrible mess in the broker logs ... there will be errors about TLS handshake etc. So I would be very careful about doing any `nc` stuff to it.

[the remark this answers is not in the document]
katheris: I hadn't considered the impact that might have on the controller logs, that could lead users to think they have a misbehaving client, when the actually don't. Looking at https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/server/ControllerServer.scala there seems to be fewer actions taking place between the controller starting to listen and accepting connections, so perhaps there is minimal benefit to having separate checks. @tombentley what do you think? Would it be reasonable to have the same liveness and readiness checks for the controller? Specifically have the check be that the controller is listening on the port.

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Proposal
+
+There are three different types of liveness/readiness probe that we need to consider, first for a controller only, then 
+for a broker only, and finally for a combined node that has both roles.
+
+The following statements describe the intent of the proposed probes:
+
+* A controller is:
+    * "alive" if it has a process running.
```
tombentley, line 54: If this is really the check we want why not use `ps`? Or an HTTP server running in the server for which the `bind`/`listen`/`accept` doesn't depend deeply on anything about the Kafka code?

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,154 @@
...
+    * "ready" when it is ready to join the controller quorum.
+* A broker is:
+    * "alive" if it has a process running.
+    * "ready" when it is ready to start accepting producer/consumer requests.
+* A combined node is:
+    * "alive" if it has a process running.
+    * "ready" when it is ready to start accepting producer/consumer requests.
```
tombentley, line 61: A sentence here about this `readiness` not blocking connections from other controllers would really help clarify, I think.

tombentley (commented):
Thanks! I _think_ (purely based on reading the code and thought experiements) that this should work. Have you tried to implement it?

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,154 @@
...
+NOT_RUNNING (0) -> STARTING (1) -> RECOVERY (2) -> RUNNING (3) -> PENDING_CONTROLLED_SHUTDOWN (6) -> SHUTTING_DOWN (7)
+
+The other possible value is 127 which represents UNKNOWN.
+
+### Current state metric and quorum
+
+Some useful things to be aware of with the current-state metric and how the controller quorum affects the broker nodes:
+
+* The [current-state](https://kafka.apache.org/documentation/#kraft_quorum_monitoring) metric indicates whether a controller is currently a leader or follower.
+If this metric shows one of these two states, this means the leader election has happened and a controller quorum has been successfully formed.
```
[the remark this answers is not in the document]
katheris: Good point, I've moved it into rejected alternatives because I think it is useful information to be aware of when reading the proposal

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,154 @@
...
+* The current-state metric is not accessible as a Yammer metric, so cannot be accessed by the KafkaAgent in the same way it currently reads the BrokerState metric.
+* The current-state metric could perhaps be read by the KafkaAgent if it called the MBean server within the JVM directly or read it from `$LOG_DIR/__cluster_metadata-0/quorum-state`.
+* The broker nodes will not move out of the STARTING state until the quorum leader election has happened.
```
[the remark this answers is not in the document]
katheris: Thanks, I've added this

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,154 @@
...
+This is similar to the current behaviour of brokers when ZooKeeper is not ready.
+
+### Combined mode
+
+The proposed probes are:
+
+* Liveness: node is listening on 9090
+* Readiness: the BrokerState metric >=3 && != 127
+
+**Note:** This means the nodes will not become ready until all the other nodes
```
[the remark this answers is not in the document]
katheris: I hadn't really though through combining pure brokers and combined-mode brokers, KRaft mode definitely opens us up to all sorts of architecture choices!

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,154 @@
...
+are up and running.
+This is acceptable because the Strimzi headless services use "publishNotReadyAddresses", which
```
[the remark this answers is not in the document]
katheris: I don't think this will be true for pure controller nodes, because they will become ready as soon as they are listening on the controller port. So that means they don't need to talk to each other until they are deemed "ready".

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,154 @@
...
+* Readiness: the BrokerState metric >=3 && != 127
+
+**Note:** This means the nodes will not become ready until all the other nodes
+are up and running.
+This is acceptable because the Strimzi headless services use "publishNotReadyAddresses", which 
+means the nodes will be able to communicate even if they are currently marked as not ready.
+
+### Impact on the KafkaRoller
+
+The existing KafkaRoller checks whether a pod is marked as ready to determine whether it needs to be rolled.
```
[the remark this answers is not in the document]
katheris: I've expanded this section to include the change about observing the state directly

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Impact on the KafkaRoller
+
+The existing KafkaRoller checks whether a pod is marked as ready to determine whether it needs to be rolled. 
+It does not take into account the state of other broker pods.
+The move to KRaft mode introduces some new requirements in terms of when a pod should or should not be rolled:
+
+* If a broker pod has not become ready, KafkaRoller should check the controller quorum is formed before rolling the pod.
+* If a combined pod has not become ready, KafkaRoller should check that all other combined pods have been scheduled 
+(i.e. not in pending state) before rolling the pod or waiting for it to become ready.
```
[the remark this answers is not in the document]
katheris: I've added a sentence for pure controllers, the only requirements I could think of was rolling the controllers that weren't the active controller first.

katheris:
@scholzj @tombentley I've made some updates to the proposal. The commit lists all the changes, but the most significant one is I've changed the controller only readiness probe to be 
```
controller is listening on the port of the first address in `controller.listener.names` (9090)
```
based on the discussions we had about noise in the logs and not much happening between the controller starting to listen and actually accepting connections.

This does mean that the controller liveness and readiness are now identical. Tom also pointed out that if the log recovery of the metadata topic takes a long time, then that would delay the controller from listening on the port and would therefore mean the liveness probe might take it out.

With this in mind, I suggest we change the liveness checks for both the controller and the combined nodes to be something else. Tom previously suggested either a basic `ps` or have an HTTP server that starts up separately from the Kafka process. I think either of these would be acceptable, I'm currently leaning towards the basic `ps` request. What do you think?

tombentley:
> Tom previously suggested either a basic ps or have an HTTP server that starts up separately from the Kafka process. I think either of these would be acceptable, I'm currently leaning towards the basic ps option. What do you think?

`ps` would be quicker to implement. But I can also see that HTTP-based might be better, given we expect to run an HTTP server in there _anyway_ (to expose metrics). I also have a (possibly unfounded) dislike for `exec` probes.

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,173 @@
...
+
+* A controller is:
+    * "alive" if it has a process running.
+    * "ready" when it is ready to join the controller quorum.
+* A broker is:
+    * "alive" if it has a process running.
+    * "ready" when it is ready to start accepting producer/consumer requests.
+* A combined node is:
+    * "alive" if it has a process running.
+    * "ready" when it is ready to start accepting producer/consumer requests.
```
ppatierno, line 56: does it mean it's already joined the controller quorum right? could we make it clearer?

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,173 @@
...
+* If a combined pod has not become ready, KafkaRoller should check that all other combined pods have been scheduled 
+(i.e. not in pending state) before rolling the pod or waiting for it to become ready.
+
+The reason for the final bullet (combined pod not being ready) might not be immediately obvious, so it is explained below.
+The brokers do not move to RUNNING until a quorum has been formed.
+For example in combined mode during normal startup the following would happen:
+
+* All the pods are started at the same time
+* The controller processes in each pod form a quorum
+* The broker processes start up and start talking to the controller quorum so the controller unfences them and they move to STARTED and are marked as ready
```
ppatierno, line 108: is `STARTED` actually the `RUNNING` state?

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,173 @@
...
+* All pods are now ready
+
+However, if the pods are started initially with affinity constraints so they are all in pending, then the constraint is removed, the following would happen:
+
+* The first pod is scheduled
+* The controller process starts up but cannot form a quorum because the other controllers are missing
+* The broker process starts up but since there is no controller quorum it does not move to STARTED, it stays in STARTING. This means the pod is not marked as ready
+* Because this pod doesn't become ready the other pods aren't scheduled, so the pod just sits in crash loop backoff
```
ppatierno, line 116: AFAIU, the above four lines sound to be an explanation of why the Readiness for "Broker only mode" and "Combined mode" is `the BrokerState metric >=3 && != 127`. If I got it right, would it make sense to provide this explanation in the previous section?

044-kraft-liveness-readiness.md
showuon: I think we should mention this is the "minimal exception", not exactly what we will do. Otherwise, it's a little confusing when seeing "we just want to check if it has a process running" here, and then seeing "we check controller is listening on the port" below.

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,173 @@
...
+* Update the KafkaAgent to check for a broker state of >= 3 && != 127.
+This will be used for the broker only and combined readiness checks later and will improve the existing ZooKeeper based broker check.
+
+### Phase 2 - Strimzi adds support for broker only and controller only modes
+* Controller only pod liveness and readiness checks are fully implemented to match this proposal.
+* Broker only pods are marked as both alive and ready if listening on port 9091.
+* Combined mode pods are marked as both alive and ready when the pod is listening on 9090 (no change from phase 1).
+
+### Phase 3 - KafkaRoller is updated
+* KafkaRoller is updated to take the existence of controller pods and status of controller quorum into account when deciding whether to roll pods.
```
showuon, line 134: I don't think there is any dependency between "supporting broker-only and controller-only mode" and "KafkaRoller updated", right? Maybe we should note that the phase 2 and phase 3 is not strictly in this order?

044-kraft-liveness-readiness.md
katheris: I've added a sentence to clarify this.

044-kraft-liveness-readiness.md
katheris: Yes good spot, I've updated it

044-kraft-liveness-readiness.md
katheris: The above four lines explains why the KafkaRoller should verify that the controller pods are not stuck in pending before waiting for broker pods to be ready. So I think it is right that it sits here. This proposal states that the KafkaRoller should stop relying on readiness to determine whether to roll pods and use it's own checks. So I think having this further up where we are discussing the readiness checks would muddle those back up again.

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,173 @@
...
+### Impact on the KafkaRoller
+
+The existing KafkaRoller checks whether a pod is marked as ready to determine whether it needs to be rolled. 
+This relies on the fact that currently the pod being ready implies that the BrokerState is >= RUNNING.
+Additionally, it does not take into account the state of other broker pods.
+The move to KRaft mode introduces some new requirements in terms of when a pod should or should not be rolled.
+This combined with the varying readiness checks means we need to change the way the KafkaRoller works:
+
+* The KafkaRoller should observe the BrokerState metric and any other metrics it needs directly, rather than inferring the state based on readiness.
+* If more than one controller pod has not become ready, KafkaRoller should try to role the controllers that aren't the current active controller first.
```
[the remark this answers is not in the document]
katheris: Good spot, fixed

044-kraft-liveness-readiness.md
katheris: I think there is a dependency. The KafkaRoller cannot be updated to understand the difference between broker and controller nodes until they are supported as part of Strimzi. These two phases could be done together in the same PR maybe but I don't think it makes sense to do phase 3 and then phase 2 as you would have to make guesses about how phase 2 would be implemented.

044-kraft-liveness-readiness.md
katheris: I'm not sure what you mean by "minimal exception". I put above that these statements "describe the intent of the proposed probes". So the intent is that the controller should be marked alive if there is some sort of process running, then lower down we describe how we actually work out that a process is running. So that could be by seeing that it is listening on a port, or as Tom says by a basic `ps` check or running some other process in the container.

katheris:
@scholzj @showuon @ppatierno @tombentley I think I've addressed everyones comments on this.

The only outstanding piece I am unsure on is in the current proposal the liveness and readiness checks for the controller only node are identical and check if the controller is listening. Tom pointed out that if log recovery takes a long time then the controller won't listen until it has completed so Kubernetes might try to restart the pod.

So I would propose we change the liveness check for controller only and combined mode to something else. The options that have been suggested so far are either to use some sort of `ps` command to see that the process has started, or to run a separate http server and use that to assert that the container is running. What does everyone think?

044-kraft-liveness-readiness.md
showuon: Make sense. Thanks for the explanation.

katheris:
> But could you explain more about the "http server"? What's it for and how we're going to create it? If it's just for liveness probe, will it overkill to start a separate http server?

Yeah it would just be an endpoint we can call from the liveness script. Since the KafkaAgent is going to be updated to host an endpoint for metrics we could perhaps use it to also host an endpoint for the liveness check. I am personally leaning towards a `ps` type check as I do think the additional endpoint might be overkill

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,174 @@
...
+| ZK-based broker | if (ready) { listening on 9091 } else { have ZK session } | if (ready) { return true } |
+
+The "ready" check in both cases is based on the existence of the file `../kafka-ready` which is 
+created by the KafkaAgent when the broker state metric has value 3 (RUNNING).
+
+## Background
+
+This section contains some background information about the ports and metrics we could use for checks.
+
+### Ports
```
scholzj, line 23: Nit ... empty line after the title will make the readability in raw mode batter.

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,174 @@
...
+In KRaft mode there are two different ports that the node could be listening on.
+If it has a controller process it should be listening on the port defined as the `controller.listener.names` which in Strimzi is hard-coded to 9090.
+If it has a broker process it should be listening on the replication port specified as the `inter.broker.listener.name` which is hard-coded to 9091.
+If the node is running as a combined mode, it should be listening on both ports
+
+### BrokerState
```
scholzj, line 29: Nit ... empty line after the title will make the readability in raw mode batter.

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,174 @@
...
+## Current situation
+
+These are the current probes used in Strimzi when ZooKeeper is present:
+
+|Component| Liveness | Readiness |
+|---|---|---|
+| ZooKeeper | "imok" from 127.0.0.1:12181 | "imok" from 127.0.0.1:12181 |
+| ZK-based broker | if (ready) { listening on 9091 } else { have ZK session } | if (ready) { return true } |
+
+The "ready" check in both cases is based on the existence of the file `../kafka-ready` which is
```
scholzj, line 16: Maybe you can clarify that both cases here means `Kafka liveness and readiness probe` and `ZooKeeper and Kafka` => I think it can be a little bit confusing with having both Kafka and Zoo in the Table right above.

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,174 @@
...
+### Combined mode
+
+The proposed probes are:
+
+* Liveness: node is listening on the port of the first address in `controller.listener.names` (9090)
+* Readiness: the BrokerState metric >=3 && != 127
+
+**Note:** This means the nodes will not become ready until a majority of the other controllers are up and running.
+This is acceptable because the Strimzi headless services use "publishNotReadyAddresses", which 
+means the nodes will be able to communicate even if they are currently marked as not ready.
```
scholzj, line 87: The part `is listening on the port of the first address in` in this section is a bit confusing. Neither `controller.listener.names` nor `inter.broker.listener.name` contain any address the broker is listening to. They contain the listener name. It is also not configurable anyway - neither the name and the port.

So this leaves me a bit confused. Does the way it is specified here mean some sophisticated way to parse out the listener address fro the configuration? Or is it really just an overcomplicated thing to say that you will grep netstat output for listening on port 9090 / 9091? If the second, please simplify it to make it a bit more clear what does it mean.

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,174 @@
...
+### Phase 1 - Strimzi only supports combined mode
+* Combined mode pods are marked as both alive and ready when the pod is listening on 9090. 
+This is an improvement on the current state where the pod is marked as alive and ready as soon as it starts up.
+* Update the KafkaAgent to check for a broker state of >= 3 && != 127.
+This will be used for the broker only and combined readiness checks later and will improve the existing ZooKeeper based broker check.
+
+### Phase 2 - Strimzi adds support for broker only and controller only modes
+* Controller only pod liveness and readiness checks are fully implemented to match this proposal.
+* Broker only pods are marked as both alive and ready if listening on port 9091.
+* Combined mode pods are marked as both alive and ready when the pod is listening on 9090 (no change from phase 1).
```
scholzj, line 132: These will likely be shell scripts -> so do they need to wait for the actual support for the broker and controller modes? OR can they be added right away just not used for the time being?

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,174 @@
...
+The proposed probes are:
+
+* Liveness: node is listening on the port of the first address in `controller.listener.names` (9090)
+* Readiness: the BrokerState metric >=3 && != 127
+
+**Note:** This means the nodes will not become ready until a majority of the other controllers are up and running.
+This is acceptable because the Strimzi headless services use "publishNotReadyAddresses", which 
+means the nodes will be able to communicate even if they are currently marked as not ready.
+
+### Impact on the KafkaRoller
```
scholzj, line 89: From here on, the proposal seems to be quite confusing in terms of what is proposes, what is being approved and what should be done in the future. I think that need to be clarified and we need to make sure there is a clear deliverable (or make it clear that there is no real deliverable).

It seems to me that, this proposal:
* Suggests how the liveness and readiness should be implemented in phase 4
* Leaves the phase 3 completely unspecified, so essentially this proposal cannot be done until the phase 4 is specified in some proposal and approved?
* The content of the `Impact on the KafkaRoller` chapter makes sense in general, but it seems IMHO too simplified to take it as something to be approved. So I think it should have some clear note that this will be covered in detail in another proposal? (at least that is what I assume?)

----

To be clear, I do not have problem with having this agreed upon before the other parts are done. It should be just more clear what are we agreeing upon and make it clear what is just outline of something what will be discussed and approved in separate proposal (e.g. the `Impact on the KafkaRoller` part).

044-kraft-liveness-readiness.md
katheris: This was a change that was proposed by @tombentley, originally it just stated the port directly as you suggest. My assumption was that in reality the implementation would just check for port 9090/9091, however I thought it was useful in the proposal to be specific that those ports actually represent e.g. the port for the controller listener, or the port for the inter broker listener.

044-kraft-liveness-readiness.md
katheris: The scripts could be written in advance, however these phases are meant to describe the actual checks that can be used, so if we did write them in advance they would have to be disabled.

I think the advantage of waiting is we will need a mechanism to determine whether the node is running in broker only or controller only or combined mode, and it might be that once we have added support for those modes it becomes clearer what mechanism that is. For example you could check the server properties, but it might be we end up needing an env var on the pod, and if so it would be simpler to use that.

044-kraft-liveness-readiness.md
katheris: I see what you mean, I've updated the wording to try to make it clearer what is part of this proposal and what parts are more for information to help inform a separate proposal for KafkaRoller in KRaft mode. I've also merged phase 3 and 4 below, since the KafkaRoller changes wouldn't directly be covered by this proposal, it is just required before the final changes can be made to the liveness and readiness checks. Hopefully that helps clarify.

katheris:
@tombentley @ppatierno @scholzj do any of you have a strong opinion on whether the controller liveness check should be done with a `ps` type call or by having some separate endpoint we stand up? I think so far Tom has specified a preference for an endpoint and both myself and Luke favour a `ps` type check.

tombentley:
@katheris I don't object to using an `exec` probe at all. I only have a slight preference for the `http` probe as I think the operator (specifically the roller) will anyway end up using an endpoint to get the broker state. Since we'll end up with an HTTP server in the broker process _anyway_, then it seemed using it for the probe wasn't much of a reach.

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,178 @@
...
+## Proposal
+
+There are three different types of liveness/readiness probe that we need to consider, first for a controller only, then 
+for a broker only, and finally for a combined node that has both roles.
+
+The following statements describe the intent of the proposed probes:
+
+* A controller is:
+    * "alive" if it has a process running.
+    * "ready" when it is ready to join the controller quorum.
```
ppatierno, line 52: just to be sure I got this right. Isn't "ready" when it joined a controller quorum? or actually when it's ready to join but not joined yet? I would expect the former, or? Maybe just a matter of how it sounds in English vs Italian? :-)

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,178 @@
...
+* All pods are now ready
+
+However, if the pods are started initially with affinity constraints so they are all in pending, then the constraint is removed, the following would happen:
+
+* The first pod is scheduled
+* The controller process starts up but cannot form a quorum because the other controllers are missing
+* The broker process starts up but since there is no controller quorum it does not move to RUNNING, it stays in STARTING. This means the pod is not marked as ready
+* Because this pod doesn't become ready the other pods aren't scheduled, so the pod just sits in crash loop backoff
+
+## Proposed order of tasks to complete this proposal
```
ppatierno, line 126: These 3 phases are now describing the "listening" mentioning the ports directly (9090 and 9091) and not what's in the `inter.broker.listener.name` and `controller.listener.names`. Would it be better having consistency?

ppatierno:
@katheris I have no strong opinion, but I would agree with @tombentley if we are going to have an HTTP server anyway at some point for the KafkaRoller.

044-kraft-liveness-readiness.md
ppatierno: Ok, makes sense now, thanks.

044-kraft-liveness-readiness.md
scholzj: I think it sounds weird and confusing this way. First, there is no address, and second, it might be read as that you do some parsing of the configs. Maybe you can then say something like this?

```
* Liveness: node is listening on the port 9090 which is configured in `controller.listener.names`
```

or something like that?

scholzj:
> @tombentley @ppatierno @scholzj do any of you have a strong opinion on whether the controller liveness check should be done with a ps type call or by having some separate endpoint we stand up? I think so far Tom has specified a preference for an endpoint and both myself and Luke favour a ps type check.

No, there is no HTTP server or anything like that proposed here. So this has to be an exec probe.

I get it that people have some plans and visions of how to develop certain features and that these might be still in preparation behind closed doors before the proposal is published. That is fine. But these cannot be leaking into existing proposals which are expected to be voted on. Each proposal that goes into the vote needs to be complete and consistent. So this proposal either has to wait for the other work to be complete and build on top of it or it has to work around it (such as by using an `exec` probe for the time being).

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,178 @@
...
+
+Since changes to the KafkaRoller and support for non-combined mode is needed to fully implement this proposal, the following phases of development are proposed:
```
scholzj, line 128: Maybe this gives a bit more clear statement:

```suggestion
The KRaft liveness and readiness probes can be implemented in 3 different phases.
Phases 1 and 2 can be implemented solely based on this proposal.
Phase 3 can be implemented only after additional improvements of the `KafkaRoller` which will be covered by a separate proposal.
```

ppatierno:
@scholzj or maybe adding the HTTP idea in the proposal just for this purpose right now and making clear that can be used for future improvements? What's wrong doing this way? I mean I agree that it should be in the proposal.

044-kraft-liveness-readiness.md
katheris: I've edited the part higher up to refer to the ports first and then mention they are connected to those specific configurations. I think here it's fine to just refer to the ports as that is what will ultimately be in the checks.

044-kraft-liveness-readiness.md
katheris: Good suggestion, I've updated it

katheris:
I've now updated this proposal to have an exec style check for the controller and combined node liveness checks. I will hold off rebasing this with the correct file name and fixing the conflict in the README.md until everyone is happy with the contents of the proposal, as renaming the file will make the comments less visible.

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,198 @@
...
+### Broker only mode
+
+The proposed probes are:
+
+* Liveness: broker is listening on the port 9091 which is configured in `inter.broker.listener.name`
+* Readiness: the BrokerState metric >=3 && != 127
+
+**Note:** This means the brokers will not become ready until a majority of the controllers 
+are up and running.
+This is similar to the current behaviour of brokers when ZooKeeper is not ready.
```
scholzj, line 78: Quick question ... With ZooKeeper it works today like this:
* When the broker is starting and ZooKeeper is not available (not running, not having a quorum), the broker crash-loops.
* When the broker was already running and ready (connected to ZooKeeper) and suddenly ZooKeeper loops quorum, Kafka will keep trying to reconnect, but it does not crash-loop or become not ready. It is still ready and does what it can without ZooKeeper.

How does this behave in KRaft - mainly in the second point? What will the readiness probe do here when it was running and ready and then the controller quorum is lost? Will it be the same behavior? Will it move into un-ready?

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,198 @@
...
+This will be used for the broker only and combined readiness checks later and will improve the existing ZooKeeper based broker check.
+
+### Phase 2 - Strimzi adds support for broker only and controller only modes
+* Controller only pod liveness and readiness checks are fully implemented to match this proposal.
+* Broker only pods are marked as both alive and ready if listening on port 9091.
+* Combined mode pods are marked as both alive and ready when there is a Java process running (no change from phase 1).
+
+### Phase 3 - KafkaRoller is updated to take the existence of controller pods and status of controller quorum into account when deciding whether to roll pods
+* Broker only readiness probe fully implemented to match this proposal.
+* Combined mode readiness probe fully implemented to match this proposal.
```
scholzj, line 163: Just formatting improvements.

```suggestion
### Phase 1 - Strimzi only supports combined mode

* Combined mode pods are marked as both alive and ready when there is a Java process running. 
  This is an improvement on the current state where the pod is marked as alive and ready as soon as the container starts up, regardless of whether the Java process started.
* Update the KafkaAgent to check for a broker state of >= 3 && != 127.
  This will be used for the broker only and combined readiness checks later and will improve the existing ZooKeeper based broker check.

### Phase 2 - Strimzi adds support for broker only and controller only modes

* Controller only pod liveness and readiness checks are fully implemented to match this proposal.
* Broker only pods are marked as both alive and ready if listening on port 9091.
* Combined mode pods are marked as both alive and ready when there is a Java process running (no change from phase 1).

### Phase 3 - KafkaRoller is updated to take the existence of controller pods and status of controller quorum into account when deciding whether to roll pods

* Broker only readiness probe fully implemented to match this proposal.
* Combined mode readiness probe fully implemented to match this proposal.
```

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,198 @@
...
+Previous versions of this proposal discussed whether we could check if this metric had a state of either follower or leader and use that to infer that the controller quorum had been formed and the controller should be marked as ready.
+This approach had a couple of problems:
+* The individual state of the controller doesn't always imply the full quorum state. 
+For example, this particular controller might be offline, but the rest of the quorum is still healthy.
+Also, during a leadership election the metric might show a state of `candidate` or `voted`.
+* In general readiness checks in Kubernetes should be based on individual nodes, not the cluster as a whole.
+* The current-state metric is not accessible as a Yammer metric, so cannot be accessed by the KafkaAgent in the same way it currently reads the BrokerState metric.
+Although it could perhaps be read by the KafkaAgent if it called the MBean server within the JVM directly or read it from `$LOG_DIR/__cluster_metadata-0/quorum-state`.
+* The broker nodes will not move out of the STARTING state until the quorum leader election has happened.
```
scholzj, line 191: Just a formatting change

```suggestion
### Use the current-state metric for controller readiness

The [current-state](https://kafka.apache.org/documentation/#kraft_quorum_monitoring) metric indicates the current state of the quorum.
Previous versions of this proposal discussed whether we could check if this metric had a state of either follower or leader and use that to infer that the controller quorum had been formed and the controller should be marked as ready.
This approach had a couple of problems:
* The individual state of the controller doesn't always imply the full quorum state. 
  For example, this particular controller might be offline, but the rest of the quorum is still healthy.
  Also, during a leadership election the metric might show a state of `candidate` or `voted`.
* In general readiness checks in Kubernetes should be based on individual nodes, not the cluster as a whole.
* The current-state metric is not accessible as a Yammer metric, so cannot be accessed by the KafkaAgent in the same way it currently reads the BrokerState metric.
  Although it could perhaps be read by the KafkaAgent if it called the MBean server within the JVM directly or read it from `$LOG_DIR/__cluster_metadata-0/quorum-state`.
* The broker nodes will not move out of the STARTING state until the quorum leader election has happened.
```

scholzj (commented):
I left one more question and some formatting suggestions. But LGTM otherwise.

ppatierno (approved):

044-kraft-liveness-readiness.md
katheris: From reading the KIP my understanding is that if the broker cannot contact the lead controller it will transition back into the RECOVERY state, which means yes it would move to un-ready. I think that is acceptable though, since the broker would also stop accepting client requests at that point. Whereas in ZooKeeper mode the broker could still accept client requests if it could not contact ZooKeeper. I'll do some tests this afternoon to confirm my understanding and make sure the broker state does transition back as I described.

This is the part of the KIP that I think is relevant:

> The Broker State Machine
Currently, brokers register themselves with ZooKeeper right after they start up.  This registration accomplishes two things: it lets the broker know whether it has been elected as the controller, and it lets other nodes know how to contact it.
>
> In the post-ZooKeeper world, brokers will register themselves with the controller quorum, rather than with ZooKeeper.
>
> Currently, if a broker loses its ZooKeeper session, the controller removes it from the cluster metadata.  In the post-ZooKeeper world, the active controller removes a broker from the cluster metadata if it has not sent a MetadataFetch heartbeat in a long enough time.
>
> In the current world, a broker which can contact ZooKeeper but which is partitioned from the controller will continue serving user requests, but will not receive any metadata updates.  This can lead to some confusing and difficult situations.  For example, a producer using acks=1 might continue to produce to a leader that actually was not the leader any more, but which failed to receive the controller's LeaderAndIsrRequest moving the leadership.
>
> In the post-ZK world, cluster membership is integrated with metadata updates.  Brokers cannot continue to be members of the cluster if they cannot receive metadata updates.  While it is still possible for a broker to be partitioned from a particular client, the broker will be removed from the cluster if it is partitioned from the controller.

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,198 @@
...
+means the nodes will be able to communicate even if they are currently marked as not ready.
+
+### Implementation of the checks
+
+To check that the Java process is running we will execute a command in the container. The proposed command is:
+```shell
+for proc in $(ls -1 /proc/ | grep [0-9]); do echo "$(ls -lh /proc/$proc/exe 2>/dev/null || true)" | grep -q java; done
+```
+This is preferable over using a tool like `ps` (`ps aux | grep -v grep | grep java`) since it does not require a change to the tools added to the container image.
+In future this could be replaced with a call to an endpoint, since other proposals discuss adding endpoints for example to expose metrics.
```
PaulRMellor, line 98:
```suggestion
In future this could be replaced with a call to an endpoint, since other proposals discuss adding endpoints, for example, to expose metrics.
```

044-kraft-liveness-readiness.md
```diff
@@ -0,0 +1,198 @@
...
+The reason for the final bullet (combined pod not being ready) might not be immediately obvious, so it is explained below.
+The brokers do not move to RUNNING until a quorum has been formed.
+For example in combined mode during normal startup the following would happen:
+
+* All the pods are started at the same time
+* The controller processes in each pod form a quorum
+* The broker processes start up and start talking to the controller quorum so the controller unfences them and they move to RUNNING and are marked as ready
+* All pods are now ready
+
+However, if the pods are started initially with affinity constraints so they are all in pending, then the constraint is removed, the following would happen:
```
PaulRMellor, line 137:
```suggestion
However, if the pods are started with affinity constraints, they are all in a pending state and the following issue arises when forming a quorum:
```

PaulRMellor: I was a bit confused by the _constraint is removed_. I assume we're saying how the final bullet is handled with affinity constraints. Are we saying this is a limitation that needs resolving with KafkaRoller first?  Maybe we should make any dependencies more explicit?

PaulRMellor (approved):
A clear and useful-sounding proposal

044-kraft-liveness-readiness.md
katheris: I've done some tests but haven't actually managed to see the BrokerState metric transition out of 3 once it got to 3. I tried:
 - A single controller and removing the controller
 - 3 controllers and taking down a controller to force an election
 - 3 controllers and taking down two controllers
 - 3 controllers and taking down all controllers

In all cases the BrokerState metric stayed at value 3

044-kraft-liveness-readiness.md
katheris: Not quite, so the below bullets describe what happens if you initially have constraints that mean the pods can't be scheduled, and then later remove that constraint. The result is basically that even though the cluster should come up, it doesn't due to the way KafkaRoller works. I've made some changes to hopefully make it clearer, see what you think.

044-kraft-liveness-readiness.md
scholzj: But did you check if you can connect with clients or not? If it indeed blocks the clients, then we probably need to transition it and it should be exposed somehow through Kafka. If the clients can keep using it with some limitations, we can probably stick with being ready as we do today.

katheris: I did some more tests this morning and it seems like the clients don't get blocked, but some requests do fail. When all controllers were down any existing producer and consumer clients were able to continue producing and consuming. However any requests that included a metadata request failed, so I couldn't start a new consumer or create a new topic. However I could create a new producer client.

scholzj: Ok, that sounds similar to ZK node. So I think we can keep it ready as we do today. Thanks for the clarification.

scholzj:
This has now enough votes to be approved. Let's give it one more day for any further comments and then we will close the vote.

044-kraft-liveness-readiness.md
PaulRMellor: Thanks @katheris. I understand now.

showuon (approved):
LGTM! Thanks.

im-konge (approved):

scholzj:
Approved with 6 binding approvals and 1 non-binding approval.

PR #70, opened 2022-12-22
Add a proposal for FSM kafka roller

devguyio:
This PR proposes changes to the Kafka roller that acts based on a Finite State Machine (FSM) of Kafka server states and possible transitions for rolling Kafka servers to a healthy state.

**TODO**

- [ ] Implementation milestones
- [ ] Compatibility
- [ ] FSM States description


045-kafka-roller-refactoring.md
```diff
@@ -0,0 +1,100 @@
...
+
+#### Observations
+
+* Kafka Roller decisions can become more accurate and better informed by having rich observations coming from different sources (e.g. Kafka admin API, Kubernetes API, metrics, etc.).
+* Creating an abstraction for the observations' collection that encapsulates different observation sources, including custom sources (e.g. a Kafka Agent). Such custom sources would help with better understanding of the Kafka server state (e.g. broker is in log recovery).
+
+#### State Classification
+
+* Kafka servers can be classified into a set of states that are determined by the collected observations (e.g. needs restart, unhealthy, stable, etc.)
+* The Kafka server states can be modeled as a state machine which reflects the Kafka Roller's understanding of the Kafka server state and determines the possible actions needed to transition the Kafka server to a healthy state.
```
tombentley, line 53: Again, we can make this more clear.

```suggestion
* The Kafka Roller's understanding of a particular Kafka server's state determines the possible actions needed to transition that server to a healthy state. In other words, the set of possible states together with the rules for the allowed actions in each state form a state machine.
```

045-kafka-roller-refactoring.md
```diff
@@ -0,0 +1,100 @@
...
+
+* Responsible for mapping observations to states.
+* Does not understand state transitions.
+
+#### The Processor
+
+* An abstraction around Kafka server management. 
+* Responsible for pod deletion and reconfiguration
+* Does not understand state transitions.
+
```
tombentley, line 82: You need to go into detail about how the processor works:

1. That it only processes a single pod/server at a time. Aside: With the right checks I don't think this is something we need to enforce directly; i.e. the acks=all availability constraint on its own might be enough to prevent bad things happening, without forcing the algorithm to be sequential. But starting with sequential seems a sensible starting point.
2. How does it determine which pod to process next? My protocol code used a total order on the states to determine order, if you recall?
3. That once we've attempted to restart/reconfigure a pod the first time we don't progress to another pod until we've see that pod enter the `stable` state. 
4. Under what circumstances does the algorithm terminate. Running out of servers that need reconfig/restart is the happy path, but what about exceptional paths? Having an algorithm that definitely terminates, even if via a timeout, is a requirement to avoid the operator getting stuck.

045-kafka-roller-refactoring.md
```diff
@@ -0,0 +1,100 @@
...
+### State Machine
+
+The following diagram illustrates the proposed Kafka server states and possible transitions
```
tombentley, line 85: 1. You should describe the states in detail.
2. For those states that imply the operator taking an action you should describe what the action is.
3. For those states that do not imply the operator taking an action you should explain what happens in the Kafka cluster that will cause a transition. This is a bit weird, in a way, because the state machine we've defined has transitions that are not caused by the operator. They're informed based on how we _expect_ the cluster to behave. But we purposely don't want to build those into the roller as _expectations_. For example, a broker can go from `stable` (i.e. running and in the ISR) to `syncing` (i.e. running but not in the ISR) because of producer load/interbroker throttling. Those are beyond the roller's ability to actively influence, but they do change the behaviour of the roller: To move on to the next pod we want to see the broker joint the ISR at least momentarily. If it later falls out of the ISR the roller doesn't directly care: It's not going to cause the roller to try restarting the broker; but it might cause us to delay restarting another broker if it would impact topic availability.

devguyio:
Won't be working on it anymore

PR #71, opened 2022-12-22
Analyse broker metrics before rolling kafka

tinaselenge:


tinaselenge:
Feedback from @SamBarker
```
I think the proposal does a nice job of summarising the plan. I'm just an interested observer so haven't seen a lot of detail or done much thinking about the problem. 

I see jetty as an implementation detail. I think the high level outline should just say add a web sever, or even light weight web server and then specify jetty in an implementation section. 

I also wonder if a servlet which only servers broker-state is overly restrictive? Maybe a suggest adding a metrics servlet which accepts a metric name or maybe a metric key as a parameter (I'd be 100% on board with the implementation only supporting a fixed value of broker-state) I see this as being an equivalent design in the short term but one that is more open to extension in the future. I imagine we will want something similar in the future to expose quorum status or similar in the kraft world.

I think the proposal should probably give some details about the format/schema it supports for metrics it exposes. Potentially supporting the Prometheus format is the cheapest way forward, again mostly thinking about being open for extension rather than current requirements. Prometheus already has a standard for exposing metrics over HTTP which we could just re-use, if its appropriate.

This could well be my lack of experience with the roller but I would sort of expect a broker in log recovery to pass a timeout check as it was still doing something useful. So I would probably be talking in terms of eagerly checking for broker-state or log-recovery rather than doing in the exceptional case.

I think what I'm really diving at is I want the proposal to have two different sections. A high level overview of the big picture changes: Add web server exposing metrics. Followed by an proposed implementation talking about Jetty and the specifics of the endpoint(s) i.e. GET requests.
```

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
+<!-- This template is provided as an example with sections you may wish to comment on with respect to your proposal. Add or remove sections as required to best articulate the proposal. -->
```
scholzj, line 1: You can probably remove this.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Proposal
+
+The proposal is a REST API endpoint on the KafkaAgent which exposes the BrokerState metric that it already uses. And modify KafkaRoller so that it queries the new endpoint to find out what state a broker is in. 
+
+1. Add a web server to KafkaAgent that exposes an endpoint which returns a value of metric
+    - Configure Jetty server to only accept SSL connections on a port 8443.
+    - Expose an endpoint `/broker-state`.
+    - Add a servlet to handle a `GET` request and return an integer value of `BrokerState` metric.
+    - It will reuse broker's own certificate, which will already be in the container.
```
scholzj, line 31: What about security beyond encryption? E.g. TLS client authentication? I wonder what value encryption adds without authentication.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+2. Update KafkaRoller to query the broker state from the KafkaAgent API
+    - Currently KafkaRoller obtains headless service for a given pod and uses it to boostrap Kafka AdminClient. It can use the same mechanism to get the address for the KafkaAgent endpoint.
```
scholzj, line 33: Just a wording ... the headless service gives the pods a stable DNS name. The operator knows the structure of the DNS name, so it just generates the address and uses it to connect to the Pod. So strictly speaking, it does not directly interact with the headless service. So maybe you can say something like `KafkaRoller will use the Pod DNS name to query the KafkaAgent.`.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+    - Instead of checking broker's readiness through `isReady()` function, utilise the new restful endpoint to determine if the broker is ready.
+    - After restarting a broker if it does not become ready within the operational timeout period, check whether the broker is performing log recovery by querying the new endpoint:
+        - If log recovery is in progress, do not restart the broker.
+        - Otherwise, force restart the broker which is the current behabviour.
+    - Add a new configurable timeout for log recovery that is 60 minutes by default.
+    - If the log recovery does not complete within the timeout period, then fail the reconciliation and report an error for the human operator to investigate further. 
+
+### For future consideration:
+- The operator might need access to more broker metrics. We can easily implement additional endpoints to expose them.
+- The readiness check could also make use of the KafkaAgent endpoint to get the broker state instead of checking for a file on disk.
```
scholzj, line 43: The readiness check would probably need an internal non-TLS endpoint. But I guess that can be elaborated on in a further proposal.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Compatibility
+
+When the operator gets updated with the KafkaRoller change, it will expect the KafkaAgent endpoint to be there. However, KafkaAgent would only get updated after the operator.
+
+1. Roll out the KafkaAgent changes for the new API.
+2. Roll out the KafkaRoller changes and gated behind a feature flag. By default, KafkaRoller will still have the current behaviour and will not use the KafkaAgent API. 
+3. User can enable KafkaRoller to use KafkaAgent API using the feature flag. The user has to make sure that they are running the version that has the KafkaAgent API before they enable the feature. 
+4. We can mark the feature flag deprecetad immediately. 
+5. In the next major release, we can remove the feature flag and KafkaRoller can always use the KafkaAgent API.
```
scholzj, line 58: I assume you mean Feature gate (as is the Strimzi terminology)? If you want to use a Feature Gate, you should suggest a name and a proposed schedule here. (check the other proposals with feature gates for examples). 

But I think you should try to figure out a better way because rolling out something like this will take a fairly long time. A very similar feature - Control plan listener was rolling out for almost 9 releases. It will be very complicated to build some further features on top of this in such case, maintain and test the code etc.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Rejected alternatives
```
scholzj, line 60: You should leave an empty line after the heading for better readability.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+2. Update KafkaRoller to query the broker state from the KafkaAgent API
+    - Currently KafkaRoller obtains headless service for a given pod and uses it to boostrap Kafka AdminClient. It can use the same mechanism to get the address for the KafkaAgent endpoint.
+    - Instead of checking broker's readiness through `isReady()` function, utilise the new restful endpoint to determine if the broker is ready.
+    - After restarting a broker if it does not become ready within the operational timeout period, check whether the broker is performing log recovery by querying the new endpoint:
+        - If log recovery is in progress, do not restart the broker.
+        - Otherwise, force restart the broker which is the current behabviour.
+    - Add a new configurable timeout for log recovery that is 60 minutes by default.
+    - If the log recovery does not complete within the timeout period, then fail the reconciliation and report an error for the human operator to investigate further. 
+
+### For future consideration:
```
scholzj, line 41: You should leave an empty line after the heading for better readability.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
+<!-- This template is provided as an example with sections you may wish to comment on with respect to your proposal. Add or remove sections as required to best articulate the proposal. -->
+
+# Analyse broker metrics before rolling kafka
```
scholzj, line 3:
```suggestion
# Analyse broker metrics before rolling Kafka
```

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+
+This proposal describes a solution for KafkaRoller's shortcoming with brokers in log recovery. 
+
+## Current situation
+
+The current logic in Kafka Roller:
+- Take a list of kafka pods and reorder them based on the readiness.
```
scholzj, line 10:
```suggestion
- Take a list of Kafka pods and reorder them based on the readiness. 
```

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+# Analyse broker metrics before rolling kafka
+
+This proposal describes a solution for KafkaRoller's shortcoming with brokers in log recovery. 
+
+## Current situation
+
+The current logic in Kafka Roller:
+- Take a list of kafka pods and reorder them based on the readiness. 
+- Roll unready pods first without considering why it could be unready.
+- Before restarting a kafka pod, consider if it is a controller or it has an impact on the availability.
```
scholzj, line 12: I will not comment further on it. But it should be fixed everywhere ;-)

```suggestion
- Before restarting a Kafka pod, consider if it is a controller or it has an impact on the availability.
```

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+1. Add a web server to KafkaAgent that exposes an endpoint which returns a value of metric
+    - Configure Jetty server to only accept SSL connections on a port 8443.
+    - Expose an endpoint `/broker-state`.
+    - Add a servlet to handle a `GET` request and return an integer value of `BrokerState` metric.
+    - It will reuse broker's own certificate, which will already be in the container.
+2. Update KafkaRoller to query the broker state from the KafkaAgent API
+    - Currently KafkaRoller obtains headless service for a given pod and uses it to boostrap Kafka AdminClient. It can use the same mechanism to get the address for the KafkaAgent endpoint.
+    - Instead of checking broker's readiness through `isReady()` function, utilise the new restful endpoint to determine if the broker is ready.
+    - After restarting a broker if it does not become ready within the operational timeout period, check whether the broker is performing log recovery by querying the new endpoint:
+        - If log recovery is in progress, do not restart the broker.
```
scholzj, line 36: If the operator detects that log recovery is in progress, it should probably issue some warning as well. Since this is a situation which should be ideally prevented.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+        - Otherwise, force restart the broker which is the current behabviour.
+    - Add a new configurable timeout for log recovery that is 60 minutes by default.
```
scholzj, line 38: Where do the 60 minutes come from? I would suggest following the usual way Strimzi approaches this:
* Wait for operations timeout (5 minutes by default)
* If it doesn't happen within that time, fail the reconciliation
* In the next loop, continue where you left off (i.e. again wait for the recovery to finish)

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+When the operator gets updated with the KafkaRoller change, it will expect the KafkaAgent endpoint to be there. However, KafkaAgent would only get updated after the operator.
+
+1. Roll out the KafkaAgent changes for the new API.
+2. Roll out the KafkaRoller changes and gated behind a feature flag. By default, KafkaRoller will still have the current behaviour and will not use the KafkaAgent API. 
+3. User can enable KafkaRoller to use KafkaAgent API using the feature flag. The user has to make sure that they are running the version that has the KafkaAgent API before they enable the feature. 
+4. We can mark the feature flag deprecetad immediately. 
+5. In the next major release, we can remove the feature flag and KafkaRoller can always use the KafkaAgent API.
+
+## Rejected alternatives
+- Why not use VertX? Strimzi is moving away from using VertX. Kafka already uses Jetty which is convenient for KafkaAgent, as there is no need to introduce a new dependency.
```
scholzj, line 61: This is not really about Strimzi's use of Vert.x. Vert.x is used in operators only. It is not part of the Kafka images anyway. So regardless of how long we keep using it, Jetty which is already part of Kafka is a more logical choice.

scholzj (commented):
Thanks for the PR. I left some comments. I think it looks mostly good, but the backwards compatibility requires some considerations. Otherwise the rollout of this to the users will take very long and the featur will not be really availbale for a very long time.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+    - After restarting a broker if it does not become ready within the operational timeout period, check whether the broker is performing log recovery by querying the new endpoint:
+        - If log recovery is in progress, do not restart the broker.
+        - Otherwise, force restart the broker which is the current behabviour.
+    - Add a new configurable timeout for log recovery that is 60 minutes by default.
+    - If the log recovery does not complete within the timeout period, then fail the reconciliation and report an error for the human operator to investigate further. 
+
+### For future consideration:
+- The operator might need access to more broker metrics. We can easily implement additional endpoints to expose them.
+- The readiness check could also make use of the KafkaAgent endpoint to get the broker state instead of checking for a file on disk.
+- The liveness and readiness checks potentially will need to access non yammer metrics for KRAFT, which could be exposed in the same manner, but would need a new mechanism to collect them.
```
PaulRMellor, line 44:
```suggestion
- The liveness and readiness checks potentially will need to access non-yammer metrics for KRAFT, which could be exposed in the same manner, but would need a new mechanism to collect them.
```

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Affected/not affected projects
```
PaulRMellor, line 46:
```suggestion
## Affected projects
```

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+* strimzi/kafka-cluster-operator
+
+## Compatibility
+
+When the operator gets updated with the KafkaRoller change, it will expect the KafkaAgent endpoint to be there. However, KafkaAgent would only get updated after the operator.
+
+1. Roll out the KafkaAgent changes for the new API.
+2. Roll out the KafkaRoller changes and gated behind a feature flag. By default, KafkaRoller will still have the current behaviour and will not use the KafkaAgent API. 
+3. User can enable KafkaRoller to use KafkaAgent API using the feature flag. The user has to make sure that they are running the version that has the KafkaAgent API before they enable the feature. 
+4. We can mark the feature flag deprecetad immediately.
```
PaulRMellor, line 57:
```suggestion
4. We can mark the feature flag deprecated immediately. 
```

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+- The readiness check could also make use of the KafkaAgent endpoint to get the broker state instead of checking for a file on disk.
+- The liveness and readiness checks potentially will need to access non yammer metrics for KRAFT, which could be exposed in the same manner, but would need a new mechanism to collect them.
+
+## Affected/not affected projects
+
+* strimzi/kafka-cluster-operator
+
+## Compatibility
+
+When the operator gets updated with the KafkaRoller change, it will expect the KafkaAgent endpoint to be there. However, KafkaAgent would only get updated after the operator.
```
PaulRMellor, line 52:
```suggestion
When the operator gets updated with the KafkaRoller change, it will expect the KafkaAgent endpoint to be there. However, KafkaAgent would only get updated after the operator.

To maintain backwards compatibility, the following process is suggested:
```

PaulRMellor (approved):
A useful-sounding proposal. I made a few minor suggestions.

scholzj:
I think one other aspect should be also considered here ... not just the upgrades from the older version without the _HTTP_ based agent. But also the future-looking upgrades. So I think it should consider two things:
* While I do not think this constitutes a public API per-se, we should have some concept of API versioning to be able to handle future evolutions and changes
* We should have some mechanism to discover the supported versions of the API

This should help to facilitate smooth upgrades also for the future.

046-analyse-broker-metrics-before-rolling-kafka.md
fvaleri: Since we now have log recovery metrics (KIP-831), it would be good to use them in a periodic info message and to check if any progress is being made.

046-analyse-broker-metrics-before-rolling-kafka.md
fvaleri: Again, using metrics from KIP-831 could help the user to see that log recovery progress is being made, despite repeated reconciliation failures.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,63 @@
...
+    - Configure Jetty server to only accept SSL connections on a port 8443.
+    - Expose an endpoint `/broker-state`.
+    - Add a servlet to handle a `GET` request and return an integer value of `BrokerState` metric.
+    - It will reuse broker's own certificate, which will already be in the container.
+2. Update KafkaRoller to query the broker state from the KafkaAgent API
+    - Currently KafkaRoller obtains headless service for a given pod and uses it to boostrap Kafka AdminClient. It can use the same mechanism to get the address for the KafkaAgent endpoint.
+    - Instead of checking broker's readiness through `isReady()` function, utilise the new restful endpoint to determine if the broker is ready.
+    - After restarting a broker if it does not become ready within the operational timeout period, check whether the broker is performing log recovery by querying the new endpoint:
+        - If log recovery is in progress, do not restart the broker.
+        - Otherwise, force restart the broker which is the current behabviour.
```
ppatierno, line 37: typo "behabviour"

046-analyse-broker-metrics-before-rolling-kafka.md
ppatierno, line 44: "KRaft"

046-analyse-broker-metrics-before-rolling-kafka.md
ppatierno: I agree, I think that we could avoid mentioning Vert.x here. The Jetty choice is the obvious one because it's already there.

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: We could periodically check the recovery metrics using the existing pollingIntervalMs (1000 by default), however that might be too often unnecessarily. 

Or we could include the value of the recovery metrics in the warning message itself so on every retry (5 minutes or more), we will be able to see the progress. 

This also means we would have to expose log recovery metrics via KafkaAgent too?

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: @scholzj I like what you suggested. So just to clarify, once the reconciliation is failed , we will use KafkaRoller's current behaviour with backoff delay, retry and timeout? I think max attempts to retry is 6 by default.

046-analyse-broker-metrics-before-rolling-kafka.md
scholzj: At least for my comment, this is more something I would expect the cluster operator to do when it sees the recovery happening in the KafkaRoller.

tombentley: I think we want a `WARN` log the first time we find we've exceeded the operation timeout, and a 2nd `WARN` message (including the total observed time from pod deletion/recreation to `RUNNING` state) once we observed the transition to `RUNNING`. This:

1. avoids spamming the logs
2. makes it easy to see that you are (or were) in this state
3. makes it easy to know how to what to reconfigure the timeout value to to avoid this in the future.

046-analyse-broker-metrics-before-rolling-kafka.md
tombentley: It should be possible to use this opportunistically: If the endpoint works then apply the new logic, if not then apply the current logic.

046-analyse-broker-metrics-before-rolling-kafka.md
scholzj: Well, I'm not an expert on the Kafka Roller. In other areas - e.g. when loadbalancer is being provisioned - we simply wait for 5 minutes. And when it is not ready in 5 minutes, the reconciliation fails. Some time later a next reconciliation will start and check again if the load balancer is ready, wait for up to 5 minutes and fail if it is still not ready. The waiting itself happens in smaller intervals - not sleeping for 5 minutes but checking periodically during the 5 minutes of course.

Ideally, you should never hang for too long in a single reconciliation. The fact that the KafkaRoller sometimes backoffs for 10s of minutes is a design issue rather than a feature. But a separate decision is needed if this is something the user should handle or if this is something where we should have some cut-off time when the broker is force-rolled. 

But in the past, I saw environments where the recovery was taking 20+ hours. So I think if there is cut-off, the time needs to be carefully considered and likely configurable. If the agent can expose progress, that might of course make it eaiser to decide if it is slow but progressing or stuck.

046-analyse-broker-metrics-before-rolling-kafka.md
scholzj: I guess that would be an option ... the thing I was wondering about is whether this is good enough with respect to distinguishing if the _agent is stuck_ or whether the broker uses the old version. 

I guess we also need to think about the `then apply the current logic`. The upgrade part is something hat will be needed for a fairly long time. So this needs to elaborate on how this will be handled in the code as well.

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: Are you suggesting that we utilise mTLS, that way we do both server and client authentications? Is mTLS enabled by default or configurable in Strimzi?

scholzj: All internal communication is secured with mTLS out of the box without any possibility to disable (at least right now).

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Proposal
+ 
+The objective of this proposal is to introduce a REST API endpoint on the KafkaAgent which exposes the BrokerState metric that it already uses. And modify KafkaRoller so that it queries the new endpoint to find out what state a broker is in and avoids unnecessary restarts.
+ 
+1. Add a RESTful endpoint to KafkaAgent that exposes an endpoint which returns a value of metric
+    - Configure Jetty server to only accept SSL connections on port 8443.
+    - It will reuse the broker's own certificate, which will already be in the container.
+    - Expose an endpoint `/broker-state`.
+    - Add a servlet to handle a `GET` request and return an integer value of the `BrokerState` metric.
```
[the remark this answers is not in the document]
tinaselenge: I suggested using the Servlet API mainly because it was already used in another Strimzi operator ( https://github.com/strimzi/kafka-access-operator/blob/main/src/main/java/io/strimzi/kafka/access/server/HealthServlet.java). However, happy to change it to  use Jetty's native ServerHandler API.

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: I had the same concern @scholzj mentioned. However, I think if the agent is stuck, it's ok to apply the current logic that would force restart the broker, that way we might recover the stuck agent. 
I have addressed the suggestion in the latest commit.

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: I have addressed the suggestions in the latest commit. 

@tombentley do you think it's worth exposing the observed time as a metric from the operator? So that when we want to take it into consideration for changing the timeout, it could be more useful to see the observed value over time, rather than looking at one event.  But this could be for future improvements.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,64 @@
...
+- Before restarting a Kafka pod, consider if it is a controller or if it has an impact on the availability.
+- If a Kafka pod does not become ready within the operational timeout (300000ms by default) after restarting, then force restart it without considering why it's taking a long time.
+
+A Kafka broker can take a long time to become ready while performing log recovery. In this case, KafkaRoller ends up force restarting the broker and this can continue indefinitely, because of the log recovery on the broker startup again. KafkaRoller currently does not have a way to know what state a broker is in before making a decision to force restart it.
+ 
+KafkaAgent currently collects the `kafka.server:type=KafkaServer,name=BrokerState` metric and creates a `kafka-ready` file on disk if the broker state value is `3`(`RUNNING`). The Kafka readiness check passes if it finds the file on disk.
+
+## Motivation
+
+To address the issues with restarts and log recovery, KafkaRoller needs to be aware of the current broker state.  If a broker is performing log recovery, KafkaRoller should wait longer instead of force restarting it.
```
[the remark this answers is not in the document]
tinaselenge: Addressed in the latest commit.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,64 @@
...
+    - Expose an endpoint `/broker-state`.
+    - Add a servlet to handle a `GET` request and return an integer value of the `BrokerState` metric.
+2. Update KafkaRoller to query the broker state from the KafkaAgent API
+    - Use the Pod DNS name to query the KafkaAgent API.
+    - Instead of checking the broker's readiness through `isReady()` function, utilise the new RESTful endpoint to determine if the broker is ready.
+    - After restarting a broker if it does not become ready within the operational timeout period (5 minutes by default), check whether the broker is performing log recovery by querying the new endpoint:
+    - If log recovery is in progress:
+        - Wait for log recovery to complete until the operation timeout reaches.
+        - First time the operation timeout is exceeded, log a WARN message and fail the reconciliation.
+        - If the log recovery completes in the next reconciliation, issue a warning message that includes the total observed time from pod deletion/recreation to `RUNNING` state. This would be helpful to figure out a more appropriate timeout value to avoid this in the future.
```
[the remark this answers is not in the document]
tinaselenge: I have taken the suggestion and included this idea, please let me know what you think :)

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,86 @@
...
+The objective of this proposal is to introduce a REST API endpoint within KafkaAgent, which exposes metrics from a Kafka broker. Furthermore, KafkaRoller would be modified so that it can query this endpoint to determine the current state of a broker, thus preventing unnecessary restarts. Additionally, KafkaRoller would also display the remaining logs and segments to recover, allowing a human operator to observe the recovery process and determine if it is making progress or is stuck.
+
+### Implementation details
+
+*KafkaAgent*:
+- Add a web server to accept SSL connections on port 8443.
+- Use Jetty for the web server since Kafka already uses it therefore avoids an additional dependency.
+- Reuse the broker's own certificate that would be in the container for TLS authentication. 
+- Configure the server to require TLS client authentication. 
+- Expose an endpoint `v1/brokermetrics` that includes the API version in the URI path.
```
scholzj, line 32: This is good. But I wonder if you need discovery as well => i.e. way to find out which APIs are supported. So that you don't have to blindly try the versions in the future in some upgrade. E.g. some `apis/` endpoint which would return a list of supported API versions ... e.g. just `["v1"]` right now. That would allow the future rollers to decide what it is talking with and use the right API during upgrades. My guess is that this might be useful one day and it would not cost much on the implementation effort (no need to any sophisticated mechanisms in the KAfkaAgent, just return a static data 🤷 ).

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,86 @@
...
+
+*Logic for handling log recovery*
+- Wait for log recovery to complete until the operation timeout (5 mins by default) reaches.
+- While waiting for log recovery to complete, check its progress by periodically sending a request to the new endpoint for `remainingLogsToRecover` and `remainingSegmentsToRecover` metrics and log them in an `INFO` message.
+- If the operation timeout is exceeded, check the value of `logRecoveryExceededTimeout` field in the CR:
+    - If false, log a `WARN` message, fail the reconciliation and set `logRecoveryExceededTimeout` to true.
+    - If true, fail the reconciliation.
+- If kafka pod becomes ready, check the value of `logRecoveryExceededTimeout`. If set to true:
+    - Calculate the time passed since the `restartTimeStamp` and log it in a WARN message. This would be helpful to figure out a more appropriate timeout value to avoid this in the future.
+    - Empty `restartTimeStamp` and `logRecoveryExceededTimeout` values.
```
scholzj, line 65: TBH, I do not get this. Can you elaborate on why this is useful? Why not keep it simple and if your wait for the broker to be ready is over and it is still recovering, just raise a WARN log message and be done with it. Or maybe I did not understand properly what exactly is the added value of this mechanism.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,86 @@
...
+*KafkaRoller*:
+- Use the Pod DNS name to query the KafkaAgent API.
+- Use the cluster operator's client certificate and keys to authenticate with the KafkaAgent API server.
+- Instead of checking the broker's readiness through `isReady()` function, utilise the new RESTful endpoint to determine if the broker is ready.
+- Include an additional field in the status section of the Kafka CR. This will hold data that will be utilised by KafkaRoller in case the reconciliation process is repeated while the log recovery process is still ongoing:
+```
+status:
+    <kafka_pod_name>:
+      restartTimeStamp: 
+      logRecoveryExceededTimeout: <true|false>
```
scholzj, line 50: If we wanna stick with this (see the comment below - I'm not sure I follow this), then there should be definitely something between `.status` and `.<kafka-pod-name>` ... the pod names should not just randomly mix into the status root. You will also not be able to model it because you would need `Map<String, Object>` for this and the status API has a given structure.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,86 @@
...
+- Use Jetty's native ServerHandler API to handle a `GET` request and return 
+    - an integer value for `BrokerState`.
+    - an integer value for  `remainingLogsToRecover`.
+    - an integer value for `remainingSegmentsToRecover`.
+- If no metric name was provided in the request or the requested metric name is not supported, return HTTP 404 (Not Found) to the client.
+
+*KafkaRoller*:
+- Use the Pod DNS name to query the KafkaAgent API.
+- Use the cluster operator's client certificate and keys to authenticate with the KafkaAgent API server.
+- Instead of checking the broker's readiness through `isReady()` function, utilise the new RESTful endpoint to determine if the broker is ready.
```
scholzj, line 44: Should it be really instead? Or in addition? Or some other combination? (e.g. after we are done with the HTTP endpoint we check the readiness)? The readiness is important so it should still be checked that it eventually gets ready.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,86 @@
...
+- Extend KafkaAgent to collect `kafka.log:type=LogManager,name=remainingLogsToRecover` and `kafka.log:type=LogManager,name=remainingSegmentsToRecover` metrics.
+- Take the metric name as a parameter in the request query e.g `v1/brokermetrics?name=BrokerState`. Use the value of the MBean's `name` field for the request  e.g. `BrokerState` and  `remainingLogsToRecover`.
+- Use Jetty's native ServerHandler API to handle a `GET` request and return 
+    - an integer value for `BrokerState`.
+    - an integer value for  `remainingLogsToRecover`.
+    - an integer value for `remainingSegmentsToRecover`.
+- If no metric name was provided in the request or the requested metric name is not supported, return HTTP 404 (Not Found) to the client.
+
+*KafkaRoller*:
+- Use the Pod DNS name to query the KafkaAgent API.
```
scholzj, line 42: You should probably also add some pre-check? E.g. that the pod is running? The readiness today covers all previous states. But it does not make sense to poll the Agent API when it is in ContainerCreating for 30 minutes because it is missing some secret or pulling container image over a slow line.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Affected projects
+
+* strimzi/kafka-cluster-operator
+
+## Compatibility
+
+After the operator gets updated with the KafkaRoller change, it will expect the KafkaAgent endpoint to be present. However, the KafkaAgent may not have been updated yet.
+
+To maintain backwards compatibility, use the new logic that handles log recovery only if KafkaRoller is able to successfully connect to the KafkaAgent endpoint. If the KafkaAgent endpoint is not available yet or the agent is not running, the current logic of restarting the broker after a timeout period should be used. This ensures compatibility with previous versions. Additionally, in the case that the KafkaAgent API is available but the agent is stuck or not running for some reason, the current logic's force restart of the broker may help recover the agent.
```
scholzj, line 81: I think this is missing more context of the whole roller. Here (but basically also in the previous section) you describe some steps. But it is not clear where exactly they fit into the KafkaRoller. I think for general understanding it might be best to describe the whole Roller process and how it works today and explain which parts you change and how.

As an example ... you suggest here that if it cannot reach the Agent API, it will be force-rolled. But that is certainly not the right approach for upgrades. In the upgrades, regardless of whether you reach the Agent API, if the pod is ready, you do not force roll it but you need to connect to it with Kafka Admin API and roll it through the regular process which follows the in-sync replicas etc.

It is not really clear if that is intended that way or not since it is not clear which part of the roller you are proposing to change.

046-analyse-broker-metrics-before-rolling-kafka.md
tombentley: I wonder if that is really better:

* `GET /v2/brokermetrics` => 404, `GET /v1/brokermetrics` => 200
* `GET /apis` => 200, `GET /v1/brokermetrics` => 200

It's two `GET` requests in both cases, but `/apis` means we need a separate endpoint on the server side,  to make a separate kind of request on the client side, and brings a requirement to model endpoints for the response (extra serialization and deserialization, even if it's trivial). I think I'd be inclined to keep it simple for now.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,86 @@
...
+- Reuse the broker's own certificate that would be in the container for TLS authentication. 
+- Configure the server to require TLS client authentication. 
+- Expose an endpoint `v1/brokermetrics` that includes the API version in the URI path.
+- Extend KafkaAgent to collect `kafka.log:type=LogManager,name=remainingLogsToRecover` and `kafka.log:type=LogManager,name=remainingSegmentsToRecover` metrics.
+- Take the metric name as a parameter in the request query e.g `v1/brokermetrics?name=BrokerState`. Use the value of the MBean's `name` field for the request  e.g. `BrokerState` and  `remainingLogsToRecover`.
+- Use Jetty's native ServerHandler API to handle a `GET` request and return 
+    - an integer value for `BrokerState`.
+    - an integer value for  `remainingLogsToRecover`.
+    - an integer value for `remainingSegmentsToRecover`.
+- If no metric name was provided in the request or the requested metric name is not supported, return HTTP 404 (Not Found) to the client.
```
tombentley, line 39: If we _really_ to pass multiple things to the CO (see later comment) it would be better, imho, to have a single endpoint. i.e. `GET /v1/recovery-state` returns the record containing all three observations together:

```json
{
  "brokerState": 2,
  "remainingLogsToRecover": 123,
  "remainingSegmentsToRecover": 456
}
```

That avoids the need to make three HTTP requests (and reduces, but does not eliminate, the possibility for races which could result in confusing observations like `remainingSegmentsToRecover > 0` but `remainingLogsToRecover == 0`).

046-analyse-broker-metrics-before-rolling-kafka.md
tombentley: +1

046-analyse-broker-metrics-before-rolling-kafka.md
tombentley: +1 to Jakubs point. Also be explicit about _when_ this is present (i.e. I assume it's present in the CR `state` only for the duration of a pod having been restarted and not observed to have progressed to RUNNING broker state, so normally this would be missing from the `status` entirely?)

046-analyse-broker-metrics-before-rolling-kafka.md
tombentley: Now I see this written down it looks pretty convoluted. What we were trying to achieve here was a mechanism to allow the user to know how long their brokers are _actually_ take to recover. On reflection they can get that easily enough from the broker metrics themselves (e.g. from prometheus/grafana). So there's no great value in futzing around trying to log something (whether once or many times): We should just fail the reconciliation and let the user figure it out. That would allow us to keep the simpler endpoint for the broker state metrics (returning a single integer), keep this stuff out of the CR `status` entirely and the user would still be able to reconfigure the CR (in the event that log recovery was really taking a long time).

Sorry for having led you astray here @tinaselenge.

046-analyse-broker-metrics-before-rolling-kafka.md
scholzj: Well, maybe I'm a bit pessimistic about the versioning ... but you can also imagine it like this 😉:
* GET /v5/brokermetrics => 404
* GET /v4/brokermetrics => 404
* GET /v3/brokermetrics => 404
* GET /v2/brokermetrics => 404
* GET /v1/brokermetrics => 200

As we need to support upgrades across versions, it is not necessarily just the current version and the previous one. You might need to support more of them.

ppatierno: But the second case will always be two `GET`(s) (getting the API version and then asking for it).
The first one could be more if the roller starts to try with the latest one, maybe `v4` but ends after 3 times of 404 to get 200 for `v1`. Or it could not happen to have a so updated roller communicating with a potential so old agent?

046-analyse-broker-metrics-before-rolling-kafka.md
ppatierno: +1 with Tom's suggestion. We should aim to reduce the HTTP calls for getting what we want. Having a single specific endpoint which consistently returns the status of the recovery looks much better to me as well.
Using a specific endpoint also copes with what you mentioned later in the proposal:

> The operator might need access to more broker metrics. We can easily implement additional endpoints to expose them

046-analyse-broker-metrics-before-rolling-kafka.md
tombentley: When we got to that point I think we'd have the justification to consider `/apis`. And to be honest I don't see the evolution of this endpoint being so rapid that going that far back in the versions was at all common (i.e. because operator and broker images tend to be upgraded together).

ppatierno: @scholzj I was with you, just clicking "submit review" later :-)

046-analyse-broker-metrics-before-rolling-kafka.md
fvaleri: I agree that this is unnecessary complexity and simply failing the reconciliation with a message that shows log recover progress is enough.

046-analyse-broker-metrics-before-rolling-kafka.md
fvaleri: In future considerations I read that readiness probe will use the agent's endpoint. I think it's better if we make it part of this proposal, while keeping the old behavior for backward compatibility.

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: Another idea is instead of logging a `WARN` and having to check if it's the first time exceeding the operation timeout, we could create a new error cause that says something like log recovery in progress. I noticed that reconciliation logging is quite noisy already and it does log error cause when the reconciliation fails.

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: I agree with this. Reconciliation failure already gets logged with a cause, so thinking about it again , we probably won't need this conditional WARN log. Also like you said, user can monitor the metrics for log recovery progress directly from the brokers and it's better to see the metric trend over time, rather than looking at one event from log. I will go ahead remove the parts about writing to the CR.

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: As agreed above, we are not storing anything in the CR. I will remove this part

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: The intention is not to have an additional logic there to force roll the broker if it can't reach the API. It would just follow the current behaviour which waits for broker to become ready and may force roll if it times out. I was trying to say, following the current behaviour is ok when the API is not reachable. If the agent is stuck or  not running, that would likely to be the action to take to get it working anyway. 

I will try make this clearer and how it fits to the KafkaRoller logic.

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: Now thinking about it, it does not make sense to change the readiness function in KafkaRoller. 

The readiness probe part was put in the future consideration because it's not really a part of solving the log recovery issue. I think once we have the agent exposing the API, it should be in a separate proposal.

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: As majority agreed, keeping it simple for now and going with 404 error for invalid version. We can implement something smarter to handle the versions when we have multiple of them.

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: we can use the existing isPodStuck() function as a pre check.

fvaleri (approved):
LGTM.

Perhaps we could update the proposal's title to better reflect the issue we are addressing. Something like "Avoid broker restart when in log recovery state".

046-analyse-broker-metrics-before-rolling-kafka.md
scholzj: > When we got to that point I think we'd have the justification to consider /apis. And to be honest I don't see the evolution of this endpoint being so rapid that going that far back in the versions was at all common (i.e. because operator and broker images tend to be upgraded together).

The point is that you cannot do it _when it comes to it_. Because then it would not be in the previous versions and you will not be able to rely on it. And rolling it out when we get to `v5` would mean for example waiting for `v10` before you can make any use for it. So it is not something _we do when we get there_. You cannot really do it when you get there.

The point of this is also not the regular rolling updates but upgrades. So regardless of `operator and broker images tend to be upgraded together`, you will need to deal with these situations when the operator from Strimzi 0.44 will be doing the upgrade from Strimzi 0.34 for example.

I do not really know how quickly will the API evolve. Maybe we will still use `v1` in 2 years. But who knows?

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,93 @@
...
+- Expose an endpoint `v1/recovery-state` that includes the API version in the URI path.
+- Extend KafkaAgent to collect `kafka.log:type=LogManager,name=remainingLogsToRecover` and `kafka.log:type=LogManager,name=remainingSegmentsToRecover` metrics.
+- Use Jetty's native ServerHandler API to handle a `GET` request and return a JSON response that looks like this:
+```
+{
+  "brokerState": 2,
+  "remainingLogsToRecover": 123,
+  "remainingSegmentsToRecover": 456
+}
+```
```
scholzj, line 47: What would it return when it is in state `3` and is running? What is the `/recovery-state` in such case? Is it empty? Or should it be `/broker-state` and always return the state and only when the state is `2` also return the progress?

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,93 @@
...
+- Return HTTP 404 (Not Found) if a request is sent to an unknown version of the API e.g. `GET /v5/recovery-state` and `v5` does not exist.
+
+**KafkaRoller**:
+
+The proposed flow of KafkaRoller
+![The proposed flow of KafkaRoller](images/046-kafka-roller-new-flow.png)
+
+If a kafka pod does not become ready within the operational timeout (5 mins by default) take the following actions:
+1. Check if the maximum number of retries has been reached. If it has, fail the reconciliation.
```
scholzj, line 56: How many retries do we have for the Pod readiness here? This seems to suggest we wait the 5 minutes repeatedly here in one reconciliation? Or does it men something else?

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,93 @@
...
+2. Ensure that the pod is not in a stuck state. The existing `isPodStuck()` can be used for this which returns true if the pod is in `CrashLoopBackOff`, `ImagePullBackOff`, `ContainerCreating` or `Pending` state. If the pod is stuck, continue with the current behaviour to retry rolling of the pod.
+3. Use the Pod DNS name to query the KafkaAgent API endpoint.
+4. If the API returns 503 (Service Unavailable), continue with the current behaviour to retry rolling of the pod.
```
scholzj, line 59: Where would you get the HTTP 503 from?

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,93 @@
...
+6. Send `GET v1/recovery-state` request to the endpoint and check the value of `brokerState` from the response.
+7. If the `brokerState` value is `2` (`RECOVERY`), follow the new logic for handling log recovery.
+ 
+*Handling log recovery*
+
+8. Set `inLogRecovery` to true. (`inLogRecovery` is a new boolean that would be added to the `RestartContext`).
+9. Log the values of `remainingLogsToRecover` and `remainingSegmentsToRecover` from the request response in an `INFO` message.
+10. Continue with the current behaviour to retry rolling of the pod.
+11. On the next retry, check `inLogRecovery` inside the `restartIfNecessary()` function.
+12. When `inLogRecovery` is set to true, call `awaitReadiness()` which periodically checks and waits for readiness until the operational timeout reaches.
```
scholzj, line 70: We already waited for 5 minutes for the readiness before this. Do we want to wait again? We ideally want to make sure to not do too many timing out _waiting for readiness_ calls in one reconciliation. Looping the reconciliation is important because it helps to inform the user.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,93 @@
...
+  "brokerState": 2,
+  "remainingLogsToRecover": 123,
+  "remainingSegmentsToRecover": 456
+}
+```
+- Return HTTP 404 (Not Found) if a request is sent to an unknown version of the API e.g. `GET /v5/recovery-state` and `v5` does not exist.
+
+**KafkaRoller**:
+
+The proposed flow of KafkaRoller
```
scholzj, line 52: How does this treat the other brokers? Assume you have one broker recovering and other 9 brokers are ready and need rolling update for reconcifiguration. Will they be rolled in the same reconciliation cycle if possible? Or will that be left for next time unless the non-ready broker recovers?

046-analyse-broker-metrics-before-rolling-kafka.md
scholzj: I mean, if we want to stick with the 404, then fine, I can deal with it. But the reasoning from @tombentley is not valid and does not make sense. That is what I'm trying to point out. When we are at v5, adding the `/apis` endpoint would not help with v1-v5. And just because we are at `v5`, the are is no guarantee that we will add more versions in the future as we maybe just found the right version. So the logic there is wrong.

scholzj (commented):
In the diagrams, there is one part which is not clear to me:
* The updated flow suggests that if `PodStuck == true` you check the Agend API. But when the Pod is stuck, the agent will nto be running. Shouldn't the true / false be reverted there?

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: The intention is to always return the broker state and KafkaRoller would check the value. The progress values would be empty I guess if the state is 3, for example:
```
{
  "brokerState": 3,
  "remainingLogsToRecover":0,
  "remainingSegmentsToRecover": 0
}
```
KafkaRoller would not check the progress values if the state is not equal to 2.

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: The default max attempt is 6. We would be following the current behaviour here, which waits for 5 minutes for pod readiness and then retry as long as we haven't hit the max attempts. So yes it already repeatedly waits for 5 minutes up to 6 times.

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: Hmm, I suppose it should be actually a timeout rather than 503?

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: We are not adding any additional wait here. On the timeout, it will simply set `inLogRecovery` to true and move on to the next retry. On the retry, it will wait for readiness right away because inLogRecovery is set to true. It already does that if `forceRestart` is set false. The retry logic on the timeout is not new.

046-analyse-broker-metrics-before-rolling-kafka.md
tinaselenge: > In the diagrams, there is one part which is not clear to me:
> 
> * The updated flow suggests that if `PodStuck == true` you check the Agend API. But when the Pod is stuck, the agent will nto be running. Shouldn't the true / false be reverted there?

Sorry, I meant to fix that.  You are right, it should be other way around :)

046-analyse-broker-metrics-before-rolling-kafka.md
scholzj: I guess it depends ... timeout when it is somehow stuck, connection refused in older versions, or when it does not listen for some reason?

046-analyse-broker-metrics-before-rolling-kafka.md
scholzj: Might be worth specifying it in the document. I also wonder:
* if `broker-state` is a better name for the endpoint if it essentially covers the whole broker state and not just the recovery phase
* Also, what HTTP code would it return? If it always returns 200, it would not be possible to use it as healthcheck later and we would need a separate endpoint for it. (no strong opinion on which way is better. But would be wort clarifying here.)

046-analyse-broker-metrics-before-rolling-kafka.md
scholzj: I guess it makes sense to stick with what we have in this proposal. But it might be worth clarifying in the proposal.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,93 @@
+# Analyse broker metrics before rolling Kafka
+
+This proposal describes a solution for KafkaRoller's shortcoming with brokers in log recovery.
+
+## Current situation
+
+The current logic in KafkaRoller:
+- Take a list of Kafka pods and reorder them based on the readiness.
+- Roll unready pods first without considering why it could be unready.
+- Before restarting a Kafka pod, consider if it is a controller or if it has an impact on the availability.
```
ppatierno, line 10: maybe it's worth making clear these two points:
- controller is rolled as last pod.
- "impact on the availability" is related to avoid under replicated partitions for example.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,93 @@
...
+The following implementation details will describe behaviour of the new API in KafkaAgent and how KafkaRoller would interact with it and handle a broker that is in log recovery.
+
+### Implementation details
+
+**KafkaAgent**:
+
+- Add a web server to accept SSL connections on port 8443.
+- Use Jetty for the web server since Kafka already uses it therefore avoids an additional dependency.
+- Reuse the broker's own certificate that would be in the container for TLS authentication. 
+- Configure the server to require TLS client authentication.
```
ppatierno, line 37: we are using mTLS terminology now, let's stick with that.

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,93 @@
...
+
+**KafkaRoller**:
+
+The proposed flow of KafkaRoller
+![The proposed flow of KafkaRoller](images/046-kafka-roller-new-flow.png)
+
+If a kafka pod does not become ready within the operational timeout (5 mins by default) take the following actions:
+1. Check if the maximum number of retries has been reached. If it has, fail the reconciliation.
+2. Ensure that the pod is not in a stuck state. The existing `isPodStuck()` can be used for this which returns true if the pod is in `CrashLoopBackOff`, `ImagePullBackOff`, `ContainerCreating` or `Pending` state. If the pod is stuck, continue with the current behaviour to retry rolling of the pod.
+3. Use the Pod DNS name to query the KafkaAgent API endpoint.
```
ppatierno, line 58: wdym by "query" here? What is described at steps 5 and 6? So actually authenticating and sending the GET request? If yes, then 5 and 6 should be put here in the right order then moving 4 down because you can get a connection refused or timeout (as suggested by Jakub) only after you try to authenticate and submit the GET request, or?

046-analyse-broker-metrics-before-rolling-kafka.md
ppatierno: Maybe we should also handle the case when Kafka Agent is running but it's not able to get the metrics from Kafka (so somehow Kafka is running but has a problem) ... in that case 503 could make sense.

046-analyse-broker-metrics-before-rolling-kafka.md
ppatierno: Wondering if, by using `broker-state` as endpoint name we could have an optional `recovery` field in the JSON when there is the need to show the remaining logs and segments to recover instead of 0.
It would be:

```json
{
  "brokerState": 2,
  "recovery": {
       "remainingLogsToRecover": 123,
       "remainingSegmentsToRecover": 456
   }
}
```

or 

```json
{
  "brokerState": 3,
}
```

or `recoveryState` instead of `recovery`.

tinaselenge: > Wondering if, by using broker-state as endpoint name we could have an optional recovery field in the JSON when there is the need to show the remaining logs and segments to recover instead of 0. 

I think we would always need those fields but only use them to log if brokerState==2, to avoid sending multiple HTTP requests.

ppatierno: Why multiple requests? If brokerState is 3, there is nothing to log. If brokerState is 2, you get the fields and log them. Everything in one request.

scholzj: > Yes, I was thinking it would return 200 but I'm not sure what you mean by using it as healthcheck later?

The HTTP probe works the way that it checks for HTTP status `>= 200 && < 400`. So, depending on the status codes, this endpoint might or might not be usable for a liveness probe or readiness probe based on the HTTP endpoint.

E.g. if it always returns 200 here, we would need a separate endpoint for readiness to return HTTP 200 only when state is 3 for example - which is not wrong per-se, just got to be clarified.

tinaselenge: For liveness and readiness probe we would probably have to create an internal non-TLS endpoint, which could then return 200 only if broker state is >3 and !=127 ?

tinaselenge: > Why multiple requests? If brokerState is 3, there is nothing to log. If brokerState is 2, you get the fields and log them. Everything in one request.

Ah I misunderstood it initially. So the server can include the recovery fields in the response only if the broker state is 2. 

I will clarify both points above in the proposal :)

scholzj (approved):
Just a nit for the formatting. But LGTM otherwise. Thanks a lot for the effort and the patience answering the questions!

046-analyse-broker-metrics-before-rolling-kafka.md
```diff
@@ -0,0 +1,112 @@
...
+
+**KafkaAgent**:
+
+- Add a web server to accept SSL connections on port 8443.
+- Use Jetty for the web server since Kafka already uses it therefore avoids an additional dependency.
+- Reuse the broker's own certificate that would be in the container for TLS authentication.
+- Configure the server with mTLS authentication.
+- Expose an endpoint `v1/broker-state` that includes the API version in the URI path.
+- Extend KafkaAgent to collect `kafka.log:type=LogManager,name=remainingLogsToRecover` and `kafka.log:type=LogManager,name=remainingSegmentsToRecover` metrics.
+- Use Jetty's native ServerHandler API to handle a `GET` request and return a JSON response that would look like this:
```
ppatierno, line 42: just for completeness with the next two returns, let's add "return HTTP 200 (OK) with a body JSON response ..."

ppatierno (approved):
Just a nit, but LGTM! Thanks for addressing all feedback! ;-)

tinaselenge:
Thanks so much everyone for reviewing the proposal and providing very good feedback. I have addressed the last 2 comments, so it should be ready to merge if there is nothing else. 🙂

scholzj:
I think we should give more time for other maintainers to review it after the last changes. For whatever reason, a lot of them were removed from the review request and I had to add them again yesterday. So I think we should not merge it yet.

CC @strimzi/maintainers + @tombentley who looks like he commented but did not approved.

im-konge (approved):

tombentley (approved):

tombentley:
Thanks for this @tinaselenge, it is an important (and long overdue) improvement.

see-quick (approved):

scholzj:
We have now 6 binding +1 votes and 1 non-biding +1 (well, 2 including the proposal author). Let's will keep it open for another day and if nobody else raises any comments we should proceed and merged this as approved.

scholzj:
Approved with 6 binding votes. Thanks for theproposal @tinaselenge

yyang48:
@tinaselenge Thanks for opening this proposal! We hit the same issues from time to time in our production. Let me know if any help, pr reviews or etc, I can make to make this land in the release.

PR #73, opened 2023-02-07
000-proposal-enhance-image-extension

aidanmelen:
# Enhance Image Extension for Custom Authentication

## Introduction

Strimzi provides an automated and managed way of running a Kafka cluster, but some teams still prefer hosting their Kafka cluster using cloud providers. The current `kafka-base` image provided by Strimzi can be easily extended for custom JARs, but the generated configuration scripts make it difficult to add custom configuration properties for those JARs. This proposal aims to enhance the image extension by allowing teams to easily add custom JARs and provide custom configuration properties for those JARs.

## Problem Statement

The current configuration generation scripts in the `kafka-base` image prevent overriding the `security.protocol` and adding custom properties without also overriding the entire generate script. This makes it difficult to configure a custom jar and authenticate with a Kafka cluster hosted by a cloud provider.

## Proposed Solution

The proposed solution is to update the generate scripts to interpolate a `<COMPONENT>_ADDITIONAL_CONFIGURATION` environment variable and append its value (a multi-line string of ini-formatted properties) to the bottom of the properties file. This way, the properties supplied by this environment variable would not be overridden by Strimzi. 

## Affected Components

The following components would be affected by this proposal:
- Kafka Connect
- MirrorMaker2
- MirrorMakerProducer
- MirrorMakerConsumer
- Kafka Bridge

## Rejected Alternatives

A `.spec.authentication.mechanism.custom` option has been proposed as a way to allow authentication overrides, but this proposal focuses on solving the problem through image extension rather than the CRD. This approach would allow the Strimzi project to focus on Strimzi Kafka and simply provide a valid extension path.


000-template.md
```diff
@@ -1,29 +0,0 @@
-<!-- This template is provided as an example with sections you may wish to comment on with respect to your proposal. Add or remove sections as required to best articulate the proposal. -->
```
scholzj, line 1: Please do not delete the template.

000-proposal-enhance-image-extension.md
```diff
@@ -0,0 +1,26 @@
+# Enhance Image Extension for Custom Authentication
```
scholzj, line 1: This should be properly modeled through a new authentication type in the CRD. That allows it to be maintainable in the long term. The `<COMPONENT>_ADDITIONAL_CONFIGURATION` does not offer proper support for a well designed API, its validation etc.

000-proposal-enhance-image-extension.md
```diff
@@ -0,0 +1,26 @@
...
+The following components would be affected by this proposal:
+- Kafka Connect
+- MirrorMaker2
+- MirrorMakerProducer
+- MirrorMakerConsumer
+- Kafka Bridge
+
+## Rejected Alternatives
+
+A `.spec.authentication.mechanism.custom` option has been proposed as a way to allow authentication overrides, but this proposal focuses on solving the problem through image extension rather than the CRD. This approach would allow the Strimzi project to focus on Strimzi Kafka and simply provide a valid extension path.
```
scholzj, line 26: Without a properly designed API, we cannot guarantee that this will work in the next release, we cannot do proper validation etc. That makes it unsupportable.

aidanmelen:
https://github.com/strimzi/strimzi-kafka-operator/issues/5473#issuecomment-1420964328

000-template.md
aidanmelen: yikes, sorry.

aidanmelen:
@scholzj Thank you for your time and the amazing operator. Cheers!

PR #76, opened 2023-03-16
Unidirectional topic operator

tombentley:
This is currently still in draft and not ready for general review. Anyone who is _especially_ interested is welcome to comment. This is based on an [earlier draft](https://hackmd.io/88umcDZAScKTDHMjFAPdmA?view). 

I'm working on a [proof of concept](https://github.com/strimzi/strimzi-kafka-operator/compare/main...tombentley:barnabas:unidirectional-topic-operator?expand=1) which has informed the design so far.


050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,680 @@
...
+> 2. The user changes the new `spec.managed` field to `False` so that the topic is no longer Kube-managed.
+> 
+>     ```yaml=
+>     metadata:
+>       generation: 124
+>       name: foo
+>       finalizer: 
+>       - strimzi.io/topic-operator
+>     spec:
+>       managed: False # do not synchronize to Kafka
```
tombentley, line 371: The user operator uses a `strimzi.io/pause-reconciliation` annotation to do a similar thing. We should use that (if the semantics are the same), or be explicit about why we've not using it (if the semantics of `spec.managed` are necessarily different)

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,680 @@
...
+>     status:
```
tombentley, line 386: We should also remove the `spec.topicName` to help indicate that there might not be a topic in Kafka for this `KafkaTopic`.

050-unidirectional-topic-operator.md
scholzj: This annotation is used in Cluster Operator as well. When set, the condition in `.status` is set to `Paused`.

050-unidirectional-topic-operator.md
scholzj: You cannot do that, because if you remove `.spec.topicName` you might not later know what the actual topic name was, or?

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,680 @@
...
+// Call out any future or backwards compatibility considerations this proposal has accounted for.
+
+To migrate, existing BTO users would have to:
+
+1. Review how they're using the BTO and whether they require birdirectionality
+2. Undeploy the BTO (they can retain their existing `KafkaTopics`)
+3. Deploy the UTO
+
+
+### Disadvantages
```
scholzj, line 651: The use of finalizers should be listed in the disadvantages as well I think. It is something that might have its justification, but it also brings issues to deletions.

050-unidirectional-topic-operator.md
tombentley: What happens if you delete a paused CR?

050-unidirectional-topic-operator.md
tombentley: That would be fine, actually. 

1. `spec.managed=false` 
2. then changing the `spec.topicName` 
3. then `spec.managed=true`

 just means you changed which topic in Kafka is being managed by the resource, just like it would if step 2 was deletion and recreation.

050-unidirectional-topic-operator.md
tombentley: Yes, done.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,717 @@
...
+* For errors when processing a ZooKeeper znode watch.
+  UTO doesn't use znode watches, so this cannot happen.
+* On attempts to change a `KafkaTopic`'s `spec.topicName`.
+  For UTO this will be reported via the status only.
+* On detection of conflicting changes between Kube and Kafka
+  For UTO such conflicts cannot happen.
+* On attempts to decreate the number of partitions.
+  For UTO this will be reported via the status only.
+
+Thus, the UTO will emit no events.
```
devguyio, line 670: I find that Kubernetes events provide a different dimension to native kube applications observability than logging, metrics, and status conditions. 
1. **Compared to logs & Metrics** : UTO logs and metrics are meant to be infrastructure status targeting the cluster administrators. In contrast, namespaced Kube events target the CR authors who might not have access to the CO logs and metrics. This becomes increasingly important for operators that provide some form of namespace policies or multi-tenancy-like operation mode such as the UTO. 
2. **Compared to status** Kube events provide more of a historical view of state change or error progression. For E.g. a current `Ready` resource with 0 events is different from a `Ready` resource with associated Kube events that show e.g. conflict -> ready -> conflict. Especially for a case when another resource with a conflicting name is created that ends up in **both** resources becoming not ready.

devguyio (commented):
I took a look at this since I'm interested in the Topic Operator. I like the move to the Unidirectional topic operator and the rational behind stopping pursuing a BTO makes sense to me.

Do you think it's beneficial to find a way to decouple the KafkaTopic CRD from the actual UTO implementation to allow multiple KafkaTopic operators/implementations for such use cases as when topic managed is done via a 3rd party API,  the `Admin Server` or even the Admin API ?

IMHO since this proposal will result in an implementation of a new operator and it includes an API change I think it might be worth considering adding support for a TopicClass pattern (i.e. such as GatewayClass, StorageClass etc) . E.g.

```yaml
kind: KafkaTopic
metadata:
  name: consumer.offsets
spec:
  # Optional and can have a cluster default value or per-namespace default value
  topicProviderClassName: foo-kafka-provider
  topicName: __consumer_offsets
  replicas: 3
  partitions: 50
  config:
    retention.ms: 123456789
```

This should be a small API change and small implementation detail in the UTO that would open the doors to more user adoption of Strimzi IMHO.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,717 @@
...
+> 
+> - Topics `foo-app.x` and `foo-app.y` can only be managed by `KafkaTopics `in namespace `team.a` (due to the `foo-app.` prefix), likewise `bar-app-x` and `bar-app-y` (due to the `bar-app-` prefix). Additionally `config-foo` can be only managed from this namespace (due to the `topicNames` list).
+> - Topics `quux-app.x` and `quux-app.y` can only be managed by `KafkaTopics `in namespace `team.b` (due to the `quux-app.` prefix).
+> Additionally `config-quux` can be only managed from this namespace (due to the `topicNames` list).
+> - Topics `config-gee`, `__consumer_offsets`, and `__transaction_state` could only be managed from the `kafka.cluster.admins` namespace.
+>
+> Rules for legal policies:
+> 
+> * All `topicNamePrefixes` must be disjoint (i.e. not given prefix is a prefix of any other prefix).
+> * Not `topicNamePrefix` maybe be a prefix of of any listed `topicNames`.
```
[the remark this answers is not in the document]
tombentley:
```suggestion
> * No given `topicNamePrefix` maybe be a prefix of of any listed `topicNames`.
```

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,717 @@
...
+<tr><td>STRIMZI_STORE_NAME				<td>Dropped
+<tr><td>STRIMZI_APPLICATION_ID				<td>Dropped
+<tr><td>STRIMZI_STALE_RESULT_TIMEOUT_MS			<td>Dropped
+<tr><td>STRIMZI_USE_ZOOKEEPER_TOPIC_STORE		<td>Dropped
+</table>
+
+Notes:
+
+1. `STRIMZI_NAMESPACE` is retained for backwards compatibility and configures the operator with a policy allowing all topics in the given namespace.
+   `STRIMZI_POLICY_JSON` configures a policy as JSON string.
```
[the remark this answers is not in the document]
tombentley: I don't have a strong preference. I went with JSON because I felt a suspicious of env vars containing newlines  which YAML would require (it seems likely that something, somewhere, eventually is going to do something with the significant whitespace).

050-unidirectional-topic-operator.md
tombentley: @devguyio that's true, but I don't see any of those existing BTO events really requiring an `Event`. The one example you give (`ResourceConflict`) can only happen as a result of someone creating a conflicting `KafkaTopic` in the same namespace as an existing one. Presumably they're looking for the `Ready` in that case and will resolve the problem more or less immediately. Those resources being in conflict doesn't have any impact on how the topic is being used by Kafka applications. So I just don't seem much value in the historical state change in that case.

I think there might be a role for `Events` in the future. For example if/when we add support for RF change then when data movement is required it can take a long time. There seems a stronger case for using events to notify about those state changes (perhaps in addition to a 2nd condition with `type: ReplicasReady` perhaps), because we don't expect anyone to waiting on the outcome. But I don't want to say that in _this_ proposal, because I don't want to bite off RF change at the same time as a whole new operator.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,717 @@
...
+
+### The "single resource" principal
+
+To avoid ambiguity we will continue to require that a single `KafkaTopic` is used to manage a single topic in a single Kafka cluster.
+
+The fact that there are legal topic names (in Kafka) that are not legal resource names (in Kube) provided the motivation for supporting `spec.topicName`, so that a topic in Kafka could be managed by a `KafkaTopic` with a different `metadata.name`.
+As such it's not possible for the Kube apiserver to enforce the single resource principal because multiple `KafkaTopic` resource might refer to the same topic (i.e. the uniqueness constraint on `metadata.name` is insufficient).
+It is therefore left to the operator to detect if there are multiple `KafkaTopics` which point to the same topic in Kafka.
+
+To do this, the operator will keep an in-memory mapping of topic name to (namespace, name)-pairs.
```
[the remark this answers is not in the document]
tombentley: We could do that, but:

* Admission webhooks are essentially global to the Kube cluster, whereas there can be multiple independent TO's in a Kube cluster. To support multiple TOs the webhook would need a global view over all those TOs and how they mapped to Kafka clusters, because it's fine to have two `KafkaTopics` with `spec.topicName=foo` if they're in different Kafka cluster.
    - even if we felt this was worth the complexity, consider too that this needs to cope with TOs coming and going, being reconfigured. So the incremental complexity for the full dynamic case is significantly more than the base complexity of the purely static case.
* It's another process/pod/deployment:
    - it would _still_ need an in-memory map AFAICS, only this time potentially a much larger map, because it's now keyed on (clusterId, topicName)-pairs and covers all the Kafka clusters in the Kube cluster. Probably it's roughly the same memory usage in total, just it's now lumped in one process, but there doesn't seem to be an _advantage_ wrt memory usage overall. 
    - it creates a new failure mode where the TO is running but the WebHook is not configured. So it doesn't provide as good safety guarantee as making the same check in the TO itself.

tombentley:
@devguyio 

> Do you think it's beneficial to find a way to decouple the KafkaTopic CRD from the actual UTO implementation to allow multiple KafkaTopic operators/implementations for such use cases as when topic managed is done via a 3rd party API, the Admin Server or even the Admin API ?
>
> IMHO since this proposal will result in an implementation of a new operator and it includes an API change I think it might be worth considering adding support for a TopicClass pattern (i.e. such as GatewayClass, StorageClass etc) 

That's an interesting idea, but I don't see how it helps with using an Admin server. A `TopicClass` could be used to specify that the operations on the CR happened via an Admin server, but that doesn't change the source of truth. The problem is: when an operator is used the `KafkaTopics` it consumes become the source-of-truth for the configuration of those topics. Any seperate Admin server (or other tooling):

* either talks directly to Kafka via an `Admin` client, in which case Kafka needs to be the source of truth, else the UTO will overwrite any changes done via the Admin (as called out in this proposal)
* or talks to the Kube apiserver, i.e. updates the `KafkaTopics` (which become the source of truth), leaving the UTO to actually update Kafka. This works without conflict, but means that 
    - either your Admin server needs to know about `KafkaTopics` (i.e. in general you're locked into using `KafkaTopic`-aware tooling, which is already mentioned as a Disadvantage in this proposal)
    - or you get into the game of intercepting the Kafka protocol requests for things like `CreateTopics` etc. That can be done, but comes with additional deployment costs and it's not ideal for producer and consumer latency to add a proxy. I suppose you could also work around that using multiple listeners and an `Authorizer` plugin so that the topic mutating RPCs were only allowed for certain listeners which would be proxied, and `Produce` and `Fetch` RPCs would use an unproxied listener. But this is now pretty complicated and well outside the scope of _this_ proposal.
* the only other alternative is to make Kafka itself the source of truth. Then to allow update of `KafkaTopics` you're playing a different game: pretending to be a Kube apiserver. Kube has [a way of doing that](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/) but it comes with a different bunch of compromises:
    - it's global to the kube cluster (so as with the admission web hook you have a problem mapping KafkaTopics to the Kafka clusters they're supposed to be in). 
    - cluster administrators typically are not at all keen on api aggregation, so while it seems an elegant solution in theory it's unlike to see little in the way of practical uptake.

But having written all of this, I'm still wondering whether I'm missing your point entirely.

devguyio:
> @devguyio
> 
> > Do you think it's beneficial to find a way to decouple the KafkaTopic CRD from the actual UTO implementation to allow multiple KafkaTopic operators/implementations for such use cases as when topic managed is done via a 3rd party API, the Admin Server or even the Admin API ?
> > IMHO since this proposal will result in an implementation of a new operator and it includes an API change I think it might be worth considering adding support for a TopicClass pattern (i.e. such as GatewayClass, StorageClass etc)
> 
> That's an interesting idea, but I don't see how it helps with using an Admin server. A `TopicClass` could be used to specify that the operations on the CR happened via an Admin server, but that doesn't change the source of truth. The problem is: when an operator is used the `KafkaTopics` it consumes become the source-of-truth for the configuration of those topics.

What I'm proposing is splitting the source of truth for the "managed" `KafkaTopic` resources across two or more operators, each reacting only to the `KafkaTopic` resources with its designated `TopicClass`.

So basically the changes to Strimzi  would be:

1. Adding an optional `topicClassName` field to the `KafkaTopic` spec.
1. An OOTB Strimzi UTO that works as follows
    1. Topic management of  `KafkaTopic` resources with an empty or default `topicClassName`  && `managed != False` according to this proposal. 
    2. Treat any `KafkaTopic` resources with `topicClassName` set to something other than default as "unmanged" and defer its management to another operator.
    
So for the admin server scenario there will exist another operator e.g. CustomerAdminTopicOperator that provides the `customer-admin-server' class and works exactly as above but only for its designated class. 

E.g.
1. User creates a `KafkaTopic` resource with `topicClassName: customer-admin-server` .
2. Strimzi OOTB UTO observes the resources class name and ignores this resources.
3. CustomerAdminTopicOperator observes the resource with correct class name and proceeds to do the validation and topic sync using an Admin API
4. The topic shows up in Kafka and the OOTB UTO considers it as unmanaged, like any other topic.
5. It's up to the CustomerAdminTopicOperator to ensure no conflict occurring between topics created via `KafkaTopics` and the Admin API externally (e.g the UI) . 

Do you think this would work?

I can imagine there can be some improvements such as moving the topic naming policies to a ConfigMap to be shared by all TopicClass operators etc.

050-unidirectional-topic-operator.md
devguyio: Yeah this makes sense to me.

tombentley:
> What I'm proposing is splitting the source of truth for the "managed" `KafkaTopic` resources across two or more operators, each reacting only to the `KafkaTopic` resources with its designated `TopicClass`.
> 
> So basically the changes to Strimzi would be (in addition to what exists in this proposal):
> 
>     1. Adding an optional `topicClassName` field to the `KafkaTopic` spec .
> 
>     2. An OOTB Strimzi UTO that works as follows
>        
>        1. Topic management of  `KafkaTopic` resources with an empty or default `topicClassName`  && `managed != False` according to this proposal.
>        2. Treat any `KafkaTopic` resources with `topicClassName` set to something other than default as "unmanged" and defer its management to another operator.
> 
> 
> So for the admin server scenario there will exist another operator e.g. CustomerAdminTopicOperator that provides the `customer-admin-server' class and works exactly as above but only for its designated class.

@devguyio: 

1. Something like this would be possible anyway because the UTO will do the same label selection on the `KafkaTopics` is consumes at the BTO does. By configuring a UTO and a `CustomerAdminTopicOperator` with disjoint label selectors you could partition the resources in a namespace between two (or more) operators. One difference is that the `topicClass` makes this much more explicit, which might be useful, but doesn't appear to be of fundamental importance.
2. The behaviour of this `CustomerAdminTopicOperator` would differ from the UTO how? By talking so an REST Admin server, rather than using `Admin` directly? That doesn't help with the source of truth problem .I don't think you were suggesting it did, but it's the SoT problem which makes the UTO incompatible with topic management via any other means, and is the biggest weakness of the current proposal IMO.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,718 @@
...
+
+1. the operator reacting to the new `KafkaTopics`.
+2. the application starting and triggering autocreation of the topics.
+
+**If the operator wins:** then auto-creation will not happen, and the user's intent is honoured.
+
+**If the application wins:** then the topics will be created using the default configuration for the cluster. The operator will later reconcile the `KafkaTopics` which will result in reconfiguration the default configuration for auto-created topics differs from the `KafkaTopic.spec`. That reconfiguration will either succeed, or fail (e.g. because a `KafkaTopic` has a lower `spec.partitions` than the auto-created topic).
+
+To avoid this it is **recommended** that the Topic Operator not be used with cluster where `auto.create.topics.enable` is `true`. 
+The operator will emit a `WARN` level log on startup if autocreation is enabled.
```
fvaleri, line 534: I've seen this happening with users more than one time and recovering from bad cluster defaults requires service disruption. In addition to the log on startup, can we also emit a Kube event with this warning?

fvaleri (commented):

050-unidirectional-topic-operator.md
tombentley: Do we know that users are more likely to notice a Kube `Event` than a warning in the logs?  

Emitting a Kube `Event` doesn't avoid any service disruption caused by changing the default cluster behaviour. It only makes the likely mistake more noticeable. And it's not _necessarily_ a mistake: It's only a problem if the user isn't creating their topics prior to application deployment. And even if it _is_ a mistake, it's not necessarily consequential (we expect the topic would get reconfigured almost immediately, so it's only a problem for certain changes, like RF change, or changing the #partitions). 

I'm not opposed to doing this, but I'm not completely convinced it would actually make a difference.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,718 @@
...
+
+To avoid ambiguity we will continue to require that a single `KafkaTopic` is used to manage a single topic in a single Kafka cluster.
+
+The fact that there are legal topic names (in Kafka) that are not legal resource names (in Kube) provided the motivation for supporting `spec.topicName`, so that a topic in Kafka could be managed by a `KafkaTopic` with a different `metadata.name`.
+As such it's not possible for the Kube apiserver to enforce the single resource principal because multiple `KafkaTopic` resource might refer to the same topic (i.e. the uniqueness constraint on `metadata.name` is insufficient).
+It is therefore left to the operator to detect if there are multiple `KafkaTopics` which point to the same topic in Kafka.
+
+To do this, the operator will keep an in-memory mapping of topic name to (namespace, name)-pairs.
+This will allow us to detect the case where multiple resources are attempting to manage the same topic in Kafka.
+When this happens both resources' `Ready` status will change to `False` with suitable `reason`.
```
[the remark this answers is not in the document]
tombentley: I guess we could do that. It's not impossible for the creationTimestamps to be the same, but very unlikely. I think this will work, but I'll confirm.

050-unidirectional-topic-operator.md
fvaleri: > It's only a problem if the user isn't creating their topics prior to application deployment. 

Agree, but this is exactly the situation depicted in the gitops vs application race, which is fairly common nowadays. I think we could be fine with the log warning and a documentation paragraph/section describing potential issues with topic auto autocreation, maybe in relation to few common use cases.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+>       The topic in Kafka will thus become unmanaged.
+>    3. Otherwise, the operator attempts to deletes the topic from Kafka.
+>        1. If deletion succeeds the operator will remove `strimzi.io/topic-operator` from the `metadata.finalizer` and Kube will remove the resource.
+>        2. If deletion failed with `UNKNOWN_TOPIC_OR_PARTITION` (i.e. the topic didn't exist) the operator will continue to remove the finalizer. I.e. this case is not treated as an error.
+>        3. If deletion failed with any other error this is reported via a condition in the `status.conditions` and the finalizer is not removed.
+---
+
+Similarly to how it is possible to "manage" an existing unmanaged topic in Kafka, we will also provide a mechanism for "unmanaging" a managed topic, via a `spec.managed` flag.
+This will allow a `KafkaTopic` to be deleted **without** the topic in Kafka being deleted.
+This can be useful operationally, for example to change the `metadata.name` of a `KafkaTopic`.
```
scholzj, line 354: Have you considered removing the finalizer from a topic with `spec.managed: false`?

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+>       name: foo
+>       finalizer: 
+>       - strimzi.io/topic-operator
+>     spec:
+>       managed: False # do not synchronize to Kafka
+>     status:
+>       observedGeneration: 123 # != generation => it is Kube-managed
+>     ```
+> 
+> 3. The user waits until the operator has observed that change, indicated by the `status.observedGeneration` matching the `metadata.generation`:
```
scholzj, line 391: For the paused reconciliation, we also set a `Paused` condition. I think we should do that here as well. It might simplify some of the automation around this and confirm that it is really unmanaged.

050-unidirectional-topic-operator.md
scholzj: I actually don't know for sure about User Operator. I expect the deletion will proceed since eventually in the next timer based reconciliation, it will not remember anymore that this was paused user. For the resources such as `Kafka`, the Kube GC will wipe the resources right away.

050-unidirectional-topic-operator.md
scholzj: Yes. But the point is that:
* You might fight with some GitOps tooling which would set the `spec.topicName` back
* Assuming the GitOps tooling did not reverse your change, the user will be unable to just edit the spec and set `spec.managed=true` to recreate the topic / have it managed again.

The procedure you describe is fine if done by the user. But I do not think the UTO should edit the `.spec` this way.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+        lastTransitionTime: 20230301T103000Z
+    ```
+
+> Note that both these `NotSupported` reasons can arise both from changes being made to the `KafkaTopic`, or changes made directly in Kafka.
+> For example, the UTO cannot tell whether the `spec.partitions` was decreased, or whether it was unchanged and the conflict arises because someone increased the number of partitions directly in Kafka so that it looks like the `KafkaTopic` is requesting a decrease. 
+> In any case, when the user takes action to fix the problem the `KafkaTopic` will (eventually) get reconciled and the `status` will revert to `Ready`.
+
+### Topic deletion
+
+We will use a [Kube finalizer](https://kubernetes.io/blog/2021/05/14/using-finalizers-to-control-deletion/) for deletion.
```
scholzj, line 279: The finalizers still make me nervous. Have you considered the possibility to allow users to disable their use and take the risk the topics might not delete in Kafka when the `KafkaTopic` is deleted?

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+<tr><td>STRIMZI_REASSIGN_THROTTLE			<td>Dropped, see note 2
+<tr><td>STRIMZI_REASSIGN_VERIFY_INTERVAL_MS		<td>Dropped, see note 2
+<tr><td>STRIMZI_TOPIC_METADATA_MAX_ATTEMPTS		<td>Dropped
+<tr><td>STRIMZI_TOPICS_PATH				<td>Dropped
+<tr><td>STRIMZI_STORE_TOPIC				<td>Dropped
+<tr><td>STRIMZI_STORE_NAME				<td>Dropped
+<tr><td>STRIMZI_APPLICATION_ID				<td>Dropped
+<tr><td>STRIMZI_STALE_RESULT_TIMEOUT_MS			<td>Dropped
+<tr><td>STRIMZI_USE_ZOOKEEPER_TOPIC_STORE		<td>Dropped
+</table>
```
scholzj, line 644: Markdown table might be more readable?

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+<tr><td>counter		<td>strimzi.reconciliations.periodical	<th>Unchanged
+<tr><td>counter		<td>strimzi.reconciliations		<th>Unchanged
+<tr><td>counter		<td>strimzi.reconciliations.failed	<th>Unchanged
+<tr><td>counter		<td>strimzi.reconciliations.successful	<th>Unchanged
+<tr><td>gauge		<td>strimzi.resources			<th>Unchanged
+<tr><td>timer		<td>strimzi.reconciliations.duration	<th>Unchanged
+<tr><td>gauge		<td>strimzi.reconciliations.paused	<th>Unchanged
+<tr><td>counter		<td>strimzi.reconciliations.locked	<th>Unchanged
+<tr><td>gauge		<td>strimzi.resource.state		<th>Unchanged
+</table>
```
scholzj, line 730: Markdown table?

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+
+Thus, the UTO will emit no events.
+Any software which consumed BTO events will not be compatible with UTO.
+
+## Migration
+
+To migrate, existing BTO users would have to:
+
+1. Review how they're using the BTO and whether they require bidirectionality
+   1. If bidirectional support is required by their usage then the UTO cannot be used.
```
scholzj, line 761: Please clarify here that we will not support the BTO going forward.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+That can prevent other resources from being removed (for example the `Namespace` containing the `KafkaTopics`).
+In these cases it is possible to remove the finalizer manually (e.g. via `kubectl edit`), which will then allow the apiserver to remove any other resources.
+
+#### Compatibility with other tooling
+
+This approach is not easily compatible with other topic management tooling. This includes things like the `kafka-configs.sh` and other Apache Kafka scripts as well as any other tooling which uses the `Admin` client, including applications themselves.
+
+<!--Even using the "scheduled auto creation" option, changes done via the admin server would be reverted by the next timed reconciliation. -->
+
+In the particular case of an "Admin server", to provide CLI or UI functionality in a Strimzi-aware way it _might_ be possible to use wrap `Admin` (decorator pattern), or write it using some abstraction over `Admin` which could also support operations on `KafkaTopics` for those interactions which would otherwise result in conflicting sources of truth.
```
scholzj, line 780: There is no _Admin server_ in Strimzi. I think this should be removed or be more generic. The Admin server was removed / archived long time ago with the UI.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+
+
+## Future work
+
+This propsal aims to set the direction and define the semantics of the UTO.
+It is not intended, by itself, to define the complete post-BTO picture.
+Future work _could_ include:
+
+* Providing tooling to conveniently create `KafkaTopics` for existing topics in Kafka.
+* Adopting the config partitioning idea to allow multiple `KafkaTopics` to manage disjoint subsets of topics' configurations.
```
scholzj, line 792: Possibly you can add here tooling to remove the finalizers?

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+>     Since the `spec.magnaged` is `False` it does **not** attempt topic deletion.
+>     1. It removes `strimzi.io/topic-operator` from the `metadata.finalizer`.
+>     2. Kube proceeds to remove the resource.
+
+Note that `spec.managed=False` means that a `KafkaTopic` can exist without there being a corresponding topic in Kafka.
+
+---
+
+
+### Multi-namespace support
```
scholzj, line 432: The proposal mostly looks good to me. But I think that this is a bit premature to be included in it. I think the multi-namespace support needs to be considered and designed in the context of the UO as well.

* It would be weird to have multi-namespace TO and not UO => it should be probably added in parallel
* It would be good to consider how the namespacing is done in TO and UO. In the UO, in the past, we considered following Kubernetes and their naming principles by suffixing the user name with the namespace as a suffix. This approach might not be suitable for TO and we might intentionally choose a different approach for each of them. But we should do so intentionally after consideration and not arrive at it randomly.
* The policy seems to introduce a parallel security aspect (parallel to Kafka ACLs / UO and to Kube RBAC). It seems to be very limited (managing topics through KafkaTopic resources) while not impacting in any way the other security parts - especially the Kafka ACLs. So, a `team.a` is seemingly an owner of topics `foo-app.*`. But `team.b` can easily create users to produce or consumer to / from these topics. It can even create users with ACLs to delete or modify them. I think this would need to be properly thought through and linked to the UO and ACL management to create some value. Doing it without it can prevent doing this in the future without compatibility issues.

That all seems to be way above this proposal. I would probably prefer to do one of the following:
* Remove the multi-namespace support from this proposal and leave it for later
* Adapt the multi-namespace support to give us more options for the future. E.g.
    * Remove the policy concept until it is properly thought through and deal with multi-namespace conflicts in the same way as for single-namespace conflicts for the time being.
    * Acknowledge the possible future differences how UO might handle multi-namespace support.

----

Assuming you would want to proceed with this as is, I think it is missing clarification on how the multi-namespace watching will be configured. And what the _default_ policy will be. Does the policy configure that? Does the example policy below mean that only namespaces `team.a`, `team.b` and `kafka.cluster.admins` will be watched? That would be inefficient as the number of watches would not scale well when using per-namespace watch.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+    livenessProbe: # probe values may need to be changed
+    readinessProbe: # probe values may need to be changed
+    startupProbe: # probe values may need to be changed
+    resources: # Resource requirements may differ
+    jvmOptions: # JVM options may need to be changed
+    logging:  # Logger names will need to be changed
+
+    # Unchanged options
+    image: 
+    watchedNamespace: # becomes optional
```
scholzj, line 680: This is already optional -> defaults to the namespace of the Kafka CR.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+  For UTO this will be reported via the status only.
+* On detection of conflicting changes between Kube and Kafka
+  For UTO such conflicts cannot happen.
+* On attempts to decrease the number of partitions.
+  For UTO this will be reported via the status only.
+
+Thus, the UTO will emit no events.
+Any software which consumed BTO events will not be compatible with UTO.
+
+## Migration
```
scholzj, line 756: I think the standalone BTO to UTO migration is quite clear. But the migration within a Kafka cluster would IMHO deserve more details. Will a feature gate drive it? Will there be a flag? Annotation?

050-unidirectional-topic-operator.md
scholzj: Perhaps calling it a tenancy model would be more matching than a security model as I referred to it in the initial comment. But I think the core of the comment stands.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+
+### Non-goals
+
+* Supporting RF change (this could be added later)
+
+### CR API
+
+We would remain schema-compatible with the existing `KafkaTopic` resource at the API level. 
+This means that all existing `spec` section could be handled by the new operator and all existing fields of the `status` would still be present for applications that consume the status.
+However, the semantics of a `KafkaTopic` would change incompatibly: The `spec` would no longer be updated when Kafka-side configuration changed.
```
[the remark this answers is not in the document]
tombentley: This is correct, actually. While the _syntax_ changes are backwards compatible, the `spec` no longer behaves in exactly the same way.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+>     ```yaml=
+>     metadata:
+>       generation: 1
+>       name: foo
+>     spec:
+>       partitions: 12
+>       replicas: 50
+>    ```
+> 
+> 2. The operator checks its in-memory map to see whether `foo` is already managed by some other resource.
```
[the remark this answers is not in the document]
tombentley: You're right that this example isn't completely clear. I'll try to reword it.

But to explain here: The operator first checks to see whether there is another resource trying to manage the same topic name, before it even knows whether a topic of that name exists in Kafka. For this example, we're describing the overall outcome where no topic of that name exists in Kafka, and here considering whether there are two resources  trying to manage the same topic.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+> 
+>     1. If so, then both resources are updated with error conditions in their `status`.
```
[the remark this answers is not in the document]
tombentley: I forgot to update this bit when I changed this from "both get put into `ResourceConflict`" (which is what I originally had) to "only the newest resource gets put into `ResourceConflict`" aka "first in wins" (which was suggested by @fvaleri). There are pros and cons either way:

Marking both as conflicted:
* means that the topic is temporarily unmanageable via either resource
* means that 
    * either the operator needs to update another resource than the one that changed, complicating the locking which we use to prevent concurrent reconciliation in the case of multi-threaded reconciliation.
    * or that we don't try to update the status of "the other" resource when reconciling "this one", leaving it to the next timed reconciliation to update the other one. The lagginess of this simpler alternative is not attractive.
* means that the error message on the condition needs to enumerate every conflicting resource, which could actually be many more than 2 in the general case.

Marking only the later-created resources is actually simpler, since updates to other resources are avoided and the conflicted resources need only point to the first-created resource. These are all relatively shallow considerations, but together they persuaded me to make the change.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+Changes thus detected will result in an error condition.
+
+#### Decreasing `spec.partitions`
+
+This will NOT be supported, because it's not supported by Kafka.
+The user will either need to revert the `spec.partitions`, or recreate the topic in Kafka (which they can do via the `KafkaTopic` or not).
+
+#### Changing `spec.relicas`
+
+This will NOT be supported using as part of this proposal.
```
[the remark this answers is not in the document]
tombentley: That would be nice, but _this_ proposal guarantee that.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+Thus, the UTO will emit no events.
+Any software which consumed BTO events will not be compatible with UTO.
+
+## Migration
+
+To migrate, existing BTO users would have to:
+
+1. Review how they're using the BTO and whether they require bidirectionality
+   1. If bidirectional support is required by their usage then the UTO cannot be used. 
+2. Undeploy the BTO (they can retain their existing `KafkaTopics`)
```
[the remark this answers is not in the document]
tombentley: It's optional and defaults to true, so it's not necessary to add it.

050-unidirectional-topic-operator.md
tombentley: Yes. If a CR with `spec.managed: false` gets deleted then the finalizer would be removed by the operator.

050-states.png shows all the different paths between the different states of the CR.

Note that, as currently specified, there is path where a CR can get deleted without the finalizer being removed by the operator. It happens if the operator's label selector initially matches the CR (so that the finalizer gets added), but the CR is then changed so that it no longer matches the operator's selector (the operator interprets this as "it's not my resource any more", assuming some other operator instance is now managing the CR. But if no other operator takes ownership then when the CR is then deleted there is nothing to delete the finalizer.

scholzj: > Yes. If a CR with spec.managed: false gets deleted then the finalizer would be removed by the operator.

What I meant here was if you considered removing the finalizer already when the `spec.managed` is set to `false` and not just when it is deleted.

> Note that, as currently specified, there is path where a CR can get deleted without the finalizer being removed by the operator. It happens if the operator's label selector initially matches the CR (so that the finalizer gets added), but the CR is then changed so that it no longer matches the operator's selector (the operator interprets this as "it's not my resource any more", assuming some other operator instance is now managing the CR. But if no other operator takes ownership then when the CR is then deleted there is nothing to delete the finalizer.

This would actually mean such resource cannot be deleted rather then `can get deleted without the finalizer being removed` as the finalizer might stay there. When the labels change and no longer match the selector, the TO will get a DELETE event. So I wonder if we should unset the finalizer at this event. But I guess that might raise conflicts with other instances and would not guarantee anything anyway. So maybe we have to live with it.

050-unidirectional-topic-operator.md
scholzj: I was not really questioning the `.spec.managed` here. I'm wondering if we should have a flag to simply disable the finalizers when the user decides so. I.e. if they would be disabled, they would not be set but the user will accept that there might be situations when deleting the `KafkaTopic` does not trigger the deletion in Kafka.

050-unidirectional-topic-operator.md
tombentley: Since the UO doesn't use finalizers if a paused `KafkaUser` is deleted then the user will be removed from Kafka. So the behaviour seems pretty similar. Maybe we should use the annotation instead to be consistent. That would also fit nicely with using the Paused annotation instead of removing the `spec.topicName`. WDYT?

scholzj: You mean instead of having the `.spec.managed` field to have some annotation `strimzi.io/managed: false`? That works for me, but not sure it makes a big difference. 

Or do you want to use the paused-reconciliation annotation for this but not delete the topic from Kafka? TBH, it would work for me as well. But it would be a different behavior from the User or Cluster operator as there will the user or cluster be really deleted.

The TO obviously works differently as it will not have any _reconcile all topics_ method as it doe snot manage all topics. In the UO, it is not the deletion of the paused KafkaUser which triggers the deletion. It is the periodical reconciliation and having the CRs as a single source of truth. Similarly, It is not the CO deleting the Kafka cluster which is paused - it is Kubernetes GC. So code-wise, the TO behavior would be the same. But from the user perspective it would differ.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,801 @@
...
+>     status:
+>       observedGeneration: 124 # == generation => it is not syncrhonized
+>     ```
+> 
+> 3. The operator notices the `metadata.deletionTimestamp` field.
+>    1. If there is more than one `KafkaTopic` which is trying to manage this topic then deletion from Kafka is not attempted;
+>       the operator will remove the finalizer and Kube will remove the resource; 
+>       upon timed reconciliation the condition on the other `KafkaTopic` will be removed and reconciliation via that other `KafkaTopic` will proceed as normal.
+>    2. If deletion is disallowed via `delete.topic.enable=false` broker configuration then the operator will remove the finalizer and Kube will remove the resource. 
+>       The topic in Kafka will thus become unmanaged.
```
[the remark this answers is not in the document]
tombentley: As above, I don't think we need to specify that here.

050-unidirectional-topic-operator.md
tombentley: On balance: 

* I think the difference between paused and unmanaged is fundamental and important, so let's not use `strimzi.io/pause-reconciliation`.
* But it's probably more consistent to use an annotation for managed than to put it in the `spec`. So I'll change it to `strimzi.io/managed` with a boolean value, defaulting to true.
* I thought too about the `Paused` condition and whether to use that or a `Managed=false` condition. But the way the operator works it will simply apply the different logic on the next reconciliation: It's not a piece of state that needs to be reconciled. i.e. it's just not possible for a resource with the annotation to behave transiently in a way that's incompatible with the annotation.

tombentley:
@ppatierno + @scholzj I think I've addressed all your comments now, if you want to take another look. Thanks!

fvaleri (commented):
Thanks Tom. I left few more comments.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,813 @@
...
+In order to avoid collisions where two or more `KafkaTopics` are trying to manage the same topic in Kafka the operator will enforce a policy for which namespaces can manage which topics. 
+
+---
+
+#### Example
+
+> Conceptually (i.e. this may or may not be configured via YAML):
+> 
+> ```yaml=
+> policy:
```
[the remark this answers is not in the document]
tombentley: Note that this multi-namespace support is commented out.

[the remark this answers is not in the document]
tombentley: I guess I'll remove it properly so reviewers don't get confused.

fvaleri (approved):
LGTM. Thanks.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,744 @@
...
+
+### Non-goals
+
+* Supporting replication factor change (this could be added later)
+
+### Maintaining compatibility of the `KafkaTopic` custom resource
+
+We would remain schema-compatible with the existing `KafkaTopic` resource at the API level. 
+This means that all the existing `spec` sections could be handled by the new operator and all existing fields of the `status` section would still be present for applications that consume the status.
+However, the semantics of a `KafkaTopic` would change incompatibly: The `spec` would no longer be updated when Kafka-side configuration changed.
```
scholzj, line 74: When you say that the semantics of the `KafkaTopic` changes, I would imagine that the fields remain the same but have a different meaning. I do not think that is the case here -> the semantics of the KafkaTopic custom resource is still the same, and all the fields still have the same meaning. The backward incompatible change is from my point of view in the TO itself, not in the CR. So I think this could be a bit confusing.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,744 @@
...
+>     status: True
+>     lastTransitionTime: 20230301T103000Z
+> ```
+> 
+---
+
+
+
+
+
```
scholzj, line 240: So many empty lines are probably not needed.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,744 @@
...
+```
+
+While the behaviour described above will be the _default_ it will be possible to opt out of this using the UTO's new `STRIMZI_ADD_FINALIZER` config parameter. 
+
+* `STRIMZI_USE_FINALIZER=true`: The presence the `strimzi.io/topic-operator` finalizer will be checked (and added, if missing) on every reconciliation where the `metadata.deletionTimestamp` isn't set, including unmanaged topics. The operator will not add or remove the finalizer on `KafkaTopics` which do not match its label selector.
+* `STRIMZI_USE_FINALIZER=false`: The operator will remove the `strimzi.io/topic-operator` finalizer if present. 
+
+---
+
+#### Example 3: `KafkaTopic` deletion propagates to Kafka
```
scholzj, line 309: Example 2 seems to be missing (=> the numbering is wrong)

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,744 @@
...
+
+### Operational considerations
+
+#### Changing `metadata.name`
+
+This WILL be supported by:
+
+1. unmanaging the topic (`strimzi.io/managed: false`).
+2. deleting the `KafkaTopic`.
+3. recreating the `KafkaTopic` with a new name.
```
scholzj, line 434: I guess this assumes there is `.spec.topicName` set. In other words, you can _rename_ this way the KafkaTopic resource, but not the Kafka topic. I think that might be worth clarifying.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,744 @@
...
+This will NOT be supported, because it's not supported by Kafka.
+The user will either need to revert the `spec.partitions`, or recreate the topic in Kafka (which they can do via the `KafkaTopic` or not).
+
+#### Changing `spec.replicas`
+
+This will NOT be supported as part of this proposal.
+
+#### Recreating topics in the Kafka cluster
+
+This WILL be supported.
```
scholzj, line 462: What exactly does this mean?

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,744 @@
...
+    image: 
+    watchedNamespace:
+    reconciliationIntervalSeconds:
+```
+
+### Changes to the KafkaTopic custom resource API
+
+The CR API is _structurally_ unchanged except:
+
+* `spec.managed` (described above), optional, is added
```
scholzj, line 619: This is not valid anymore, or?

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,744 @@
...
+## Migration
+
+### Standalone case
+
+Existing users of the standalone BTO would have to:
+
+1. Review how they're using the BTO and whether they require bidirectionality
+   1. If bidirectional support is required by their usage then the UTO cannot be used. The user will have no way of managing topics using `KafkaTopic` resources once ZooKeeper support is removed by Apache Kafka and/or Strimzi.
+2. Undeploy the BTO (they can retain their existing `KafkaTopics`)
+3. Deploy the UTO (reconfiguring their `Kafka` CR, or raw `Deployment` in the standalone case).
```
scholzj, line 694: This is standalone case -> so the `Kafka` CR mention is not needed here?

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,744 @@
...
+In these cases it is possible to remove the finalizer manually (e.g. via `kubectl edit`), which will then allow the apiserver to remove any other resources.
+
+#### Compatibility with other tooling
+
+This approach is not easily compatible with other topic management tooling.
+This includes things like the `kafka-configs.sh` and other Apache Kafka scripts as well as any other tooling which uses the `Admin` client, including Kafka applications themselves.
+
+In the particular case of a hypthetical "Admin server" that was providing CLI or UI functionality in a Strimzi-aware way, it _might_ be possible to use wrap `Admin` (decorator pattern), or write it using some abstraction over `Admin` which could also support operations on `KafkaTopics` for those interactions which would otherwise result in conflicting sources of truth.
+
+Alternatively it might be possible to use a Kafka-aware proxy to redirect Kafka protocol messages which changed topic state to the Kube API, along with a Topic Operator which had a priviliged connection that was not proxied in this way.
```
scholzj, line 724: You say _it might be possible to use a Kafka-aware proxy_ ... but given no proxy with any capability to intercept Kafka topics and manage them through custom resource exists, you it be better to say _it might be possible to develop a Kafka-aware proxy_?

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,744 @@
...
+## Future work
+
+This propsal aims to set the direction and define the semantics of the UTO.
+It is not intended, by itself, to define the complete post-BTO picture.
+Future work _could_ include:
+
+* Providing tooling to conveniently create `KafkaTopics` for existing topics in Kafka.
+* Providing tooling to conveniently remove finalizers from `KafkaTopics`
+
+## Rejected alternatives
```
scholzj, line 737: Should you add `Using polling via the Kafka `Admin` client.` as well here? You mention it at the beginning as considered alternative. This is an obvious alternative and people asked about why not do that in the past. So it might be worth listing it here as well.

scholzj (approved):
LGTM. Thanks for the proposal. I left some more comments -> mostly just some details and clarifications.

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,744 @@
...
+It deviates from the standard operator pattern by being bidirectional.
+That is, it will reconcile changes to a `KafkaTopic` resource both to _and from_ a Kafka cluster.
+The bidirectionality means applications and users can continue to use Kafka-native APIs (such as the `Admin` client) and tooling (e.g. the scripts provided by Apache Kafka) as required.
+The KafkaTopic's `spec` will be updated by the operator to reflect those changes.
+
+The Strimzi community often abbreviates the topic operator as TO, but to avoid ambiguity in this document we will refer to the existing bidirectional topic operator as the BTO, and the proposed unidirectional topic operator as the UTO.
+
+The BTO makes use of ZooKeeper znode watches to know about changes to topic state within the Kafka cluster.
+
+But Apache Kafka is removing its ZooKeeper-dependence which means that the BTO won't work with Kafka clusters that use it's replacement, known as KRaft.
```
ppatierno, line 17:
```suggestion
But Apache Kafka is removing its ZooKeeper-dependence which means that the BTO won't work with Kafka clusters that use its replacement, known as KRaft.
```

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,744 @@
...
+> Note that both these `NotSupported` reasons can arise both from changes being made to the `KafkaTopic`, or changes made directly in Kafka.
+> For example, the UTO cannot tell whether the `spec.partitions` was decreased, or whether it was unchanged and the conflict arises because someone increased the number of partitions directly in Kafka so that it looks like the `KafkaTopic` is requesting a decrease. 
+> In any case, when the user takes action to fix the problem the `KafkaTopic` will (eventually) get reconciled and the `status` will revert to `Ready`.
+
+### Topic deletion
+
+By default we will use a [Kube finalizer](https://kubernetes.io/blog/2021/05/14/using-finalizers-to-control-deletion/) for deletion.
+This means that deletion via the kube REST API will first mark the resource as scheduled for deletion by setting the `metadata.delectionTimestamp`, allowing the operator to handle the deletion and then update the `KafkaTopic`, removing its `metadata.finalizer` so that the resource is actually removed from `etcd`.
+
+Without a finalizer then the following would be posible:
```
ppatierno, line 281:
```suggestion
Without a finalizer then the following would be possible:
```

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,744 @@
...
+Let's consider each of the public APIs of the BTO in turn and describe the compatiblity story.
+
+### Operator Configuration
+
+#### Standalone case
+
+
+| Env var						| Status                        |
+|-------------------------------------------------------|-------------------------------|
+| `STRIMZI_NAMESPACE`					|				|
```
ppatierno, line 547: I guess it's Unchanged as well, or?

ppatierno (approved):
Left some nits but LGTM, thanks!

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,744 @@
...
+
+* `status.conditions` may have new values for the `reason`.
+
+The semantics of the `KafkaTopic` custom resource API will also change:
+
+1. Unidirectional reconciliation
+2. The absence of fields will not imply that they take their default values, only that the value is not specified in this CR.
+
+> The reason for 2 is to leave the door open to a further refinement of the UTO in the future.
+> Consider the case where a Kafka cluster is centrally managed by an infrastruture team.
```
im-konge, line 631:
```suggestion
> Consider the case where a Kafka cluster is centrally managed by an infrastructure team. 
```

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,744 @@
...
+Using finalizers on `KafkaTopics` means that the operator has to be running (to remove the finalizer) before the resources will actually be deleted.
+That can prevent other resources from being removed (for example the `Namespace` containing the `KafkaTopics`).
+In these cases it is possible to remove the finalizer manually (e.g. via `kubectl edit`), which will then allow the apiserver to remove any other resources.
+
+#### Compatibility with other tooling
+
+This approach is not easily compatible with other topic management tooling.
+This includes things like the `kafka-configs.sh` and other Apache Kafka scripts as well as any other tooling which uses the `Admin` client, including Kafka applications themselves.
+
+In the particular case of a hypthetical "Admin server" that was providing CLI or UI functionality in a Strimzi-aware way, it _might_ be possible to use wrap `Admin` (decorator pattern), or write it using some abstraction over `Admin` which could also support operations on `KafkaTopics` for those interactions which would otherwise result in conflicting sources of truth.
```
im-konge, line 722:
```suggestion
In the particular case of a hypothetical "Admin server" that was providing CLI or UI functionality in a Strimzi-aware way, it _might_ be possible to use wrap `Admin` (decorator pattern), or write it using some abstraction over `Admin` which could also support operations on `KafkaTopics` for those interactions which would otherwise result in conflicting sources of truth.
```

050-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,744 @@
...
+
+Alternatively it might be possible to use a Kafka-aware proxy to redirect Kafka protocol messages which changed topic state to the Kube API, along with a Topic Operator which had a priviliged connection that was not proxied in this way.
+This would result in Kube being the effective source of truth for topic state.
+
+
+## Future work
+
+This propsal aims to set the direction and define the semantics of the UTO.
```
im-konge, line 730:
```suggestion
This proposal aims to set the direction and define the semantics of the UTO.
```

im-konge (approved):
LGTM, this looks really great! :)

050-unidirectional-topic-operator.md
tombentley: Good question. I don't recall now what I meant by it. I'll remove it..

tombentley:
Thanks for the comments and reviews. I have resolved all remaining open comments. 

This proposal has 5 binding votes from maintainers, plus my own as proposer, plus a further non-binding vote.

If anyone has any more comments or votes, please do so by Friday. If there is nothing new I will close the vote and merge this PR on Monday.

PR #77, opened 2023-03-31
Kafka Node Pools

scholzj:
This is the initial version of the Kafka Node Pools proposal. It is based on my prototyping work (which does not implement everything, but it has the core parts). 

There are still some open points which are up for review and discussion and hopefully we can address them in the comments phase. They are described in the _Open Points_ section and are mainly about the migration and backwards compatibility.

### DIFF
--- a/050-Kafka-Node-Pools.md
+++ b/050-Kafka-Node-Pools.md
@@ -1,0 +1,806 @@
+# Kafka Node Pools
+
+## Current situation
+
+Strimzi currently allows to have one set of ZooKeeper nodes and one set of Kafka nodes.
+All nodes within the given set have very similar configuration and differ only in small details:
+
+* Each Pod might have a PVC using different storage class
+* The Kafka pods have (when using `StrimziPodSets`) each their own Config Map with their Kafka configuration.
+  The configurations differ between the different nodes in the advertised addresses.
+
+## Problem description
+
+Strimzi does not support nodes with:
+
+* Different resources configuration
+* Different storage configuration (the nodes might have different storage class, but not storage size or a different number of disks)
+* Different scheduling rules (affinity, tolerations, ...)
+* Different Kafka configuration (regardless whether the option is configurable per-node or per-cluster in Kafka, we always manage it per-cluster)
+* Running in different Kubernetes clusters
+
+While from time to time these limitations were raised by the users, they were for a long time acceptable and not in general demand.
+However, lately there seems to be increased demand for various features which require different configurations for different nodes.
+This proposal does not aim to solve all of these but rather to provide a basis for solving them.
+
+### ZooKeeper-less Kafka
+
+Kafka is slowly finishing the implementation of the KRaft mode.
+The KRaft mode removes the ZooKeeper dependency and replaces it with Kafka's own controller nodes based on the Raft protocol (_KRaft = Kafka + Raft_).
+In the KRaft mode, the Kafka cluster will consist only from Kafka nodes.
+But they will have different roles - they will be either controllers, or brokers, or will combine both of them.
+The controller role means the node is responsible for maintaining the cluster quorum, for the elections, discovery etc.
+The broker role means the node is responsible for handling messages as a regular broker in the old ZooKeeper based clusters.
+And finally, the combined role means the node is responsible for both at the same time.
+
+The KRaft based Kafka clusters will be typically used in 3 different architectures:
+
+1. One or more nodes combining the controller and broker roles
+2. One or more nodes with the controller roles and one or more nodes with the broker roles
+3. One or more nodes with the combined role and one or more nodes with the broker roles
+
+In Strimzi, we want to support all of these architectures.
+But we also want to make it possible to transition between them.
+Since Apache Kafka plans to remove ZooKeeper support in its 4.0 release, supporting KRaft mode in its different form is not really an option for us but something we have to do.
+
+### Stretch clusters
+
+There seems to be increased demand Kafka clusters stretched across multiple Kubernetes clusters.
+Apache Kafka itself has latency limitations and might not run well when stretched over distant clusters with big latency.
+But there are many use cases where stretched clusters might be useful.
+For example:
+
+* When running closely colocated Kubernetes clusters
+* When migrating from old Kubernetes cluster to new Kubernetes cluster
+
+## Proposal
+
+This proposal suggests introduction of Kafka _node pools_.
+_Node pools_ are groups of Kafka nodes which will have the same configuration.
+A Kafka cluster can consist of one or more pools.
+Each node pool can belong to only one Kafka cluster.
+
+### `KafkaNodePool` custom resource
+
+A new custom resource called `KafkaNodePool` will be added to Strimzi.
+One `KafkaNodePool` resource will represent one _node pool_.
+It will take over some of the configuration from the `Kafka` CR, reuse other parts from it and add some new configuration options as well.
+The `KafkaNodePool` resource will be a _public_ resource.
+It will be created and managed directly by users (unlike for example the `StrimziPodSet` resource which is used only internally by Strimzi and should not be touched by the users).
+
+The _cluster-wide_ configuration options will remain part of the `Kafka` CR.
+While the _pool-specific_ options will be taken from `KafkaNodePool` resources
+
+For the initial implementation, the following options are proposed for the `KafkaNodePool` resource and its `.spec` section:
+
+* `replicas` indicating the number of nodes in the pool
+* `storage` indicating the storage of the nodes in this pool
+* `resources` indicating the amount of resources (requests / limits) assigned to the nodes in this pool
+* `template` for customizing the pods and other resources which are part of the nodes in this pool
+* `jvmOptions` for customizing the JVM configuration
+
+For example:
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaNodePool
+metadata:
+  name: my-pool
+  # ...
+spec:
+  replicas: 3
+  resources:
+    requests:
+      memory: 8Gi
+      cpu: 2
+    limits:
+      memory: 8Gi
+      cpu: 4
+  jvmOptions:
+    -Xms: 2048m
+    -Xmx: 2048m
+  storage:
+    type: jbod
+    volumes:
+      - id: 0
+        type: persistent-claim
+        size: 100Gi
+        deleteClaim: true
+  template:
+    podSet:
+      metadata:
+        labels:
+          key1: label1
+    pod:
+      terminationGracePeriodSeconds: 60
+```
+
+The replicas and storage sections will be required in `KafkaNodePool`.
+The other fields will be optional.
+At the same time, the replicas and storage sections will be made optional in the `Kafka` resource since when using node pools, they will not be used in the `Kafka` custom resource anymore.
+
+When the optional field is not present in `KafkaNodePool` and is present in `Kafka.spec.kafka`, it will default to the configured value from the `Kafka` CR.
+However, this will be applied only on the highest level.
+For example:
+
+* If the `KafkaNodePool.spec.jvmOptions` does not exist but `Kafka.spec.kafka.jvmOptions` does exist, the values from `Kafka.spec.kafka.jvmOptions` will be used.
+* When `KafkaNodePool.spec.jvmOptions` exists and has `-Xmx: 1024m` and `Kafka.spec.kafka.jvmOptions` exists as well with `-Xms: 512m`, the operator will completely ignore the value from `Kafka.spec.kafka.jvmOptions` and use only the value from `KafkaNodePool.spec.jvmOptions`.
+* When `KafkaNodePool.spec.template` contains only the field `podSet.metadata.labels`, the operator will completely ignore the corresponding values from `Kafka.spec.kafka.template` (for example `podSet.metadata.annotations`, but also for example `pod.metadata.labels`).
+
+In the future, additional fields might be moved to the `KafkaNodePool` and its `.spec` section according to the needs.
+This proposal tries to minimize the effort and the scope and therefore picks up only some of the fields for the initial implementation.
+Since the existing fields in the `Kafka` CR will be used as the _defaults_, new fields can be added to `KafkaNodePool` resource in the future without any backwards compatibility concerns.
+
+### Resource naming
+
+Each node pool will be represented by its own `StrimziPodSet` and pods running the actual nodes.
+The `StrimziPodSet` will be created automatically by the operator based on the `KafkaNodePool` definition - it will not be created by the users.
+The `StrimziPodSet` will be named `<ClusterName>-<PoolName>` - e.g. `my-cluster-big-nodes`.
+The resulting pods will be named `<ClusterName>-<PoolName>-<ID>` - e.g. `my-cluster-big-nodes-5`.
+The naming of the related resources such as Services, Config Maps, or PVCs will follow the pod names.
+
+That means that when a node moves from one pool to another, all the resources will be recreated including the per-node services or storage.
+It is not possible to re-use these things not just because of the naming concerns, but also because the different pools might have different configurations and the old resources might not be re-usable.
+So a node which moves will start from _zero_ with empty disk and will need to re-sync all the data.
+
+### Matching the node pools to the Kafka cluster
+
+A particular `KafkaNodePool` resource can belong only to a single Kafka cluster.
+The node pools will use the same mechanism to define to which cluster they belong as we use today for `KafkaConnector`, `KafkaTopic` or `KafkaUSer` resources.
+`strimzi.io/cluster` label will be set on the `KafkaNodePool` and its value will be the name of the `Kafka` custom resource to which it belongs.
+For example, the label `strimzi.io/cluster: my-cluster` will mean that this node pool belongs to a Kafka cluster named `my-cluster`.
+The `Kafka` and `KafkaNodePool` resources always have to be in the same namespace.
+
+#### Example
+
+The following example shows how the `Kafka` and `KafkaNodePool` resources might combine:
+
+```yaml=
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+  namespace: my-namespace
+spec:
+  kafka:
+    version: 3.4.0
+    listeners:
+      - name: plain
+        port: 9092
+        type: internal
+        tls: false
+      - name: tls
+        port: 9093
+        type: internal
+        tls: true
+    config:
+      offsets.topic.replication.factor: 3
+      transaction.state.log.replication.factor: 3
+      transaction.state.log.min.isr: 2
+      default.replication.factor: 3
+      min.insync.replicas: 2
+      inter.broker.protocol.version: "3.3"
+  # ...
+---
+
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaNodePool
+metadata:
+  name: big-nodes
+  namespace: my-namespace
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  replicas: 3
+  resources:
+    requests:
+      memory: 64Gi
+      cpu: 8
+    limits:
+      memory: 64Gi
+      cpu: 12
+  storage:
+    type: jbod
+    volumes:
+    - id: 0
+      type: persistent-claim
+      size: 1000Gi
+      class: fast-storage
+    - id: 1
+      type: persistent-claim
+      size: 1000Gi
+      class: fast-storage
+---
+
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaNodePool
+metadata:
+  name: small-nodes
+  namespace: my-namespace
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  replicas: 3
+  resources:
+    requests:
+      memory: 4Gi
+      cpu: 1
+    limits:
+      memory: 4Gi
+      cpu: 2
+  storage:
+    type: jbod
+    volumes:
+    - id: 0
+      type: persistent-claim
+      size: 100Gi
+      class: regular-storage
+```
+
+#### Protecting `KafkaNodePool` resources from being shared between multiple Kafka clusters
+
+The `strimzi.io/cluster` label should guarantee that the node pool will match only one Kafka cluster.
+Additional protection will be made against undesired moves between different Kafka clusters when the `strimzi.io/cluster` label is modified.)
+The `.status` section of the `KafkaNodePool` resource will contain the Kafka cluster ID of the cluster it belongs to.
+For example:
+
+```yaml=
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaNodePool
+metadata:
+  name: my-pool
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  # ...
+status:
+  # ...
+  clusterId: cABpUKeDTqi7q_XOtBMgUv
+```
+
+The cluster ID is already part of the `Kafka` CR `.status` section.
+The operator will check the cluster IDs in every reconciliation and if the cluster IDs don't match between the `Kafka` and `KafkaNodePool` CRs, it will expect that there is some kind of an misconfiguration.
+It will throw an `InvalidresourceException` and wait for the user to fix the issue.
+
+### Scaling the node pools
+
+The `KafkaNodePool` resource will also support the `scale` sub-resource.
+Support for the `scale` sub-resource should allow integration with Kubernetes Horizontal Pod Autoscaler.
+That on its own doesn't really provide fully functional Kafka auto-scaling mechanism.
+For example, if the load is coming from a single partition which is handling too many messages, adding more nodes will not help.
+So a proper auto-scaling mechanism would need to understand the configuration or topics and partitions and their load to properly scale the broker.
+But the `scale` sub-resource would make it easier to develop such solutions in the future.
+
+### KRaft support
+
+To support the Kafka KRaft mode, a new field `role` will be added to the `KafkaNodePool` as well.
+It will contain a list of roles (`controller` or `broker`) which the nodes in given pool should have.
+This will allow to use different configurations.
+For example, for a small cluster with shared broker and controller responsibilities:
+
+```yaml=
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+  namespace: my-namespace
+spec:
+  kafka:
+    version: 3.4.0
+    listeners:
+      - name: plain
+        port: 9092
+        type: internal
+        tls: false
+      - name: tls
+        port: 9093
+        type: internal
+        tls: true
+    config:
+      offsets.topic.replication.factor: 3
+      transaction.state.log.replication.factor: 3
+      transaction.state.log.min.isr: 2
+      default.replication.factor: 3
+      min.insync.replicas: 2
+      inter.broker.protocol.version: "3.3"
+  # ...
+---
+
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaNodePool
+metadata:
+  name: mixed
+  namespace: my-namespace
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  role:
+    - controller
+    - broker
+  replicas: 3
+  resources:
+    requests:
+      memory: 8Gi
+      cpu: 2
+    limits:
+      memory: 8Gi
+      cpu: 4
+  storage:
+    type: jbod
+    volumes:
+    - id: 0
+      type: persistent-claim
+      size: 1000Gi
+    - id: 1
+      type: persistent-claim
+      size: 1000Gi
+```
+
+Or for a cluster with separate controller and broker pools:
+
+```yaml=
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+  namespace: my-namespace
+spec:
+  kafka:
+    version: 3.4.0
+    listeners:
+      - name: plain
+        port: 9092
+        type: internal
+        tls: false
+      - name: tls
+        port: 9093
+        type: internal
+        tls: true
+    config:
+      offsets.topic.replication.factor: 3
+      transaction.state.log.replication.factor: 3
+      transaction.state.log.min.isr: 2
+      default.replication.factor: 3
+      min.insync.replicas: 2
+      inter.broker.protocol.version: "3.3"
+  # ...
+---
+
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaNodePool
+metadata:
+  name: controllers
+  namespace: my-namespace
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  role:
+    - controller
+  replicas: 3
+  resources:
+    requests:
+      memory: 8Gi
+      cpu: 2
+    limits:
+      memory: 8Gi
+      cpu: 4
+  storage:
+    type: jbod
+    volumes:
+    - id: 0
+      type: persistent-claim
+      size: 100Gi
+---
+
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaNodePool
+metadata:
+  name: brokers
+  namespace: my-namespace
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  role:
+    - broker
+  replicas: 6
+  resources:
+    requests:
+      memory: 64Gi
+      cpu: 8
+    limits:
+      memory: 64Gi
+      cpu: 12
+  storage:
+    type: jbod
+    volumes:
+    - id: 0
+      type: persistent-claim
+      size: 1000Gi
+    - id: 1
+      type: persistent-claim
+      size: 1000Gi
+```
+
+### Handling of node IDs
+
+Node IDs (broker IDs) uniquely identify each of the Kafka cluster nodes.
+Node IDs will be assigned sequentially:
+
+* They will start with 0.
+* By default, when a new node ID is needed, the lowest possible available number will be used regardless the pool where the node will be added:
+    * In most cases, the node ID will be assigned from 0 up: 0, 1, 2, 3.
+    * If any gap will exist in the existing node IDs, it will be filled in.
+      For example when the existing nodes have IDs 0, 1, 4, 5, 6; the next node will get ID 2.
+* By default, when scaling down a pool, a highest used node ID will be remove.
+    * For example in Kafka cluster with two pool, where the first pool has IDs 0, 1 and 2 and the second pool has IDs 3, 4 and 5 ... when you scale down the first pool, the node with ID 2 will be removed and when you scale down the second pool node 5 will be removed.
+* The pods will reflect the node IDs in their name in the same way as they do today.
+
+The fact that the node IDs will not be assigned to the pools means the Pods will be mixed.
+In a cluster with two pools, they might look like this:
+
+* `my-cluster-big-nodes-0`
+* `my-cluster-big-nodes-1`
+* `my-cluster-small-nodes-2`
+* `my-cluster-small-nodes-3`
+* `my-cluster-big-nodes-4`
+* `my-cluster-small-nodes-5`
+
+This is anyway unavoidable if we want to allow nodes to move between the pools.
+
+To allow the operator to keep track of the assigned node IDs, it will keep track of them in the `.status` section of the `KafkaNodePool` resource.
+That way, it will be able to know at any point in time which IDs are used and which are available.
+It will be also able to identify the scale-down or scale-up events using this information.
+
+```yaml=
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaNodePool
+metadata:
+  name: big-nodes
+  # ...
+spec:
+  replicas: 3
+  # ...
+status:
+  nodeIds:
+    - 0
+    - 1
+    - 2
+  clusterId: cABpUKeDTqi7q_XOtBMgUv
+  # replicas and labelSelector fields are needed for the scale subresource
+  replicas: 3
+  labelSelector: >-
+    strimzi.io/cluster=my-cluster,strimzi.io/name=m-cluster-big-nodes,strimzi.io/kind=Kafka
+```
+
+User will be able to use annotations to indicate to the operator what should be the next ID when scaling up or what should be the node ID which should be removed when scaling down through annotations.
+Setting an annotation `strimzi.io/next-node-ids` will tell the operator what should be the next ID used in scale-up.
+The value might be a single number, list of numbers or one or more ranges.
+E.g. `[3]`, `[3, 4, 5]` or `[1000-1010]`.
+Similarly, setting an annotation `strimzi.io/remove-node-ids` will allow to configure IDs of node which should be removed.
+This will support only an ordered list of nodes IDs without ranges.
+When the operator is adding or removing a node from the pool, it will look at these annotations and if they are set, it will use them to override the default mechanism for picking up the next node ID or the node to remove.
+These annotations will be ignored when scaling is not requested by changing `.spec.replicas`.
+
+The annotations will be used only when scaling up or down or moving nodes.
+The operator will validate them only in such situations and not during every reconciliation.
+It is expected that when using auto-scaling, the annotations will not be updated after every scale-up or scale-down event.
+So when the use specifies with annotation for example that new nodes should come from a range `[1000-1010]`, such annotation is considered valid even when some of the IDs from the range are already in use.
+This way, the node pool can auto-scale up and down within the range without changing the annotation.
+
+#### Moving nodes between node pools
+
+In some situations, it might be good to be able to move a node between different node pools.
+The annotations can be used for that as well.
+The process to move node with ID 3 from pool named `small-nodes` to pool names `big-nodes` will be following:
+1) Annotate `KafkaNodePool` named `small-nodes` with annotation `strimzi.io/remove-node-ids: [3]`
+2) Annotate `KafkaNodePool` named `big-nodes` with annotation `strimzi.io/next-node-ids: [3]`
+3) Scale `KafkaNodePool` named `small-nodes` down by changing its `.spec.replicas` field
+4) Scale `KafkaNodePool` named `big-nodes` up by changing its `.spec.replicas` field
+
+The operator will proceed and remove the node 3 from the `small-nodes` pool and create a new node with ID 3 in the `big-nodes` pool.
+The moved now in its new pod will start from zero and will have to replicate all the data from other nodes in the cluster.
+
+This process has several risks:
+* The moved node starts from 0 with empty disk and will need to re-sync all data from other nodes.
+  Kafka nodes might store in some environments TBs of data, so re-syncing the data might take a long time.
+* While the node is re-syncing, the replicas hosted by it are not in-sync.
+  So with typical configurations (such as replication factor set to 3 and minimal in-sync replicas 2) any problem with one of the other node would mean loss of availability and cause that producers will not be able to produce messages anymore.
+* Since the replica will not be in-sync, it would also mean that it might not be possible to proceed with rolling updates and other operations.
+
+Due to these risks and limitations, users should instead of moving a node consider moving the partition replicas.
+They should:
+1) Scale up the target node pool to have a new node created
+2) Reassign the partition replicas from the old node to the new node
+3) Once the reassignment is complete, scale down the node pool with the old node
+
+While this will cause a change of the node ID, it will be a more graceful process.
+The reassignment process will first create a new replica and start copying the data.
+Only when the new replica on the new node is in-sync, the old replica will be deleted.
+Thanks to that, the availability is still guaranteed and other changes such as rolling updates are not affected.
+This process should be also recommended to the users in the documentation.
+
+In the future, we should try to improve the Cruise Control support to make this process easier (maybe with a new rebalance mode `move-broker` or something similar).
+This is however not part of this proposal.
+
+Despite the risks and limitations, having the ability to be able to move the nodes between node pools is still desired and might be useful.
+So it should still be supported, even if it might not be part of the initial implementation.
+
+### Deletion of the `KafkaNodePool` resource
+
+When the `KafkaNodePool` resource belonging to an existing Kafka cluster is deleted, all the resources belonging to it and managed by it will be deleted.
+This will also include the Pods and other resources.
+The PVCs will be deleted only if the `deleteClaim` flag was set to `true` in the storage configuration.
+While it is possible that the user deleted the `KafkaNodePool` by mistake, once the resource is deleted, there is not much the operator can do about it.
+This behavior does not differ from what happens if the `Kafka` custom resource is deleted, so it does not introduce any new behavior.
+Users can use other mechanisms such as RBAC or finalizers to prevent deleting the custom resources by mistake.
+
+### Implementation
+
+#### Feature Gate
+
+The node pools will be introduced gradually using a Feature Gate.
+The new Feature gate will be named `KafkaNodePools`.
+The following table shows the expected graduation of the `KafkaNodePools` feature gate:
+
+| Phase | Strimzi versions       | Default state                                          |
+|:------|:-----------------------|:-------------------------------------------------------|
+| Alpha | 0.35 - 0.37            | Disabled by default                                    |
+| Beta  | 0.38 - 0.40            | Enabled by default                                     |
+| GA    | 0.41 and newer         | Enabled by default (without possibility to disable it) |
+
+The main purpose of the feature gate is in this case to protect users from a feature which is not mature enough and might be removed in future versions.
+Only small parts of the code will actually depend on the feature gate being enabled or disabled.
+
+#### Virtual node pool
+
+To avoid two parallel implementations in the Strimzi source code, the operator will use _virtual node pool_ when the feature gate is disabled.
+It will internally (the _virtual_ node pool will not exist as a Kubernetes resource) create the a node pool based on the `Kafka` custom resource and create the resources through this node pool with the same names and configurations as before.
+This significantly simplifies the implementation and testing since the same code will be used all the time.
+
+### Backwards compatibility
+
+The node pools are not backwards compatible, but provide a simple migration path.
+When the feature gate is enabled, users have to use the `KafkaNodePool` resources.
+But they can easily convert the existing `Kafka` custom resource into a matching `Kafka` and `KafkaNodePool` resources.
+For example the following `Kafka` CR which can be used with the feature gate disabled:
+
+```yaml=
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    version: 3.3.1
+    replicas: 3
+    listeners:
+      - name: plain
+        port: 9092
+        type: internal
+        tls: false
+      - name: tls
+        port: 9093
+        type: internal
+        tls: true
+    config:
+      offsets.topic.replication.factor: 3
+      transaction.state.log.replication.factor: 3
+      transaction.state.log.min.isr: 2
+      default.replication.factor: 3
+      min.insync.replicas: 2
+      inter.broker.protocol.version: "3.3"
+    resources:
+      requests:
+        memory: 64Gi
+        cpu: 8
+      limits:
+        memory: 64Gi
+        cpu: 12
+    storage:
+      type: jbod
+      volumes:
+        - id: 0
+          type: persistent-claim
+          size: 1000Gi
+          class: fast-storage
+  # ...
+```
+
+Is equivalent to this combination of `Kafka` and `KafkaNodePool` resources:
+
+```yaml=
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    nodePoolSelector:
+      matchLabels:
+        cluster: my-cluster
+    version: 3.3.1
+    listeners:
+      - name: plain
+        port: 9092
+        type: internal
+        tls: false
+      - name: tls
+        port: 9093
+        type: internal
+        tls: true
+    config:
+      offsets.topic.replication.factor: 3
+      transaction.state.log.replication.factor: 3
+      transaction.state.log.min.isr: 2
+      default.replication.factor: 3
+      min.insync.replicas: 2
+      inter.broker.protocol.version: "3.3"
+  # ...
+---
+
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaNodePool
+metadata:
+  name: kafka
+  labels:
+    cluster: my-cluster
+spec:
+  replicas: 3
+  resources:
+    requests:
+      memory: 64Gi
+      cpu: 8
+    limits:
+      memory: 64Gi
+      cpu: 12
+  storage:
+    type: jbod
+    volumes:
+    - id: 0
+      type: persistent-claim
+      size: 1000Gi
+      class: fast-storage
+```
+
+Because the `KafkaNodePool` is named `kafka`, all the resources created by this configuration will be exactly the same and have the same names as today.
+
+## Open Points
+
+### Migration
+
+The backwards compatibility / migration story seems weak.
+The code already supports handling the original `Kafka` CR through the _virtual node pool_.
+Maybe we can use this and just find a way how to indicate if the user wants to use the `KafkaNodePool` resources or use the `Kafka` with the virtual node pool.
+That would allow to enable the feature gate without converting the custom resource and make it backwards compatible.
+Originally, the node pool selector in the `Kafka` CR was supposed to play this role.
+But that was abandoned for other reasons.
+We should try to find a new way how to indicate this.
+E.g. some `useNodePools: true` flag?
+If we use such flag, do we still need the feature gate?
+Maybe es, but we can graduate it more quickly.
+
+### Using it per-cluster
+
+The feature gate is enabled / disabled per-operator.
+That means it is disabled or enabled for all Kafka clusters managed by given operator.
+That is not nice if it means converting all of the custom resources to enable it.
+Solving the migration issue described in previous section might solve this one as well.
+
+## Risks
+
+### Eventual consistency
+
+Using multiple resources in Kubernetes is tricky as we have to rely on eventual consistency.
+Users can store multiple YAML documents in a single file and apply them with a single command.
+But their creation is not a single operation.
+It can therefore happen that the operator first sees only the `KafkaNodePool` resources but not the `Kafka` resource or the other way around.
+The basic situations can be handled by the operator code.
+But it is not completely clear how many problems and various race conditions this might be causing when updating the node pools for existing clusters etc.
+
+## Not impacted
+
+This proposal does not impact in any way how the ZooKeeper nodes, Kafka Exporter, Cruise Control or Entity Operator are configured.
+It also does not impact configuration of the `KafkaConnect`, `KafkaBridge` or other custom resources.
+
+## Rejected alternatives
+
+### KafkaNodePool selector in `.spec.kafka` section of the `Kafka` custom resource
+
+One of the options considered was to have a label selector in the `Kafka` custom resource in `.spec.kafka`.
+Using the `strimzi.io/cluster` label was chosen at the end as the preferred way.
+It makes it easier to match the `KafkaNodePool` resources to the `Kafka` cluster.
+This significantly simplifies the code - especially the watch which is triggered when 
+It also helps to ensure that one node pool will always belong only to a single Kafka cluster since one node pool resource can have only one `strimzi.io/cluster` label.
+
+### Configuring pools inside the `Kafka` custom resource
+
+One of the considered options was configuring the node pools inside the `Kafka` custom resource.
+For example:
+
+```yaml=
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    nodePoolSelector:
+      matchLabels:
+        cluster: my-cluster
+    version: 3.3.1
+    listeners:
+      - name: plain
+        port: 9092
+        type: internal
+        tls: false
+      - name: tls
+        port: 9093
+        type: internal
+        tls: true
+    config:
+      offsets.topic.replication.factor: 3
+      transaction.state.log.replication.factor: 3
+      transaction.state.log.min.isr: 2
+      default.replication.factor: 3
+      min.insync.replicas: 2
+      inter.broker.protocol.version: "3.3"
+  # ...
+  nodePools: 
+    - name: pool1
+      replicas: 3
+      resources:
+        requests:
+          memory: 64Gi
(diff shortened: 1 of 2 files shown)


050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,806 @@
...
+
+A particular `KafkaNodePool` resource can belong only to a single Kafka cluster.
+The node pools will use the same mechanism to define to which cluster they belong as we use today for `KafkaConnector`, `KafkaTopic` or `KafkaUSer` resources.
+`strimzi.io/cluster` label will be set on the `KafkaNodePool` and its value will be the name of the `Kafka` custom resource to which it belongs.
+For example, the label `strimzi.io/cluster: my-cluster` will mean that this node pool belongs to a Kafka cluster named `my-cluster`.
+The `Kafka` and `KafkaNodePool` resources always have to be in the same namespace.
+
+#### Example
+
+The following example shows how the `Kafka` and `KafkaNodePool` resources might combine:
```
ppatierno, line 156: I would clarify "... to set up a Kafka cluster with 6 brokers in total", just to make clear to users that now they have to familiarize with not just one `replicas` field if using more node pools but taking into account the sum of them.

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,806 @@
...
+
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaNodePool
+metadata:
+  name: mixed
+  namespace: my-namespace
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  role:
```
ppatierno, line 317: maybe using the plural `rules` as it is in the Kafka configuration?

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,806 @@
...
+  # replicas and labelSelector fields are needed for the scale subresource
+  replicas: 3
+  labelSelector: >-
+    strimzi.io/cluster=my-cluster,strimzi.io/name=m-cluster-big-nodes,strimzi.io/kind=Kafka
+```
+
+User will be able to use annotations to indicate to the operator what should be the next ID when scaling up or what should be the node ID which should be removed when scaling down through annotations.
+Setting an annotation `strimzi.io/next-node-ids` will tell the operator what should be the next ID used in scale-up.
+The value might be a single number, list of numbers or one or more ranges.
+E.g. `[3]`, `[3, 4, 5]` or `[1000-1010]`.
```
ppatierno, line 478: Assuming the user have a node pool with 3 brokers (so ids 0,1,2) and `strimzi.io/next-node-ids: [3]`.
He scales up from 3 to 4 through the `replicas` field so operator will create broker id 3.
Then with same `strimzi.io/next-node-ids` annotation, the user will scale up again ... now ... because 3 is used, will the operator raise a warning about it but using id = 4 or failing the reconcile asking the user to fix the annotation? I think that not having consistency between current max id and the annotation value can drive to confusion.

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,806 @@
...
+2) Reassign the partition replicas from the old node to the new node
+3) Once the reassignment is complete, scale down the node pool with the old node
+
+While this will cause a change of the node ID, it will be a more graceful process.
+The reassignment process will first create a new replica and start copying the data.
+Only when the new replica on the new node is in-sync, the old replica will be deleted.
+Thanks to that, the availability is still guaranteed and other changes such as rolling updates are not affected.
+This process should be also recommended to the users in the documentation.
+
+In the future, we should try to improve the Cruise Control support to make this process easier (maybe with a new rebalance mode `move-broker` or something similar).
```
ppatierno, line 522: While reading above I was wondering the same, so we should make clear that all the process is possible manually using the reassignment tool.

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,806 @@
...
+
+Despite the risks and limitations, having the ability to be able to move the nodes between node pools is still desired and might be useful.
+So it should still be supported, even if it might not be part of the initial implementation.
+
+### Deletion of the `KafkaNodePool` resource
+
+When the `KafkaNodePool` resource belonging to an existing Kafka cluster is deleted, all the resources belonging to it and managed by it will be deleted.
+This will also include the Pods and other resources.
+The PVCs will be deleted only if the `deleteClaim` flag was set to `true` in the storage configuration.
+While it is possible that the user deleted the `KafkaNodePool` by mistake, once the resource is deleted, there is not much the operator can do about it.
```
ppatierno, line 533: if the users delete by mistake all node pool custom resources, what happen to the `Kafka` CR?
Or in general, when in the end we'll have migrated to node pools only, the user creates `Kafka` CR but forget to create corresponding `KafkaNodePool`(s) ... does the operator prints a warning, update status, nothing?

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,806 @@
...
+
+Because the `KafkaNodePool` is named `kafka`, all the resources created by this configuration will be exactly the same and have the same names as today.
+
+## Open Points
+
+### Migration
+
+The backwards compatibility / migration story seems weak.
+The code already supports handling the original `Kafka` CR through the _virtual node pool_.
+Maybe we can use this and just find a way how to indicate if the user wants to use the `KafkaNodePool` resources or use the `Kafka` with the virtual node pool.
```
ppatierno, line 673: Without using the feature gate, can we think about something like this? 

A new Strimzi release starts to use the virtual node pool by default so that the code is handling node pools but still driven through only the `Kafka` CR.
If the user wants to explicitly move to use node pools, he creates one or more `KafkaNodePool` custom resources and on next reconcile, the operator will detect their existence and starts using them instead of the virtual one.

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,806 @@
...
+          cpu: 12
+      storage:
+        type: jbod
+        volumes:
+          - id: 0
+            type: persistent-claim
+            size: 1000Gi
+```
+
+This would have some disadvantages:
```
ppatierno, line 781: these are advantages I guess :-)

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,806 @@
...
+However, lately there seems to be increased demand for various features which require different configurations for different nodes.
+This proposal does not aim to solve all of these but rather to provide a basis for solving them.
+
+### ZooKeeper-less Kafka
+
+Kafka is slowly finishing the implementation of the KRaft mode.
+The KRaft mode removes the ZooKeeper dependency and replaces it with Kafka's own controller nodes based on the Raft protocol (_KRaft = Kafka + Raft_).
+In the KRaft mode, the Kafka cluster will consist only from Kafka nodes.
+But they will have different roles - they will be either controllers, or brokers, or will combine both of them.
+The controller role means the node is responsible for maintaining the cluster quorum, for the elections, discovery etc.
```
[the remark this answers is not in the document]
scholzj: Sorry, I'm not sure what you mean with this.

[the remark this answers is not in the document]
ppatierno: I meant that controllers are in charge of handling cluster metadata instead of ZooKeeper.

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,806 @@
...
+    pod:
+      terminationGracePeriodSeconds: 60
+```
+
+The replicas and storage sections will be required in `KafkaNodePool`.
+The other fields will be optional.
+At the same time, the replicas and storage sections will be made optional in the `Kafka` resource since when using node pools, they will not be used in the `Kafka` custom resource anymore.
+
+When the optional field is not present in `KafkaNodePool` and is present in `Kafka.spec.kafka`, it will default to the configured value from the `Kafka` CR.
+However, this will be applied only on the highest level.
```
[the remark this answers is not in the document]
scholzj: The examples below that line should explain it. It essentially means that the check will be done only at the highest level. For example ... when `KafkaNodePool.spec.jvmOptions` exists and has `-Xmx: 1024m` and `Kafka.spec.kafka.jvmOptions` exists as well with `-Xms: 512m`, the `KafkaNodePool.spec.jvmOptions` will be used and `Kafka.spec.kafka.jvmOptions` will be ignored. They will not be merged together.

050-Kafka-Node-Pools.md
scholzj: Good point. We seem to use the plural form for list fields in our APIs as well. Thanks.

050-Kafka-Node-Pools.md
scholzj: This is not part of my prototype - at least not yet. But the annotations are expected to be treated as an indication of what should be used next. So if we cannot fulfill them, we would pick up the next available ID and use it. This goes better with the ideas of autoscaling etc. It also IMHO better fits the concept of a list where some might be taken and some not etc.

050-Kafka-Node-Pools.md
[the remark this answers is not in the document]
ppatierno: I missed one thing in the example ... `-Xmx` vs `-Xms` ... it was not the same parameter to be overridden. 
I see what you meant now but "... highest level" is not clear. Maybe we should make clear that any `KafkaNodePool.spec.<field>` takes precedence over the corresponding `Kafka.spec.kafka.<field>`, or something like that?

050-Kafka-Node-Pools.md
ppatierno: Ok makes sense. Thanks.

050-Kafka-Node-Pools.md
ppatierno: > So if we cannot fulfill them, we would pick up the next available ID and use it

That's fine with me. At least just printing a WARNING like "ID 3 is already in use, picking the next available one 4"

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,821 @@
...
+    pod:
+      terminationGracePeriodSeconds: 60
+```
+
+The replicas and storage sections will be required in `KafkaNodePool`.
+The other fields will be optional.
+At the same time, the replicas and storage sections will be made optional in the `Kafka` resource since when using node pools, and they will ignored if set.
+
+When the optional field is not present in `KafkaNodePool` and is present in `Kafka.spec.kafka`, it will default to the configured value from the `Kafka` CR.
+So when `KafkaNodePool.spec.<field>` exists, `KafkaNodePool.spec.<field>` will be ignored.
```
ppatierno, line 123: I guess the second one you meant `Kafka.spec.kafka`.<field>` ?

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,821 @@
...
+      metadata:
+        labels:
+          key1: label1
+    pod:
+      terminationGracePeriodSeconds: 60
+```
+
+The replicas and storage sections will be required in `KafkaNodePool`.
+The other fields will be optional.
+At the same time, the replicas and storage sections will be made optional in the `Kafka` resource since when using node pools, and they will ignored if set.
```
tombentley, line 120: I think there must be some words missing in "since when using node pools, and they will ignored"

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,821 @@
...
+This proposal tries to minimize the effort and the scope and therefore picks up only some of the fields for the initial implementation.
+Since the existing fields in the `Kafka` CR will be used as the _defaults_, new fields can be added to `KafkaNodePool` resource in the future without any backwards compatibility concerns.
+
+### Resource naming
+
+Each node pool will be represented by its own `StrimziPodSet` and pods running the actual nodes.
+The `StrimziPodSet` will be created automatically by the operator based on the `KafkaNodePool` definition - it will not be created by the users.
+The `StrimziPodSet` will be named `<ClusterName>-<PoolName>` - e.g. `my-cluster-big-nodes`.
+The resulting pods will be named `<ClusterName>-<PoolName>-<ID>` - e.g. `my-cluster-big-nodes-5`.
+The naming of the related resources such as Services, Config Maps, or PVCs will follow the pod names.
```
tombentley, line 140: I wonder if we should mention that this puts a bound on how long the combined `<ClusterName>-<PoolName>-<ID>` can be (due to the constraints that Kube has on resource names)?

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,821 @@
...
+    - id: 0
+      type: persistent-claim
+      size: 100Gi
+      class: regular-storage
+```
+
+#### Protecting `KafkaNodePool` resources from being shared between multiple Kafka clusters
+
+The `strimzi.io/cluster` label should guarantee that the node pool will match only one Kafka cluster.
+Additional protection will be made against undesired moves between different Kafka clusters when the `strimzi.io/cluster` label is modified.)
```
tombentley, line 243:
```suggestion
Additional protection will be made against undesired moves between different Kafka clusters when the `strimzi.io/cluster` label is modified.
```

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,821 @@
...
+spec:
+  # ...
+status:
+  # ...
+  clusterId: cABpUKeDTqi7q_XOtBMgUv
+```
+
+The cluster ID is already part of the `Kafka` CR `.status` section.
+The operator will check the cluster IDs in every reconciliation and if the cluster IDs don't match between the `Kafka` and `KafkaNodePool` CRs, it will expect that there is some kind of an misconfiguration.
+It will throw an `InvalidresourceException` and wait for the user to fix the issue.
```
tombentley, line 263:
```suggestion
It will throw an `InvalidResourceException` and wait for the user to fix the issue.
```

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,821 @@
...
+User will be able to use annotations to indicate to the operator what should be the next ID when scaling up or what should be the node ID which should be removed when scaling down through annotations.
+Setting an annotation `strimzi.io/next-node-ids` will tell the operator what should be the next ID used in scale-up.
+The value might be a single number, list of numbers or one or more ranges.
+E.g. `[3]`, `[3, 4, 5]` or `[1000-1010]`.
+Similarly, setting an annotation `strimzi.io/remove-node-ids` will allow to configure IDs of node which should be removed.
+This will support only an ordered list of nodes IDs without ranges.
+When the operator is adding or removing a node from the pool, it will look at these annotations and if they are set, it will use them to override the default mechanism for picking up the next node ID or the node to remove.
+These annotations will be ignored when scaling is not requested by changing `.spec.replicas`.
+If these annotations cannot be fulfilled (for example because the ID is already taken and already in use) they will be ignored and the next suitable ID will be picked up.
+The operator will raise a warning in its log in such case.
```
tombentley, line 484: 1. I wonder if we should list exactly the cases which we'll cover. 
2. Is it enough just to log a warning?
3. I assume it's not allowed for these things to overlap between pools?

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,821 @@
...
+Until this is implemented, the partition replicas can be moved also using the `kafka-reassign-partitions.sh` tool.
+
+Despite the risks and limitations, having the ability to be able to move the nodes between node pools is still desired and might be useful.
+So it should still be supported, even if it might not be part of the initial implementation.
+
+### Deletion of the `KafkaNodePool` resource
+
+When the `KafkaNodePool` resource belonging to an existing Kafka cluster is deleted, all the resources belonging to it and managed by it will be deleted as well.
+This will also include the Pods and other resources.
+The PVCs will be deleted only if the `deleteClaim` flag was set to `true` in the storage configuration.
```
tombentley, line 535: Mention the default?

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,821 @@
...
+      type: persistent-claim
+      size: 100Gi
+      class: regular-storage
+```
+
+#### Protecting `KafkaNodePool` resources from being shared between multiple Kafka clusters
+
+The `strimzi.io/cluster` label should guarantee that the node pool will match only one Kafka cluster.
+Additional protection will be made against undesired moves between different Kafka clusters when the `strimzi.io/cluster` label is modified.)
+The `.status` section of the `KafkaNodePool` resource will contain the Kafka cluster ID of the cluster it belongs to.
```
devguyio, line 244: I think also having the node pools reflected in the `Kafka` status can be very handy to have a single view of the cluster. I can think also if it would reflect the ignored fields from the nodepools that can help users troubleshoot undesired Kafka setups. It can be limited to the case where there's a conflict to reflect the effective value.

050-Kafka-Node-Pools.md
scholzj: I have reworked this section.

050-Kafka-Node-Pools.md
scholzj: What exactly would you want to have reflected there? I'm open to ideas. But keep in mind that there are size limitations to what we can put there both from the CRD schema perspective as well as from the resource size limits.

ppatierno:
@devguyio ...

> The Kafka reconciler will create two NodePool resources out of the NodePoolClass, each has the same spec as the proposed NodePool here and also supports the scale subresource.

The concept of class means something that can be instantiated multiple times with parameters. Using the same spec of `KafkaNodePool` would not make sense to me because, for example, the `replicas` field is one of such parameters and not part of a class (doesn't make sense a class representing a pool with 3 replicas). Thinking about Kubernetes storage classes, for example, the disk size is out of that.

> The NodePool resource can be edited by users and supports the /status subresource, and if deleted by the user, it will always get recreated as long as Kafka resource still references its class.

Tbh everything created by an operator should be never touched or modified by a user. The rule should be the operator reverts any change, as it happens today for operands (because the node pool class would be an operand imho).

scholzj:
@devguyio I'm not sure I follow the added value against the complexity TBH. Plus the Class concept as used by other Kubernetes resources also goes against what you propose. The Class resource essentially just defines instance which should handle it or the mechanism but the exact configuration with things such as storage, resources or number of replicas would need to live somewhere else. 

Also the usecase here is different IMHO from Ingress, Storage or Gateway API. The separation between centrally managed infrastructure and end-user resources using that infrastructure is different for something like Kafka. For things like storage or Ingress, you have centrally managed infrastructure and end-users requesting the storage provisioning or the exposing of a service. In Kafka, you would either manage the Kafka cluster centrally or not. But I do not see how would you create the `Kafka` CR locally but have the `KafkaNodePoolClass` resources managed centrally.

050-Kafka-Node-Pools.md
scholzj: > 1. I wonder if we should list exactly the cases which we'll cover.

Sorry, what exactly you mean here? what would be the cases?

> 2. Is it enough just to log a warning?

TBH, none of this is implemented so far. The prototype just picks the next ID right now. But my plan was to treat these as an indication and not fail when they don't fit. I think failing with error is a bit dangerous if you combine it with some auto-scaling mechanisms where the automated scale-up could possibly lead to reconciliation errors. But I guess its _open for discussion_.

> 3. I assume it's not allowed for these things to overlap between pools?

I do not see overlap as an issue. I can for example imagine that you would want to have your controller nodes from the 0-99 range and all your broker pools from 100-1000 range. I don't think the overlap causes any issues - since the node IDs are picked serially, it will simply always get the next free one.

devguyio:
@ppatierno 

> @devguyio ...
> 
> > The Kafka reconciler will create two NodePool resources out of the NodePoolClass, each has the same spec as the proposed NodePool here and also supports the scale subresource.
> 
> The concept of class means something that can be instantiated multiple times with parameters. Using the same spec of `KafkaNodePool` would not make sense to me because, for example, the `replicas` field is one of such parameters and not part of a class (doesn't make sense a class representing a pool with 3 replicas). Thinking about Kubernetes storage classes, for example, the disk size is out of that.

The replica value in the `NodePoolClass`  is here provide default or initial values. 

> 
> > The NodePool resource can be edited by users and supports the /status subresource, and if deleted by the user, it will always get recreated as long as Kafka resource still references its class.
> 
> Tbh everything created by an operator should be never touched or modified by a user. The rule should be the operator reverts any change, as it happens today for operands (because the node pool class would be an operand imho).

That depends on the reconciler I believe? Since

* KafkaReconciler
  * Validates an existing NodePoolClass exists.
  * If a matching NodePool resources doesn't exist, it creates one filling the default values from the `NodePoolClass`

So in this sense, it's not "reconciling" the NodePool resource. It's not checking the desired spec vs status and reconciling it. It's more or less validating the link between the `Kafka` resource and then exposing to the user the APIs to dynamically update something like the replicas.

devguyio:
@scholzj 
> @devguyio I'm not sure I follow the added value against the complexity TBH. Plus the Class concept as used by other Kubernetes resources also goes against what you propose. The Class resource essentially just defines instance which should handle it or the mechanism but the exact configuration with things such as storage, resources or number of replicas would need to live somewhere else.
> 
> Also the usecase here is different IMHO from Ingress, Storage or Gateway API. The separation between centrally managed infrastructure and end-user resources using that infrastructure is different for something like Kafka. For things like storage or Ingress, you have centrally managed infrastructure and end-users requesting the storage provisioning or the exposing of a service. In Kafka, you would either manage the Kafka cluster centrally or not. But I do not see how would you create the `Kafka` CR locally but have the `KafkaNodePoolClass` resources managed centrally.

So trying to think deeper about what I'm trying to address, I guess the biggest thing I'm trying to address is to "whitelist" NodePools. That the actually creation of new sets of Kafka servers that didn't exist before, take place after the used edits the `Kafka` resource, not after creating a `NodePool` resource if that makes sense.

Did you consider that first users create `NodePool` resources, then add them to the Kafka resource `spec.nodepools` by name ?

ppatierno:
@devguyio ...

> That depends on the reconciler I believe?

If you have a Deployment with 3 replicas so the corresponding operand ReplicaSets will have 3 replicas, if you try to set 4 replicas on the ReplicaSets, the built in Deployment controller in Kubernetes will terminate the new pod (started by the ReplicaSets controller) because ReplicaSet is an "operand" in this case. This is how the Strimzi operator or in general operators work, or?

scholzj:
> So trying to think deeper about what I'm trying to address, I guess the biggest thing I'm trying to address is to "whitelist" NodePools. That the actually creation of new sets of Kafka servers that didn't exist before, take place after the used edits the Kafka resource, not after creating a NodePool resource if that makes sense.
> 
> Did you consider that first users create NodePool resources, then add them to the Kafka resource spec.nodepools by name ?

I did not consider that. But as I said, not sure I see the value over the complexity. I never thought about creating an additional node pool by mistake as something to be overly concerned with (plus they would anyway need to have the matching label anyway).

scholzj:
@devguyio Please also keep in mind that all of this is happening within a single namespace (the `Kafka` and `KafkaNodePool` resources have to be in the same namespace). And the general recommendation is to have one Kafka cluster per namespace. So I think that also impacts the risks here.

devguyio:
> @devguyio ...
> 
> > That depends on the reconciler I believe?
> 
> If you have a Deployment with 3 replicas so the corresponding operand ReplicaSets will have 3 replicas, if you try to set 4 replicas on the ReplicaSets, the built in Deployment controller in Kubernetes will terminate the new pod (started by the ReplicaSets controller) because ReplicaSet is an "operand" in this case. This is how the Strimzi operator or in general operators work, or?

For that case that is correct, but that's because the "owning" controller (i.e. deployment controller) sets up watches on the replicaset updates. The deployment controller doesn't expect updates on the RepliacSet resource.

That's not the case for all controllers, some controllers creates resources and ensures the resources existence only, but don't set up watches for the updates are expected to take place via the user directly on the resource. It's kinda like exposing an API, or exposing some knobs for the user once a precondition is available. E.g. Once a `ManagedServiceConnection` is created with Paolo's valid credentials, the controller creates a `ManagedServiceUserAccount` resource for Paolo's account representing all his user account data that he can use to update his username, address etc. 

So for what I proposed, the `KafkaReconciler` won't set watches on the `NodePool` resource updates (i.e. will ignore the updates) , only will make sure it still exists as long as it's in the `Kafka.spec.nodepools` array, or deleted otherwise.

The `NodePoolReconciler` will be the one responsible of reconciling `NodePools` uptade to make sure that the desired state in `NodePool.spec` matches the current state.

devguyio:
@scholzj 

> I did not consider that. But as I said, not sure I see the value over the complexity. I never thought about creating an additional node pool by mistake as something to be overly concerned with (plus they would anyway need to have the matching label anyway).

I guess the actual added value to the UX I see is that the `Kafka` resource spec still reflects to some sense all the desired state of the Kafka cluster. I.e since the user is forced to add the NodePool to the `spec.nodepools` list, the mental model of the resulting servers is easier to grasp from that resource alone. Instead of having to go through the extra step of also listing all `NodePool` with the matching label in the current namespace.

However, I can see it's a compromisation call and if the `NodePools` are all within the same namespace as the Kafka resource as you mentioned that wouldn't make it a terrible UX compared to if they were cluster wide resource .

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,832 @@
...
+    type: jbod
+    volumes:
+    - id: 0
+      type: persistent-claim
+      size: 1000Gi
+      class: fast-storage
+  # ...
+```
+
+Because the `KafkaNodePool` is named `kafka`, all the resources created by this configuration will be exactly the same and have the same names as today.
```
ppatierno, line 699: Is it possible changing the name from `kafka` to something else after the migration?

scholzj: Well, it depends on how you define _renaming_. You cannot rename Kubernetes resources. So I would say no, you cannot rename it. But of course, once you migrated to Node Pools, you can for example create a new node pool `akfak`, migrate all data to it, and delete the `kafka` node pool. You are not required to keep using the `kafka` node pool forever. But it is a bit more complicated then renaming it.

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,832 @@
...
+It will be created and managed directly by users (unlike for example the `StrimziPodSet` resource which is used only internally by Strimzi and should not be touched by the users).
+
+The _cluster-wide_ configuration options will remain part of the `Kafka` CR.
+While the _pool-specific_ options will be taken from `KafkaNodePool` resources
+
+For the initial implementation, the following options are proposed for the `KafkaNodePool` resource and its `.spec` section:
+
+* `replicas` indicating the number of nodes in the pool
+* `storage` indicating the storage of the nodes in this pool
+* `resources` indicating the amount of resources (requests / limits) assigned to the nodes in this pool
```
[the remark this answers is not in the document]
scholzj: I did not consider this. From my experience, the majority of Strimzi users do not run large fleets of uniform clusters. They tend to either have a limited number of centrally managed clusters or many different clusters with different configurations. So I'm not sure this factory approach fits - I would be afraid it increases the complexity for both users as well as on the code level without too much added value.

You can probably also combine other tools such as GitOps to create a factory-style approach like this - for example with a Helm Chart which limits the configurability of some parts to decrease the complexity / increase the uniformity

devguyio:
> @scholzj
> 
> > I did not consider that. But as I said, not sure I see the value over the complexity. I never thought about creating an additional node pool by mistake as something to be overly concerned with (plus they would anyway need to have the matching label anyway).
> 
> I guess the actual added value to the UX I see is that the `Kafka` resource spec still reflects to some sense all the desired state of the Kafka cluster. I.e since the user is forced to add the NodePool to the `spec.nodepools` list, the mental model of the resulting servers is easier to grasp from that resource alone. Instead of having to go through the extra step of also listing all `NodePool` with the matching label in the current namespace.
> 
> However, I can see it's a compromisation call and if the `NodePools` are all within the same namespace as the Kafka resource as you mentioned that wouldn't make it a terrible UX compared to if they were cluster wide resource IMHO .

@scholzj after some thought into this I trying to see what's the complexity that this model would provide, I wanted to share it outloud and see if I'm missing something cause I think it'd be a straight forward implementation:

1. Create a `NodePool` resources, The same spec as the proposed here without the annotiation.
2. The `NodePool` reconciler will validate the resource and see that NodePool `status.kafka` is not set and so will not do anything other than the validation ( and maybe set the status `bound` condition to `False`).
3. Create a `Kafka` resource and reference the node pool in a `spec.nodePools` list
```yaml
kind: Kafka
metadata:
  name: my-cluster
spec:
  kafka:
    nodePoolSelector:
      matchLabels:
 ...
  # ...
  nodePools: 
    - pool1
    - pool2
```
4. If the `NodePool` exists, the Kafka reconciler will "bind" the Kafka to the node pool. It will patch the node pool resources' status with something like `NodePool.status.kafka: my-cluster` and set the `Kafka.status,nodepools` to reflect the bound node pools.
5. If the `NodePool` doesn't exist, it will be ignored and not show up in the `Kafka.status,nodepools`, pending a future reconciliation. 
6. The `NodePool` reconciler will get the update, see that the `status.kafka` is now set, and will start reconciling similar to this proposal. 


Not sure how do you feel about that? Any obvious added complexity? It has three added values IMHO

1. There's no race condition between `Kafka` and `NodePool` resources reconciliation. NodePool will be eventually bound and reconciled.
2. No big addition to the CRD or status, so no size worries.
3. Makes the binding explicit in the `Kafka` CRD, so it's still the single source of truth of the Kafka's desired and current state , even if by explicitly linking to other resources for such desired state.

scholzj:
@devguyio It does not solve any eventual consistency issues. There is no eventual consistency issue if you first create the node pools and then the Kafka resource. The eventual consistency might happen when you create all resources from one YAML file. And it can happen in the current proposal as much as in yours.

devguyio:
@scholzj what about the rest of my points about `Kafka` being a single source of truth of the Kafka's desired and current state , even if by explicitly linking to other resources for such desired state ? Don't you think it's more intuitive to specify the `node pools` in the Kafka resource ?

scholzj:
> TBH I can't see how will it happen with my proposal. If the Kafka gets reconciled first, it'll not fail, it'll just set in the status that those nodepools are not bound.

If you have one YAML file with Kafka CR and 2 pools and you deploy it you have exactly the same issues happening as in my proposal:
* They are all created in time and are there => All OK
* The NodePools create before the Kafka CR => Same as mine proposal, All OK
* The Kafka creates first but the node pools it refers to are missing => Same issue as my proposal

So no, it does not solve anything from my PoV.

> what about the rest of my points about Kafka being a single source of truth of the Kafka's desired and current state , even if by explicitly linking to other resources for such desired state ? Don't you think it's more intuitive to specify the nodePools in the Kafka resource ?

It is not a single source of truth when you have the configuration across multiple custom resources. Not with my proposal nor when you list the node pools in the Kafka CR as you suggest.

scholzj:
PS: If you want to bring it closer to source of truth, I think it might be more interesting to think about what should be in the status section of the Kafka CR then to add more selectors. Maybe having. list of known node pools in the status and their generations could be useful for example.

devguyio:
> > TBH I can't see how will it happen with my proposal. If the Kafka gets reconciled first, it'll not fail, it'll just set in the status that those nodepools are not bound.
> 
> If you have one YAML file with Kafka CR and 2 pools and you deploy it you have exactly the same issues happening as in my proposal:
> 
> * They are all created in time and are there => All OK
> * The NodePools create before the Kafka CR => Same as mine proposal, All OK
> * The Kafka creates first but the node pools it refers to are missing => Same issue as my proposal
> 
> So no, it does not solve anything from my PoV.

I just noticed the missing piece that I was didn't specify, the Kafka CR won't get ready until all `nodePools` in the list are bound. So the reconciler won't deploy the Kafka pods cause it's aware this is not a stand alone Kafka, but it's linked to other node pools that are not available currently. In this proposal, there's no way the Kafka reconciler knows that and has no choice but to deploy the pods if no node pools exist.

devguyio:
>  The Kafka CR won't get ready until all nodePools in the list are bound

Kube does the same behavior when you create a deployment with e.g. a service account that doesn't exist. The pods won't get created as the replicaset will not be ready and will reflect that the SA doesn't exist in the status conditions as show below.

```yaml
status:
  conditions:
  - lastTransitionTime: "2023-04-07T10:14:39Z"
    message: 'pods "my-kroxy-58f4bddcb9-" is forbidden: error looking up service account
      default/hello-world: serviceaccount "hello-world" not found'
    reason: FailedCreate
    status: "True"
    type: ReplicaFailure
  observedGeneration: 1
  replicas: 0
```

Yet once that service account is created and after a few seconds, the replicaset becomes ready scaled.

```yaml
status:
  availableReplicas: 3
  fullyLabeledReplicas: 3
  observedGeneration: 1
  readyReplicas: 3
  replicas: 3
```

> I think it might be more interesting to think about what should be in the status section of the Kafka CR than to add more selectors. Maybe having. list of known node pools in the status and their generations could be useful for example.

Yeah that's definitely useful as well and it's why I suggested it in [my first comment](https://github.com/strimzi/proposals/pull/77/files#r1157273853) . 

The reason I'm discussing the change of adding a list of `nodePools` and this binding concept and trying to get to the bottom of it is because it's still a proposal and it seems like you rejected similar alternatives like embedding the whole node pools in the Kafka CRD but rejected them for valid reasons, and also highlighted in the risks the eventual consistency issue when applying all resources in a single file. 

So I'm trying to see if that was considered and there's a valid approach that fixes those two issues (as shown in my previous comment) what other reasons might make listing the node pools this Kafka belongs not an improvement to this proposal?

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,832 @@
...
+The cluster ID is already part of the `Kafka` CR `.status` section.
+The operator will check the cluster IDs in every reconciliation and if the cluster IDs don't match between the `Kafka` and `KafkaNodePool` CRs, it will expect that there is some kind of an misconfiguration.
+It will throw an `InvalidResourceException` and wait for the user to fix the issue.
+
+### Scaling the node pools
+
+The `KafkaNodePool` resource will also support the `scale` sub-resource.
+Support for the `scale` sub-resource should allow integration with Kubernetes Horizontal Pod Autoscaler.
+That on its own doesn't really provide fully functional Kafka auto-scaling mechanism.
+For example, if the load is coming from a single partition which is handling too many messages, adding more nodes will not help.
```
fvaleri, line 291: Should we keep the scale sub-resource support out of this proposal? 

I think that use cases for the HPA integration are not clear, instead a vertical autoscaler may help with the single partition issue and rebalances. The auto scaling feature deserves its own proposal and it may include the NodePool scale sub-resource improvement. In this proposal we could simply mention that possibility.

fvaleri (commented):
Thanks for this proposal. I left some initial comments.

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,832 @@
...
+        labels:
+          key1: label1
+    pod:
+      terminationGracePeriodSeconds: 60
+```
+
+The replicas and storage sections will be required in `KafkaNodePool`.
+The other fields will be optional.
+
+When the optional field is not present in `KafkaNodePool` and is present in `Kafka.spec.kafka`, it will default to the configured value from the `Kafka` CR.
```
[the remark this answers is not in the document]
scholzj: I do not find this useful. It just adds more complexity while I think nobody will really follow this. Also, what if the options is configured by the default default? (i.e. the default value for the default value in Kafka CR?)

050-Kafka-Node-Pools.md
scholzj: I think you are mixing different things. Scale subresource does not mean autoscaling on its own. It means the ability to scale the resource independently without understanding its structure which is the basic building block of autoscaling solution. We already support the scale subresource almost everywhere apart from the `Kafka` CR because that has 2 replicas fields and `KafkaTopic` as partition count cnnot be decreased. So I think having the scale subresource is in sync with the rest of Strimzi.

050-Kafka-Node-Pools.md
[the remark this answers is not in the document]
fvaleri: > Also, what if the options is configured by the default default? (i.e. the default value for the default value in Kafka CR?)

I think you are referring to things like LimiRanges. That's different because they are only applied at the pod admission stage and changing values would only have an effect on new pods. Instead, changing Kafka resources would cause cause a rolling update.

050-Kafka-Node-Pools.md
fvaleri: OK, but if scale support is the default, why do we need to mention HPA and auto-scaling in this context? I think this is confusing.

050-Kafka-Node-Pools.md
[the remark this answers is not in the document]
scholzj: No. There are fields in the Kafka CR which do have their own defaults when not set. So I think what you suggest here is not really as straightforward and understandable. In any case, it is definitely not something what is key to the proposal. If we see the need for this or some misunderstandings of what values are used, this can be added at any point later.

050-Kafka-Node-Pools.md
scholzj: The scale subresource is a key building block for any auto-scaling solution and auto-scaling is something what was asked for by the users in the past. It also plays an important role for the motivation why to use separate custom resource for the node pools. That is why it is worth mentioning in the proposal. But at the same time, I feel like it is important to clarify that this does not mean that this proposal somehow solves auto-scaling. It just respects it as one of the possible future improvements which can build on top of this.

At the same time, the scale subresource is cheap implementation wise. So if nobody uses it because there is no feasible way to autoscale Kafka, then I don't think we care.

050-Kafka-Node-Pools.md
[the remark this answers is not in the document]
fvaleri: I agree that this is not key to the proposal and the above status would be confusing, but yet this is a behavior that could be unexpected for some users and make troubleshooting more complicated (you need to combine two different resources). An alternative could be to simply log the KafkaNodePool applied configuration at startup. WDYT?

fvaleri:
> @devguyio It does not solve any eventual consistency issues. There is no eventual consistency issue if you first create the node pools and then the Kafka resource. The eventual consistency might happen when you create all resources from one YAML file. And it can happen in the current proposal as much as in yours.

I also think that the race condition is there in both cases and it is pretty common in K8s land. The @devguyio configuration adds the information of which KafkaNodePools are expected to be bound to Kafka, but in the end I tend to favor consistency and the label way of modeling the resource binding is already used in many places inside Strimzi and it's the same mechanism that K8s uses for the EndpointSlice->Service binding.

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,832 @@
...
+Only small parts of the code will actually depend on the feature gate being enabled or disabled.
+
+The annotation `strimzi.io/node-pools: enabled` on the `Kafka` custom resource will be also required in order to use the `KafkaNodePool` resources on a given Kafka cluster.
+
+Since the KRaft mode will be supported only with `KafkaNodePool` resources, the `UseKRaft` feature gate will be configured as dependent on the `KafkaNodePools` feature gate.
+It will not be possible to enable `UseKRaft` feature gate without also having `KafkaNodePools` feature gate enabled.
+
+#### Virtual node pool
+
+To avoid two parallel implementations in the Strimzi source code, the operator will use _virtual node pool_ when the feature gate is disabled.
```
fvaleri, line 591: This will be a virtual node pool named `kafka` right?

devguyio:
> > @devguyio It does not solve any eventual consistency issues. There is no eventual consistency issue if you first create the node pools and then the Kafka resource. The eventual consistency might happen when you create all resources from one YAML file. And it can happen in the current proposal as much as in yours.
> 
> I also think that the race condition is there in both cases and it is pretty common in K8s land. The @devguyio configuration adds the information of which KafkaNodePools are expected to be bound to Kafka,

@fvaleri Thanks for chiming in! 

I'm not sure if you've read [my last two comments](https://github.com/strimzi/proposals/pull/77#issuecomment-1499526259) ? If so, can you please help me understand how will the race condition happen? As I explained, showed that Kafka won't readiness will be false and pods won't be deployed even if the Kafka gets reconciled before the NodePool. I'm afraid I can't see how will that race happen.

> but in the end I tend to favor consistency and the label way of modeling the resource binding is already used in many places inside Strimzi and it's the same mechanism that K8s uses for the EndpointSlice->Service binding.

I agree that consistency generally is favorable, but notice that EndpointSlice is an internal resource that's not created by the user and is by design not exposed in the `Service` API. Users aren't expected to manage them. Yet it is still primarily [tracked by an owner reference](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#ownership) and the label is used merely for easier lookup since it can't by design be listed in the Service resource.

devguyio:
Also note that we can still keep the label for lookup and consistency. IMHO it's an orthogonal point to the list of node pools, both can exist together.

fvaleri:
@devguyio np, I'm also trying to understand the design, so I think it's a helpful discussion.

> how will the race condition happen?

When Kafka and NodePool CRs are created at the same time (a single YAML file), there is no guarantee on the ordering in which these resources will be created by K8s and thus reconciled by the operator. This means that the same cluster configuration can produce different system behaviors. This is true regardless of how you handle the missing NodePool resources. In either case, we are switching from a strong consistent Kafka model to an eventual consistent Kafka model. Not a problem in this specific context.

> If the Kafka gets reconciled first, it'll not fail, it'll just set in the status that those nodepools are not bound.

Failure or not the end result would be the same: a Kafka cluster which is NotReady after one or more reconciliations. What if I forget to create the NodePools or there is a mismatch between the desired NodePool name in Kafka CR and the actual NodePool name? Is there a timeout after which you would switch from a "waiting for binding" state to a "binding failure" state?

Additionally, the current prototype has just one KafkaReconciler to handle both Kafka and NodePool CRs, while your design would require two reconcilers communicating through status updates, which I think would add some complexity.

050-Kafka-Node-Pools.md
scholzj: Yes, it will be named `kafka`. That way, all the resources will have the same names as today.

050-Kafka-Node-Pools.md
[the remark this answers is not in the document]
scholzj: I do not see the confusion TBH. But the advantage of this being not a key part of the proposal is that if there actually turns out to be any confusion by the users, then we can decide what to add and where exactly.

devguyio:
Just for the sake of completeness, and fwiw I'm gonna put down the change to the proposal that I'm suggesting in full here again in full cause it's scattered across multiple comments. 

1. Explicitly declare the desired node pools in the `Kafka` CR by adding a `nodePools` list to the `Kafka` CR 
```yaml
kind: Kafka
metadata:
  name: my-cluster
spec:
  kafka:
    nodePoolSelector:
      matchLabels:
 ...
  # ...
  nodePools: 
    - poolclass1
    - poolclass2
```

2. If any of the node pools don't exist, Kafka readiness will be set to false and no change will happen. This means
    1. If this a new creation of `Kafka`  : 
        1. Set Kafka readiness to False and set conditions.
        2. Don't create any pods for any node pool since we know the user expects this Kafka to belong to multiple node pools, but we can't achieve the full desired setup due to some missing node pools.
        3. Either the user made a mistake, so we wait for them to fix it. Or it's an eventual consistency situation in which all node pools will be available on a following reconciliation. 
    2. If this is an update
        1. Set Kafka readiness to False and set conditions.
        2. The existing pods will keep running and no rolling out or termination will take place.
        3. Either the user made a mistake, so we wait for them to fix it. Or it's an eventual consistency situation in which all node pools will be available on a following reconciliation. 


That's the core of the change I'm suggesting. Whether we do a "binding" or a separate NodePool reconciler are all implementation details that were just me trying to explain the main change, but they're irrelevant and orthogonal to the suggestion. 

The core of the suggestion is to make things simple and less clever by having an explicit list of node pools in the Kafka CR. 

> This means that the same cluster configuration can produce different system behaviors. This is true regardless of how you handle the missing NodePool resources. In either case, we are switching from a strong consistent Kafka model to an eventual consistent Kafka model.

@fvaleri I think since no change will take place if even one nodepool is missing means that the same cluster configuration must always produce the same system behavior.

>  What if I forget to create the NodePools or there is a mismatch between the desired NodePool name in Kafka CR and the actual NodePool name? Is there a timeout after which you would switch from a "waiting for binding" state to a "binding failure" state?

Regardless of `binding` which as I said above is not the core of the change, nevertheless if a user made a mistake in a node pool name, I believe it's better to not presume: 1) Do not do any change, don't create any new pods, don't terminate any pods, nothing. 2) Make it visible that we can't reconcile the desired state the user asked us to. 3) Wait for the user to fix the problem.

I believe that's better than doing a "partial" deployment. 

Imagine without the change I'm suggesting the user creates two node pools, controllers and brokers but puts the wrong name in the controllers NodePool `cluster` label. The brokers will be rolled out without the controllers. So Kafka will go in a broken state. Or even if there's a measure to prevent that specific scenario, imagine the two pools are `smallpool` and `bigpool` but the name is wrong in `bigpool` , Kafka will be deployed and be marked as ready even though it's the wrong, undesired deployment.

I think in such scenarios it's way better to not do any change, set Kafka as not Ready, and wait for a user intervention to fix the problem, hence my proposed change.

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,832 @@
...
+The `KafkaNodePool` resource will be a _public_ resource.
+It will be created and managed directly by users (unlike for example the `StrimziPodSet` resource which is used only internally by Strimzi and should not be touched by the users).
+
+The _cluster-wide_ configuration options will remain part of the `Kafka` CR.
+While the _pool-specific_ options will be taken from `KafkaNodePool` resources
+
+For the initial implementation, the following options are proposed for the `KafkaNodePool` resource and its `.spec` section:
+
+* `replicas` indicating the number of nodes in the pool
+* `storage` indicating the storage of the nodes in this pool
```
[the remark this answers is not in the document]
scholzj: And what would be the actual benefit? It would be quite complex in terms of implementation. Today, the proposal is changing the 1:N relationship (one Kafka CR to multiple pods) to N:M relationship (multiple NodePools against multiple pods). Having a separate storage configuration would change it to N:M:O relationship (multiple NodePools against multiple pods against multiple storage pools). 

It would also further complicate the user experience as even a very basic cluster (3 KRaft controllers + 3 brokers) would likely require at least two storage pools (different size for controllers versus brokers) and two node pools (controllers and brokers).

I'm not sure I understand the advantages it would bring. WDYM with vertical scaling in this case? The node pool can be vertically scaled -> you can just change the resources and/or storage size without any limitations.

050-Kafka-Node-Pools.md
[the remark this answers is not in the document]
SamBarker: Yeah my background is definitely showing through :D 

Defaulting to the factory model probably would be a small increment in complexity for users managing small numbers of clusters and potentially pointing them towards a different pattern, which is probably more opinionated than the project want to be. I do wonder if we can find a middle ground, this ties in with [my thinking](https://github.com/strimzi/proposals/pull/77#discussion_r1162112642) around `stroagePools`, and support either an inline definition or a pointer to a `spec`. The default option would be an inline definition. 

To turn from the user complexity to the implementation complexity for a moment, again I don't actually think it's that large, or rather doesn't need to be. In my mental model the reconciliation would effectively split into two phases. An assembly phase stitching everything together (i.e. following the pointers to the `spec` definitions so as to appear as if inline) which would output a fully rendered model to the second phase of the reconciler which would then operate on that rendered model and thus wouldn't care if it contained pointers or inline definitions were used. 

I guess that would lead to questions about validation of the separate `spec` resources. I think, initially at least, they could just be handled like the inline version would be in the current proposal however there would be the option to validate them in a dedicated reconciler.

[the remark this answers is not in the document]
scholzj: Sorry, I'm getting a bit lost and I'm still not sure I understand what would be the actual benefit.

The storage configuration is inlined already today and as part of this proposal. I also don't follow what you mean by the _factory model_. There is no factory here. The resources belong to a single unique cluster in a single namespace. There is no reuse. Especially with storage, which is mostly configurable at creation only, it would be quite hard to share the configuration and the changes would very quickly get completely out of sync without any real possibility to fix it.

For example - imagine someone creates the storage configuration for `gp2` volumes and uses it to create cluster A. And then changes the _shared_ storage configuration to `gp3` and creates cluster B. Suddenly, you have a single storage configuration that is always invalid for either cluster A or cluster B.

ppatierno:
@devguyio just to clarify my understanding across a lot of comments ... with your last proposal, who creates the Node pool, user or the operator? It seems you changed your mind from operator to user.

devguyio:
> @devguyio just to clarify my understanding across a lot of comments ... with your last proposal, who creates the Node pool, user or the operator? It seems you changed your mind from operator to user.

@ppatierno yeah quite a lot of discussions and back and froth indeed. So I did drop the proposal of `NodePoolClass` and `NodePool` indeed after a few comments from Jakub for two reasons:

1. To focus first on my biggest concern, representing the desired set of nodepools this Kafka belongs to in the kafka CR, which can be achieved without the `NodePoolClass` and auto-creation of `NodePools`.
2. IMHO node pool classes can be a separate addition and separate discussion later. 

That's why I tried to put the proposed change again in full again in my previous comment.

050-Kafka-Node-Pools.md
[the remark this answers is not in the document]
SamBarker: I miss have miss understood, I had taken away the impression that the increase the resources available the node pool needed replaced as that's not the case it doesn't bring any advantage.I miss have miss understood, I had taken away the impression that the increase the resources available the node pool needed replaced as that's not the case it doesn't bring any advantage.

050-Kafka-Node-Pools.md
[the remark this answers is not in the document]
scholzj: > While I don't have the data to back up the hunch I'd guess people operating diverse sets of Kafka clusters have very similar Zookeeper clusters (Zookeeper having very different scaling needs to Kafka and I suspect the same holds for Kraft controllers) and would thus benefit from having a standard definition of a controller cluster while maintaining the flexibility to run very different scales of broker cluster.

There will be probably similar pieces of configuration. But I think they might differ and different users might want different pieces to be _reusable_. You talked about storage and resources. Others would want to reuse affinity / tolerations / topology spread constraints rules, etc. I don't think you can solve this by splitting the Strimzi configuration into many small pieces. I think the solution is to solve it on the upper levels.

So, if you are running a large and centrally managed fleet of Kafka clusters, you will probably rely on other means anyway. For example on a custom operator for managing the clusters. Or on some GItOps flows and custom Helm Charts which allows you to decide which configuration should be shared and which should be up to customization. I think leaving the unification of some parts of the custom resources to this layer is much better then creating a separate resource for every single field someone wants share for their fleet of Kafka clusters.

Frawless (approved):

im-konge (approved):
This looks really great, thanks a lot Jakub :)

ppatierno:
@devguyio if having the list of pools in the Kafka CR, what's the purpose of still having the cluster label in the KafkaNodePool CR? Is this bidirectional link really useful?
I think that both are a way to define a owner-owned relationship.
The pool list defines "owner to owned" while the label more a "owned to owner" which is what we usually have in Strimzi for all others operand resources (which is more in general how Kube works with selectors, or?) 
Tbh it is not still clear to me which issue this change is going to solve apart from a UX one?
Isn't reporting the matched node pools in the status of the Kafka CR so making the user aware of what's happening?

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,865 @@
...
+They can keep using their existing `Kafka` resources without any changes.
+The operator will continue to use the _virtual node pool_ as described in the previous section.
+
+When they decide to start using the actual `KafkaNodePool` resources, they can either:
+* Deploy a brand new Kafka cluster with one or more `KafkaNodePool` resources
+* Convert their existing Kafka cluster to use node pools
+
+#### Converting existing Apache Kafka clusters
+
+To convert an existing Kafka cluster to use node pools, the user has to do the following two steps:
```
PaulRMellor, line 640:
```suggestion
To convert an existing Kafka cluster to use node pools, the user has to do the following:
```

PaulRMellor: as a minimum? or is there more they can do in terms of config at this stage?

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,865 @@
...
+    type: jbod
+    volumes:
+    - id: 0
+      type: persistent-claim
+      size: 1000Gi
+      class: fast-storage
+  # ...
+```
+
+Because the `KafkaNodePool` is named `kafka`, all the resources created by this configuration will be exactly the same and have the same names as today.
```
PaulRMellor, line 732: do we need to say anything about what happens if another name is used?

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,865 @@
...
+But it is not completely clear how many problems and various race conditions this might be causing when updating the node pools for existing clusters etc.
+
+The eventual consistency also shows up in situations when you for example decide to scale two of your node pools.
+The scale events will be independent and will almost never be caught by the same reconciliation.
+So you will have one reconciliation to scale the first pool followed by the second to scale the second one.
+
+## Not impacted
+
+This proposal does not impact in any way how the ZooKeeper nodes, Kafka Exporter, Cruise Control or Entity Operator are configured.
+It also does not impact configuration of the `KafkaConnect`, `KafkaBridge` or other custom resources.
```
PaulRMellor, line 762: is there any expectation that other components might be handled in the same way?

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,865 @@
...
+
+## Rejected alternatives
+
+### KafkaNodePool selector in `.spec.kafka` section of the `Kafka` custom resource
+
+One of the options considered was to have a label selector in the `Kafka` custom resource in `.spec.kafka`.
+Using the `strimzi.io/cluster` label was chosen at the end as the preferred way.
```
PaulRMellor, line 769:
```suggestion
Using the `strimzi.io/cluster` label was preferred in the end.
```

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,865 @@
...
+It makes it easier to match the `KafkaNodePool` resources to the `Kafka` cluster.
+This significantly simplifies the code - especially the watch which is triggered when
```
PaulRMellor, line 771:
```suggestion
This significantly simplifies the code - especially the watch which is triggered when changes are made.
```

PaulRMellor: Sentence is unfinished. Guessed the ending!

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,865 @@
...
+And of course also some disadvantages:
+* It would not possible to support the `scale` sub-resource and support autoscaling in the future.
+* In the past, we had issues with the size of our CRDs.
+  Bundling the node pools into the `Kafka` CRD would make it bigger.
+
+## Not addressed
+
+### Migration to KRaft / ZooKeeper-less Kafka
+
+While this proposal defines how the node pools will provide support for running Apache Kafka in the KRaft mode, they do not focus on the migration of ZooKeeper-based Kafka clusters to KRaft based Kafka clusters.
```
PaulRMellor, line 854:
```suggestion
While this proposal defines how the node pools will provide support for running Apache Kafka in the KRaft mode, it does not focus on the migration of ZooKeeper-based Kafka clusters to KRaft-based Kafka clusters.
```

PaulRMellor (approved):
The implementation was clearly outlined and the addition of node pools seems like a great feature for Strimzi

rareddy:
TBH, I only spent less than 10 minutes in reading through the proposal, apologize if I say something very ignorant

1) Why is it called `KafkaNodePools`?  "node" is kube level resource, I thought "KafkaBrokerPool" should have been a more fitting resource name. Unless a Broker and Node are synonymous in this profile.
2) I saw this as externalizing the configuration of "set of" brokers to fulfill the request for them to have different configurations in a given single Kafka cluster.  So, looks more like `KafkaBrokerProfile` or `KafkaBrokerConfiguration`
3)  I see u have debated about this at length above, but having one or more `KafkaNodePools` CRs along with Kafka CR to represent one logical "Kafka cluster" makes me think as broker resources are auxiliary resources like "topic" CR than part of the core Kafka cluster, which they are not.  Are there any specific "KafkaNodePool" operations you are thinking about which do not apply to other brokers in general that are reasons behind it? 
4) Having all `KafkaNodePools` belong to a single cluster and maintaining co-relation to `Kafka` cluster using labels/annotations seems like an additional burden on the user.

scholzj:
> Why is it called KafkaNodePools? "node" is kube level resource, I thought "KafkaBrokerPool" should have been a more fitting resource name. Unless a Broker and Node are synonymous in this profile.

Node is a generic term. _Broker_ is only one specific type of Kafka node. When running with ZooKeeper, all Kafka nodes are brokers. But with ZooKeeper-less Kafka, you need to distinguish broker nodes and controller nodes (or mixed nodes which are both broker and controller at the same time) which are both represented by the node pools. So calling it `KafkaBrokerPool` would be misleading.

> I saw this as externalizing the configuration of "set of" brokers to fulfill the request for them to have different configurations in a given single Kafka cluster. So, looks more like KafkaBrokerProfile or KafkaBrokerConfiguration

No, this is not externalizing configuration of a set of brokers. The `KafkaNodePool` represents the set of brokers (or rather nodes as explained above). You can see it for example from defining the number of brokers and their names and IDs.

> I see u have debated about this at length above, but having one or more KafkaNodePools CRs along with Kafka CR to represent one logical "Kafka cluster" makes me think as broker resources are auxiliary resources like "topic" CR than part of the core Kafka cluster, which they are not. Are there any specific "KafkaNodePool" operations you are thinking about which do not apply to other brokers in general that are reasons behind it?

The `KafkaNodePool` resources have similarities with `KafkaTopic` or `KafkaUser` resources. But there are obviously many differences as well. As you can see in the proposal, the possibility of having the node pools defined inside the `Kafka` CR instead of using separate `KafkaNodePool` resources was considered but (at least for the time being) was rejected. It would close doors to interesting options such as better support for auto-scaling etc.

I'm not sure what exactly you mean by _operations_ in this context. But the node pools are separate entities that can be added or removed as needed.

> Having all KafkaNodePools belong to a single cluster and maintaining co-relation to Kafka cluster using labels/annotations seems like an additional burden on the user.

As you can see in the proposal, other methods were considered. The current one was chosen as a good mix of user and implementation friendliness.

I agree that it certainly has increased complexity for the user compared to having only one resource and having the node pools fully configured inside the `Kafka` CR. But as mentioned above, it also brings advantages that a single resource cannot offer. It is not always possible to find a single perfect solution for every feature. So we either need to abandon some of them or find some reasonable middle ground 🤷.

rareddy:
>I'm not sure what exactly you mean by operations in this context. But the node pools are separate entities that can be >added or removed as needed.

By `Operations` I meant, for example, only rolling brokers in a certain node pool, or fixing a disk skew across given single or multiple node pools, any such operations that can be performed currently across all brokers, be restricted to a NodePool.

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,876 @@
...
+The `KafkaNodePool` resource will also support the `scale` sub-resource.
+Support for the `scale` sub-resource should allow integration with Kubernetes Horizontal Pod Autoscaler.
+That on its own doesn't really provide a fully functional Kafka auto-scaling mechanism.
+For example, if the load is coming from a single partition which is handling too many messages, adding more nodes will not help.
+So a proper auto-scaling mechanism would need to understand the configuration of topics and partitions and their load to properly scale the broker.
+But the `scale` sub-resource would make it easier to develop such solutions in the future.
+
+### KRaft support
+
+To support the Kafka KRaft mode, a new `roles` filed will be added to the `KafkaNodePool`.
```
tombentley, line 326:
```suggestion
To support the Kafka KRaft mode, a new `roles` field will be added to the `KafkaNodePool`.
```

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,876 @@
...
+    - 1
+    - 2
+  clusterId: cABpUKeDTqi7q_XOtBMgUv
+  # replicas and labelSelector fields are needed for the scale subresource
+  replicas: 3
+  labelSelector: >-
+    strimzi.io/cluster=my-cluster,strimzi.io/name=m-cluster-big-nodes,strimzi.io/kind=Kafka
+```
+
+Users will be able to use annotations to indicate to the operator what should be the next ID when scaling up or what should be the node ID which should be removed when scaling down through annotations.
```
tombentley, line 540: Which resources will these annotations be on? I assume it's the node pools themselves, but it's best to be explicit.

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,876 @@
...
+Users can store multiple YAML documents in a single file and apply them with a single command.
+But their creation is not a single operation.
+It can therefore happen that the operator first sees only the `KafkaNodePool` resources but not the `Kafka` resource or the other way around.
+The basic situations can be handled by the operator code.
+But it is not completely clear how many problems and various race conditions this might be causing when updating the node pools for existing clusters etc.
+
+The eventual consistency also shows up in situations when you for example decide to scale two of your node pools.
+The scale events will be independent and will almost never be caught by the same reconciliation.
+So you will have one reconciliation to scale the first pool followed by the second to scale the second one.
+
```
tombentley, line 767: I guess we should be clear in this section that it's up to the users of the operator to take account of this. We should also mention that the Strimzi documentation will reflect this, for example, by procedures changing individual resources and waiting for readiness, so that the CR changes in any procedure have deterministic serialization.

050-Kafka-Node-Pools.md
```diff
@@ -0,0 +1,876 @@
...
+            size: 1000Gi
+```
+
+This would have some advantages:
+
+* Having a single resource is easier to work with as it has a clear consistency.
+  Using multiple resources means you have to deal with Kubernetes's eventual consistency when updating multiple resources at the same time
+
+And of course also some disadvantages:
+* It would not possible to support the `scale` sub-resource and support autoscaling in the future.
```
tombentley, line 857:
```suggestion
* It would not be possible to support the `scale` sub-resource and support autoscaling in the future.
```

tombentley (approved):
Left a few minor comments, but LGTM. Thanks Jakub!

scholzj:
> By Operations I meant, for example, only rolling brokers in a certain node pool, or fixing a disk skew across given single or multiple node pools, any such operations that can be performed currently across all brokers, be restricted to a NodePool.

@rareddy In general, these kinds of operations focus on the underlying Kubernetes resources and not on our CRs when possible. (It is for example not possible with the `KafkaConnector` or `KafkaRebalance` resources which do not have any representation in Kubernetes resources). So for example to trigger manual rolling update, you annotate the corresponding PodSet. So these operations will be supported based for a specific node pool only (it is actually already done in the prototype). But this is done through the internal representation in different StrimziPodSet resources which will be annotated. So it does not depend on the actual representation in the Strimzi `Kafka` or `KafkaNodePool` custom resources.

scholzj:
This proposal now has 5 +1 votes (approvals) from maintainers (+ my own as an author). Thanks everyone for the comments. 

If anyone has any more comments or votes, please do so till Friday. Otherwise if there is nothing new I will close the vote over the weekend.

fvaleri (approved):
LGTM.

see-quick (approved):
LGTM. Thanks for this proposal, Jakub!

PR #78, opened 2023-05-11
Update to  proposal #48

tinaselenge:
When implementing the second part of the proposal, I realised the order in the current flow is slightly different, therefore the proposed changes would look different as well. However, the core of the proposal is still the same so is the new logic for handling the broker in log recovery. The diagrams describing the flows have been updated accordingly.


048-avoid-broker-restarts-when-in-recovery.md
```diff
@@ -65,24 +69,19 @@ The proposed flow of KafkaRoller
...
 
-As described in the graph above, the following steps are added before the retry step:
+As described in the graph above, the following steps are added after checking the connection to the broker:
 
-1. Ensure that the Kafka pod is not in a stuck state. The existing `isPodStuck()` can be used for this. If the pod is stuck, continue with the current behaviour to retry rolling of the pod.
-2. Use the Pod DNS name to connect to the KafkaAgent API endpoint. If it cannot connect to the endpoint, continue with the current behaviour to retry rolling of the pod. (The connection might timeout or refused because KafkaAgent is stuck or running older version).
+1. If the AdminClient connection failed, check the broker state instead of force restarting the pod.
+2. Use the Pod DNS name to connect to the KafkaAgent API endpoint. If it cannot connect to the endpoint (the connection might timeout or refused because KafkaAgent is stuck or running older version), continue with the current behaviour.
```
scholzj, line 75: Isn't it easier to check if the broker is running but not ready and then check the Agent instead of trying to connect with Admin API first??

tinaselenge: Yes that could be simpler. We could check the broker state after the initial wait for readiness times out (https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java#L371). That would mean pod is running but not ready. If broker is in log recovery after the timeout, throw an exception to skip on rest of the checks. Then next time we retry the pod, again would wait for readiness and check the broker state after the timeout. 

Initially I was going to check the broker state here https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/KafkaRoller.java#L512 but I think what you suggested makes more sense.

048-avoid-broker-restarts-when-in-recovery.md
```diff
@@ -4,17 +4,21 @@ This proposal describes a solution for KafkaRoller's shortcoming with brokers in
 
 ## Current situation
 
-The current logic in KafkaRoller:
+The current logic in KafkaRoller from high level:
```
PaulRMellor, line 7:
```suggestion
The following list outlines the high-level logic currently implemented in `KafkaRoller` for rolling pods
```

048-avoid-broker-restarts-when-in-recovery.md
```diff
@@ -4,17 +4,21 @@ This proposal describes a solution for KafkaRoller's shortcoming with brokers in
...
+- If rolling the pod does affect the availibility, then restart the pod and wait for pod readiness.
+- If pod does not became ready within the operational timeout (300000ms by default) after restarting, then retry until the maximum attempt is reached. 
 
 The current flow of KafkaRoller
 ![The current flow of KafkaRoller](images/048-kafka-roller-current-flow.png)
 
-A Kafka broker can take a long time to become ready while performing log recovery. In this case, KafkaRoller could end up force restarting the broker and this can continue indefinitely, because of the log recovery on the broker startup again. KafkaRoller currently does not have a way to know what state a broker is in before making a decision to force restart it.
+A Kafka broker can take a long time to become ready while performing log recovery. While broker is in log recovery, AdminClient connection cannot be made to the broker. In this case, KafkaRoller force restarts the broker and this can continue indefinitely, because of the log recovery on the broker startup again. KafkaRoller currently does not have a way to know what state a broker is in before making a decision to force restart it.
```
PaulRMellor, line 21:
```suggestion
A Kafka broker can take a long time to become ready while performing log recovery. While the broker is in log recovery, AdminClient connection cannot be made to the broker. In this case, KafkaRoller force restarts the broker and this can continue indefinitely, because of the log recovery on the broker startup again. KafkaRoller currently does not have a way to know what state a broker is in before making a decision to force restart it.
```

048-avoid-broker-restarts-when-in-recovery.md
```diff
@@ -65,24 +69,19 @@ The proposed flow of KafkaRoller
...
-
-7. Set `inLogRecovery` to true for the pod  (`inLogRecovery` is a new boolean that would be added to the `RestartContext`).
-8. Log the values of `remainingLogsToRecover` and `remainingSegmentsToRecover` from the JSON response in an `INFO` message.
-9. Continue with the current behaviour to retry rolling of the pod.
-10. On the next retry, check `inLogRecovery` inside the `restartIfNecessary()` function.
-11. When `inLogRecovery` is set to true, call `awaitReadiness()` which periodically checks and waits for readiness until the operational timeout reaches.
-13. If the operational timeout is reached, repeat from step1.
-14. If the maximum number of attempts has been reached, include log recovery in the reason for failure.
+5. If the HTTP return code is not 200, continue with the current behaviour.
+6. If the `brokerState` value is `2` (`RECOVERY`)
```
PaulRMellor, line 79:
```suggestion
6. If the `brokerState` value is `2` (`RECOVERY`):
```

048-avoid-broker-restarts-when-in-recovery.md
```diff
@@ -65,24 +69,19 @@ The proposed flow of KafkaRoller
...
+     - Log the values of `remainingLogsToRecover` and `remainingSegmentsToRecover` from the JSON response in an `INFO` message.
+     - Call `awaitReadiness()` which periodically checks and waits for readiness until the operational timeout reaches.
+7.  If the `brokerState` value is not `2` (`RECOVERY`), continue with the current behaviour of force restarting the pod and waiting for readiness until the operational timeout reaches.
```
PaulRMellor, line 82:
```suggestion
7. If the `brokerState` value is not `2` (`RECOVERY`), continue with the current behaviour of force restarting the pod and waiting for readiness until the operational timeout reaches.
```

048-avoid-broker-restarts-when-in-recovery.md
```diff
@@ -65,24 +69,19 @@ The proposed flow of KafkaRoller
...
+8. If the operational timeout is reached, continue with the current behaviour of rescheduling the rolling of the pod. This will result in repeating from step 1 if broker is still in log recovery.
```
PaulRMellor, line 83:
```suggestion
8. If the operational timeout is reached, continue with the current behaviour of rescheduling the rolling of the pod. This will result in repeating from step 1 if the broker is still in log recovery.
```

PaulRMellor (approved):
Thanks. Looks good. Left a couple of suggestions

scholzj (approved):
Thanks for the update @tinaselenge

048-avoid-broker-restarts-when-in-recovery.md
```diff
@@ -1,20 +1,25 @@
 # Avoid broker restart when in log recovery state
+> Note: This proposal has been updated since the time it was merged.
```
tombentley, line 2:
```suggestion
> Note: This proposal has been updated since its initial merge.
```

048-avoid-broker-restarts-when-in-recovery.md
```diff
@@ -1,20 +1,25 @@
...
 ## Current situation
 
-The current logic in KafkaRoller:
+The following list outlines the high-level logic currently implemented in `KafkaRoller` for rolling pods:
 - Take a list of Kafka pods and reorder them based on the readiness.
-- Roll unready pods first without considering why it could be unready.
 - If the pod is the controller, roll it last.
-- Do not roll the pod, if it has an impact on the availability such as causing under replicated partitions.
-- If a Kafka pod does not become ready within the operational timeout (300000ms by default) after restarting, then it may force restart it without considering why it's taking a long time.
+- If pod is not ready before rolling, wait for it to become ready until the operation timeout reaches.
```
tombentley, line 11:
```suggestion
- If pod is not ready before rolling, wait for it to become ready until the operation timeout is reached.
```

048-avoid-broker-restarts-when-in-recovery.md
```diff
@@ -1,20 +1,25 @@
...
+- If pod is in a stuck state e.g CrashLoopBackOff, force restart the pod.
```
tombentley, line 12:
```suggestion
- If pod is in a stuck state e.g `CrashLoopBackOff`, force restart the pod.
```

048-avoid-broker-restarts-when-in-recovery.md
```diff
@@ -1,20 +1,25 @@
...
+- If pod is unschedulable, fail the reconcilation. 
+- If the AdminClient connection to pod failed, force restart the pod. 
+- Check if rolling the pod would have an impact on the availability such as causing under replicated partitions.
+- If rolling the pod does affect the availibility, then restart the pod and wait for pod readiness.
```
tombentley, line 16: A missing not here?

```suggestion
- If rolling the pod does not affect the availibility, then restart the pod and wait for pod readiness. Otherwise consider other pods for rolling first.
```

tombentley (commented):
These modifications make sense, but it looks like the maximum time we might wait for a broker that's in log recover is determined by the same configuration used for the other cases. Maybe I misunderstood, but if not, perhaps we want to make the log recovery timeout a separate thing than the timout used in the cases where the pod is not ready or is unreachable.

tinaselenge:
> These modifications make sense, but it looks like the maximum time we might wait for a broker that's in log recover is determined by the same configuration used for the other cases. Maybe I misunderstood, but if not, perhaps we want to make the log recovery timeout a separate thing than the timeout used in the cases where the pod is not ready or is unreachable.

@tombentley  that's right. I remember this being mentioned on the initial PR for the proposal and the agreement was to use the existing operation timeout for the log recovery rather than a separate timeout because the log recovery could take from minutes to hours and it's hard to know what the default or configured timeout should be.

ppatierno (approved):
LGTM. Thanks!

tinaselenge:
Thanks everyone for reviewing the PR!

Frawless (approved):

im-konge (approved):

tinaselenge:
Hi, @scholzj @tombentley if there is no further comments, can this be merged now please?

tombentley:
> the agreement was to use the existing operation timeout for the log recovery rather than a separate timeout because the log recovery could take from minutes to hours and it's hard to know what the default or configured timeout should be.

Just because we don't know what the default or configured timeout should be is not an argument for not having a separate configuration parameter for it. The user won't initially know either, but will be able to bump up the `logRecoveryTimeout` based on experience they gain while using Strimzi. The point is that by giving them a separate parameter they can express their willingness to wait 12 hours for log recovery, but only 5 minutes for an unreachable pod.

scholzj:
@tombentley My understanding is that the timeout you are talking about is about how long we hang there before essentially failing the reconciliation. I.e. it is the same as when waiting for readiness in other places -> we will not kill the pod after the operationsTimeout, but just failed the reconciliation and retry in next reconciliation. If that is the case, I do not see why there should be a custom timeout because it is not about how long we want before killing it but essentially in how long blocks are we going to wait and I do not think this should be configurable separately. 

Or did I get it wrong and we plan to kill the pod after the timeout? I thought we want to avoid that as the main purpose of the proposal.

tinaselenge:
> Just because we don't know what the default or configured timeout should be is not an argument for not having a separate configuration parameter for it. The user won't initially know either, but will be able to bump up the `logRecoveryTimeout` based on experience they gain while using Strimzi. The point is that by giving them a separate parameter they can express their willingness to wait 12 hours for log recovery, but only 5 minutes for an unreachable pod.

I agree, and perhaps that was not the right reason. I think we discussed that since reconciliation loop will repeatedly check the broker state (and log) and wait, configuring a separate timeout was not necessary. Instead of blocking a single reconciliation for 12 hours, we would let it fail after the operation timeout and repeat the process until the broker recovery completes or the human operator takes action.

tinaselenge:
> @tombentley My understanding is that the timeout you are talking about is about how long we hang there before essentially failing the reconciliation. I.e. it is the same as when waiting for readiness in other places -> we will not kill the pod after the operationsTimeout, but just failed the reconciliation and retry in next reconciliation. If that is the case, I do not see why there should be a custom timeout because it is not about how long we want before killing it but essentially in how long blocks are we going to wait and I do not think this should be configurable separately.
> 
> Or did I get it wrong and we plan to kill the pod after the timeout? I thought we want to avoid that as the main purpose of the proposal.

You are correct, we  would only fail the reconciliation, not kill the pod after the timeout.

048-avoid-broker-restarts-when-in-recovery.md
```diff
@@ -63,26 +68,20 @@ The proposed flow of KafkaRoller
...
+1. Wait for pod to become ready if it's not in ready state before checking if it can be rolled.
+2. If pod does not become ready within the operation timeout, check the broker state.
+3. Use the Pod DNS name to connect to the KafkaAgent API endpoint. If it cannot connect to the endpoint (the connection might timeout or refused because KafkaAgent is stuck or running older version), continue with the current behaviour.
+4. Use the cluster operator's client certificate and keys for mTLS authentication.
+5. Send `GET v1/broker-state` request to the endpoint.
+6. If the HTTP return code is not 200, continue with the current behaviour.
+7. If the `brokerState` value is `2` (`RECOVERY`):
+     - Log the values of `remainingLogsToRecover` and `remainingSegmentsToRecover` from the JSON response in an `INFO` message.
+     - Return an exception to repeat from step 1 until the maximum number of attempts is reached.
+     - If the maximum number of attempts has been reached, include log recovery in the reason for failure.
```
tombentley, line 83:
```suggestion
     - If the maximum number of attempts has been reached then wait the reconciliation including log recovery in the reason for failure.
```

tombentley (approved):
Thanks for the clarification about what "failure" means, let's add it to the text.

see-quick (approved):
LGTM.

PR #82, opened 2023-07-07
Added proposal to get KRaft controller configuration via Kafka Agent

ppatierno:
This proposal is about providing a way for the Kafka Roller to get `controller` node configuration (when in KRaft mode) in order to have the ability to roll nodes (on configuration changes) and moving forward with the Zookeeper to KRaft migration support.
This aims to be a temporary solution when the main problems will be address in the Kafka project upstream as described in the proposal itself.

### DIFF
--- a/056-kafka-agent-controller-configuration.md
+++ b/056-kafka-agent-controller-configuration.md
@@ -73,6 +73,8 @@ We could think about having a dedicated additional parameter for that, but what 
 
 During the Kafka start up script, the node configuration is generated and saved on the disk in the `/tmp/strimzi.properties` file.
 As soon as the Kafka Agent runs, it can read this file and load its content in memory in a JSON format.
+The top level JSON value is an object with all string-typed values (i.e. even if they are numbers).
+Following, an example of the JSON object bringing the `controller` configuration.
 
 ```json
 {
--- a/056-kafka-agent-controller-configuration.md
+++ b/056-kafka-agent-controller-configuration.md
@@ -1,0 +1,116 @@
+# KRaft controller configuration via Kafka Agent
+
+This proposal is about enabling the current Kafka Agent to get the configuration from KRaft `controller` nodes and make it available to the Kafka Roller in the Strimzi Cluster Operator.
+This way the Kafka Roller can argue about the need to roll the `controller` nodes on configuration changes.
+
+## Current situation
+
+The Kafka Agent is a Java agent, running alongside the Kafka process, which provides the following features:
+
+* check that the broker is connected to ZooKeeper and it is also in a valid running state and make this information available to the Kubernetes platform via the liveness and readiness probes (it creates some specific files on the broker disk).
+* expose an HTTP endpoint, securely reachable by the Kafka Roller, to provide information about the current broker state and ongoing log recovery operations (more details in the proposal [#48](https://github.com/strimzi/proposals/blob/main/048-avoid-broker-restarts-when-in-recovery.md)).
+
+With the above features, the Kubernetes platform can deal with restarting the Kafka container if it's not ready or alive and the Kafka Roller can roll the Kafka brokers depending on their state.
+
+Furthermore, the Kafka Roller is already able to connect to Kafka brokers in order to get their configuration and check if they need an update, dynamically or by rolling them, in order to get a new configuration.
+
+As today, the Kafka Agent is not started when the cluster is deployed in KRaft mode.
+
+## Motivation
+
+The current Kafka Agent, together with the Kafka Roller, are good for dealing with updating and/or rolling Kafka brokers when working with ZooKeeper-based deployments.
+With Kafka moving to KRaft and the community deprecating ZooKeeper, the Strimzi project has to provide the best experience to the users for a seamlessly migration.
+As today, the Strimzi Cluster Operator already supports the KRaft mode (behind a feature gate).
+In KRaft mode, nodes can have `controller` or `broker` (or both) roles, but currently the Kafka Roller is able to connect to `broker`(s) only when it comes to get the configuration and arguing about dynamic updates and/or rolling.
+The current Kafka implementation of the `controller` role has some limitations instead:
+
+* it doesn't support the `METADATA` API, so it's not possible to specify a `controller` address for bootstrapping a Kafka Admin client (as it is used in the Kafka Roller for the `broker`(s))
+* it doesn't support the `DESCRIBE_CONFIGS` API, so even if an Admin client was able to connect, it cannot retrieve the current `controller` configuration (as it is available in the Kafka Roller for the `broker`(s))
+* unless hardcoded in the Kafka codebase, it's not well made available which configuration properties are pure `broker`, pure `controller` or applicable to both roles.
+
+Due to the above limitations, the Kafka Roller is not able to connect to a `controller` node by using a Kafka Admin client instance, retrieving the current configuration and dealing with a dynamic update and/or rolling.
+
+> Despite the missing support for the above APIs, the `controller` role supports `INCREMENTAL_ALTER_CONFIGS` API, so assuming the Kafka Roller was able to connect, getting configuration and dealing with changes, it would be able to update it.
+
+The best solution would have above limitations being addressed in the Kafka upstream project.
+The [KIP-919](https://cwiki.apache.org/confluence/display/KAFKA/KIP-919%3A+Allow+AdminClient+to+Talk+Directly+with+the+KRaft+Controller+Quorum) aims to address the first one, enabling a Kafka Admin client to connect directly to a `controller` node.
+Once connected, getting the configuration from the `controller` node is not part of the KIP so something to discuss in the upstream community (maybe raising it in the KIP-919 or opening a new one).
+The problem related to the `controller` only configuration properties is not taken into account yet.
+
+While the KRaft mode is already defined as "production ready" and the migration process from ZooKeeper-mode is getting improvements and stability, the Strimzi project has to keep the pace in order to provide the best experience to the users.
+Having the best support and fix the issues in the Kafka upstream project could take long time, maybe not landing on time in the next Kafka 3.6.0 release.
+For this reason, a temporary solution, built into the Strimzi project, could unlock the next chunk of work for supporting an automatic ZooKeeper-mode to KRaft migration within the Strimzi Cluster Operator.
+
+## Proposal
+
+### Running Kafka Agent when KRaft mode enabled
+
+As today, the Kafka Agent doesn't run when KRaft mode is enabled (based on a condition in the `kafka_run.sh` bash script).
+On one side, it makes sense because for Kubernetes liveness and readiness probes we are using a different approach for KRaft (more details on proposal [#46](https://github.com/strimzi/proposals/blob/main/046-kraft-liveness-readiness.md)) and not creating some files on the brokers.
+On the other side, we are not providing the broker and log recovery status to the Kafka Roller when KRaft mode is enabled.
+
+The Kafka Agent gets some parameters (separated by `:`).
+
+```shell
+KAFKA_OPTS="${KAFKA_OPTS} -javaagent:$(ls "$KAFKA_HOME"/libs/kafka-agent*.jar)=/var/opt/kafka/kafka-ready:/var/opt/kafka/zk-connected:$KEY_STORE:$CERTS_STORE_PASSWORD:$TRUST_STORE:$CERTS_STORE_PASSWORD"
+```
+
+The first two are the paths to the files to be created on the file system to set the Kafka broker as ready and connected to ZooKeeper (used by the Kubernetes liveness and readiness probes).
+The rest are key and certificate stores (with related password), to allow the Kafka Roller to connect to the Kafka Agent via HTTPS.
+The first two are not needed when in KRaft mode so they could be passed as the string `"null"`.
+
+```shell
+KAFKA_OPTS="${KAFKA_OPTS} -javaagent:$(ls "$KAFKA_HOME"/libs/kafka-agent*.jar)=null:null:$KEY_STORE:$CERTS_STORE_PASSWORD:$TRUST_STORE:$CERTS_STORE_PASSWORD"
+```
+
+This would be recognized by the Kafka Agent as a "running cluster in KRaft mode".
+This way the Kafka Agent would not take care of handling these files for the Kubernetes liveness and readiness probes in ZooKeeper-mode.
+We could think about having a dedicated additional parameter for that, but what we aim with this proposal is anyway a temporary solution and also we are moving away from ZooKeeper, so at some point the first two parameters would be totally removed.
+
+### Exposing KRaft `controller` configuration
+
+During the Kafka start up script, the node configuration is generated and saved on the disk in the `/tmp/strimzi.properties` file.
+As soon as the Kafka Agent runs, it can read this file and load its content in memory in a JSON format.
+
+```json
+{
+    "broker.id": "0",
+    "node.id": "0",
+    "process.roles": "controller",
+    "controller.listener.name": "CONTROL_PLANE-9090",
+    ...
+    ...
+}
+```
+
+Such a configuration is returned through the `/v1/node-configuration` HTTP endpoint.
+Because of this file never changing, even if the node has its configuration dynamically updated, there is no need for the Kafka Agent to watch the file.
+It is enough to load its content on the Kafka Agent startup and keep it as cached in memory.
+The same content will be just returned every time the Kafka Roller is asking for it.
+
+### Kafka Roller usage
+
+The usage of the `controller` node configuration in the Kafka Roller is not a goal of this proposal, but for better understanding it is useful to mention it.
+
+The Kafka Roller queries the Kafka Agent to get the `controller` node configuration in JSON format via HTTPS on the `/v1/node-configuration` endpoint.
+It also has an hardcoded list of the `controller` only configuration properties (which has to be defined taking a look at the Kafka codebase).
+When there is a change in the `Kafka` custom resource, the Kafka Roller checks if some of the `controller` specific parameters are changed, between the desired configuration in the custom resource and the one got from the Kafka Agent.
+If there is any difference, the Kafka Roller just rolls the `controller` node.
+
+## Affected/not affected projects
+
+The main affected project is the Kafka Agent one.
+Also the bash script for running Kafka is affected in order to start the Kafka Agent when KRaft mode is enabled.
+The Kafka Roller is not impacted because the part related to it using the information exposed by the Kafka Agent is not a goal of this proposal.
+
+## Compatibility
+
+This proposal is not going to break any backward compatibility.
+The Kafka Agent will run in KRaft mode as well but not handling the files used by the Kubernetes platform for checking liveness and readiness of the Kafka broker.
+It will still provide broker and log recovery status on the corresponding HTTP endpoint.
+It is going to add one more HTTP endpoint for providing `controller` node configuration.
+
+## Rejected alternatives
+
+Waiting for the missing APIs related issues to be addressed in the Kafka upstream project first and then moving forward with the work on the automated migration within Strimzi.
+This would push the migration support from ZooKeeper-mode to KRaft too far in the future.
--- a/056-kafka-agent-controller-configuration.md
+++ b/056-kafka-agent-controller-configuration.md
@@ -1,7 +1,7 @@
 # KRaft controller configuration via Kafka Agent
 
 This proposal is about enabling the current Kafka Agent to get the configuration from KRaft `controller` nodes and make it available to the Kafka Roller in the Strimzi Cluster Operator.
-This way the Kafka Roller can argue about the need to roll the `controller` nodes on configuration changes.
+This way the Kafka Roller can determine the need to roll the `controller` nodes on configuration changes.
 
 ## Current situation
 
@@ -20,26 +20,28 @@ As today, the Kafka Agent is not started when the cluster is deployed in KRaft m
 
 The current Kafka Agent, together with the Kafka Roller, are good for dealing with updating and/or rolling Kafka brokers when working with ZooKeeper-based deployments.
 With Kafka moving to KRaft and the community deprecating ZooKeeper, the Strimzi project has to provide the best experience to the users for a seamlessly migration.
-As today, the Strimzi Cluster Operator already supports the KRaft mode (behind a feature gate).
-In KRaft mode, nodes can have `controller` or `broker` (or both) roles, but currently the Kafka Roller is able to connect to `broker`(s) only when it comes to get the configuration and arguing about dynamic updates and/or rolling.
+Currently, the Strimzi Cluster Operator already supports the KRaft mode (behind a feature gate).
+In KRaft mode, each node's `process.roles` must be one of `controller` or `broker` or `controller,broker`, but currently the Kafka Roller is only able to get the configuration of nodes having the `broker` role.
+The current configuration is needed to determine whether any action is necessary and if so whether a dynamic configuration update is sufficient, or whether restarting the node is required.
 The current Kafka implementation of the `controller` role has some limitations instead:
 
 * it doesn't support the `METADATA` API, so it's not possible to specify a `controller` address for bootstrapping a Kafka Admin client (as it is used in the Kafka Roller for the `broker`(s))
 * it doesn't support the `DESCRIBE_CONFIGS` API, so even if an Admin client was able to connect, it cannot retrieve the current `controller` configuration (as it is available in the Kafka Roller for the `broker`(s))
-* unless hardcoded in the Kafka codebase, it's not well made available which configuration properties are pure `broker`, pure `controller` or applicable to both roles.
+* furthermore beyond simply reading the Kafka source code there's no way of knowing (either via the Kafka Admin client, or otherwise) which configuration properties are pure `broker`, pure `controller` or applicable to both roles.
 
 Due to the above limitations, the Kafka Roller is not able to connect to a `controller` node by using a Kafka Admin client instance, retrieving the current configuration and dealing with a dynamic update and/or rolling.
 
 > Despite the missing support for the above APIs, the `controller` role supports `INCREMENTAL_ALTER_CONFIGS` API, so assuming the Kafka Roller was able to connect, getting configuration and dealing with changes, it would be able to update it.
+> For this proposal we will not be supporting the dynamic reconfiguration of pure `controller` nodes.
 
 The best solution would have above limitations being addressed in the Kafka upstream project.
 The [KIP-919](https://cwiki.apache.org/confluence/display/KAFKA/KIP-919%3A+Allow+AdminClient+to+Talk+Directly+with+the+KRaft+Controller+Quorum) aims to address the first one, enabling a Kafka Admin client to connect directly to a `controller` node.
-Once connected, getting the configuration from the `controller` node is not part of the KIP so something to discuss in the upstream community (maybe raising it in the KIP-919 or opening a new one).
+Once connected, getting the configuration from the `controller` node is not currently part of the KIP so something to discuss in the upstream community (maybe raising it in the KIP-919 or opening a new one).
 The problem related to the `controller` only configuration properties is not taken into account yet.
 
 While the KRaft mode is already defined as "production ready" and the migration process from ZooKeeper-mode is getting improvements and stability, the Strimzi project has to keep the pace in order to provide the best experience to the users.
 Having the best support and fix the issues in the Kafka upstream project could take long time, maybe not landing on time in the next Kafka 3.6.0 release.
-For this reason, a temporary solution, built into the Strimzi project, could unlock the next chunk of work for supporting an automatic ZooKeeper-mode to KRaft migration within the Strimzi Cluster Operator.
+For this reason, a temporary solution, built into the Strimzi project, will unlock the next chunk of work for supporting an automatic ZooKeeper-mode to KRaft migration within the Strimzi Cluster Operator.
 
 ## Proposal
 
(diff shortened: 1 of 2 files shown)


ppatierno:
FYI @tinaselenge I will add you as reviewer as soon as you accept the invite to the "contributors" team I sent to you ;-)

scholzj (commented):
Should this instead cover the whole issue of rolling controllers? It is not just the configuration bt also finding out if they are in sync etc.

ppatierno:
> Should this instead cover the whole issue of rolling controllers? It is not just the configuration bt also finding out if they are in sync etc.

Can you elaborate a little bit more please? 
Do you mean exposing some KRaft related metrics (https://kafka.apache.org/documentation/#kraft_monitoring) to the Kafka Roller so it can argue if the metadata topic is replicated/sync on all the controllers?

scholzj:
Yes, I think those are the other parts missing.

052-kafka-agent-controller-configuration.md
```diff
@@ -0,0 +1,116 @@
...
+Also the bash script for running Kafka is affected in order to start the Kafka Agent when KRaft mode is enabled.
+The Kafka Roller is not impacted because the part related to it using the information exposed by the Kafka Agent is not a goal of this proposal.
+
+## Compatibility
+
+This proposal is not going to break any backward compatibility.
+The Kafka Agent will run in KRaft mode as well but not handling the files used by the Kubernetes platform for checking liveness and readiness of the Kafka broker.
+It will still provide broker and log recovery status on the corresponding HTTP endpoint.
+It is going to add one more HTTP endpoint for providing `controller` node configuration.
+
```
tombentley, line 112: Is it worth mentioning what will happen if (or what prevents) the operator talking to an older kafka image where the endpoint is not present (and so gets a 404)? I don't think there's anything we can reasonably do in this case. Failing the reconciliation with a suitable error message would be enough.

tombentley (commented):
Thanks Paolo. I left lots of little nits, but overall this accurately describes what we've discussed. I agree with Jakub that we should specify the changes in the `KafkaRoller` in this proposal too, so that readers can understand how the complete solution will work. We should also think a bit about what happens if, for some reason, we cannot get the config from the endpoint: Does the roller restart the node immediately, or eventually?

ppatierno:
> I agree with Jakub that we should specify the changes in the KafkaRoller in this proposal too, so that readers can understand how the complete solution will work

@tombentley I have got something different from Jakub comment. I mean ... you are talking about the "changes in the KafkaRoller" and I think I covered them (as an overall picture) in the "Kafka Roller usage" section.
Jakub is talking about exposing information about how much in sync are the controllers, so not just configuration but some additional metrics (i.e. high-watermark, log end offset, ...) related to metadata topic that the Kafka Roller can use to argue about the need for rolling.

052-kafka-agent-controller-configuration.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+This would be recognized by the Kafka Agent as a "running cluster in KRaft mode".
+This way the Kafka Agent would not take care of handling these files for the Kubernetes liveness and readiness probes in ZooKeeper-mode.
+We could think about having a dedicated additional parameter for that, but what we aim with this proposal is anyway a temporary solution and also we are moving away from ZooKeeper, so at some point the first two parameters would be totally removed.
+
+### Exposing KRaft `controller` configuration
+
+During the Kafka start up script, the node configuration is generated and saved on the disk in the `/tmp/strimzi.properties` file.
+As soon as the Kafka Agent runs, it can read this file and load its content in memory in a JSON format.
```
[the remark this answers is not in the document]
ppatierno: > top level JSON value would always be an object

Do you mean having ...

```json
{
    "config": {
          "broker.id": "0",
          ....
    }
}
```

or is it fine with you using the no-named root JSON object containing the configuration fields directly?
Or maybe I mislead and you just mean that I have to explain in the text what I have got as an example :-)

052-kafka-agent-controller-configuration.md
```diff
@@ -0,0 +1,116 @@
...
+Because of this file never changing, even if the node has its configuration dynamically updated, there is no need for the Kafka Agent to watch the file.
+It is enough to load its content on the Kafka Agent startup and keep it as cached in memory.
+The same content will be just returned every time the Kafka Roller is asking for it.
+
+### Kafka Roller usage
+
+The usage of the `controller` node configuration in the Kafka Roller is not a goal of this proposal, but for better understanding it is useful to mention it.
+
+The Kafka Roller queries the Kafka Agent to get the `controller` node configuration in JSON format via HTTPS on the `/v1/node-configuration` endpoint.
+It also has an hardcoded list of the `controller` only configuration properties (which has to be defined taking a look at the Kafka codebase).
```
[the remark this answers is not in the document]
ppatierno: Yeah sure, I will fix it.

052-kafka-agent-controller-configuration.md
```diff
@@ -0,0 +1,116 @@
...
+When there is a change in the `Kafka` custom resource, the Kafka Roller checks if some of the `controller` specific parameters are changed, between the desired configuration in the custom resource and the one got from the Kafka Agent.
```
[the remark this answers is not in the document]
ppatierno: I don't know the roller well enough, but I would say yes. I will double check in the code and add more info here.

052-kafka-agent-controller-configuration.md
ppatierno: Yeah I just missed that written down, but it was in my notes ... 404 was the idea.

052-kafka-agent-controller-configuration.md
```diff
@@ -0,0 +1,116 @@
...
+
+The Kafka Agent gets some parameters (separated by `:`).
+
+```shell
+KAFKA_OPTS="${KAFKA_OPTS} -javaagent:$(ls "$KAFKA_HOME"/libs/kafka-agent*.jar)=/var/opt/kafka/kafka-ready:/var/opt/kafka/zk-connected:$KEY_STORE:$CERTS_STORE_PASSWORD:$TRUST_STORE:$CERTS_STORE_PASSWORD"
+```
+
+The first two are the paths to the files to be created on the file system to set the Kafka broker as ready and connected to ZooKeeper (used by the Kubernetes liveness and readiness probes).
+The rest are key and certificate stores (with related password), to allow the Kafka Roller to connect to the Kafka Agent via HTTPS.
+The first two are not needed when in KRaft mode so they could be passed as the string `"null"`.
```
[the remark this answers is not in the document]
ppatierno: I am totally fine with this, not a strong opinion (even because anyway it's temporary :-)).

052-kafka-agent-controller-configuration.md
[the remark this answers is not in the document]
tombentley: I'm fine with 

```json
{
    "broker.id": "0",
    "node.id": "0",
    "process.roles": "controller",
    "controller.listener.name": "CONTROL_PLANE-9090",
    ...
    ...
}
```

my point was that the values are strings any not another other JSON type, even when the Kafka type would be int, long, list etc.

tombentley (commented):
> Jakub is talking about exposing information about how much in sync are the controllers, so not just configuration but some additional metrics (i.e. high-watermark, log end offset, ...) related to metadata topic that the Kafka Roller can use to argue about the need for rolling.

Right. So you should today be able to use `Admin.describeMetadataQuorum()` to get the current leader (`QuorumInfo.leaderId()`) and other information about the quorum. I don't think it's the lag of the followers that we're interested in though. The lag could be >0 but the follower is still functioning. I think what we're interested in is `QuorumInfo.ReplicaState.lastCaughtUpTimestamp()`. When this is less than [controller.quorum.fetch.timeout.ms](https://kafka.apache.org/documentation/#brokerconfigs_controller.quorum.fetch.timeout.ms) (from the active controller's PoV) then the follower is "still a member of the cluster".  We need a precondition that prevents restarting a controller node while:

1. Any of the expected controller nodes are not in the voters, and
2. Any of them have `leader.lastCaughtUpTimestamp() - node.lastCaughtUpTimestamp() > `valueOf("controller.quorum.fetch.timeout.ms")`

I think that would be enough to prevent loss of quorum during a rolling restart (right @fvaleri, @mimaison, @showuon?)

The fly in the ointment with this is it currently requires at least one live broker node (which will forward the request to the active controller). KIP-919 would change that. Equivalent information doesn't seem to be available from the documented metrics. So we'd need to specifically have handling for this no-broker possibility, even if it was more heuristic (e.g. "wait at least N minutes once all controller nodes seem to be running). It is less than ideal, but I don't see a ready alternative right now. In practice we don't expect anyone running in production to end up in a state where they have no brokers up and are needing to restart controllers. And if they are then by definition they have no broker availability anyway, so temporary loss of quorum would not be as much of a problem as it would normally be. Unless @fvaleri, @mimaison, @showuon have no better compromises?

fvaleri:
@tombentley 

> 1. Any of the expected controller nodes are not in the voters

I think we can relax a bit saying "majority + 1 of the expected voters". Wdyt?

> 2. Any of them have leader.lastCaughtUpTimestamp() - node.lastCaughtUpTimestamp() > valueOf("controller.quorum.fetch.timeout.ms")`

Should be greater than or equal?

> don't see a ready alternative right now.

The KafkaRoller knows if the cluster has no live broker node right? So it should simply block any controller restart. I would be the 0 condition in your list.

showuon:
>  1.  Any of the expected controller nodes are not in the voters, and

I don't know what kind of situation this is. The `voters` list returned from `DESCRIBE_QUORUM` is from the voter config, so I don't understand what `the expected controller nodes are not in the voters` mean. Could you elaborate more on it?
 
>  2. Any of them have leader.lastCaughtUpTimestamp() - node.lastCaughtUpTimestamp() > valueOf("controller.quorum.fetch.timeout.ms")

I think this is too strict. `leader.lastCaughtUpTimestamp() - node.lastCaughtUpTimestamp() > valueOf("controller.quorum.fetch.timeout.ms")` means the follower node might have only 1 chance to fetch data from the leader, which might still not be able to catch up with the leader. 

On the other hand, what do we expect for restarting the slow follower? Suppose this following is just slow due to network, we might not interrupt it under this case, right? So, maybe we restart the node only when it is stuck, like this:
`LogEndOffset of the node is not increasing AND leader.lastCaughtUpTimestamp() - node.lastCaughtUpTimestamp() > valueOf("controller.quorum.fetch.timeout.ms")`

WDYT?

052-kafka-agent-controller-configuration.md
```diff
@@ -0,0 +1,125 @@
...
+If the Kafka Roller is talking to a previous version of the Kafka Agent where the `/v1/node-configuration` HTTP endpoint doesn't exist, it will get a `404 (NOT FOUND)` error.
+In this case, it will log an error message and fail the reconciliation.
+
+### Kafka Roller usage
+
+The usage of the `controller` node configuration in the Kafka Roller is not a goal of this proposal, but for better understanding it is useful to mention it.
+
+When going through the available nodes, as a set of `NodeRef` records, the Kafka Roller is able to get the role of each of them by using the corresponding boolean `controller` and `broker` fields.
+Detecting a `controller`, the Kafka Roller queries the Kafka Agent to get the `controller` node configuration in JSON format via HTTPS on the `/v1/node-configuration` endpoint.
+The Kafka Roller also has an hardcoded list of the `controller` only configuration properties (which has to be defined taking a look at the Kafka codebase).
```
ShubhamRwt, line 104:
```suggestion
The Kafka Roller also has a hardcoded list of the `controller` only configuration properties (which needs to be defined taking a look at the Kafka codebase).
```

tinaselenge:
> > 1. Any of the expected controller nodes are not in the voters, and
> 
> I don't know what kind of situation this is. The `voters` list returned from `DESCRIBE_QUORUM` is from the voter config, so I don't understand what `the expected controller nodes are not in the voters` mean. Could you elaborate more on it?
> 
> > 2. Any of them have leader.lastCaughtUpTimestamp() - node.lastCaughtUpTimestamp() > valueOf("controller.quorum.fetch.timeout.ms")
> 
> I think this is too strict. `leader.lastCaughtUpTimestamp() - node.lastCaughtUpTimestamp() > valueOf("controller.quorum.fetch.timeout.ms")` means the follower node might have only 1 chance to fetch data from the leader, which might still not be able to catch up with the leader.
> 
> On the other hand, what do we expect for restarting the slow follower? Suppose this following is just slow due to network, we might not interrupt it under this case, right? Like when a partition is not catching up to the leader, we will not restart this node. So, maybe we restart the node only when it is stuck, like this: `LogEndOffset of the node is not increasing AND leader.lastCaughtUpTimestamp() - node.lastCaughtUpTimestamp() > valueOf("controller.quorum.fetch.timeout.ms")`
> 
> That is, this node is stuck and not sending fetch request within `controller.quorum.fetch.timeout.ms`
> 
> WDYT?

I think this makes sense. It does seem like voters list does not reflect if controllers are alive or not, although observers list does get updated when a broker shuts down, which is an odd inconsistency. My thought on the KafkaRoller is:

1. When a reconciler starts up, the KafkaRoller should make sure that we have majority of the voters alive and at least one of the follower is up to date with the leader (lag of 0?) in case there is a re-election during the rolling. Since we can't take voters list into an account, I guess we can check that majority of the voters are `leader.lastCaughtUpTimestamp() - node.lastCaughtUpTimestamp() < valueOf("controller.quorum.fetch.timeout.ms")`. 
2. Should we then follow the current logic of restarting the unhealthy/unready controllers first? In which case, besides checking the Kubernetes readiness, we can also check the logEndOffset not increasing while being fallen behind the leader as @showuon suggested.
3. Once we roll a controller, before we move on to the next controller to roll, what's the safety criteria here? We can wait for the rolled controller to catch up to the leader (when the lag becomes 0?) but also do we check again that we have the majority of the controllers alive before rolling the next controller or not since we already checked this in the beginning of reconciler? 

These also depend on the answers to some of @katheris 's questions above, which order we are rolling the brokers and controller and what level of risk we can accept.

katheris:
@tinaselenge and I have reviewed the configuration options that you can supply to a Kafka node and this is the list of configurations we suggest should be used to trigger a controller restart. Any other configuration changes would be ignored by the new logic. @tombentley @showuon @mimaison does this list look sufficient to you? Are there any that should be removed from or added to the list?

```
        static final List<String> CONTROLLER_CONFIGS = List.of(
                "authorizer.class.name",
                "auto.create.topics.enable",
                "background.threads",
                "broker.heartbeat.interval.ms",
                "broker.session.timeout.ms",
                "connection.failed.authentication.delay.ms",
                "connections.max.idle.ms",
                "connections.max.reauth.ms",
                "controlled.shutdown.enable",
                "controlled.shutdown.max.retries",
                "controlled.shutdown.retry.backoff.ms",
                "controller.listener.names",
                "controller.quorum.append.linger.ms",
                "controller.quorum.election.backoff.max.ms",
                "controller.quorum.election.timeout.ms",
                "controller.quorum.fetch.timeout.ms",
                "controller.quorum.request.timeout.ms",
                "controller.quorum.retry.backoff.ms",
                "controller.quorum.voters",
                "controller.quota.window.num",
                "controller.quota.window.size.seconds",
                "controller.socket.timeout.ms",
                "create.topic.policy.class.name",
                "delete.topic.enable",
                "early.start.listeners",
                "initial.broker.registration.timeout.ms",
                "kafka.metrics.polling.interval.secs",
                "kafka.metrics.reporters",
                "leader.imbalance.check.interval.seconds",
                "leader.imbalance.per.broker.percentage",
                "listener.security.protocol.map",
                "listeners",
                "log.dir",
                "log.dirs",
                "max.connection.creation.rate",
                "max.connections.per.ip.overrides",
                "max.connections.per.ip",
                "max.connections",
                "metadata.log.dir",
                "metadata.log.max.record.bytes.between.snapshots",
                "metadata.log.max.snapshot.interval.ms",
                "metadata.log.segment.bytes",
                "metadata.log.segment.min.bytes",
                "metadata.log.segment.ms",
                "metadata.max.idle.interval.ms",
                "metadata.max.retention.bytes",
                "metadata.max.retention.ms",
                "metric.reporters",
                "metrics.num.samples",
                "metrics.recording.level",
                "metrics.sample.window.ms",
                "node.id",
                "num.io.threads",
                "num.network.threads",
                "principal.builder.class",
                "process.roles",
                "sasl.enabled.mechanisms",
                "sasl.kerberos.kinit.cmd",
                "sasl.kerberos.min.time.before.relogin",
                "sasl.kerberos.principal.to.local.rules",
                "sasl.kerberos.service.name",
                "sasl.kerberos.ticket.renew.jitter",
                "sasl.kerberos.ticket.renew.window.factor",
                "sasl.login.callback.handler.class",
                "sasl.login.class",
                "sasl.login.connect.timeout.ms",
                "sasl.login.read.timeout.ms",
                "sasl.login.refresh.buffer.seconds",
                "sasl.login.refresh.min.period.seconds",
                "sasl.login.refresh.window.factor",
                "sasl.login.refresh.window.jitter",
                "sasl.login.retry.backoff.max.ms",
                "sasl.login.retry.backoff.ms",
                "sasl.mechanism.controller.protocol",
                "sasl.oauthbearer.clock.skew.seconds",
                "sasl.oauthbearer.expected.audience",
                "sasl.oauthbearer.expected.issuer",
                "sasl.oauthbearer.jwks.endpoint.refresh.ms",
                "sasl.oauthbearer.jwks.endpoint.retry.backoff.max.ms",
                "sasl.oauthbearer.jwks.endpoint.retry.backoff.ms",
                "sasl.oauthbearer.jwks.endpoint.url",
                "sasl.oauthbearer.scope.claim.name",
                "sasl.oauthbearer.sub.claim.name",
                "sasl.oauthbearer.token.endpoint.url",
                "sasl.server.callback.handler.class",
                "sasl.server.max.receive.size",
                "security.providers",
                "server.max.startup.time.ms",
                "socket.connection.setup.timeout.max.ms",
                "socket.connection.setup.timeout.ms",
                "socket.listen.backlog.size",
                "socket.receive.buffer.bytes",
                "socket.request.max.bytes",
                "socket.send.buffer.bytes",
                "ssl.cipher.suites",
                "ssl.client.auth",
                "ssl.enabled.protocols",
                "ssl.endpoint.identification.algorithm",
                "ssl.engine.factory.class",
                "ssl.key.password",
                "ssl.keymanager.algorithm",
                "ssl.keystore.certificate.chain",
                "ssl.keystore.key",
                "ssl.keystore.location",
                "ssl.keystore.password",
                "ssl.keystore.type",
                "ssl.principal.mapping.rules",
                "ssl.protocol",
                "ssl.provider",
                "ssl.secure.random.implementation",
                "ssl.trustmanager.algorithm",
                "ssl.truststore.certificates",
                "ssl.truststore.location",
                "ssl.truststore.password",
                "ssl.truststore.type",
                "unclean.leader.election.enable"
        );
```

showuon:
> First, I agree that there are things about how the KafkaRoller will roll controller nodes which haven't been defined here. However, I think those things need to be agreed whether this proposal is implemented, or whether the APIs defined in KIP 919 are used. So with that in mind does it make sense to have a separate proposal that discusses how the KafkaRoller should handle controller nodes?

It makes sense to have another proposal when KIP-919 is implemented upstream and we're going to adopt the APIs there. 

> Second, what is the impact of the quorum being lost during a rolling restart? Would it be acceptable to risk a loss of quorum when this proposal is implemented with the expectation that later down the line the KafkaRoller would be updated to properly handle controller rolls to prevent loss of quorum?

The impact of quorum being lost is the no metadata update is available, ex: create topic, expand ISR, change config...etc. We should try to avoid loss of quorum in this proposal IMO.

> Finally, apart from KafkaRoller checks for the quorum, should the KafkaRoller also be opinionated about what order it rolls the nodes in? For example at the moment it rolls the controller last. Should it roll all controller nodes, then all broker nodes? Should it roll the controller quorum leader last? I wonder if that also needs to be agreed somewhere, whether that's here or in a different proposal.

The order of rolling nodes, good question! I don't think it'd make any difference we roll controller first or broker first. And the active controller last while rolling controller is for sure. I'm thinking about the configs related to the connection between broker <-> controllers, ex: `controller.listener.names`, either we roll controller first or broker first, there will be a period of time broker cannot talk to controllers, which is not what we want. We should discuss it separately.

showuon:
> @tinaselenge and I have reviewed the configuration options that you can supply to a Kafka node and this is the list of configurations we suggest should be used to trigger a controller restart. Any other configuration changes would be ignored by the new logic. @tombentley @showuon @mimaison does this list look sufficient to you? Are there any that should be removed from or added to the list?

        static final List<String> CONTROLLER_CONFIGS = List.of(
                "authorizer.class.name",
                "auto.create.topics.enable",  // broker-only
                "background.threads",
                "broker.heartbeat.interval.ms",
                "broker.session.timeout.ms",
                "connection.failed.authentication.delay.ms",
                "connections.max.idle.ms",
                "connections.max.reauth.ms",
                "controlled.shutdown.enable",
                "controlled.shutdown.max.retries",
                "controlled.shutdown.retry.backoff.ms",
                
                // Below will be forbidden for users to change
                "controller.listener.names", 
                "controller.quorum.append.linger.ms",
                "controller.quorum.request.timeout.ms",
                "controller.quorum.retry.backoff.ms",
                "controller.quorum.voters",
                "controller.quota.window.num",
                "controller.quota.window.size.seconds",
                "controller.socket.timeout.ms",
                // end of controller.* configs

               // exceptions for controller.*
               "controller.quorum.election.backoff.max.ms",
                "controller.quorum.election.timeout.ms",
                "controller.quorum.fetch.timeout.ms",
               // end of exception for controller.*

                "create.topic.policy.class.name",  
                "delete.topic.enable",
                "early.start.listeners",
                "initial.broker.registration.timeout.ms",
                "kafka.metrics.polling.interval.secs",
                "kafka.metrics.reporters",
                "leader.imbalance.check.interval.seconds",
                "leader.imbalance.per.broker.percentage",
                 
                // forbidden prefix with listener.* and listeners
                "listener.security.protocol.map", 
                "listeners",
                // forbidden prefix with listener.* and listeners

                // forbidden for log.dir
                "log.dir",
                "log.dirs",
                // forbidden for log.dir
         
                "max.connection.creation.rate",
                "max.connections.per.ip.overrides",
                "max.connections.per.ip",
                "max.connections",
                "metadata.log.dir",  // should be forbidden, filed https://github.com/strimzi/strimzi-kafka-operator/pull/9031 to fix it
                "metadata.log.max.record.bytes.between.snapshots",
                "metadata.log.max.snapshot.interval.ms",
                "metadata.log.segment.bytes",
                "metadata.log.segment.min.bytes",
                "metadata.log.segment.ms",
                "metadata.max.idle.interval.ms",
                "metadata.max.retention.bytes",
                "metadata.max.retention.ms",
                "metric.reporters",
                "metrics.num.samples",
                "metrics.recording.level",
                "metrics.sample.window.ms",

                "node.id", // forbidden for node.id
                
                "num.io.threads",
                "num.network.threads",
                "principal.builder.class",

                "process.roles", // forbidden for process.roles
                // forbidden for sasl.*
                "sasl.enabled.mechanisms",
                "sasl.kerberos.kinit.cmd",
                "sasl.kerberos.min.time.before.relogin",
                "sasl.kerberos.principal.to.local.rules",
                "sasl.kerberos.service.name",
                "sasl.kerberos.ticket.renew.jitter",
                "sasl.kerberos.ticket.renew.window.factor",
                "sasl.login.callback.handler.class",
                "sasl.login.class",
                "sasl.login.connect.timeout.ms",
                "sasl.login.read.timeout.ms",
                "sasl.login.refresh.buffer.seconds",
                "sasl.login.refresh.min.period.seconds",
                "sasl.login.refresh.window.factor",
                "sasl.login.refresh.window.jitter",
                "sasl.login.retry.backoff.max.ms",
                "sasl.login.retry.backoff.ms",
                "sasl.mechanism.controller.protocol",
                "sasl.oauthbearer.clock.skew.seconds",
                "sasl.oauthbearer.expected.audience",
                "sasl.oauthbearer.expected.issuer",
                "sasl.oauthbearer.jwks.endpoint.refresh.ms",
                "sasl.oauthbearer.jwks.endpoint.retry.backoff.max.ms",
                "sasl.oauthbearer.jwks.endpoint.retry.backoff.ms",
                "sasl.oauthbearer.jwks.endpoint.url",
                "sasl.oauthbearer.scope.claim.name",
                "sasl.oauthbearer.sub.claim.name",
                "sasl.oauthbearer.token.endpoint.url",
                "sasl.server.callback.handler.class",
                // forbidden for sasl.*
                "sasl.server.max.receive.size", // exception for sasl.*

                
                "security.providers", // forbidden for security.*
                "server.max.startup.time.ms",
                "socket.connection.setup.timeout.max.ms",
                "socket.connection.setup.timeout.ms",
                "socket.listen.backlog.size",
                "socket.receive.buffer.bytes",
                "socket.request.max.bytes",
                "socket.send.buffer.bytes",

                // forbidden for ssl.*
                
                "ssl.client.auth",
                "ssl.endpoint.identification.algorithm",
                "ssl.engine.factory.class",
                "ssl.key.password",
                "ssl.keymanager.algorithm",
                "ssl.keystore.certificate.chain",
                "ssl.keystore.key",
                "ssl.keystore.location",
                "ssl.keystore.password",
                "ssl.keystore.type",
                "ssl.principal.mapping.rules",
                "ssl.provider",
                "ssl.trustmanager.algorithm",
                "ssl.truststore.certificates",
                "ssl.truststore.location",
                "ssl.truststore.password",
                "ssl.truststore.type",
                // forbidden for ssl.*

                 // exception for ssl.*
                 "ssl.cipher.suites",
                 "ssl.protocol",
                "ssl.enabled.protocols",
                "ssl.secure.random.implementation",
                // end exception for ssl.*

                "unclean.leader.election.enable". // broker-only config. In Controller, it uses Raft protocol to do leader election

                // in v3.6.0, delegation token will be supported in KRaft
                "delegation.token.master.key"
                "delegation.token.secret.key"
                "delegation.token.max.lifetime.ms"
                "delegation.token.expiry.time.ms"
                "delegation.token.expiry.check.interval.ms"

                // the queued configs are also used in controller 
                "queued.max.requests"
                "queued.max.request.bytes"

               "alter.config.policy.class.name" // controller also works


        );

showuon:
Read [this](https://kafka.apache.org/documentation/#replace_disk) while checking the docs:

> To determine if the majority of the controllers have the committed data, run the kafka-metadata-quorum.sh tool to describe the replication status
> Check and wait until the Lag is small for a majority of the controllers. If the leader's end offset is not increasing, you can wait until the lag is 0 for a majority; otherwise, you can pick the latest leader end offset and wait until all replicas have reached it. Check and wait until the LastFetchTimestamp and LastCaughtUpTimestamp are close to each other for the majority of the controllers. At this point it is safer to format the controller's metadata log directory. 

So, the suggestion from the doc about how to determine if the majority of the controllers have the committed data: 
1. Check and wait until the Lag is small for a majority of the controllers
  1.1. If the leader's end offset is not increasing, you can wait until the lag is 0 for a majority (Basically this won't happen because there is a periodically heartbeat commit by default)
  1.2. you can pick the latest leader end offset and wait until all replicas have reached it. <-- I think this is a good method

2. Check and wait until the LastFetchTimestamp and LastCaughtUpTimestamp are close to each other for the majority of the controllers.  <-- This is the way we proposed. 


So, are we going to adopt (1.2) + (2)?  Or just (2)? If we want to adopt (1.2), the rolling will be slower since we need multiple rounds of reconciliation to verify if follower caught up with the expected offset, but of course, it provides more accuracy.

mimaison:
It seems `initial.broker.registration.timeout.ms` is a broker only configuration too.

056-kafka-agent-controller-configuration.md
```diff
@@ -0,0 +1,99 @@
...
+}
+```
+
+Such a configuration is returned through the `/v1/node-configuration` HTTP endpoint.
+Because this file never changing once the broker/controller process has started, even if the node has its configuration dynamically updated, there is no need for the Kafka Agent to watch the file.
+It is enough to load its content on the Kafka Agent startup and keep it as cached in memory.
+The same content will be just returned every time the Kafka Roller is asking for it.
+
+If the Kafka Roller is talking to a previous version of the Kafka Agent where the `/v1/node-configuration` HTTP endpoint doesn't exist, it will get a `404 (NOT FOUND)` error.
+In this case, it will log an error message and fail the reconciliation.
```
scholzj, line 71: How will this information be secured? It contains possibly sensitive information such as passwords etc.

scholzj (commented):
As I raised before in a different place -> this seems like unnecessary effort. Until the KAfka APIs are implemented, we can do the diff using the existing means -> for example by a checksum of the configuration stored as a pod annotation and used to roll it. This is already used for many other options affecting the Kafka nodes. So it seems to make sense to me to reuse it and not invent something new.

056-kafka-agent-controller-configuration.md
ppatierno: The Kafka Agent endpoint is exposed via HTTPS as it's already for the broker state endpoint.

scholzj: https://github.com/strimzi/strimzi-kafka-operator/pull/8892 adds 8080 port for the probes.

ppatierno: but only for the probes. The broker state is still on HTTPS. And the config endpoint will be the same.

ppatierno: Well even from the code you linked you see ...

```java
server.setConnectors(new Connector[] {httpsConn, httpConn});
```

so Jetty is able to handle both HTTP and HTTPS at the same time and having different endpoints on them.
So again broker state on HTTPS and probes on HTTP.

scholzj:
After a discussion with @ppatierno @katheris @tinaselenge, we decided to replace this proposal with https://github.com/strimzi/strimzi-kafka-operator/pull/9125 PR that uses an already existing mechanism for detecting the controller configuration changes. The mechanism it is based on is already used for example to roll pods when plugin configurations change. As it is not a new mechanism, it does not need the proposal. The same caveat applies to it as to this proposal -> it serves as a stop-gap mechanism that will be replaced once Kafka provides operationalized Kafka controller nodes (e.g. the KIP-919 and possible follow-ups).

ppatierno:
Thanks Jakub. I am going to close this proposal. I reviewed and approved the PR you provided.

PR #84, opened 2023-07-24
feat: Proposal for reconcile and Kafka version stored in Kafka Status

samuel-hawker:
Proposal for reconciled version metadata to be included in the CRs recording which operator has finished reconciling
This allows better user story for confirming an upgrade has completed.


052-record-reconciled-versions.md
```diff
@@ -0,0 +1,203 @@
...
+- The operator updates the `strimzi.io/reconciling` annotation value on reconciliation startup without overriding the `strimzi.io/reconciled` annotation. If the operator errors during reconcile it will be clear that the currently reconciled version does not match the version that the operator is applying. A two stage commit mechanism here gives much more information regarding whether it started the reconciliation versus whether it finished successfully.
+
+An important note here, in order to avoid unnecessary extra reconcilliations, and to avoid constant re-adding and removal of the annotation, this proposal suggests that `strimzi.io/reconciling` remains, even after a reconcile, and that the annotation is only ever updated once a new operator version attempts the reconcile.
+
+
+The following two sections assume that along with status updates, the operator now also patches the annotations alongside to reflect the new annotation state, which will require a patch/update to the CR, rather than a simple `updateStatus` call.
+
+## `Kafka` Custom Resource
```
scholzj, line 46: The Kafka CR is a public resource. We should not touch its specification as that could cause many issues with causing loops between the operator and the user applications adding and removing the annotations.

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,203 @@
...
+- It signals that a reconcile on version Y has started, and that the operator is functional without checking the logs of the operator.
+- The operator updates the `strimzi.io/reconciling` annotation value on reconciliation startup without overriding the `strimzi.io/reconciled` annotation. If the operator errors during reconcile it will be clear that the currently reconciled version does not match the version that the operator is applying. A two stage commit mechanism here gives much more information regarding whether it started the reconciliation versus whether it finished successfully.
+
+An important note here, in order to avoid unnecessary extra reconcilliations, and to avoid constant re-adding and removal of the annotation, this proposal suggests that `strimzi.io/reconciling` remains, even after a reconcile, and that the annotation is only ever updated once a new operator version attempts the reconcile.
+
+
+The following two sections assume that along with status updates, the operator now also patches the annotations alongside to reflect the new annotation state, which will require a patch/update to the CR, rather than a simple `updateStatus` call.
```
scholzj, line 44: TBH, this seems to overcomplicate this quite significantly. It also significantly increases the number of API calls. 

Have you considered to:
* Use just one annotation
* Roll it out to StrimziPodSets and Pods as part of their regular reconciliation?

That should give you exactly the same information about the progress with significantly less effort: 
* You will know how far did the new version of the operator got by looking at where the new annotations are:
  * Did it reconciled the StrimziPodSet? If yes, it has the new version. If not, it has the old version
  * Did it managed to roll out the pods? If yes, the pods will have the new versions. If not, it still has the old version

This is also much more accurate information. Because if the operator is failing in the middle, you will know exactly how far it got. If you do the annotations on the beginning and at the end, you do not have the information about how far did it got only that it is _in the process_. But this way, you can from the resources know exactly that it is stuck for example _in between Kafka and ZooKeeper_.

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,203 @@
...
+Though this works, from a user standpoint it is far from ideal as it assumes a level of user knowledge about the cluster that may not be appropriate for all users.
+
+## Motivation
+
+This proposal allows a user to rely on the `Kafka` custom resource as their single source of truth, instead of watching/querying the individual kafka/zookeeper pods.
+This proposal lays out all the information we need and at what points in the reconcile. This information is already available within the Reconcile loop, so orchestrating and figuring out if a component has been fully upgraded will be propagated into the Custom Resource via metadata.
+
+## Proposal
+
+This proposal will cover the `Kafka` and `StrimziPodSet` custom resources, but could be extended to include other Custom Resources if there would be value in it. This proposal starts with the `Kafka` Custom Resource as this is the core component of Strimzi, and its reliance on `StrimziPodSet` means we should implement it for this CR type as well.
```
scholzj, line 19: I think that for the consistency reasons, this should be at least applied to:
* Connect based deployments (Connect / MM2) as they use StrimziPodSets
* But also to all Deployments as they also cover running components and correspond in that to the StrimziPodSets.

samuel-hawker:
After going your review comments and a small slack discussion about this proposal, I have reworked it to be in the status now. I think this is now a lot simpler, and though we could propogate annotations through to the pods in the SPS's to track upgrade, i don't think this is necessary due to how the `KafkaReconciler` will wait on all pods to be ready post rolling update.

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,160 @@
...
+`status.versions.reconciled=0.35.0` would remain if a reconcilliation on `0.36.0` failed as the information that a prior reconcilliation passed could be useful to an end user.
+
+### status.versions.managedBy
+The `status.versions.managedBy` field is patched to the CR at the start of reconcile in the `initialStatus` and signals to a user that the operator is attempting to reconcile the CR at version `Y`, regardless of success or failure.
+```
+status:
+  versions:
+    managedBy: '0.37.0'
+```
+
```
ppatierno, line 45: I have a general comment about the names for these fields. These are Strimzi operator versions, which I would make clear in naming those fields.

ppatierno: Yeah the last ones are too verbose, taking `Operator` out is fine with me. But another question I have is about the higher level `versions` field which sounds to be generic. Do you envisage the usage of it for something else? Or do we need it?

samuel-hawker: I don't think it is _needed_, I thought to better distinguish that these are version fields, and to have a base to build off of for potential future additions might be useful.
I.e. extra version metadata, though I can't give any solid examples.

I can cut out the versions field and instead simplify to:
```
status:
  strimziReconciled: 0.37.0
  strimziManagedBy: 0.37.0
```
or something else like:
status:
  reconciledVersion: 0.37.0
  managedByVersion: 0.37.0
```

Comment (ppatierno) [reply 052-record-reconciled-versions.md:45]: It was just a questions not a statement to remove it. I think having `versions` would allow to add more later so I would keep it. But then we need to be precise about which version we are talking about so one of your previous suggestions could be fine:

```yaml
status:
  versions:
    strimziReconciled: 0.37.0
    strimziManagedBy: 0.37.0
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,160 @@
...
+```
+status:
+  versions:
+    strimziReconciled: '0.37.0'
+```
+If a reconcile does not reach completion, this field is not added or updated, however in the case where it already exists it is left at the previous value.
+i.e.
+`status.versions.strimziReconciled=0.35.0` would remain if a reconcilliation on `0.36.0` failed as the information that a prior reconcilliation passed could be useful to an end user.
+
+### status.versions.strimziManagedBy
```
scholzj, line 38: I have concerns about the usability of this. I do not think the difference between those two is really clear. If the reconciliation gets stuck at reconciling ZooKeeper or even before at the wrong version of Kafka for example, the `strimziManagedBy` will not really mean much as the operator does not really manage the cluster when it gets stuck right at the beginning. Similarly, if the reconciliation fails because of an issue with Kafka Exporter, the `strimziReconciled` might give the old version. But essentially everything other than the Kafka Exporter is being reconciled by the new version already.

So as an expert, you might understand these nuances. But then you should be able to figure it out from the resources and logs. And if you are just a regular user, you will not understand them and these two fields will be meaningless or confusing.

It also makes me a bit wonder of what are we trying to achieve:
* Is it the upgrade tracking? What exactly do you want to track? For example, if the upgrade jumps Kafka versions, the first successful reconciliation might not mean the end of it as the Kafka upgrade is multistage. Or if you want to track the 
* Is it trying to track which operator version is managing the cluster now? Maybe we should simplify it by having one version in the status and make it clear that the user needs to figure out from the logs / conditions how far did it got. But it would allow to see that this is now managed by a new version of the operator (successfully or not).

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,160 @@
...
+The `status.versions.strimziManagedBy` field is patched to the CR at the start of reconcile in the `initialStatus` and signals to a user that the operator is attempting to reconcile the CR at version `Y`, regardless of success or failure.
+```
+status:
+  versions:
+    strimziManagedBy: '0.37.0'
+```
+
+This field is required for the following reasons:
+- It signals that a reconcile on version Y has started, and that the operator reconciler & watcher are functional without checking the logs of the operator.
+- The operator updates the `status.versions.strimziManagedBy` field on reconciliation startup without overriding the `status.versions.strimziReconciled` field. If the operator errors during reconcile it will be clear that the currently reconciled version does not match the version that the operator is applying. A two stage commit mechanism here gives much more information regarding whether it started the reconciliation versus whether it finished successfully.
```
scholzj, line 48: As I said above -> I think it does not give much more information. It is meaningless unless you figure out the progress which it made. I think comparing it to two-stage commit is misleading. This is not a binary all or nothing state. It will be stuck somewhere on the long road between the two versions where some parts are reconciled by the new version and some parts are not.

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,160 @@
...
+
+An important note here, in order to avoid unnecessary extra reconcilliations, and to avoid constant re-adding and removal of the field, this proposal suggests that `status.versions.strimziManagedBy` remains, even after a reconcile, and that the annotation is only ever updated once a new operator version attempts the reconcile.
+
+
+## `Kafka` Custom Resource changes
+- The update to `status.versions.strimziManagedBy` would happen in the `initialStatus` [method](https://github.com/strimzi/strimzi-kafka-operator/blob/2a1fdf9d8695bb22a2bf977b0ba5414291530207/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java#L292)
+- The update to `status.versions.strimziReconciled` would happen in the `createOrUpdate` [method](https://github.com/strimzi/strimzi-kafka-operator/blob/2a1fdf9d8695bb22a2bf977b0ba5414291530207/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java#L137) alongside the `Ready` update where the operator version would be updated into the field.
```
scholzj, line 55: You need to manage both versions in both methods. Otherwise they will be appearing and disappearing which would make them useless.

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,160 @@
...
+```
+status: {}
+```
+
+### Fresh Install
+
+- On a fresh install, a user creates a Kafka CR:
+  ```
+  strimziReconciled: null
+  strimziManagedBy: null
```
scholzj, line 76: You probably don't want to list the null values anywhere.

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,160 @@
...
+  strimziManagedBy: 0.38.0
+  ```
+  and patches `StrimziPodSet` CRs
+- StrimziPodSet reconciler reconciles the `SPS`s with help of `KafkaRoller` and `ZookeeperRoller`
+- Kafka Reconciler picks up that `SPS` is now Ready and finished upgrading.
+  Kafka Reconciler continues to install other components such as entity operator.
+  Once all components are upgraded, update `Kafka` CR with
+  ```
+  strimziReconciled: 0.38.0
+  strimziManagedBy: 0.38.0
```
scholzj, line 142: Did you abandon the idea of annotating the StrimziPodSets and Deployments? It is a bit confusing to me as you mention the rollers here or the SPS / Deployments int he intro. But there are no more details about them.

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,160 @@
...
+  ```
+- User can now verify `Kafka.status.versions.strimziReconciled` to check if it has reconciled to correct/expected level.
+- if there is an error at any point during reconcile `Kafka.status.versions.strimziReconciled` will not be updated.
+
+
+## Affected/not affected projects
```
scholzj, line 148:
```suggestion

## Affected/not affected projects

```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,160 @@
...
+## Proposal
+
+This proposal will cover the `Kafka` custom resource but could apply to other CR types that use `StrimziPodSets` such as `KafkaConnect` and `KafkaMirrorMaker2`, but could be extended to include other Custom Resources if there would be value in it.
+This proposal starts with the `Kafka` Custom Resource as the base example as it is the core component of Strimzi, and it deploys both `StrimziPodSet`s and `Deployment`s.
+
+This proposal assumes that all key value pairs mentioned will be writen to the `status` field as this is managed by the operator and can be added to without impacting end users.
+
+The following is the proposed new fields:
+
+### status.versions.strimziReconciled
```
scholzj, line 26:
```suggestion

### status.versions.strimziReconciled

```

052-record-reconciled-versions.md
scholzj, line 38:
```suggestion

### status.versions.strimziManagedBy

```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+This field is required for the following reasons:
+- It signals that a reconcile on version Y has started, and that the operator reconciler & watcher are functional without checking the logs of the operator.
+- The operator updates the `status.versions.strimziManagedBy` field on reconciliation startup without overriding the `status.versions.strimziReconciled` field. If the operator errors during reconcile it will be clear that the currently reconciled version does not match the version that the operator is applying. A two stage commit mechanism here gives much more information regarding whether it started the reconciliation versus whether it finished successfully.
+
+An important note here, in order to avoid unnecessary extra reconcilliations, and to avoid constant re-adding and removal of the field, this proposal suggests that `status.versions.strimziManagedBy` remains, even after a reconcile, and that the annotation is only ever updated once a new operator version attempts the reconcile.
+
+
+## `Kafka` Custom Resource changes
```
scholzj, line 53:
```suggestion

## `Kafka` Custom Resource changes

```

052-record-reconciled-versions.md
samuel-hawker: I did abandon the idea somewhat after it was made clear to me, after our conversation that the KafkaAssembly had all the info it needed to make the determination whether a full roll/reconcile was finished (if this is not the case though, please correct me on this).

Technically we could continue to propagate the annotation through the pods, but considering the main aim of the proposal is to have info in the status to reflect whether the upgrade was complete, I think this metadata should only be propagated to the deployments and SPS if it is required as part of the mechanism of determining whether a full reconcile has finished.

The purpose is to make an upgrades Readiness + finished state more clear to a user in just the source of truth, the `Kafka` CR rather than relying on checking the pods directly.

052-record-reconciled-versions.md
scholzj: So, in that case, maybe you could keep only one version field and update it at the successful reconciliation only. But for best understanding, maybe we should name it accordingly? `lastSuccessfulReconcilitationBy`? Or something similar to make it clear it talks about successful reconciliations? I think that would have some value, the name would make it easier to understand and with one version only it would be in general more understandable as there would not be the _How do they differ_ effect. So it is something I could live with for sure.

That said, my personal preference would be to keep the `managedBy` field and set it regardless of the reconciliation result. And your check that the upgrade completed would be that `managedBy` has new version and the Kafka CR is `Ready`. It would also let you see that `managedBy` has new version and the Kafka CR is **not** `Ready` in which case you basically know that the upgrade of the operator happened, but something went wrong when upgrading the operand. On the other hand, if the upgrade succeeded and then it failed for an unrelated reason, while you went away from the keyboard, you will not know that the reconciliation passed before. So, I guess it has different points of view. But as I said, I would be fine with the above as well - just sharing for consideration.

---

The Kafka version was not mentioned so far. I think it would be useful to have it - especially since the `.spec.kafka.version` is not mandatory. I guess the challenge is:
* When do you update it? With respect to the reconciliation progress - after the reconciliation succeeds or at some point when KafkaReconciler succeeds?
* Is this something for Kafka CR? Should we consider having the Kafka version defined per node pool? It was not discussed in detail. But for example some users do upgrade through rolling out new brokers with new Kafka versions, reassigning the partitions and then delete the old brokers. So we should consider how likely it is that we will support this in the future and if yes, what do we do with this field. Otherwise I think it would be useful.

samuel-hawker: This has given me a lot of food for thought, especially regarding the KafkaNodePools upgrade option.
I think we could also add something like a `status.versions.kafka` to the `KafkaNodePool` CR as well, and this would be reflected also in the Kafka CR, since the KafkaReconciler should 'know' about which version of Kafka it is switching to through the Node Pools.

I think I will edit the proposal to now cover 3 status fields (names not yet confirmed):
```
status.versions.lastSuccessfulReconcilitationBy
status.versions.managedBy
status.versions.kafka
```
These will be included in the `Kafka` CR.

And then a user combining and checking `status.versions.lastSuccessfulReconcilitationBy` , `status.versions.kafka` and `status.phase==Ready` would cover a great deal of upgrade scenarios being 'finished'/successful.

the `status.versions.managedBy` field being updated would assure a user that an upgrade was atleast starting, but the value of this field can still be debated.

scholzj: But why mix both status.versions.lastSuccessfulReconcilitationBy and status.versions.managedBy together? Wouldn't one be enough? Two introduce the confusion of what is the difference. As for the node pools, I would not add anything there right now as they are not reconciled separately. And the Strimzi operator will be always the same as for the Kafka resource. But we need to consider if the Kafka version in the Kafka CR is future proof or not.

samuel-hawker: I will rewrite, combining status.versions.lastSuccessfulReconcilitationBy and status.versions.managedBy
If I come up with a compelling reason in the flow to add it back in, I will.
Behaviour for `lastSuccessfulReconcilitationBy` will be, update on success, leave it as-is if present on failure.

scholzj: It is not about compelling reasons. It is about usability. You can add 10 different versions. But then it will be completely confusing what they mean, what is the difference, which one they need to care about and so on. So I think you should pick one.

samuel-hawker:
I have posted another update:

- remove `managedBy` field
- added a kafka version field to identify multi-reconcile upgrades
- added potential extra fields for inter broker protocol version and log format version as these could be relevant (discussion to be had)
- add clarity to examples and make clear why and how status fields wont be removed when set

scholzj:
@samuel-hawker Inter broker protocol version and log format version are basically deprecated as far as I understood. `log.message.format.version` is ignored is `inter.broker.protocol.version` is `3.0` or higher. `inter.broker.protocol.version` is not deprecated formally. But it is not used in KRaft mode. So it will be essentially removed in Kafka 4.0. So not sure I would include those given they cannot be easily removed later.

samuel-hawker:
A great point. I will remove.

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,185 @@
...
+  versions:
+    lastSuccessfulReconciliationBy: '0.37.0'
+```
+If a reconcile does not reach completion, this field is not added or updated, however in the case where it already exists it is left at the previous value.
+i.e.
+`status.versions.lastSuccessfulReconciliationBy=0.35.0` would remain if a post-operator-upgrade reconciliation on `0.36.0` failed, as the information that a prior reconciliation passed would be useful to an end user.
+
+### status.versions.kafka
+
+The `status.versions.kafka` field is updated in the CR at the end of the KafkaReconciler in the `podsReady` method and signals to a user that the KafkaReconciler has finished reconciling the Kafka brokers to Kafka version `Y`, regardless of success or failure later in the reconcile.
```
scholzj, line 44: I assume it will just update the field in the status and have it be written at the end of the reconciliation rather then having a separate status update API call?

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,185 @@
...
+
+If the operator errors during the Kafka reconciliation this field will not be updated. 
+If Kafka was successfully updated, the field will also be updated, meaning errors later in the reconcile, such as failing to deploy the `entity-operator` or `kafka-exporter` will still show `status.versions.kafka` correctly up to date, but `lastSuccessfulReconciliationBy` will not be updated, and the CR will no longer be in a `Ready` state.
+
+
+## `Kafka` Custom Resource & Reconciler changes
+
+- Update `KafkaStatus` with new `versions` field and sub fields as agreed on this proposal.
+
+- Update `initialStatus` [method](https://github.com/strimzi/strimzi-kafka-operator/blob/0c4e96de69e20cb80e86fd8ccb3ab8baca431822/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaAssemblyOperator.java#L292) to seed new status with the `status.versions` field if it is already set on the CR. This is so that any status update leaves these fields unchanged, unless they are explicitly modified as part of reconcile.
```
scholzj, line 61: I don't think I follow this part. The _initia status_ basically only says it is _deploying_ a new cluster. A new cluster means the status is empty. So what is there to _seed_ at this point? It does not have a Kafka version yet and it was never successfully reconciled.

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,185 @@
...
+
+- The update to `status.versions.kafka` would happen after the `podsReady` [method](https://github.com/strimzi/strimzi-kafka-operator/blob/0c4e96de69e20cb80e86fd8ccb3ab8baca431822/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaReconciler.java#L817) updating the Kafka version once all pods are ready. The Kafka version can be determined from the `StrimziPodSet` which has the information available in the `strimzi.io/kafka-version` annotation.
```
scholzj, line 63: This seems related to my comment above. Why not store it and update it later when the reconciliation ends (successfully or not)?

Also, why do you need to know the version from the PodSet? The operator has to know the version it uses there, so there should be no need to query the pod set annotation.

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,185 @@
...
+
+
+### Upgrade (from no mechansim)
+
+Same as fresh install, upgrade from 0.37.0 to 0.38.0 with Kafka cersion 3.5.1
+- On an upgrade, a user has a Kafka CR:
+  ```
+  # the status.versions is not set, but treat it as empty here
+  status:
+    versions: {}
```
scholzj, line 106: The version fill would not be there in the old version.

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,185 @@
...
+      kafka: 3.5.1
+  ```
+- User can now verify `Kafka.status.versions.lastSuccessfulReconciliationBy` and `Kafka.status.versions.kafka` to check if it has reconciled to correct/expected levels.
+- if there is an error at any point prior to the Kafka Broker rollout `Kafka.status.versions.kafka` will not be updated.
+- if there is an error at any point during reconcile `Kafka.status.versions.lastSuccessfulReconciliationBy` will not be updated.
+
+## `KafkaNodePool` considerations
+
+Some users use new `KafkaNodePool` CRs to facilitate upgrades through rolling out new brokers with new Kafka versions, reassigning the partitions and then delete the old brokers.
+This proposal should work with this method od upgrade, since the Kafka version + readiness checks will be track by the `StrimziPodSet`, so that when a user had finished upgrading via this method, the new `status.versions.kafka` should be updated accordingly.
```
scholzj, line 166: I think the future consideration to make here is what if different node pools have different Kafka versions. And thus there will be no single version for the `status.versions.kafka` field (as multiple of them will be used). And this is not answered in this paragraph.

_Note, we might never support something like this in reality. But I expect that this will be raised by some users in the future._

scholzj (commented):
I think the general idea with the version fields is fine with me. Ileft some more comments about the technical details / implementation.

samuel-hawker:
Thanks for the review comments.
I have removed the sections that just confused matters, and elaborated on the node pool case, proposing that we choose the minimum of all the node pool versions, but I suspect you may have thoughts as to whether this is a good solution or not.

scholzj:
> I have removed the sections that just confused matters, and elaborated on the node pool case, proposing that we choose the minimum of all the node pool versions, but I suspect you may have thoughts as to whether this is a good solution or not.

TBH, I have only one idea with regard to the Kafka version and Node pools. We could make it list => today we can list there only one version. If we ever support multiple versions, we could list all used versions. Not sure it is good, but it would allow us to extend it to multiple versions if needed 🤷‍♂️

samuel-hawker:
A list is sensible - alternatively if we are quite convinced the scenario is unlikely, we can keep the field as a single string, and in the unlikely event multiple kafka versions are needed we just comma delimiter them i.le.
```
status:
  versions:
    kafka: "3.5.0,3.5.1"
```
this is probably not ideal, but if we genuinely think Strimzi supporting such configurations is unlikely it might be better, since I think it would be marginally easier for a user to verify a string field contains a version, rather than parsing a yaml/json array and checking for its presence as a element, but I could be persuaded either way here.

scholzj:
@samuel-hawker I guess that would work for me. Maybe you can add it to the proposal and add the proper list to the rejected alternatives and we will see what other reviewers think?

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,197 @@
...
+This proposal starts with the `Kafka` Custom Resource as the base example as it is the core component of Strimzi, and it deploys both `StrimziPodSet`s and `Deployment`s.
+
+This proposal assumes that all key value pairs mentioned will be writen to the `status` field as this is managed by the operator and can be added to without impacting end users.
+
+The following is the proposed new fields:
+
+### status.versions.lastSuccessfulReconciliationBy
+
+The `status.versions.lastSuccessfulReconciliationBy` field is patched at the end of a **successful** reconcile, and signals to a user that the operator at version `X` was able to reconcile to successful completion, meaning it reached the end of a reconcile without error.
+This would be the field a user would query against to check whether an upgrade was complete - or atleast a complete roll of the Kafka brokers on the latest operator version. Details on how to verify that an upgrade has completed when the `Kafka.spec.kafka.version` is updated is covered in the next section.
```
scholzj, line 30:
```suggestion
This would be the field a user would query against to check whether an operator upgrade was complete - or at least a complete roll of the Kafka brokers on the latest operator version. Details on how to verify that an upgrade has completed when the `Kafka.spec.kafka.version` is updated is covered in the next section.
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,197 @@
...
+      lastSuccessfulReconciliationBy: 0.38.0
+      kafka: 3.5.1
+  ```
+- User can now verify `Kafka.status.versions.lastSuccessfulReconciliationBy` and `Kafka.status.versions.kafka` to check if it has reconciled to correct/expected levels.
+- if there is an error at any point prior to the Kafka Broker rollout `Kafka.status.versions.kafka` will not be updated.
+- if there is an error at any point during reconcile `Kafka.status.versions.lastSuccessfulReconciliationBy` will not be updated.
+
+## `KafkaNodePool` considerations
+
+Some users use the new `KafkaNodePool` CRs to facilitate upgrades through rolling out new brokers with new Kafka versions, reassigning the partitions and then delete the old brokers.
```
scholzj, line 164: It is not that someone uses it today as this might suggest. It is more something to consider. So maybe this would be better wording?

```suggestion
It might happen that in the future each `KafkaNodePool` CR supports having its own Kafka version which might differ from other `KafkaNodePools` in the same Kafka cluster.
```

scholzj (commented):
Looks good to me. Just two more nits about the wording.

scholzj (approved):
Thanks @samuel-hawker

052-record-reconciled-versions.md
ppatierno: I see this proposal being updated with using `kafka` and `lastSuccessfulReconciliationBy`. While the former sounds clear, the latter is not. Why did you decide to remove "strimzi" as part of the name?

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,197 @@
...
+status:
+  versions:
+    kafka: '3.5.1'
+```
+
+Note: This will update the field in the reconcilers internal status object and have it be written at the end of the reconciliation rather then having a separate status update API call each time this is updated.
+
+This field is required as it signals that a reconcile to Kafka version `Y` has finished, and this extra information is important and needs to be available, as a user just checking `status.versions.lastSuccessfulReconciliationBy` is not sufficient for upgrades where multiple rolls of the brokers are required, such as a Kafka version update alongside an upgrade.
+
+If the operator errors during the Kafka reconciliation this field will not be updated.
```
ppatierno, line 55: It could be important to highlight that if you have more than one broker (as it should be), the Kafka version could be updated on some of them but failing on the others and if there is a blocking issue it will stay this way.
So this `kafka` version field doesn't exactly reflect this situation. It will stay with the old version (because not all pods are ready) but in reality some brokers have new version, others the old one.

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,197 @@
...
+  versions:
+    kafka: '3.5.1'
+```
+
+Note: This will update the field in the reconcilers internal status object and have it be written at the end of the reconciliation rather then having a separate status update API call each time this is updated.
+
+This field is required as it signals that a reconcile to Kafka version `Y` has finished, and this extra information is important and needs to be available, as a user just checking `status.versions.lastSuccessfulReconciliationBy` is not sufficient for upgrades where multiple rolls of the brokers are required, such as a Kafka version update alongside an upgrade.
+
+If the operator errors during the Kafka reconciliation this field will not be updated. 
+If Kafka was successfully updated, the field will also be updated, meaning errors later in the reconcile, such as failing to deploy the `entity-operator` or `kafka-exporter` will still show `status.versions.kafka` correctly up to date, but `lastSuccessfulReconciliationBy` will not be updated, and the CR will no longer be in a `Ready` state.
```
ppatierno, line 56: even cruise control is part of the Kafka CR, it could fail as well. It is worth mentioning it alongside EO and Kafka Exporter.

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,197 @@
...
+      kafka: 3.5.0
+  ```
+- User can now verify `Kafka.status.versions.lastSuccessfulReconciliationBy` and `Kafka.status.versions.kafka` to check if it has reconciled to correct/expected levels.
+- if there is an error at any point prior to the Kafka Broker rollout `Kafka.status.versions.kafka` will not be updated.
+- if there is an error at any point during reconcile `Kafka.status.versions.lastSuccessfulReconciliationBy` will not be updated.
+
+
+### Upgrade (from no mechansim)
+
+Same as fresh install, upgrade from 0.37.0 to 0.38.0 with Kafka cersion 3.5.1
```
ppatierno, line 101:
```suggestion
Same as fresh install, upgrade from 0.37.0 to 0.38.0 with Kafka version 3.5.1
```

052-record-reconciled-versions.md
samuel-hawker: @ppatierno I did rewrite it to use `lastSuccessfulStrimziReconciliationBy` but I was thinking this was getting too long as a field name, so removed the `strimzi` from the name as I _felt_ it was implicit that Strimzi was reconciling it - however I can re-add it if you are in favour of it.
The objective of this field is to clear in what it is displaying to a user, and I am unsure if `lastSuccessfulStrimziReconciliationBy` is anymore clear than `lastSuccessfulReconciliationBy`, but curious on your thoughts here as we need to get the field name right!

scholzj: It is just Strimzi reconciling it. So what value does it really add to have the `strimzi` in the field name?

ppatierno: > It is just Strimzi reconciling it. So what value does it really add to have the strimzi in the field name?

That's true, I was just wondering for consistency. Btw no really strong opinion I can live with the current field name.

ppatierno (approved):
LGTM. Thanks Sam!

see-quick (approved):
Thanks for the proposal and happy to help you with the integration of such logic inside our STs.

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,198 @@
...
+This proposal lays out all the information that is required, and at what points in the reconcile this information will be used. This information is already readily available within the Reconcile loop, so orchestrating and figuring out if a component has been fully upgraded will be a case of propagating this information into the Custom Resource via status fields.
+
+## Proposal
+
+This proposal will cover the `Kafka` custom resource but could apply to other CR types that use `StrimziPodSets` such as `KafkaConnect` and `KafkaMirrorMaker2`, but could be extended to include other Custom Resources if there would be value in it.
+This proposal starts with the `Kafka` Custom Resource as the base example as it is the core component of Strimzi, and it deploys both `StrimziPodSet`s and `Deployment`s.
+
+This proposal assumes that all key value pairs mentioned will be writen to the `status` field as this is managed by the operator and can be added to without impacting end users.
+
+The following is the proposed new fields:
```
tombentley, line 25:
```suggestion
The following are the proposed new fields:
```

tombentley (commented):
I'm on board with the basic functionality. But I have a few questions:

1. No `status.versions.zookeeper` or distinguishing the version of the KRaft quorum? I know we wouldn't support people using a different kraft version than broker version, but there's a window during upgrade where they will differ.
2. Should we also publish `status.versions.cruiseControl` too, similarly for other components? While I'm not advocating for or against that, I think we should be explicit why we are/are not doing that.
3. Having the single `status.versions` mixes operator versions with component versions. To me those are quite distinct. Did you consider `status.operatorLastSuccessfulVersion` and `status.kafkaVersion` (or possibly `status.componentVersions.kafka` if we see a lot of value in the grouping)?

tombentley:
@samuel-hawker I should have add: "Thanks for the proposal"!

samuel-hawker:
Thanks for the review @tombentley
To answer your questions:

> 1. No status.versions.zookeeper or distinguishing the version of the KRaft quorum? I know we wouldn't support people using a different kraft version than broker version, but there's a window during upgrade where they will differ.

I have not included zookeeper here as soon it will be removed, but additionally, from a user perspective I think a user would be more inclined to care about their kafka version, than their zookeeper version, since this is what they will be tracking during upgrade.
This proposal is solving the following problem: `For the `Kafka` Custom Resource it is often unclear when an upgrade has truly finished, whether all the brokers have rolled and to which version of the operator a cluster has been reconciled with and which version of Kafka.` and I am not personally of the opinion that the zookeeper version adds anything to this goal.

> 2. Should we also publish status.versions.cruiseControl too, similarly for other components? While I'm not advocating for or against that, I think we should be explicit why we are/are not doing that.

Similar justification to point 1, since a user will be tracking upgrade through the `kafka` and `lastSuccessfulReconciliationBy` fields, I feel like `lastSuccessfulReconciliationBy` specifically covers all components and the level of granular version checking is not required for all components, besides Kafka which is a special case due to its multi-roll situation.

> 3 Having the single status.versions mixes operator versions with component versions. To me those are quite distinct. Did you consider status.operatorLastSuccessfulVersion and status.kafkaVersion (or possibly status.componentVersions.kafka if we see a lot of value in the grouping)?

I did consider this, however I then decided that mixing all of these versions was reasonable as a user might want them all in one place (the `versions.<x>` fields, however I can definitely be persuaded to change it to: `status.operatorLastSuccessfulVersion` (or `lastSuccessfulReconciliationVersionBy`) and `status.kafkaVersion` 

Given my answers to 1 and 2 though I am unsure about the `status.componentVersions.kafka` grouping as you suggested.

I hope this answers your questions!

tombentley:
@samuel-hawker thanks. Your answers to 1 and 2 make sense. For 3, I don't feel _very_ strongly, but I think I prefer `status.operatorLastSuccessfulVersion` and `status.kafkaVersion`. I'm interested to hear what others think.

scholzj:
> @samuel-hawker thanks. Your answers to 1 and 2 make sense. For 3, I don't feel _very_ strongly, but I think I prefer `status.operatorLastSuccessfulVersion` and `status.kafkaVersion`. I'm interested to hear what others think.

I would not have a problem with having them in the `.status` directly.

ppatierno:
> @samuel-hawker thanks. Your answers to 1 and 2 make sense. For 3, I don't feel very strongly, but I think I prefer status.operatorLastSuccessfulVersion and status.kafkaVersion. I'm interested to hear what others think.

My preference is having `status.versions` but if the majority feels it's better having all versions directly under `status` I can live with it.

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,198 @@
...
+    versions:
+      lastSuccessfulReconciliationBy: 0.37.0
+      kafka: 3.5.0
+  ```
+- User can now verify `Kafka.status.versions.lastSuccessfulReconciliationBy` and `Kafka.status.versions.kafka` to check if it has reconciled to correct/expected levels.
+- if there is an error at any point prior to the Kafka Broker rollout `Kafka.status.versions.kafka` will not be updated.
+- if there is an error at any point during reconcile `Kafka.status.versions.lastSuccessfulReconciliationBy` will not be updated.
+
+
+### Upgrade (from no mechansim)
```
im-konge, line 100:
```suggestion
### Upgrade (from no mechanism)
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,198 @@
...
+    versions:
+      lastSuccessfulReconciliationBy: 0.38.0
+      kafka: 3.5.1
+  ```
+- User can now verify `Kafka.status.versions.lastSuccessfulReconciliationBy` and `Kafka.status.versions.kafka` to check if it has reconciled to correct/expected levels.
+- if there is an error at any point prior to the Kafka Broker rollout `Kafka.status.versions.kafka` will not be updated.
+- if there is an error at any point during reconcile `Kafka.status.versions.lastSuccessfulReconciliationBy` will not be updated.
+
+
+### Upgrade (with new mechansim)
```
im-konge, line 130:
```suggestion
### Upgrade (with new mechanism)
```

im-konge (approved):
LGTM, thanks for the proposal :) just two nits

samuel-hawker:
Given the pre-existing fields in the status do not already set a precedent for nesting, I _think_ it makes sense for the versions fields to sit at the same level as `clusterID`, i.e. directly in the `status` field.

I will rename the fields to:
```
status.operatorLastSuccessfulVersion
status.kafkaVersion
```

FYI @scholzj @tombentley 

@ppatierno does this sound reasonable to you? Happy to tweak the names, or leave with the `versions` field if any new conversations/reasoning come to light.

tombentley (approved):
LGTM, thanks @samuel-hawker!

Frawless (approved):

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+Though this works, from a user standpoint it is far from ideal as it assumes a level of user knowledge about the cluster that may not be appropriate for all users.
+
+## Motivation
+
+This proposal allows a user to rely on the `Kafka` custom resources as their single source of truth, instead of watching/querying the individual kafka/zookeeper pods.
+This proposal lays out all the information that is required, and at what points in the reconcile this information will be used. This information is already readily available within the Reconcile loop, so orchestrating and figuring out if a component has been fully upgraded will be a case of propagating this information into the Custom Resource via status fields.
+
+## Proposal
+
+This proposal will cover the `Kafka` custom resource but could apply to other CR types that use `StrimziPodSets` such as `KafkaConnect` and `KafkaMirrorMaker2`, but could be extended to include other Custom Resources if there would be value in it.
```
PaulRMellor, line 20:
```suggestion
This proposal will cover the `Kafka` custom resource, but could apply to other CR types that use `StrimziPodSets` such as `KafkaConnect` and `KafkaMirrorMaker2`. It could be extended to include other custom resources if there is value in it.
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+This proposal starts with the `Kafka` Custom Resource as the base example as it is the core component of Strimzi, and it deploys both `StrimziPodSet`s and `Deployment`s.
```
PaulRMellor, line 21:
```suggestion
This proposal starts with the `Kafka` custom resource as the base example as it is the core component of Strimzi, and it deploys `StrimziPodSet` and `Deployment` resources.
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+status:
+  operatorLastSuccessfulVersion: '0.37.0'
+```
+If a reconcile does not reach completion, this field is not added or updated, however in the case where it already exists it is left at the previous value.
+i.e.
+`status.operatorLastSuccessfulVersion=0.35.0` would remain if a post-operator-upgrade reconciliation on `0.36.0` failed, as the information that a prior reconciliation passed would be useful to an end user.
+
+### status.kafkaVersion
+
+The `status.kafkaVersion` field is updated in the CR at the end of the KafkaReconciler in the `podsReady` method and signals to a user that the KafkaReconciler has finished reconciling the Kafka brokers to Kafka version `Y`, regardless of success or failure later in the reconcile.
```
PaulRMellor, line 43:
```suggestion
The `status.kafkaVersion` field is updated in the CR at the end of the KafkaReconciler in the `podsReady` method and signals to a user that the `KafkaReconciler` has finished reconciling the Kafka brokers to Kafka version `Y`, regardless of success or failure later in the reconcile.
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+```
+status:
+  kafkaVersion '3.5.1'
+```
+
+Note: This will update the field in the reconcilers internal status object and have it be written at the end of the reconciliation rather then having a separate status update API call each time this is updated.
```
PaulRMellor, line 49:
```suggestion
Note: This will update the field in the reconciler's internal status object and have it be written at the end of the reconciliation rather then having a separate status update API call each time this is updated.
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+
+This field is required as it signals that a reconcile to Kafka version `Y` has finished, and this extra information is important and needs to be available, as a user just checking `status.operatorLastSuccessfulVersion` is not sufficient for upgrades where multiple rolls of the brokers are required, such as a Kafka version update alongside an upgrade.
+
+If the operator errors during the Kafka reconciliation this field will not be updated, and will instead remain set to its prior set value, or unset if no previous value was set.
+This does mean that during an upgrade the value set in the `status.kafkaVersion` field will not reflect the kafka version correctly for all brokers, but this is desired, as the operator should only update the status field to reflect a finished upgrade so that a user can query the field to understand when an upgrade is finished and not for a breakdown broker-by-broker, if a user wants an understanding this granular this proposal does not cover this, and would instead recommend they check the kafka broker versions by querying the pods.
```
PaulRMellor, line 54:
```suggestion
During an upgrade, the value in the `status.kafkaVersion` field may not accurately reflect the Kafka version for all brokers. However, this behavior is intentional. The operator updates the status field only upon completing an upgrade, allowing users to determine upgrade completion at the cluster level, rather than on a broker-by-broker basis. For a more granular check, users seeking individual broker versions can query the Kafka broker versions through pod queries.
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+If Kafka was successfully updated, the field will also be updated, meaning errors later in the reconcile such as failing to deploy the `entity-operator`, `cruise-control`, `kafka-exporter`, or any other components added in the future to the `KafkaAssembly`; will still show `status.versions.kafka` correctly up to date, but `operatorLastSuccessfulVersion` will not be updated, and the CR will no longer be in a `Ready` state.
```
PaulRMellor, line 55:
```suggestion
If Kafka was successfully updated, the field will also be updated, meaning errors later in the reconcile such as failing to deploy the `entity-operator`, `cruise-control`, `kafka-exporter`, or any other components added in the future to the `KafkaAssembly`, will still show `status.versions.kafka` correctly up to date.  However, `operatorLastSuccessfulVersion` will not be updated, and the CR will no longer be in a `Ready` state.
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+  status:
+    operatorLastSuccessfulVersion: 0.37.0
+    kafkaVersion: 3.5.0
+  ```
+- User can now verify `Kafka.status.operatorLastSuccessfulVersion` and `Kafka.status.kafkaVersion` to check if it has reconciled to correct/expected levels.
+- if there is an error at any point prior to the Kafka Broker rollout `Kafka.status.kafkaVersion` will not be updated.
+- if there is an error at any point during reconcile `Kafka.status.operatorLastSuccessfulVersion` will not be updated.
+
+
+### Upgrade (from no mechanism)
```
PaulRMellor, line 96: This heading isn't displaying properly in the preview

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+## Examples
+
+### Fresh Install
+
+- On a fresh install, a user creates a Kafka CR:
+  ```
+  # the status is not actually set, but treat it as empty here
+  status: {}
+  ```
+- The KafkaReconciler reconciles the CR and creates `StrimziPodSet` CRs
```
PaulRMellor, line 76:
```suggestion
- The `KafkaReconciler` reconciles the CR and creates `StrimziPodSet` CRs
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+- StrimziPodSet reconciler reconciles the `SPS`s creating the Kafka Brokers
```
PaulRMellor, line 77:
```suggestion
- StrimziPodSet reconciler reconciles the StrimziPodSets creating the Kafka brokers
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+- Kafka Reconciler picks up that the Kafka Brokers managed by the SPS are now Ready and finished installing.
+  Updating the status with the reconciled kafka versions:
+  ```
+  status:
+    operatorLastSuccessfulVersion: 0.37.0
+    kafkaVersion: 3.5.1
+  ```
+- Kafka Reconciler continues to install other components such as entity operator.
```
PaulRMellor, line 143:
```suggestion
- Kafka reconciler continues to install other components such as Entity Operator.
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+  Once all components are deployed and updated, update `Kafka` CR with
```
PaulRMellor, line 144:
```suggestion
  Once all components are deployed and updated, update `Kafka` CR with:
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+  ```
+  status:
+    operatorLastSuccessfulVersion: 0.38.0
+    kafkaVersion: 3.5.1
+  ```
+- User can now verify `Kafka.status.operatorLastSuccessfulVersion` and `Kafka.status.kafkaVersion` to check if it has reconciled to correct/expected levels.
```
PaulRMellor, line 150:
```suggestion
- User can now verify `Kafka.status.operatorLastSuccessfulVersion` and `Kafka.status.kafkaVersion` to check if it has reconciled to the expected versions.
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+- if there is an error at any point prior to the Kafka Broker rollout `Kafka.status.kafkaVersion` will not be updated.
+- if there is an error at any point during reconcile `Kafka.status.operatorLastSuccessfulVersion` will not be updated.
```
PaulRMellor, line 152:
```suggestion
- if there is an error at any point prior to the Kafka broker rollout, `Kafka.status.kafkaVersion` will not be updated.
- if there is an error at any point during reconcile, `Kafka.status.operatorLastSuccessfulVersion` will not be updated.
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+
+## `KafkaNodePool` considerations
+
+It might be that in the future each `KafkaNodePool` CR supports having its own Kafka version which might differ from other `KafkaNodePools` in the same Kafka cluster.
+This proposal should work with this method of upgrade, since the Kafka version + readiness checks will be track by the `StrimziPodSet`, so that when a user had finished upgrading via this method, the new `status.kafkaVersion` should be updated accordingly.
```
PaulRMellor, line 157:
```suggestion
This proposal should work with this method of upgrade, since the Kafka version + readiness checks will be tracked by the `StrimziPodSet`, so that when a user had finished upgrading via this method, the new `status.kafkaVersion` should be updated accordingly.
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+In the potential case where a user is using multiple `KafkaNodePool`s for a cluster using different Kafka Versions, it is my recommendation that `status.kafkaVersion` is a comma delimitered (sequentially ordered) list of all these Kafka Versions rather than the minimum or maximum kafka version.
```
PaulRMellor, line 158:
```suggestion
In the potential case where a user is using multiple `KafkaNodePool`s for a cluster using different Kafka versions, it is my recommendation that `status.kafkaVersion` is a comma delimitered (sequentially ordered) list of all these Kafka Versions rather than the minimum or maximum Kafka version.
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+To the best of my knowledge only the `strimzi-kafka-operator` would be affected by this mechanism, including the code changes to the reconcilers and the doc changes.
+
+## Compatibility
+
+This mechanism is backwards compatible, in the sense that it works even if the status fields were not set on prior versions or reconciles.
+No compatibility issues, unless this feature was later removed in which case a user would no longer be able to rely on this status field.
+
+## Rejected alternatives
+
+- Using two new metadata annotations both `strimzi.io/reconciled` and `strimzi.io/reconciling` annotations, which would be a two staged update, so as to distinguish between a start of a reconcile and a completed one on the current operator version. However annotations can be overwritten by users.
```
PaulRMellor, line 172:
```suggestion
- Using `strimzi.io/reconciled` and `strimzi.io/reconciling` annotations, two new metadata annotations which would be a two-stage update, so as to distinguish between a start of a reconcile and a completed one on the current operator version. However annotations can be overwritten by users.
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+
+- Just checking the pods from the StrimziPodSets, this is the current mechanism, but is hardly ideal, given there can be quite a few pods, and the KafkaAssembly already has a step to wait for all pods to be rolled, we can add the logic for updating the status to happen after this.
```
PaulRMellor, line 174:
```suggestion
- Just checking the pods from the StrimziPodSets, this is the current mechanism, but is hardly ideal, given there can be quite a few pods, and the `KafkaAssembly` already has a step to wait for all pods to be rolled, we can add the logic for updating the status to happen after this.
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+
+- Having a field `Kafka.status.versions.managedBy` that is updated at the beginning of reconcile, this has little value to the user and might just cause more confusion.
```
PaulRMellor, line 176:
```suggestion
- Having a field `Kafka.status.versions.managedBy` that is updated at the beginning of reconcile, which has little value to the user and might just cause more confusion.
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+
+- It may be useful to also include versions for the `inter-broker-protocol-version` and `log-message-format-version` in the status, such as `status.versions.kafkaInterBrokerProtocol` and `status.versions.kafkaLogMessageFormat` as these too could cause multiple reconciles and a user might want to verify that these have been picked up by the brokers. However these are not used in KRaft as `log.message.format.version` is ignored if `inter.broker.protocol.version` is 3.0 or higher. `inter.broker.protocol.version` is not deprecated formally. But it is not used in KRaft mode. So it will be essentially removed in Kafka `4.0`. So removing it from this proposal.
+
+- The `Kafka.status.kafkaVersion` field being an array of `String`s for the purposes of covering the multiple-kafka versions scenario in the [KafkaNodePool section](#-`KafkaNodePool`-considerations).
+  i.e
```
PaulRMellor, line 181:
```suggestion
  For example:
```

052-record-reconciled-versions.md
```diff
@@ -0,0 +1,190 @@
...
+  ```
+  status:
+    kafkaVersion:
+    - '3.5.0'
+    - '3.5.1'
+  ```
+  since this is unlikely to be supported, and reading a string rather than an array is easier, a single string for the field seems preferable.
```
PaulRMellor, line 188:
```suggestion
Since this is unlikely to be supported, and reading a string rather than an array is easier, a single string for the field seems preferable.
```

samuel-hawker:
@PaulRMellor thank you so much for such a thorough review, addressed every comment.
Functionally no changes to proposal.
We have so many approvals now - are there any more thoughts or can we move this towards a merge?

scholzj:
@samuel-hawker I usually say something like `If no new comments are received until tomorrow EOB, it will be merged` and give one more day or so to everyone to have a look. Please do not forget to rename the file (52 is already taken) and add it to the index in README.md before merging it 😉.

samuel-hawker:
Rebased branch + renamed file + added to index - thanks for the reminder Jakub!

**If no new comments are received until tomorrow end of day, it will be merged** 

Thank you everyone for helping to refine this proposal!

samuel-hawker:
Merged. Thanks again everyone.
Raised an issue: https://github.com/strimzi/strimzi-kafka-operator/issues/9002 and self-assigned

PR #86, opened 2023-08-04
Ability to create CC API users

kyguy:
This proposal is about giving developers the ability to create REST API users for the Cruise Control REST API. This would allow roles and permissions to be defined to allow developers and third-party applications to access the Cruise Control REST API without having to disable HTTP basic authentication.


052-cruise-control-api-users.md
```diff
@@ -0,0 +1,204 @@
...
+* `heathcheck` - Used by Strimzi for checking the readiness of the Cruise Control application.
+
+The access to these API users are limited to Strimzi.
+So when HTTP basic authentication is enabled for Cruise Control, no one can access the the Cruise Control API other than Strimzi.
+
+## Motivation
+
+There are a couple use cases where developers would want to access the Cruise Control API without disabling API security:
+
+* Debugging Cruise Control in a production environment
```
scholzj, line 20: What does this mean? Debugging in production sounds like a bad idea in the first place. So why should it be supported?

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,204 @@
...
+* Running the Cruise Control UI alongside a Strimzi-managed Kafka cluster
```
scholzj, line 21: You should probably explain what does the UI give to the users over the KafkaRebalance resources and / or Grafana dashabords. I.e., why should one want it?

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,204 @@
...
+So when HTTP basic authentication is enabled for Cruise Control, no one can access the the Cruise Control API other than Strimzi.
+
+## Motivation
+
+There are a couple use cases where developers would want to access the Cruise Control API without disabling API security:
+
+* Debugging Cruise Control in a production environment
+* Running the Cruise Control UI alongside a Strimzi-managed Kafka cluster 
+
+In order to allow developers and third-party applications to access the Cruise Control API without needing to disable API security, we would need to provide a method of creating API users.
```
scholzj, line 23: Again, why would they want/need it? What will they be able to do over the API?

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,204 @@
...
+
+Although this would simplify the overall design it would also shift a lot of the maintanence responsibility to the developer.
+The developer would need to create a valid Cruise Control API user configuration and encode it into a secret.
+
+## Rejected alternatives
+
+### Move API user configuration ConfigMap
+
+Having the API users configuration exposed in a `ConfigMap` would significantly reduce the code complexity of configuration management.
+That being said, the API users configuration contains sensitive information such as passwords and therefore should not be exposed via a ConfigMap.
```
scholzj, line 194: Why is ConfigMap an option here? I think the right Kubernetes resource is Secret which will have the same complexity but not the sensitivity drawback. So I do not follow why was this considered.

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,204 @@
...
+USER role: has access to all the GET endpoints except bootstrap and train.
+ADMIN role: has access to all endpoints.
+```
+
+The problem with this approach is that it would add more complexity to the code base since it would require:
+* A new Strimzi `CruiseControlUser` custom resource
+* A watch on the new `CruiseControlUser` resource
+* A reconciler to manage the `CruiseControlUser` resource.
+
+### Developer managed API user configuration
```
scholzj, line 166: This is probably something I personally would prefer.

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,204 @@
...
+These resources are watched and managed by the the Strimzi User Operator, which creates Kafka users based on the descriptions provided by the resources.
+
+For more details, checkout the Strimzi documentation [here](https://strimzi.io/docs/operators/latest/deploying#proc-configuring-kafka-user-str)
+
+#### KafkaConnect users
+
+At this time, Strimzi doesn't offer any method of creating or configuring KafkaConnect API users.
+That being said, the direction we take with creating and managing Cruise Control API users could be used as a precedent for its design in the future.
+
+## Alternatives
```
scholzj, line 123: What does this mean? Are you proposing to implement all of these in parallel? Or did you reject these? I think this should have some more clarity.

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,204 @@
...
+spec:
+  cruiseControl:
+    apiUsersConfig:
+      secret:
+        secretName: cruise-control-api-users-secret
+     ...
+```
+
+Although this would simplify the overall design it would also shift a lot of the maintanence responsibility to the developer.
+The developer would need to create a valid Cruise Control API user configuration and encode it into a secret.
```
scholzj, line 187: I do not see this really as a downside. Our chosen way to support Cruise Control is through the operator:
* Via the KafkaRebalance custom resource
* Via automated rebalancing at scale-down / scale-up
* Via automated rebalancing when some threshold of _unbalancedness_ is reached

You proposal does not dispute this -> so it would not be a primary way of interacting with Cruise Control and the effort and complexity on Strimzi side should correspond to it.

scholzj (commented):
Thanks for the proposal. I would like it to see address in more detail following aspects:
* What are the use cases and the value for this. You list things which someone might be able to do with it. But do not explain the added value of this feature, what new things will it enable compared to the other ways to use it.
* It is not clear what the risks and possible conflicts are between users using the API and the operator. What if user is running a rebalance through the API while the operator tries to run another one etc.?
* You should explain how does this align with our preferred way of using Cruise Control -> i.e. through the operator.

I think these things should explain whether this makes sense and how much of the complexity should be on the user side versus on the Strimzi side.

ppatierno:
Thanks for the proposal Kyle.
I remember that in the past we already had discussions about allowing users accessing API just as a `VIEWER` in order to be able to use the Cruise Control UI but just as a "visual" way to get information instead of looking at `KafkaRebalance` custom resources. With this proposal we are enabling the creation of an `ADMIN` user as well who could run a rebalance without using `KafkaRebalance` custom resource which could make also a conflicts if two rebalances are started in two different ways. Why are we going to allow something like this with this proposal?

052-cruise-control-api-users.md
fvaleri: As others say, if we enable admin operations for third-party, there will be a race condition between Strimzi and these applications, much like the race condition we have today between CC and KafkaRoller. The big difference is that we don't control the third-party applications.

fvaleri (commented):
Hi @kyguy, thanks for the proposal. I also have some doubts about enabling users with ADMIN role.

This is what you say in your blog post:

> If using the Cruise Control UI with a Strimzi-managed Kafka cluster, it is best to stick with using it solely for monitoring your cluster, not for operating on it.

Assuming the CC UI can run with just the USER role, this is something we could provide without any side effect.

052-cruise-control-api-users.md
kyguy: > For example, when the issue is environmental, and can't be reproduces in lower environments.

I know it is a weak justification but this was what I was getting at.

> The thing I don't see is which debugging data would be enabled, that are currently not available in other forms (status, logs, events), and how you can use them?

Some of the statistical information surrounding cluster and partition load and user tasks is either only available or more easily gathered through API GET requests.  I have updated the the proposal to address this feedback, let me know what you think!

052-cruise-control-api-users.md
kyguy: Some of the statistical information surrounding cluster and partition load and user tasks is only available through the Cruise Control API,  I have updated the the proposal to address this feedback, let me know what you think!

052-cruise-control-api-users.md
kyguy: Agreed, just added some more justifications in recent changed, let me know what you think!

052-cruise-control-api-users.md
kyguy: You are right, removed

052-cruise-control-api-users.md
kyguy: I have read all the feedback and thought about this a lot and I understand your reasoning and agree that this implementation would make more sense given the circumstances. I have just updated the proposal to highlight this a the suggested approach.

052-cruise-control-api-users.md
kyguy: No no, I just wasn't sure which direction to take, but given the feedback, I have moved all alternatives to the rejected alternatives section

052-cruise-control-api-users.md
kyguy: This make sense

kyguy:
Thank you for the feedback @scholzj @fvaleri @ppatierno, it is much appreciated. I agree that we should better balance the implementation complexity with the significance of the feature. Since  accessing the Cruise Control API directly is not the support or recommend method of interacting with Cruise Control, we should probably keep the implementation of the feature as simple as possible

Also, we probably should have the `ADMIN` role disabled given the risks that is poses, however, in the latest edits of the proposal I left this choice a open ended in case anyone had any other thoughts. Would it be so bad if we allowed users to experiment with the ADMIN role with a caveat? Maybe for users that knew what they were doing and wanted more control?

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,220 @@
...
+### Goals
+
+* A method of creating Cruise Control API users.
+
+### Developer managed API user configuration
+
+Having the developer manage a single secret with their custom Cruise Control API user configuration would greatly reduce the complexity of the proposed design.
+Strimzi could add the contents of this secret to the Cruise Control API user configuration file containing Strimzi's `admin` and `healthcheck` API users.
+
+The schema could look something like this:
```
scholzj, line 42: You should propose the API ... not something like the API ;-)

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,220 @@
...
+```yaml=
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  ...
+spec:
+  cruiseControl:
+    apiUsersConfig:
+      secret:
+        secretName: cruise-control-api-users-secret
```
scholzj, line 53: Please provide details of what will be inside the secret.

scholzj, line 53: To have more flexibility for the future, you can add maybe some type to it. E.g.

```suggestion
    apiUsers:
      type: htaccess # Or whatever the right name would be for what is in the secret.
      secret:
        secretName: cruise-control-api-users-secret
```

This will leave it open to add other `type` in the future with some other options.

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,220 @@
...
+However, given that accessing the Cruise Control API directly is neither the supported or primary method of interacting with Strimzi Cruise Control, this method would not be worth the complexity that it would add to the code base.
+
+## Future work
+
+This proposal aims to set the direction and define the semantics of Cruise Control API users and maybe set the precedent for creating API users for other Kafka components like KafkaConnect.
+It is not intended, by itself, to define the complete picture.
+Future work _could_ include:
+
+* Configurable API user roles
+* Password renewals for generated passwords
```
scholzj, line 219: Is this still relevant for the new method for managing the users? I would assume that the user just changes the secret to _renew_ passwords.

scholzj (commented):
I left some more comments. But I feel good about the direction. As for the ADMIN role, I'm indifferent -> as it is now more clear that this is an _expert_ option and assuming some documentation warnings, I would fine to have it there. But if others think it should be disabled, it would be fine for me as well.

052-cruise-control-api-users.md
fvaleri: +1

fvaleri: The type is tricky here. The mechanism is similar to htaccess, but not the same. It's basically a properties file with Jetty's HashLoginService's credentials file format (user: password,role). We could use `jetty-realm` as type id.

fvaleri (approved):
We clearly documented the risks of using the ADMIN role. 

Added few minor comments, but LGTM.

052-cruise-control-api-users.md
scholzj: TBH, I do not really care about the name of the type -> as I said in my comment, it should be adjusted to what it really is. `htaccess` was just an example. But it could be `type: fromSecret` or whatever else fits what it does.

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,220 @@
...
+So when HTTP basic authentication is enabled for Cruise Control, no one can access the the Cruise Control API other than Strimzi.
+
+## Motivation
+
+There are a couple use cases where developers would want to access the Cruise Control API without disabling API security:
+
+* Monitoring a Strimzi-managed Kafka cluster with the Cruise Control UI.
+* Gathering Cruise Control specific statistical information that is not currently available via Strimzi Operator or Cruise Control sensor metrics e.g. detailed information surrounding cluster and partition load and user tasks.
+* Experiment with Cruise Control operations not currently supported by Strimzi.
+* Debugging Cruise Control in a secured environments.
```
tombentley, line 23:
```suggestion
* Debugging Cruise Control in a secured environment.
```

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,220 @@
...
+### Risks
+
+Of course, supporting such a feature could come with some risks.
+If we were to support developers to create Cruise Control API users with the `ADMIN` role, those users would have access to potentially destructive cluster operations.
+For example, an API user with the ADMIN role would be able to execute a rebalance opeation while the Strimzi Operator is in the middle of preforming a rolling update of the Kafka brokers.
+
+That being said, supporting developers to create Cruise Control API users with non-`ADMIN` roles like `USER` and `VIEWER` roles would mitigate this risk completely.
+This is because the `USER` and `VIEWER` roles are read-only roles and would not conflict with Strimzi cluster operations.
+
+We could support the creation of API users with `ADMIN` roles with a caveat so that developers could experiment with Cruise Control write operations or we could simply disable it completely.
```
tombentley, line 71: A proposal should be written in terms of what you _actually propose doing_. You can elaborate on why that's what's being proposed if necessary. And you can use the Rejected Alternatives section to forestall obvious variations which people might have.

052-cruise-control-api-users.md
tombentley: +1. It's completely unclear to me, right now, how we associate a role with each user. If that's not part of this secret please explain how that would work.

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+
+There are a couple use cases where developers would want to access the Cruise Control API without disabling API security:
+
+* Monitoring a Strimzi-managed Kafka cluster with the Cruise Control UI.
+* Gathering Cruise Control specific statistical information that is not currently available via Strimzi Operator or Cruise Control sensor metrics e.g. detailed information surrounding cluster and partition load and user tasks.
+* Experiment with Cruise Control operations not currently supported by Strimzi.
+* Debugging Cruise Control in a secured environment.
+
+At this time, Strimzi Cruise Control integration offers limited access to Cruise Control data and features.
+Although this limited access protects developers from preforming potentially dangerous cluster operations it also prevents developers from accessing harmless monitoring data.
```
scholzj, line 26: `preventing them from dangerous cluster operations` sounds like preventing users from using advanced features. But my understanding was that there are serious concerns that this will interfere with the Strimzi Operators. I think these are two different things where as any conflicts with Strimzi are significantly more serious. So this deserves a clarification.

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+
+## Proposal
+
+### Goals
+
+* A method of creating Cruise Control API users.
+
+### Developer managed API user configuration
+
+Having the developer manage a single secret with their custom Cruise Control API user configuration would greatly reduce the complexity of the proposed design.
```
scholzj, line 39: I'm not sure I follow this sentence. Maybe it can be removed?

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  ...
+spec:
+  cruiseControl:
+    apiUsers:
+      type: fromSecret (1)
+      secret:
+        secretName: cruise-control-api-users-secret
```
scholzj, line 89: I guess it should have also a reference to the field under which the file is stored?

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+     ...
+```
+(1) A type field is added here for configuration flexibility in the future.
+
+Strimzi would decode and use the contents of this secret to populate the Cruise Control API auth credentials file which would be mounted and used by the Cruise Control pod.
+
+Although this would simplify the overall design it would also shift a lot of the maintanence responsibility to the developer.
+The developer would need to create a valid Cruise Control API user configuration and encode it into a secret.
+That being said, this is still a reasonable amount of responsibility to leave to the developer.
+This is because accessing the Cruise Control API directly is neither the supported or primary way of interacting with Cruise Control for a Strimzi-managed Kafka cluster.
```
scholzj, line 99: Maybe like this, it would sound better? _Supported_ is a bit confusing term in OSS project. We do not provide any support as a _helpline_. Everything what the code allows is basically supported and unsupported in the same time.

```suggestion
This is because accessing the Cruise Control API directly is not the primary way of interacting with Cruise Control for a Strimzi-managed Kafka cluster and is expected to be used only by advanced users in special use-cases.
```

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+
+Cruise Control's auth credential files follow Jetty's HashLoginService's file format: `username: password [,rolename ...]`.
+
+By default Cruise Control defines three roles: `VIEWER`, `USER` and `ADMIN`.
+```
+VIEWER role: has access to the most lightweight kafka_cluster_state, user_tasks and review_board endpoints.
+USER role: has access to all the GET endpoints except bootstrap and train.
+ADMIN role: has access to all endpoints.
+```
+For more details checkout the [Cruise Control Wiki](https://github.com/linkedin/cruise-control/wiki/Security#authorization)
```
scholzj, line 49: Should this say that at this point only the USER and VIEWER roles will be supported if that is the plan?

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+If we were to support developers to create Cruise Control API users with the `ADMIN` role, those users would have access to potentially destructive cluster operations.
+For example, an API user with the ADMIN role would be able to execute a rebalance opeation while the Strimzi Operator is in the middle of preforming a rolling update of the Kafka brokers.
+
+For the above reasons, this proposal advocates for disabling the creation of API users with `ADMIN` roles for now.
+
+We can always enable the `ADMIN` role at a later date with a caveat in the documentation so that advanced developers could experiment with Cruise Control write operations at their own risk.
+
+That being said, supporting developers to create Cruise Control API users with non-`ADMIN` roles like `USER` and `VIEWER` roles would mitigate this risk completely.
+This is because the `USER` and `VIEWER` roles are read-only roles and would not conflict with Strimzi cluster operations.
+So this proposal if for enabling the `USER` and `VIEWER` roles only.
```
scholzj, line 114: Should this be moved to rejected alternatives explaining that the ADMIN role was considered but rejected (at least for now)?

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+      ...
+```
+These resources are watched and managed by the the Strimzi User Operator, which creates Kafka users based on the descriptions provided by the resources.
+
+For more details, checkout the Strimzi documentation [here](https://strimzi.io/docs/operators/latest/deploying#proc-configuring-kafka-user-str)
+
+#### KafkaConnect users
+
+At this time, Strimzi doesn't offer any method of creating or configuring KafkaConnect API users.
+That being said, the direction we take with creating and managing Cruise Control API users could be used as a precedent for its design in the future.
```
scholzj, line 150: The flow of this seems a bit weird as you already covered how it should be handled in CC. But I guess it might be useful for some readers? Should we move it to the _Current Situation_ section? Or should we frame it more how does the CC user management compare with other parts of Strimzi?

scholzj (commented):
I left some more comments. Mostly about the phrasing, etc. But I think in general the proposal looks good.

052-cruise-control-api-users.md
PaulRMellor: Are we introducing the proposal here?  In which case, maybe something like: "_The proposal is to incorporate a developer-managed approach for configuring a single secret with customized Cruise Control API user settings._"

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Goals
+
+* A method of creating Cruise Control API users.
+
+### Developer managed API user configuration
+
+Having the developer manage a single secret with their custom Cruise Control API user configuration would greatly reduce the complexity of the proposed design.
+
+Cruise Control's auth credential files follow Jetty's HashLoginService's file format: `username: password [,rolename ...]`.
```
PaulRMellor, line 41:
```suggestion
Cruise Control's authentication credential files adhere to Jetty's HashLoginService's file format: `username: password [,rolename ...]`.
```

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+
+By default Cruise Control defines three roles: `VIEWER`, `USER` and `ADMIN`.
+```
+VIEWER role: has access to the most lightweight kafka_cluster_state, user_tasks and review_board endpoints.
+USER role: has access to all the GET endpoints except bootstrap and train.
+ADMIN role: has access to all endpoints.
+```
+For more details checkout the [Cruise Control Wiki](https://github.com/linkedin/cruise-control/wiki/Security#authorization)
+
+So a developer could create a Cruise Control API auth credentials file, `cruise-control-auth.txt`, that looked like this:
```
PaulRMellor, line 51:
```suggestion
So a developer could create a Cruise Control API authentication credentials file named `cruise-control-auth.txt`, with the following content:
```

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cruise-control-api-users-secret
+type: Opaque
+data:
+  key: dXNlck9uZTogcGFzc3dvcmRPbmUsIFVTRVIKdXNlclR3bzogcGFzc3dvcmRUd28sIFZJRVdFUgo=
+```
+
+The developer could then reference the secret in an `spec.cruiseControl.apiUsers` section of the `Kafka` resource.
```
PaulRMellor, line 75:
```suggestion
The developer could then reference the secret in a `spec.cruiseControl.apiUsers` section of the `Kafka` resource.
```

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+```
+For more details checkout the [Cruise Control Wiki](https://github.com/linkedin/cruise-control/wiki/Security#authorization)
+
+So a developer could create a Cruise Control API auth credentials file, `cruise-control-auth.txt`, that looked like this:
+```
+userOne: passwordOne, USER
+userTwo: passwordTwo, VIEWER
+```
+
+Then the developer could create a secret using that file:
```
PaulRMellor, line 57:
```suggestion
Then the developer can use this file to create a secret using the following command:
```

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+
+```
+kubectl create secret generic cruise-control-api-users-secret  --from-file=key=cruise-control-auth.txt
+```
+
+This would create a Secret:
```
PaulRMellor, line 63:
```suggestion
The command would create a secret:
```

052-cruise-control-api-users.md
PaulRMellor: I don't know if we want to say anything here about the need to document the approach and underlining in the documentation that this is for advanced Cruise Control operations?

Something like: "However, with proper supporting documentation, this new feature will give developers a greater degree of control and assistance with certain use-cases."

052-cruise-control-api-users.md
PaulRMellor, line 114:
```suggestion
So this proposal is for enabling the `USER` and `VIEWER` roles only.
```

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+USER role: has access to all the GET endpoints except bootstrap and train.
+ADMIN role: has access to all endpoints.
+```
+
+The problem with this approach is that it would add significant complexity to the code base since it would require:
+* A new Strimzi `CruiseControlUser` custom resource
+* A watch on the new `CruiseControlUser` resource
+* A reconciler to manage the `CruiseControlUser` resource.
+ 
+Given that accessing the Cruise Control API directly is neither the supported or primary method of interacting with Strimzi Cruise Control, this method would not be worth the complexity that it would add to the code base.
```
PaulRMellor, line 195: Maybe make the change suggested earlier by Jakub regarding _supported_ here too

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+For more details checkout the [Cruise Control Wiki](https://github.com/linkedin/cruise-control/wiki/Security#authorization)
+```
+VIEWER role: has access to the most lightweight kafka_cluster_state, user_tasks and review_board endpoints.
+USER role: has access to all the GET endpoints except bootstrap and train.
+ADMIN role: has access to all endpoints.
+```
+
+Strimzi would then aggregate the Cruise Control API auth credentials into a centralized configuration, then store it in a secret where it would be mounted and used by Cruise Control.
+
+However, given that accessing the Cruise Control API directly is neither the supported or primary method of interacting with Strimzi Cruise Control, this method would not be worth the complexity that it would add to the code base.
```
PaulRMellor, line 253: Maybe make the change suggested earlier by Jakub regarding _supported_ here too

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+  ...
+spec:
+  cruiseControl:
+    apiUsers:
+      type: fromSecret (1)
+      secret:
+        secretName: cruise-control-api-users-secret
+     ...
+```
+(1) A type field is added here for configuration flexibility in the future.
```
PaulRMellor, line 92: How might this be used? Do we want to say anything about that here?

052-cruise-control-api-users.md
PaulRMellor: Yeah, I might be tempted to move it between Current Situation and Motivation. But say that the proposal is only about CC for now in the moved section.

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,261 @@
...
+We can always enable the `ADMIN` role at a later date with a caveat in the documentation so that advanced developers could experiment with Cruise Control write operations at their own risk.
+
+That being said, supporting developers to create Cruise Control API users with non-`ADMIN` roles like `USER` and `VIEWER` roles would mitigate this risk completely.
+This is because the `USER` and `VIEWER` roles are read-only roles and would not conflict with Strimzi cluster operations.
+So this proposal if for enabling the `USER` and `VIEWER` roles only.
+
+### How API user management is handled elsewhere
+
+Cruise Control is not the only Kafka component which has API users.
+For consitency across the code base, we should also consider the following components when designing our solution:
```
PaulRMellor, line 119:
```suggestion
For consistency across the code base, we also considered the following components when designing our solution:
```

PaulRMellor (approved):
Nice, clear proposal Kyle. Docs will be important here :+1: 
Few minor suggestions

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,262 @@
...
+kind: Kafka
+metadata:
+  ...
+spec:
+  cruiseControl:
+    apiUsers:
+      type: fromSecret (1)
+      secret:
+        secretName: cruise-control-api-users-secret
+        secretKey: key
```
scholzj, line 125: Sorry for not noticing it before. But what about this API which maybe better corresponds to the content and to the other APIs:

```suggestion
    apiUsers:
      type: hashloginservice
      valueFrom:
        secretKeyRef:
          name: cruise-control-api-users-secret
          key: key
```

This way, you:
* Name the type after the type of content (i.e. how do the data you get look like, which is Jetty HashLoginService). And because of using the `type` field, more types can be added in the future
* You have a `valueFrom` construct which allows you to add more sources in the future if needed. This is also a pattern used in other Strimzi APIs already (e.g. logging configuration, password configuration in KafkaUser resuorces, metrics configuration, etc.)

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,262 @@
...
+
+Although this would simplify the overall design it would also shift a lot of the maintanence responsibility to the developer.
+The developer would need to create a valid Cruise Control API user configuration and encode it into a secret.
+That being said, this is still a reasonable amount of responsibility to leave to the developer.
+This is because accessing the Cruise Control API directly is not the primary way of interacting with Cruise Control for a Strimzi-managed Kafka cluster and is expected to be used only by advanced developers in special use-cases.
+However, with proper supporting documentation, this new feature will give developers a greater degree of control and assistance with special use-cases.
+
+### Risks
+
+Supporting developers to create Cruise Control API users with non-`ADMIN` roles like `USER` and `VIEWER` roles would mitigate this risk completely.
```
scholzj, line 140: Is this missing some previous sentence that defines the `this risk`? Or maybe you should instead of `this risk` reference the risks described int he _Motivation_ section.

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Current situation
+
+At this time, Strimzi creates two Cruise Control API users:
+
+* `admin`      - Used by Strimzi for Cruise Control admin operations.
+* `heathcheck` - Used by Strimzi for checking the readiness of the Cruise Control application.
+
+Access to these API users is limited to Strimzi.
+So when HTTP basic authentication is enabled for Cruise Control, no one can access the the Cruise Control API other than Strimzi.
```
PaulRMellor, line 14:
```suggestion
So when HTTP basic authentication is enabled for Cruise Control, no one can access the Cruise Control API other than Strimzi.
```

scholzj (commented):
Two more nit, sorry.

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,262 @@
...
+spec:
+  cruiseControl:
+    apiUsers:
+      type: fromSecret (1)
+      secret:
+        secretName: cruise-control-api-users-secret
+        secretKey: key
+     ...
+```
+(1) A type field is added here for configuration flexibility in the future. e.g. Gives us the option to have other methids of configuring API users other than using `secrets`
```
PaulRMellor, line 128:
```suggestion
(1) A type field is added here for configuration flexibility in the future. For example, it gives us the option to have other methods of configuring API users other than using secrets.
```

PaulRMellor (approved):
Couple of nits, but like the changes :+1:

scholzj (approved):
Thanks @kyguy. Nice work.

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,269 @@
...
+
+There are a couple use cases where developers would want to access the Cruise Control API without disabling API security:
+
+* Monitoring a Strimzi-managed Kafka cluster with the Cruise Control UI.
+* Gathering Cruise Control specific statistical information that is not currently available via Strimzi Operator or Cruise Control sensor metrics e.g. detailed information surrounding cluster and partition load and user tasks.
+* Experimenting with Cruise Control operations not currently supported by Strimzi.
+* Debugging Cruise Control in a secured environment.
+
+At this time, Strimzi Cruise Control integration offers limited access to Cruise Control data and features.
+This limited access protects developers from preforming cluster operations that coud interfere with Strimzi Operator operations and cause damage to their clusters.
```
ppatierno, line 62:
```suggestion
This limited access protects developers from performing cluster operations that could interfere with Strimzi Operator operations and cause damage to their clusters.
```

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,269 @@
...
+
+In order to allow developers and third-party applications to access the Cruise Control API without needing to disable API security, we would need to provide a method of creating API users.
+
+## Proposal
+
+### Goals
+
+* A method of creating Cruise Control API users.
+
+### Developer managed API user configuration
```
tombentley, line 73: In this section you're talking a lot about "developers". It's not really what kind of developer you're really thinking of. The motivation section, and the reliance on the `Kafka` CR makes it seem like this more aimed at "Kafka cluster admin", than "Kafka application developers" (i.e. people writing producer and consumer applications)

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,269 @@
...
+This is because accessing the Cruise Control API directly is not the primary way of interacting with Cruise Control for a Strimzi-managed Kafka cluster and is expected to be used only by advanced developers in special use-cases.
+However, with proper supporting documentation, this new feature will give developers a greater degree of control and assistance with special use-cases.
+
+### Risks
+
+Direct calls to the Cruise Control API like the write operations enabled by the `ADMIN` role could interfere with Strimzi Operator operations and cause damage to a Strimzi-managed Kafka cluster.
+
+Supporting developers to create Cruise Control API users with non-`ADMIN` roles like `USER` and `VIEWER` roles would mitigate this risk completely.
+This is because the `USER` and `VIEWER` roles are read-only roles and would not conflict with Strimzi cluster operations.
+So this proposal is for enabling the `USER` and `VIEWER` roles only.
```
tombentley, line 149: I'd like for the proposal to define:

* What will happen if the user specifies a file with the `ADMIN` role? 
* What will happen if the user specifies a corrupt file (not in the expected format)?

tombentley (commented):
Thanks @kyguy, I left a few minor comments.

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,269 @@
...
+* `admin`      - Used by Strimzi for Cruise Control admin operations.
+* `heathcheck` - Used by Strimzi for checking the readiness of the Cruise Control application.
+
+Access to these API users is limited to Strimzi.
+So when HTTP basic authentication is enabled for Cruise Control, no one can access the Cruise Control API other than Strimzi.
+
+### How API user management is handled elsewhere
+
+Cruise Control is not the only Kafka component which has API users.
+Although this proposal is stictly about Cruise Control API users, for consistency across the code base, we also considered the following components when designing our solution:
```
im-konge, line 19:
```suggestion
Although this proposal is strictly about Cruise Control API users, for consistency across the code base, we also considered the following components when designing our solution:
```

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,269 @@
...
+Supporting developers to create Cruise Control API users with non-`ADMIN` roles like `USER` and `VIEWER` roles would mitigate this risk completely.
+This is because the `USER` and `VIEWER` roles are read-only roles and would not conflict with Strimzi cluster operations.
+So this proposal is for enabling the `USER` and `VIEWER` roles only.
+
+## Rejected alternatives
+
+### Supporting `ADMIN` role
+
+If we were to support developers to create Cruise Control API users with the `ADMIN` role, those users would have access to potentially destructive cluster operations.
+For example, an API user with the ADMIN role would be able to execute a rebalance opeation while the Strimzi Operator is in the middle of preforming a rolling update of the Kafka brokers.
```
im-konge, line 156:
```suggestion
For example, an API user with the ADMIN role would be able to execute a rebalance operation while the Strimzi Operator is in the middle of performing a rolling update of the Kafka brokers.
```

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,269 @@
...
+For more details checkout the [Cruise Control Wiki](https://github.com/linkedin/cruise-control/wiki/Security#authorization)
+```
+VIEWER role: has access to the most lightweight kafka_cluster_state, user_tasks and review_board endpoints.
+USER role: has access to all the GET endpoints except bootstrap and train.
+ADMIN role: has access to all endpoints.
+```
+
+Strimzi would then aggregate the Cruise Control API auth credentials into a centralized configuration, then store it in a secret where it would be mounted and used by Cruise Control.
+
+Given that accessing the Cruise Control API directly is not the primary way of interacting with Cruise Control for a Strimzi-managed Kafka cluster and is expected to be used only by advanced users in special use-casesi, this method would not be worth the complexity that it would add to the code base.
```
im-konge, line 261:
```suggestion
Given that accessing the Cruise Control API directly is not the primary way of interacting with Cruise Control for a Strimzi-managed Kafka cluster and is expected to be used only by advanced users in special use-cases, this method would not be worth the complexity that it would add to the code base.
```

im-konge (approved):

ShubhamRwt (approved):
LGTM apart from the pending comments

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,269 @@
...
+At this time, Strimzi doesn't offer any method of creating or configuring KafkaConnect API users.
+That being said, the direction we take with creating and managing Cruise Control API users could be used as a precedent for its design in the future.
+
+## Motivation
+
+There are a couple use cases where developers would want to access the Cruise Control API without disabling API security:
+
+* Monitoring a Strimzi-managed Kafka cluster with the Cruise Control UI.
+* Gathering Cruise Control specific statistical information that is not currently available via Strimzi Operator or Cruise Control sensor metrics e.g. detailed information surrounding cluster and partition load and user tasks.
+* Experimenting with Cruise Control operations not currently supported by Strimzi.
```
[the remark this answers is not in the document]
kyguy: Originally was referencing POST operations before we decided to strip them out for this proposal. Going to remove this line now since we are not supporting ADMIN operations

052-cruise-control-api-users.md
kyguy: I found "user", "operator", and "admin" to be overloaded terms, terms that conflicted with key parts of the proposal. Yes this proposal is more aimed at Strimzi users, Strimzi operators, Strimzi cluster admins, or we could even say Strimzi devops but not Strimzi or Kafka application developers. However, it seemed that the term "developers" would cause the least confusion here.

ppatierno (approved):
LGTM. Thanks!

052-cruise-control-api-users.md
fvaleri: Yeah. How these error conditions are reflected in the Kafka CR status?

kyguy: @fvaleri They wouldn't be for now, for a feature like this it would not be worth the extra effort and code complexity. If needed, we could always add it at a later date

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,274 @@
...
+(1) A `type` field is added here to describe the format of the data and for configuration flexibility in the future. 
+This gives us the option to add different data format types in the future. 
+In this example we use Jetty's HashLoginService format since that is the format which Cruise Control uses for its API user configuration.
+
+(2) The `valueFrom` construct allows us to add more sources in the future if needed. 
+This is also a pattern used in other Strimzi APIs already, for example Strimzi's logging configuration, password configuration in KafkaUser resources, metrics configuration, and more.
+
+Strimzi would decode and use the contents of this secret to populate the Cruise Control API auth credentials file which would be mounted and used by the Cruise Control pod.
+
+Its worth noting that if a developer were to provide a corrupt configuration file - one which specifies the `ADMIN` role or an incorrect format, the Strimzi Operator would ignore the custom configuration and log a warning.
```
tombentley, line 140: I'm not really convinced that logging a warning is enough. The user has specified some set of CC users and the `status` section is telling them that everything is OK, so they would expect to be able to use those credentials. You can't assume users are looking at the logs. That is the approach we take fairly consistently elsewhere IIRC, and I don't see a justification for deviating from it here.

tombentley (commented):
This looks good, but there's one last point which I think needs resolving.

052-cruise-control-api-users.md
kyguy: Either way is fine with me, if adding the condition will save the user from necessary confusion/pain, I don't mind putting in the extra work. Since @fvaleri already mentioned this point, let me just add it to the proposal, I'll figure it out

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,274 @@
...
+(1) A `type` field is added here to describe the format of the data and for configuration flexibility in the future. 
+This gives us the option to add different data format types in the future. 
+In this example we use Jetty's HashLoginService format since that is the format which Cruise Control uses for its API user configuration.
+
+(2) The `valueFrom` construct allows us to add more sources in the future if needed. 
+This is also a pattern used in other Strimzi APIs already, for example Strimzi's logging configuration, password configuration in KafkaUser resources, metrics configuration, and more.
+
+Strimzi would decode and use the contents of this secret to populate the Cruise Control API auth credentials file which would be mounted and used by the Cruise Control pod.
+
+It's worth noting that if a developer were to provide a corrupt configuration file - one which specifies the `ADMIN` role or an incorrect format, the Strimzi Operator would ignore the custom configuration, log a warning, and add a warning condition to the `Kafka` custom resource.
```
tombentley, line 140: Can you show an example snippet of YAML, so as ot be explicit about what the API is?

fvaleri (commented):
Looks good overall. Thanks for the updates @kyguy. 

As @tombentley mentioned, I would just replace "developers" with "advanced user". You also use this term towards the end of the proposal.

052-cruise-control-api-users.md
fvaleri: Thanks.

052-cruise-control-api-users.md
kyguy: @tombentley  Just added an example, let me know what you think!

052-cruise-control-api-users.md
```diff
@@ -0,0 +1,274 @@
...
+For more details checkout the [Cruise Control Wiki](https://github.com/linkedin/cruise-control/wiki/Security#authorization)
+```
+VIEWER role: has access to the most lightweight kafka_cluster_state, user_tasks and review_board endpoints.
+USER role: has access to all the GET endpoints except bootstrap and train.
+ADMIN role: has access to all endpoints.
+```
+
+Strimzi would then aggregate the Cruise Control API auth credentials into a centralized configuration, then store it in a secret where it would be mounted and used by Cruise Control.
+
+Given that accessing the Cruise Control API directly is not the primary way of interacting with Cruise Control for a Strimzi-managed Kafka cluster and is expected to be used only by advanced users in special use-cases, this method would not be worth the complexity that it would add to the code base.
```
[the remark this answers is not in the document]
scholzj: I think _advanced users_ seems like a great term for the target group of this feature. Developers is completely unspecific ... it is not even clear who is meant by that. Strimzi developers? I.e. developers developing Strimzi? Or developers developing some Kafka applications? None of them is really the target group of this.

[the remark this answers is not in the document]
kyguy: Alright alright, this is third reviewer to raise this, I have just updated the proposal to changed "developer" -> "advance user". Let me know what you think

052-cruise-control-api-users.md
```diff
@@ -137,32 +137,61 @@ This is also a pattern used in other Strimzi APIs already, for example Strimzi's
...
 
-Although this would simplify the overall design it would also shift a lot of the maintanence responsibility to the developer.
-The developer would need to create a valid Cruise Control API user configuration and encode it into a secret.
-That being said, this is still a reasonable amount of responsibility to leave to the developer.
-This is because accessing the Cruise Control API directly is not the primary way of interacting with Cruise Control for a Strimzi-managed Kafka cluster and is expected to be used only by advanced developers in special use-cases.
-However, with proper supporting documentation, this new feature will give developers a greater degree of control and assistance with special use-cases.
+For example, if an advanced user were to provide a corrupt capacity configuration file, a warning condition like the following would be added to the status section of the `Kafka` custom resource :
```
tombentley, line 142: What's a capacity configuration file? I would have thought that the warning message in the example should be complaining about a "user config file".

kyguy: Argh, sorry I am mixing up another PR I am working on, it is fixed now!

tombentley (approved):
LGTM. Thanks Kyle!

scholzj:
Let's wait till Wednesday end-of-business and if there are no new comments, we should proceed and merge it as approved.

scholzj:
Approved with 5 binding and 2 non-binding +1 votes. Thanks @kyguy.

PR #87, opened 2023-08-06
feat: Proposal for accepting custom names for KafkaUsers

eslam-gomaa:
Proposal to support accepting custom names for KafkaUsers, this will give more flexibility to migrate KafkaACLs to Strimzi user-operator KafkaUsers with same users names

### DIFF
--- a/052-custom-kafka-user-name.md
+++ b/052-custom-kafka-user-name.md
@@ -1,0 +1,105 @@
+
+Add support for custom kafka users names with "`.spec.userName`" 
+
+## Current situation
+
+Currently the KafkaUser name is taken from the Kubernetes resource name (which forces specific naming for Kafka users .. no Uppercase letters or underscores)
+
+That's acceptible for new clusters, however there are cases that having the option to define custom KafkaUsers names wil be more efficeint
+
+
+## Motivaion (Use Case)
+
+_*Please consider the following use case:*_
+
+While prepairing for Migrating Kafka clusters to be managed by Strimzi, migrating the ACLs from the source cluster to the Strimzi clsuter was required.
+
+We've developed a small tool that reads the ACLs from the old cluster & generates Strimzi user-operator CRDs. 
+
+That solved migrating the ACLs, however there is a limitation .. The old cluster may have a Users with names that are not Valid for Kubernetes objects naming .. forcing us to change the migrated users names (Uppercase to lowercase && underscores to dashes)
+
+Here is an Example:
+
+> Waiting for approval to opensource the tool 
+
+
+```bash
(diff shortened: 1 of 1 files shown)


scholzj (commented):
Thanks for the proposal.

My personal view is that the use case does not justify the complexity and maintenance. Unlike topics, migration of users should be reasonably straight forward. You will likely anyway need to modify the configuration of the clients with regards to the bootstrap address and possibl also things such as TLS. So updating the username as part of it should be fairly straight forward. You also already seem to have a tool to migrate the ACLs into the KafkaUser resources. So I think it can lso easily normalize the usernames to valid Kubernetes names as part of it. Also, as a result of migrating the usernames, you will end up with much cleaner environment for the _years to come_, because using the Kubernetes names as usernames will significantly simplify working with the users. You will not need to understand and deal with some special logic of how the users are named, you will be able to easily use Kubernetes tools such as `kubectl` to work with the users, you will have much cleaner overview of what users exist in your system because the names will be part of Kubernetes itself or it will be for example much easier to onboard new users since the Kubernetes name will be likely already a familiar concept to them.

That said, if I set aside what I think about the use case, I think the proposal is missing the technical details of how it will be implemented. For example how will it deal with duplicates or conflicting resources and how would it parallelise the processing which are the parts where the singificant complexity will come from. So if nothing else, at least this should be added to it.

tombentley:
Thanks for the proposal.

Based on my experience with the topic operator, from an implementation point of view, supporting an alternative name field does add _significant_ complexity:

1. The operator needs to deal with multiple Kubernetes resources wanting to control the same user (define which one should win; maintain state to ensure they correct one does).
2. The operator needs to deal with possibility of someone changing the `spec.userName` (should it be allowed and treated as a delete+add, or be prohibited?)
3. It also makes it harder for someone using kubernetes to know which resource controls a given user in Kafka, or to know the full set of users defined in Kafka. 

So overall I agree with Jakub that the additional complexity would require a really strong use case to justify it. So far I don't recall the lack for support for this being a showstopper for many potential Strimzi users.

For what it's worth, the reason we support this for the topic operator is that users can use the Kafka `Admin` client to create topics in Kafka _anyway_, and applications like Kafka Streams make use of this. So we _have_ to support it in order for topics to be manageable by the TO. The situation with users is usually different: If people want to use `KafkaUsers` to manage accounts then they need to migrate existing applications anyway, and such migration is much less likely to be intermediated by 3rd party code with opinions about how users are named.

eslam-gomaa:
Thank you @scholzj @tombentley for your clarification,
I'm convinced by your point of view,  no need introduce complexity 

closing the PR

PR #88, opened 2023-08-09
Proposal for stopping connectors 

mimaison:
Proposal for https://github.com/strimzi/strimzi-kafka-operator/issues/8713

Kafka 3.5 introduced the STOPPED state for connectors via a new endpoint `PUT /connectors/<CONNECTOR>/stop`. Compared to PAUSED where the connector and tasks stay instantiated, in STOPPED only the configuration of the connector is kept but nothing is actually running.

A connector is PAUSED state still uses some resources but it's faster to resume. So using pause is great if you want to stop data flowing for a short duration. Otherwise it's interesting to fully stop the connector with the new API. Also the STOPPED state will be required to use the offset delete/reset endpoints that should come with Kafka 3.6.


052-connector-state.md
```diff
@@ -0,0 +1,77 @@
...
+
+A connector paused still uses some resources but it's faster to resume. So using pause is great if you want to halt data flowing for a short duration. For longer durations it's interesting to fully stop the connector with the new API. Also the stopped state will be required to use the offset delete/reset endpoints that should come with Kafka 3.6. For these reasons, Strimzi should support both pausing and stopping connectors.
+
+The `PUT /connectors/{connector}/resume` REST endpoint is used to restart both paused and stopped connectors.
+
+## Proposal
+
+Since a connector can't be paused and stopped at the same time, the proposal is to replace the current `pause` property and add a new property, `state`, in the `AbstractConnectorSpec` schema. That new property will accept the `pause`, `stop` or `run` values.
+
+In the next Strimzi release the `pause` property will be marked as deprecated. In the following release, the `pause` field will be deleted.
```
scholzj, line 34: You can mark it as deprecated together with this change. But it cannot be removed until the next API version.

052-connector-state.md
```diff
@@ -0,0 +1,77 @@
...
+```yaml
+spec:
+  class: org.apache.kafka.connect.file.FileStreamSourceConnector
+  config:
+    file: /opt/kafka/LICENSE
+    topic: my-topic
+  pause: true
+  state: pause
+  tasksMax: 1
+```
```
scholzj, line 64: We do not really do it this way as there is no simple way to modify the resource without risks to just end up in a loop with some other tools. So normally you would do something like this:
* The new field is preferred when both are set to some value
* When the old field is set in parallel with the new field, you will issue a warning about a conflcit
* When the new field is not set (i.e. is null), it will mean `run` if `pause == false` or if `pause` is not set. Or it will equal `pause` if `pause==true`. But this will not be done as a change to the custom resource. You will just handle it that way in the code.

This might stay like this until the next APi version which removes the `pause` field. Or you can say that both fields will be supported e.g. until Strimzi 0.40. In which case, in 0.41, we can change the code to completely ignore the `pause` field in the CRs ... but we still need to keep it in the custom resource definition as deprecated.

052-connector-state.md
```diff
@@ -0,0 +1,77 @@
...
+  config:
+    file: /opt/kafka/LICENSE
+    topic: my-topic
+  pause: true
+  tasksMax: 1
+```
+
+## Motivation
+
+Since Kafka 3.5, Kafka Connect connectors can also be stopped. This is done by using a new REST endpoint `PUT /connectors/{connector}/stop`. Compared to paused where the connector and tasks stay instantiated, when stopped only the configuration of the connector is kept but nothing is actually running.
```
scholzj, line 24: How will it be handled in Kafka 3.4?

scholzj (commented):
Thanks for the proposal. I think it is good in general -> I just left some comments abotu how API changes like this can be done.

052-connector-state.md
```diff
@@ -0,0 +1,77 @@
...
+
+A connector paused still uses some resources but it's faster to resume. So using pause is great if you want to halt data flowing for a short duration. For longer durations it's interesting to fully stop the connector with the new API. Also the stopped state will be required to use the offset delete/reset endpoints that should come with Kafka 3.6. For these reasons, Strimzi should support both pausing and stopping connectors.
+
+The `PUT /connectors/{connector}/resume` REST endpoint is used to restart both paused and stopped connectors.
+
+## Proposal
+
+Since a connector can't be paused and stopped at the same time, the proposal is to replace the current `pause` property and add a new property, `state`, in the `AbstractConnectorSpec` schema. That new property will accept the `pause`, `stop` or `run` values.
```
ppatierno, line 32: Using a field in the spec let me think the different approach we took with the `KafkaRebalance` resource where we are using an annotation for pause, stop, refresh ...
Are we sure we want to go in a different direction here? 
Or we should change the `KafkaRebalance` to have the same approach?
It's more as a general discussion about setting a common pattern for the users from a UX pov about how to interact with custom resources which "run" operations which could be paused, stopped, restarted and so on.

052-connector-state.md
ppatierno: Should we also add a "plan" reporting Strimzi versions where these changes will happen?

052-connector-state.md
scholzj: I do not think this is the same. AFAIK most of the rebalance annotations are commands. This is a declarative management of the state.

052-connector-state.md
mimaison: Kafka 3.4 does not have this endpoint so connectors can't be stopped. Will the next Strimzi version still support Kafka < 3.5.0?

scholzj: In general, we support two last minor Kafka versions. So once we add 3.6 support, we will likely drop 3.4 support. (I guess depending on Kafka schedules, but likely sometime shortly before the release of Strimzi 0.38)

I guess waiting is one option. Other option is simply documenting it as a known issue and handling it in the code so that it does not error.

052-connector-state.md
katheris: I agree with Jakub, the KafkaRebalance annotations are more commands to ask the operator to do something, rather than describing the current expected state of the resource.

In terms of the values, given the new field is called `state`, should the values be `paused`, `stopped`, and `running`

052-connector-state.md
```diff
@@ -0,0 +1,53 @@
...
+This only affects `strimzi-kafka-operator`. This will be usable by regular connectors (via `KafkaConnectorSpec`) and by MirrorMaker connectors  (via `KafkaMirrorMaker2ConnectorSpec`).
+
+## Compatibility
+
+If `state` is not set, it will:
+- default to `run` if `pause` is `false` or if `pause` is not set
+- default to `pause` if `pause` is `true`
+But this will not be done as a change to the custom resource, it will just be handled that way internally.
+
+If `state` is set it will take precedence over the old `pause `field. If both fields are set, a warning will be emitted to notify the user of a potential conflict.
```
katheris, line 49: For the warning is it worth both logging it and adding it as a condition to the KafkaConnector resource?

052-connector-state.md
mimaison: Is it possible to prevent setting `state` to `stop` is the Kafka version is < 3.5? For example like what happens if you try setting `pause` to a value different than `true` or `false`?

052-connector-state.md
mimaison: Sorry I'm not familiar with the semantic versioning used by Strimzi. Is 0.40 the next API version change?

052-connector-state.md
ppatierno: Not sure, isn't it a command for the operator to invoke an API to Kafka Connect and ask to run an action (pause, stop and re-start a connector)?

052-connector-state.md
scholzj: Not really. The CRD exists only once. So you can prevent it from being set to `stop` -> but for all versions only.

052-connector-state.md
scholzj: No, it describes the desired state. Compare it for example with `strimzi.io/rebalance=refresh` => you are nto asking it to be refreshed every reconciliation (ie. constantly refreshing) but just once.

052-connector-state.md
scholzj: No. There is currently no exact Strimzi version when that happens. Next APi version will be basically when the `v1beta2` version of the custom resources moves to `v1` or `v1beta3`. I think it is unlikely to happen before KRaft is done and ZooKeeper gone.

052-connector-state.md
mimaison: I'm happy to use `paused`, `stopped` and `running` as values. Then it's exactly the same as the `status.connectorStatus.connector.state` field which show the current state (apart as it's in uppercase in this field).

052-connector-state.md
mimaison: I'm not sure I fully understand. Can you clarify what you mean by `adding it as a condition to the KafkaConnector resource`?

052-connector-state.md
mimaison: I pushed an update to mention that if this is merged before support for Kafka 3.4 is dropped, the docs will mention it as a known issue.

052-connector-state.md
```diff
@@ -0,0 +1,55 @@
+<!-- This template is provided as an example with sections you may wish to comment on with respect to your proposal. Add or remove sections as required to best articulate the proposal. -->
+
+# Support stopping connectors
+
+This proposal aims at adding support for stopping Kafka Connect connectors.
```
tombentley, line 5: To join the dots, can you link to the KIP where this was discussed in Apache Kafka?

052-connector-state.md
```diff
@@ -0,0 +1,55 @@
...
+    topic: my-topic
+  pause: true
+  tasksMax: 1
+```
+
+## Motivation
+
+Since Kafka 3.5, Kafka Connect connectors can also be stopped. This is done by using a new REST endpoint `PUT /connectors/{connector}/stop`. Compared to paused where the connector and tasks stay instantiated, when stopped only the configuration of the connector is kept but nothing is actually running.
+
+A connector paused still uses some resources but it's faster to resume. So using pause is great if you want to halt data flowing for a short duration. For longer durations it's interesting to fully stop the connector with the new API. Also the stopped state will be required to use the offset delete/reset endpoints that should come with Kafka 3.6. For these reasons, Strimzi should support both pausing and stopping connectors.
```
tombentley, line 26: I think we need slight more justification than "interesting". (I'm picking on your choice of words; I don't doubt the usefulness of the feature).
```suggestion
A paused connector still uses some resources but it's faster to resume. So using pause is well-suited if you want to halt data flowing for a short duration. For longer durations it could be beneficial to free up memory and other resources by fully stopping the connector with the new API. The stopped state will be required to use the offset delete/reset endpoints that should come with Kafka 3.6. For these reasons, Strimzi should support both pausing and stopping connectors.
```

052-connector-state.md
```diff
@@ -0,0 +1,55 @@
...
+
+The `PUT /connectors/{connector}/resume` REST endpoint is used to restart both paused and stopped connectors.
+
+## Proposal
+
+Since a connector can't be paused and stopped at the same time, the proposal is to replace the current `pause` property and add a new property, `state`, in the `AbstractConnectorSpec` schema. That new property will accept the `pause`, `stop` or `run` values.
+
+The `pause` property will be marked as deprecated. In the next Strimzi API version, the `pause` field will be deleted.
+
+This feature requires at least Kafka 3.5 to work. In case it is merged when Strimzi still support older releases, a note in the documentation will be added to notify of this limitation.
```
tombentley, line 36: You should specify how the operator would handle `stop` in this case. Should it fail the reconciliation, or ignore the state entirely, or treat `stop` as `pause`? Personally I think treating stopped as paused would be acceptable in this case since that's functionally equivalent.

tombentley (approved):
LGTM! Thanks @mimaison.

052-connector-state.md
mimaison: Added a link to KIP-875.

052-connector-state.md
mimaison: Good point. I took your suggestion and added that when trying to stop a connector with 3.4, it will be paused.

052-connector-state.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Proposal
+
+Since a connector can't be paused and stopped at the same time, the proposal is to replace the current `pause` property and add a new property, `state`, in the `AbstractConnectorSpec` schema. That new property will accept the `pause`, `stop` or `run` values.
+
+The `pause` property will be marked as deprecated. In the next Strimzi API version, the `pause` field will be deleted.
+
+This feature requires at least Kafka 3.5 to work. In case it is merged when Strimzi still support older releases, a note in the documentation will be added to notify of this limitation. When trying to stop a connector with Kafka 3.4, the connector will instead be paused.
+
+This proposal does not intend to address deleting/resetting offsets. As explained in the motivation section, stopping connectors has value on its own. Adding support for deleting/resetting connector offsets will be tackled in a separate proposal once Strimzi adopts a Kafka version that supports this feature (expected to release in 3.6).
```
scholzj, line 38: Why did the example YAML disappear? I thought that was useful.

mimaison: My bad. I added it back.

scholzj (approved):
Thanks for the proposal.

052-connector-state.md
im-konge: I think that @katheris means something like 
```
  conditions:
  - lastTransitionTime: "xxxxxxxxxxx"
    message: 'Both `state` and `pause` fields were set, there is a possibility of a conflict'
    reason: ...
    status: "True"
    type: Ready
 ```

+1 from me for the adding it as a condition to KC resource

File: 052-connector-state.md [hunk 10]
```diff
@@ -0,0 +1,66 @@
...
+
+A paused connector still uses some resources but it's faster to resume. So using pause is well-suited if you want to halt data flowing for a short duration. For longer durations it could be beneficial to free up memory and other resources by fully stopping the connector with the new API. The stopped state will be required to use the offset delete/reset endpoints that should come with Kafka 3.6. For these reasons, Strimzi should support both pausing and stopping connectors.
+
+The `PUT /connectors/{connector}/resume` REST endpoint is used to restart both paused and stopped connectors.
+
+## Proposal
+
+Since a connector can't be paused and stopped at the same time, the proposal is to replace the current `pause` property and add a new property, `state`, in the `AbstractConnectorSpec` schema. That new property will accept the `pause`, `stop` or `run` values.
+
+The `pause` property will be marked as deprecated. In the next Strimzi API version, the `pause` field will be deleted.
```
ppatierno, line 34: I am not sure we can delete the field, it would make "invalid" any `KafkaConnector` already in place. We should just ignore it. @scholzj wdyt? Isn't it similar to the removal of  `jaeger` type for tracing?

ppatierno (approved):
LGTM. I left just a nit.

052-connector-state.md
scholzj: In the next API version, we can delete both `pause` as well as Jaeger tracing.

052-connector-state.md
katheris: @ppatierno @scholzj @tombentley what was your thought on changing the proposed values to `paused`, `stopped` and `running`, rather than `pause`, `stop` and `run`? I think the former makes more sense

052-connector-state.md
katheris: Yes, the K8s docs discuss them here: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions

052-connector-state.md
scholzj: I'm fine with `paused`, `stopped` and `running`. It should describe the desired state, so it probably fits better.

ppatierno: +1 from me as well.

tombentley: While I can't think of anywhere else that I've seen the gerund in resource-based APIs, I don't have a strong opinion. 

TBH the word `state` bothers me more. While it comes from the terminology used by connect itself, it's a bit more overloaded in a kube context, where we sometimes talk about the "desired state" (as a synonym for `spec`), and also use `status` as a way of communicating the connector's "state" more generally. I can't think of anything that is unambiguously better though. 
* Synonyms like `disposition` lose the connection to standard connect terminology
* Something like `connectorState` could be confused for a "KafkaConnector's `status`" . 
* Something like `targetState` might work (it's the term used in the KIP) but doesn't feel entirely natural because the `spec` is always expressing what's desired.

Frawless (approved):

see-quick (approved):

052-connector-state.md
katheris: I agree that the word `state` isn't ideal, however I do like that it aligns with the Connect output from the REST API and there isn't anything else that comes to mind as being a better fit.

052-connector-state.md
```diff
@@ -0,0 +1,66 @@
...
+
+This feature requires at least Kafka 3.5 to work. In case it is merged when Strimzi still support older releases, a note in the documentation will be added to notify of this limitation. When trying to stop a connector with Kafka 3.4, the connector will instead be paused.
```
PaulRMellor, line 36:
```suggestion
This feature requires at least Kafka 3.5 to work. In case it is merged when Strimzi still supports older releases, a note in the documentation will be added to notify of this limitation. When trying to stop a connector with Kafka 3.4, the connector will instead be paused.
```

PaulRMellor (approved):
Nice, clear proposal.

mimaison:
Thanks everyone for the comments and suggestions. I've renamed the values for the `state` field to `paused`, `stopped` and `running`. I've also incorporated @katheris suggestion about updating the message of the resource condition in case both the old and new fields are set.

katheris (approved):
LGTM

scholzj:
This has now 6 binding and one non-binding +1 votes. So I guess we can maybe leave it till tomorrow's EOB and if there are no other comments we can close it as approved?

im-konge (approved):

PR #91, opened 2023-09-12
Proposal for KRaft Support in Kafka Roller

tinaselenge:
There is a draft [PR #9146](https://github.com/strimzi/strimzi-kafka-operator/pull/9146) for the implementation.


056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,129 @@
...
+If KafkaRoller fails to roll the controller/combined nodes, it still tries to roll the brokers before reporting an error.
+This is required because until [KIP-919](https://cwiki.apache.org/confluence/display/KAFKA/KIP-919%3A+Allow+AdminClient+to+Talk+Directly+with+the+KRaft+Controller+Quorum) 
+is implemented in order to check the quorum state when rolling controllers at least one broker needs to be running.
+
+### Triggers
+The following will trigger an attempt to roll a KRaft controller or combined node:
+- Pod is annotated for manual rolling update
+- Pod is unready
+
+The triggers for broker remain the same in KRaft mode.
```
katheris, line 77:
```suggestion
The triggers for broker remain the same in ZooKeeper mode.
```

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,129 @@
...
+- Ready broker-only nodes
+
+If KafkaRoller fails to roll the controller/combined nodes, it still tries to roll the brokers before reporting an error.
+This is required because until [KIP-919](https://cwiki.apache.org/confluence/display/KAFKA/KIP-919%3A+Allow+AdminClient+to+Talk+Directly+with+the+KRaft+Controller+Quorum) 
+is implemented in order to check the quorum state when rolling controllers at least one broker needs to be running.
+
+### Triggers
+The following will trigger an attempt to roll a KRaft controller or combined node:
+- Pod is annotated for manual rolling update
+- Pod is unready
```
katheris, line 75: Should we still list a configuration change here as a trigger? Since we do expect that to result in pods rolling, it just might be implemented in different ways depending on whether the KIP is complete

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,129 @@
...
+- Broker's configuration has changed and cannot be updated dynamically
+
+### Rollability checks
+ - Attempts to connect an admin client to the broker and if it can't connect rolls the broker
+ - Does not force roll a broker performing log recovery
+ - Does not roll a broker if doing so would take the in-sync replicas count below `min.insync.replicas`
+
+### Configuration changes
+ - Compares the current configuration with expected configuration to see if a broker needs rolling
```
scholzj, line 37: I don't think it is really clear what configuration means here. The Kafka configuration itself? This might deserve more detail about how it works as it will highlight the differences.

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,129 @@
...
+
+In KRaft mode the KafkaRoller currently skips controller only nodes, but performs the above steps on any combined or broker only nodes.
+This is causing a problem in combined mode because if the quorum has not formed due to some of the nodes not being ready 
+the KafkaRoller will wait still try to contact the broker via the admin client.
+This call fails because the quorum is not formed, so in some cases this results in the cluster being stuck with some nodes 
+in a pending state.
+
+## Motivation
+
+When moving to KRaft mode the controller pods need to be rolled if configuration changes occur, also at the moment the
```
scholzj, line 48:
```suggestion
When running in the KRaft mode the controller pods need to be rolled if configuration changes occur, also at the moment the 
```

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,129 @@
...
+Roll in order:
+- Unready controller/combined nodes
+- Ready controller/combined nodes
+- Active controller
+- Unready broker-only nodes
+- Ready broker-only nodes
+
+If KafkaRoller fails to roll the controller/combined nodes, it still tries to roll the brokers before reporting an error.
+This is required because until [KIP-919](https://cwiki.apache.org/confluence/display/KAFKA/KIP-919%3A+Allow+AdminClient+to+Talk+Directly+with+the+KRaft+Controller+Quorum) 
+is implemented in order to check the quorum state when rolling controllers at least one broker needs to be running.
```
scholzj, line 67: If no brokers are running, they can still be evaluated for rolling (e.g. if they are pending or crash looping, they should be rolled if their spec changed). So not sure I follow what this sentence exactly means.

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,129 @@
...
+Roll in order:
+ - Unready brokers
+ - Ready brokers
+ - Current controller
+
+### Triggers
+The following triggers an attempt to roll a broker:
+- Pod is annotated for manual rolling update
+- Pod is unready
+- Broker's configuration has changed and cannot be updated dynamically
```
scholzj, line 29: I think there are some additional triggers. Changes to the Pod spec, pod and PVC deletion, certificate renewals etc. Not sure it is needed to list them all. But maybe the list should not suggest it lists all possibilities.

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,129 @@
...
+
+If KafkaRoller fails to roll the controller/combined nodes, it still tries to roll the brokers before reporting an error.
+This is required because until [KIP-919](https://cwiki.apache.org/confluence/display/KAFKA/KIP-919%3A+Allow+AdminClient+to+Talk+Directly+with+the+KRaft+Controller+Quorum) 
+is implemented in order to check the quorum state when rolling controllers at least one broker needs to be running.
+
+### Triggers
+The following will trigger an attempt to roll a KRaft controller or combined node:
+- Pod is annotated for manual rolling update
+- Pod is unready
+- Controller's configuration has changed and cannot be updated dynamically.
```
scholzj, line 73: Same as above. We should make it a bit more clear that this is just a list of some examples.

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Rollability checks
+
+The checks made by the KafkaRoller in different modes is described below.
+The checks include a new check for controllers to verify that rolling the node does not affect the quorum health.
+The proposed check is:
+- Fetch `lastCaughtUpTimestamp` for every node
+- A node is marked as caught up if `leaderLastCaughtUpTimestamp - lastCaughtUpTimestamp < controllerQuorumFetchTimeoutMs`
+- Count each node that is caught up, plus the leader (`numOfCaughtUpVoters`)
+- Can roll if: `numOfCaughtUpVoters > (numOfVoters)/2`
```
scholzj, line 87: I think properly explaining these values would be useful
* Is number of voters the same as number of controller replicas? 
* What if there is a scale-up / scale-down event? Is it the old or new number of replicas?
* Assuming `numOfVoters` is number of controllers ... would the formula be `numOfCaughtUpVoters > ceil( (double) numOfVoters / 2)`?
* Would it be safer to just go always only by one node? That might allow some non-voluntary disruptions to happen in parallel.
* What role does the Pod readiness play in this?

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,129 @@
...
+
+#### Separate brokers and controllers
+For controller-only:
+- Does not force roll a controller performing log recovery
+- Does not roll a ready controller if doing so would take the number of caught-up voters (inc leader) to less than half of the quorum
+- When rolling an unready controller, does not check quorum state before rolling
```
scholzj, line 93: Should we go into a bit more details about when an unready controller would be rolled?

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,129 @@
...
+- Does not roll a ready combined node if doing so would take the in-sync replicas count below `min.insync.replicas`
+- When rolling an unready combined node, does not check quorum state, or in-sync replicas count, or try to connect an admin client to the broker before rolling
+
+### Configuration changes
+For all kinds of nodes:
+- Compares the current configuration with expected configuration to see if a node needs rolling
+- Performs dynamic configuration updates if possible, otherwise rolls node on configuration change
+
+**NOTE:** At the time of writing this proposal dynamic configuration updates are not possible for controller nodes.
+However [KIP 919](https://cwiki.apache.org/confluence/display/KAFKA/KIP-919%3A+Allow+AdminClient+to+Talk+Directly+with+the+KRaft+Controller+Quorum) looks to address this
```
scholzj, line 113: So, what does that mean for the proposal? Does it expect to wait for it to be implemented? If not what workarounds / limitations will it take?

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,129 @@
+# Kafka Roller KRaft Support
+
+This proposal describes the actions that the KafkaRoller should take when operating 
+against a Strimzi cluster in KRaft mode.
+The proposal omits details about exactly how this should be implemented because 
+the logic of the KafkaRoller is currently being rewritten.
+This proposal only describes the expected behaviour so that it can be implemented 
+in either the current KafkaRoller or a future iteration of the KafkaRoller.
```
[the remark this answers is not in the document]
tinaselenge: I thought the discussion on how to configure the controller would be discussed and agreed in https://github.com/strimzi/proposals/pull/82. I tried to put a bit more detail on the configuration and this PR. If you think that's not sufficient, I can certainly add more technical details.

056-kafka-roller-kraft.md
tinaselenge: I removed this sentence, as it does  sound a bit confusing. I explained kip-919 in another section.

056-kafka-roller-kraft.md
tinaselenge: - yes it is, I used `controller` instead of `voters`.
- I believe it would be the current number of replicas as we would be getting this number from the quorum data returned via admin client API. I can try make this a bit clearer. 
- Yes, agree on this. 
- Not sure if I follow this question, could you please elaborate little more? To be honest I think the formula needs little more thought. 
- The quorum would not be checked if the pod is unready. Similar to how it works today with zookeeper based brokers. Their availability does not get checked if pod is unready, instead it gets force rolled. Again I can try make this clearer.

056-kafka-roller-kraft.md
tinaselenge: I tried making this little more clear. Some part would be resolved by the workaround and some part would not be done until this KIP is implemented.

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,131 @@
...
+
+The KafkaRoller behaviour should be unchanged when operating against a ZooKeeper based cluster.
+
+The behaviour when operating against a KRaft cluster is described below.
+
+### Order to roll nodes
+Roll in order:
+- Unready controller/combined nodes
+- Ready controller/combined nodes
+- Active controller
```
scholzj, line 61: This applies also to active controller in combined mode as well I assume?

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,131 @@
...
+**NOTE** There is a related proposal describing how to diff controller configuration currently being discussed: [PR #82](https://github.com/strimzi/proposals/pull/82)
+
+### Rollability checks
+
+The checks made by the KafkaRoller in different modes is described below.
+The checks include a new check for controllers to verify that rolling the node does not affect the quorum health.
+The proposed check is:
+- Fetch `lastCaughtUpTimestamp` for every controller node. `lastCaughtUpTimestamp` is the last millisecond timestamp at which a replica controller was known to be caught up with the quorum leader
+- Identify the `lastCaughtUpTimestamp` of the node that is the quorum leader .
+- Retrieve the current value of Kafka configuration `controller.quorum.fetch.timeout.ms`
```
scholzj, line 83: How / from where will it be retrieved?

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,131 @@
...
+The triggers for broker remain the same as Zookeeper mode.
+
+**NOTE** There is a related proposal describing how to diff controller configuration currently being discussed: [PR #82](https://github.com/strimzi/proposals/pull/82)
+
+### Rollability checks
+
+The checks made by the KafkaRoller in different modes is described below.
+The checks include a new check for controllers to verify that rolling the node does not affect the quorum health.
+The proposed check is:
+- Fetch `lastCaughtUpTimestamp` for every controller node. `lastCaughtUpTimestamp` is the last millisecond timestamp at which a replica controller was known to be caught up with the quorum leader
```
scholzj, line 81: How will you fetch it? Will the agent provide it? Is that already done or described somewhere?

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,131 @@
...
+- Identify the `lastCaughtUpTimestamp` of the node that is the quorum leader .
```
scholzj, line 82: Same as above ... How do you know who is the leader? Through the agent?

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,131 @@
...
+- Retrieve the current value of Kafka configuration `controller.quorum.fetch.timeout.ms`
+- Mark a node as caught up if `leaderLastCaughtUpTimestamp - replicaLastCaughtUpTimestamp < controllerQuorumFetchTimeoutMs`
+- Count each controller node that is caught up, including the leader (`numOfCaughtUpControllers`)
+- Can roll if: `numOfCaughtUpControllers > ceil(double) totalNumOfControllers / 2`
```
scholzj, line 86: We want to check if at least 50% + 1 controller are caught up, or? So we need to round it up? E.g. with 3 controllers, all 3 need to be in sync to be able to roll the next one. With 5 controllers, at least 4 need to be caught up.

```suggestion
- Can roll if: `numOfCaughtUpControllers > ceil((double) totalNumOfControllers / 2)`
```

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,131 @@
...
+
+In order to perform quorum check for controller, KafkaRoller creates admin client connection to the brokers. This is required until [KIP-919](https://cwiki.apache.org/confluence/display/KAFKA/KIP-919%3A+Allow+AdminClient+to+Talk+Directly+with+the+KRaft+Controller+Quorum) is implemented. If KafkaRoller cannot connect to any of the brokers, reconcilation for the controllers will be marked as failed due to not being able to determine the quorum health. KafkaRoller would then try to reconcile the brokers, which may help restoring admin client connections to them. The controllers quorum check will be retried in the next round of reconcilation.
```
scholzj, line 88: `will be retried in the next round of reconcilation` -> does that mean you will re-queue them and try them again after checking the brokers? Or that you will leave it to next reconciliation? Leaving it to the next recocniliation might possible leave it broken forever. But I'm not sure if it is something we can fix at the same time 🤔

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,131 @@
...
+
+
+## Current situation
+
+When operating on a ZooKeeper based cluster the KafkaRoller has the behaviour 
+described below.
+
+### Order to roll nodes
+Roll in order:
+ - Unready brokers
```
fvaleri, line 21: I think a numbered list would be better to represent ordering.

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,131 @@
...
+ - Ready brokers
+ - Current controller
+
+### Triggers
+The following are some of the triggers that roll a broker:
+- Pod is annotated for manual rolling update
```
fvaleri, line 27:
```suggestion
- Pod or its StrimziPodSet is annotated for manual rolling update
```

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,131 @@
...
+- Pod is unready because it is in `CrashLoopBackOff`, `ImagePullBackOff`, `ContainerCreating` or `Pending` and `Unschedulable`,
```
fvaleri, line 28: With ContainerCrating and Pending I think you mean "stuck in that state". Maybe we can create a dedicated bullet point for them.

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,131 @@
...
+ - Does not force roll a broker performing log recovery
+ - Does not roll a broker if doing so would take the in-sync replicas count below `min.insync.replicas`
+
+### Configuration changes
+ - Retrieves the current Kafka configurations of the broker via admin client and compares it with the desired configurations specified in the Kafka CR. 
+ - Performs dynamic configuration updates if possible, otherwise rolls broker on configuration change
+
+In KRaft mode the KafkaRoller currently skips controller only nodes, but performs the above steps on any combined or broker only nodes.
+This is causing a problem in combined mode because if the quorum has not formed due to some of the nodes not being ready 
+the KafkaRoller will wait still try to contact the broker via the admin client.
```
fvaleri, line 42:
```suggestion
the KafkaRoller will still try to contact the broker via the admin client.
```

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Proposal
+
+The KafkaRoller behaviour should be unchanged when operating against a ZooKeeper based cluster.
+
+The behaviour when operating against a KRaft cluster is described below.
+
+### Order to roll nodes
+Roll in order:
+- Unready controller/combined nodes
```
fvaleri, line 59: Again, I think that a numbered list would be better.

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,131 @@
...
+- Ready controller/combined nodes
+- Active controller
+- Unready broker-only nodes
+- Ready broker-only nodes
+
+
+### Triggers
+The following are some of the triggers that would roll a KRaft controller or combined node:
+- Pod is annotated for manual rolling update
```
fvaleri, line 68:
```suggestion
- Pod or its StrimziPodSet is annotated for manual rolling update
```

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,131 @@
...
+- Pod is unready because it is in `CrashLoopBackOff`, `ImagePullBackOff`, `ContainerCreating` or `Pending` and `Unschedulable`,
```
fvaleri, line 69: With ContainerCreating and Pending I think you mean "stuck in that state". Maybe we can create a dedicated bullet point for them.

056-kafka-roller-kraft.md
fvaleri: I also think that we should fail the reconciliation after a number of failed retries, in order to inform the user that there is some serious issue going on.

fvaleri (commented):
Hi @tinaselenge, thanks for working on this. Left some comments.

056-kafka-roller-kraft.md
tinaselenge: This is tricky to answer actually,  once [Proposal in PR #82](https://github.com/strimzi/proposals/pull/82) is implemented, we would be able to get the current value for
this configuration. Until then we would have to use the hard-coded default value which I'm not sure how acceptable is. I'm hoping both proposals would be implemented around similar time.

056-kafka-roller-kraft.md
tinaselenge: yes, I will mention that.

056-kafka-roller-kraft.md
tinaselenge: So we can return `UnforceableProblem` to the scheduler, which would re-queue, delay and retry the controllers until maxAttempt is reached (in the meantime, the connections to brokers could also get resolved?). However we would not check the brokers until KafkaRoller finished reconciling the controllers.  This is to get the ordering correct (so that we can re-queue the active controller to the end of controller list, not the entire list including brokers). Once brokers are checked and attempted for rolling, the reconciliation will complete, we would not check the controllers again. In the next round of reconciliation, the controllers will be retried.  

@fvaleri yes, we already do that. Reconciliation fails if any of the pods has reached its max retry attempts or reported any other failure.

scholzj:
What about the situations when the process roles are changing? Does that need some special considerations here?

056-kafka-roller-kraft.md
tinaselenge: Added more information on how the leader is identified

056-kafka-roller-kraft.md
tinaselenge: We would be using admin API to get quorum data, I've included this in the proposal now.

tinaselenge:
@scholzj @fvaleri thank you both 

> What about the situations when the process roles are changing? Does that need some special considerations here?

Good question. Would that be considered as a configuration change therefore controllers would just get rolled? Although I would have to go and test this with our POC.

tinaselenge:
We have pushed some more updates to reflect the recent changes for the controller configuration and also raised a [PR](https://github.com/strimzi/strimzi-kafka-operator/pull/9146) draft implementing the proposal.

056-kafka-roller-kraft.md
tinaselenge: As [Proposal in PR #82](https://github.com/strimzi/proposals/pull/82) is closed, we have changed the proposal to get the value of this property from the desired configurations instead. https://github.com/orgs/strimzi/discussions/9137 also discusses the justification for this in detail.

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Rollability checks
+The checks made by the KafkaRoller in different modes is described below.
+
+#### The new quorum check
+The checks include a new check for controllers to verify that rolling the node does not affect the quorum health.
+The proposed check is:
+- Create admin client connection to the brokers and call `describeMetadataQuorum` API
+- If failed to connect to the brokers, return `UnforceableProblem` which will result in delay and retry for the pod until the maximum attempts is reached.
+- From the quorum info returned from the admin API, fetch `lastCaughtUpTimestamp` for every controller except the current controller we are considering to roll. `lastCaughtUpTimestamp` is the last millisecond timestamp at which a replica controller was known to be caught up with the quorum leader.
```
scholzj, line 94: Do you really have to fetch the value separately for each node? Or is it one fetch that returns all the data? That is quite important because if you acquire the data through multiple network-based API calls, it will clearly have inconsistencies.

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+The triggers for broker remain the same as Zookeeper mode.
+
+### Rollability checks
+The checks made by the KafkaRoller in different modes is described below.
+
+#### The new quorum check
+The checks include a new check for controllers to verify that rolling the node does not affect the quorum health.
+The proposed check is:
+- Create admin client connection to the brokers and call `describeMetadataQuorum` API
+- If failed to connect to the brokers, return `UnforceableProblem` which will result in delay and retry for the pod until the maximum attempts is reached.
```
scholzj, line 93: What is the timeout? Is there some kind of readiness check? When the cluster breaks, you do not want to wait for some long timeouts to fix things.

scholzj (commented):
Looks mostly good within the existing limitations of Apache Kafka. I left few (hopefully) last nits.

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+- From the quorum info returned from the admin API, fetch `lastCaughtUpTimestamp` for every controller except the current controller we are considering to roll. `lastCaughtUpTimestamp` is the last millisecond timestamp at which a replica controller was known to be caught up with the quorum leader.
+- Check the quorum leader id using the quorum info and identify the `lastCaughtUpTimestamp` of the quorum leader
+- Retrieve value of the Kafka property `controller.quorum.fetch.timeout.ms` from the desired configurations specified in the Kafka CR. If this property does not exist in the desired configurations, then use the hard-coded default value for it which is `2000`. The reason for this is explained further in the **NOTE** below.
+- Mark a node as caught up if `leaderLastCaughtUpTimestamp - replicaLastCaughtUpTimestamp < controllerQuorumFetchTimeoutMs`.
+- Count each controller node that is caught up, including the leader (`numOfCaughtUpControllers`).
+- Can roll if: `numOfCaughtUpControllers > ceil((double) totalNumOfControllers / 2)`.
```
showuon, line 99: Is this correct? 
If totalNumOfControllers = 3, then `ceil((double) totalNumOfControllers / 2)` will be 2.
Then, we need 3 CaughtUpControllers to be able to roll?

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+
+
+#### Separate brokers and controllers
+For controller-only:
+ 1. Does not force roll a controller performing log recovery
+ 2. Force rolls a broker if the pod is in stuck state
+ 3. Attempts to connect an admin client to any of the brokers, if cannot connect, skip checks in `4` and `5` 
+ 4. Rolls a controller if the controller configuration has changed
+ 5. Does not roll a controller if doing so would take the number of caught-up controllers (inc leader) to less than half of the quorum
```
showuon, line 110: "half of the quorum" is not accurate. "majority of the quorum" is better.

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+
+For broker-only:
+ 1. Does not force roll a broker performing log recovery
+ 2. Force rolls a broker if the pod is in stuck state
+ 3. **NEW** Attempts to connect an admin client to a broker and if it can't connect rolls the broker 
+ (the order of the rollability checks is different, we only attempt admin client connection and force roll if the pod is not stuck)
```
showuon, line 116: Could you explain how we do `only attempt admin client connection`?

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+ 4. Rolls a broker if the broker configuration has changed and cannot be updated dynamically
+ 5. Does not roll a broker if doing so would take the in-sync replicas count below `min.insync.replicas`
+
+
+#### Combined mode
+1. Does not force roll a combined node performing log recovery
+2. Force rolls a combined node if the pod is in stuck state
+3. Attempts to connect an admin client to any of the brokers, if cannot connect, skip checks in `4`, `5` and `6`
+4. Rolls a combined node if the controller configuration has changed OR if the broker configuration has changed and cannot be updated dynamically
+6. Does not roll a ready combined node if doing so would take the number of caught-up controllers (inc leader) to less than half of the quorum
```
showuon, line 126: nit: majority of the quorum

056-kafka-roller-kraft.md
fvaleri: I think this is correct. If you roll with 2 controllers out of 3, the active controller will start a new election after quorum.fetch.timeout.ms has passed with no fetch requests from a majority of the quorum.

056-kafka-roller-kraft.md
tinaselenge: That is just poor choice of word. We would fetch quorum info only once which includes all nodes' lastCaughtUpTimestamps. I will reword this part.

056-kafka-roller-kraft.md
tinaselenge: We would be using the existing logic for handling `UnforceableProblem` and retrying the controller to see if it can connect to the brokers. It shouldn't wait for a long time. There is a readiness check at the beginning, for every pod.  If the controllers still can't connect to the brokers after the max num of retries, KafkaRoller would try reconcile the brokers same as usual.

056-kafka-roller-kraft.md
tinaselenge: The idea was actually to roll as long as there are 2 other controllers that are up to date with the leader if totalNumOfControllers = 3. So it's based on majority, rather than every node. It should be `numOfCaughtUpControllers.get() >= ceil((double) totalNumOfControllers / 2)`.

056-kafka-roller-kraft.md
tinaselenge: Currently KafkaRoller checks whether admin client is created successfully and if not, it immediately force rolls the broker. This sentence is trying to say, we execute this step only if pod is not stuck. I will reword it little bit better.

056-kafka-roller-kraft.md
fvaleri: I agree @tinaselenge, otherwise you won't roll at all, but you have to set `controller.quorum.fetch.timeout.ms` to something reasonably high while rolling. For example, this would avoid an unnecessary election when you have 2 caught up controllers, and 1 controller is down in a 3 controllers setup.

tinaselenge: `Maximum time without a successful fetch from the current leader before becoming a candidate and triggering an election for voters; Maximum time without receiving fetch from a majority of the quorum before asking around to see if there's a new epoch for leader`

This does not seem to say the quorum leader would trigger an election if it hasn't received fetch requests from the majority. What's majority for it in 3 controllers case though? Does it consider itself part of the majority? Only the followers could trigger an election if it can't successfully fetch so as long as the remaining follower is happily fetching from the leader, it shouldn't trigger an election?

fvaleri: >  Does it consider itself part of the majority? 

Yes, it is.

> Only the followers could trigger an election if it can't successfully fetch so as long as the remaining follower is happily fetching from the leader

I think the active controller (KRaft leader) will step down and trigger a new election instead. This also described in the classic Raft algorithm to avoid zombie leaders in case of network partitions.

In KIP-595 we have the so called "progress timeout". When a majority of voters are not fetching from the leader, the current leader will start election to reach out to other followers and re-establish the leadership or learn about the new leader. This is also the cause of an known issue described in the draft KIP-650 (that can be solved with pre-vote).

tinaselenge: I see. Thanks for clarifying that. 

So regarding your earlier point, are you saying when 1 controller is down and 1 remaining follower is fetching happily from the leader, the leader is receiving fetch requests from the remaining follower, that's considered majority of voters not fetching from the leader and would trigger re-election?

fvaleri: > when 1 controller is down and 1 remaining follower is fetching happily from the leader, the leader is receiving fetch requests from the remaining follower, that's considered majority of voters not fetching from the leader and would trigger re-election

No. When you have 2 out of 3 controllers that are in sync, that's ok (quorum is healthy). That said, when you want to roll the cluster for whatever reason, you would start by restarting the remaining follower (leader last policy). If this node restart lasts for more than `controller.quorum.fetch.timeout.ms` (2 seconds by default), then the leader will start a new election.

If you increase `controller.quorum.fetch.timeout.ms` to include the node clean restart period, then no election should be triggered during a rolling update as the follower/leader will come back before the other node realizes it was alone. Unfortunately, this also means that you would increase leader failure/partition detection. Given that this timeout can't be changed dynamically, we would need to find a reasonable value, which may be tricky. Wdyt?

cc @showuon @tombentley @mimaison

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+
+The behaviour when operating against a KRaft cluster is described below.
+
+### Order to roll nodes
+Roll in order:
+1. Unready controller/combined nodes
+2. Ready controller/combined nodes
+3. Active controller (applies to both pure controller and combined node)
+4. Unready broker-only nodes
+5. Ready broker-only nodes
```
tombentley, line 75: This order LGTM, but you've not stated the source of truth for knowing whether a particular pod is a pure controller node, combined node or pure broker node. The operator has its view (which is available from the `NodeRef`), and Kafka has its own view. We expect these to be consistent, but AFAIK that's not guaranteed (maybe @scholzj will correct me if there is such a guarantee).

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+
+> NOTE: Until [KIP-919](https://cwiki.apache.org/confluence/display/KAFKA/KIP-919%3A+Allow+AdminClient+to+Talk+Directly+with+the+KRaft+Controller+Quorum) is implemented, KafkaRoller cannot create an admin connection to the controller directly to describe its configuration or the quorum state. Therefore KafkaRoller checks the desired configurations to get the value of `controller.quorum.fetch.timeout.ms` and creates admin client connection to the brokers for the quorum check. If KafkaRoller cannot connect to any of the brokers after the maximum attempts to retry, the controllers will be marked as failed to reconcile because of not being able to determine the quorum health. KafkaRoller would then try to reconcile the brokers, which may help restoring admin client connections to them. In this scenario, the reconcilation will be completed with failure and reported to the operator. The controllers quorum check will be retried in the next round of reconcilation.
+
+
+#### Separate brokers and controllers
+For controller-only:
+ 1. Does not force roll a controller performing log recovery
+ 2. Force rolls a broker if the pod is in stuck state
```
tombentley, line 107: controller, not broker?

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+ 3. Attempts to connect an admin client to any of the brokers, if cannot connect, skip checks in `4` and `5`
```
tombentley, line 108: 4 and 5 aren't really checks (which suggests to me that they don't have side effects), they're predicated actions. When you say "skips" it's not clear what then happens. Does this controller get reconsidered for restarting later?

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+ 2. Force rolls a broker if the pod is in stuck state
+ 3. Attempts to connect an admin client to any of the brokers, if cannot connect, skip checks in `4` and `5` 
+ 4. Rolls a controller if the controller configuration has changed
+ 5. Does not roll a controller if doing so would take the number of caught-up controllers (inc leader) to less than majority of the quorum
+
+For broker-only:
+ 1. Does not force roll a broker performing log recovery
+ 2. Force rolls a broker if the pod is in stuck state
+ 3. **NEW** Attempts to connect an admin client to a broker and if it can't connect rolls the broker 
+ (the order of the rollability checks will change, we will attempt admin client connection and force roll only if the pod is not stuck)
```
tombentley, line 116: It is worth highlighting that this behaviour change will affect ZK-mode

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+ 4. Rolls a broker if the configuration has changed and cannot be updated dynamically
+ 5. Does not roll a broker if doing so would take the in-sync replicas count below `min.insync.replicas`
+
+
+#### Unready pod
+If pod is unready, KafkaRoller awaits its readiness before doing anything. If it hasn't become ready still, it checks whether it is stuck and considers to force roll. A pod is considered stuck if it is in one of following states:
+- `CrashLoopBackOff`
+- `ImagePullBackOff`
+- `ContainerCreating`
+- `Pending` and `Unschedulable`
```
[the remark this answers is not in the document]
ppatierno: Out of curiosity, does the KafkaRoller evaluate how long a pod is "stuck" before rolling? I mean, the `Pending` could be temporary as well as `ContainerCreating` which I am not sure really means it's stuck ... this is a state where a pod always transition to before being ready.

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+In KRaft mode the KafkaRoller currently skips controller only nodes, but performs the above steps on any combined or broker only nodes.
+This is causing a problem in combined mode because if the quorum has not formed due to some of the nodes not being ready 
+the KafkaRoller will still try to contact the broker via the admin client.
+This call fails because the quorum is not formed, so in some cases this results in the cluster being stuck with some nodes 
+in a pending state.
+
+## Motivation
+
+When running in the KRaft mode the controller pods need to be rolled if configuration changes occur, also at the moment the 
+existing logic is blocking the full implementation of liveness and readiness checks as described in [proposal #46](https://github.com/strimzi/proposals/blob/main/046-kraft-liveness-readiness.md).
```
ppatierno, line 61: would be worth adding that it's also blocking ZooKeeper to KRaft migration where we need controllers to be rolled more times during the process? FYI this is the proposal https://github.com/strimzi/proposals/pull/93

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+ 3. **NEW** Attempts to connect an admin client to a broker and if it can't connect rolls the broker 
+ (the order of the rollability checks will change, we will attempt admin client connection and force roll only if the pod is not stuck)
+ 4. Rolls a broker if the broker configuration has changed and cannot be updated dynamically
+ 5. Does not roll a broker if doing so would take the in-sync replicas count below `min.insync.replicas`
+
+
+#### Combined mode
+1. Does not force roll a combined node performing log recovery
+2. Force rolls a combined node if the pod is in stuck state
+3. Attempts to connect an admin client to any of the brokers, if cannot connect, skip checks in `4`, `5` and `6`
```
ppatierno, line 124: skip checks but then? rolls?

README.md
```diff
@@ -7,63 +7,64 @@
...
-|  9  | [Roadmap to using the CRD v1 API](./009-crd-v1-roadmap.md)  |
-|  8  | [TLS encrypting the Kafka Connect REST API](./008-tls-encrypt-the-kafka-connect-rest-api.md)  |
-|  7  | [Restarting Kafka Connect connectors and tasks](./007-restarting-kafka-connect-connectors-and-tasks.md) |
-|  6  | [Design Documentation](./006-design-docs.md) |
-|  5  | [Improving configurability of Kafka listeners](./005-improving-configurability-of-kafka-listeners.md) |
-|  4  | [GitHub repository restructuring](./004-github-repository-restructuring.md) |
-|  3  | [Remove deprecated Topic Operator deployment in Kafka CRD](./003-remove-deprecated-topic-operator-from-kafka-crd.md) |
-|  2  | [Documentation improvements](./002-documentation-improvements.md) |
-|  1  | [Move Strimzi Kafka operators to Java 11](./001-move-strimzi-kafka-operators-to-java-11.md) |
+| #  | Title                                                                 |
```
ppatierno, line 10: is the change to the whole file necessary? I see you removed a space from the proposal number column, I would leave it. We are prepared to reach 100 proposals soon ;-)

056-kafka-roller-kraft.md
ppatierno: > I believe it would be the current number of replicas as we would be getting this number from the quorum data returned via admin client API. I can try make this a bit clearer.

As per some tests I did in the past, it seems that the `describeMetadataQuorum` call doesn't update the list of voters on scale up/scale down. When I raised this, I think @showuon was going to open a KIP to fix this problem. Is it still not fixed? Is it still the case?

056-kafka-roller-kraft.md
tinaselenge: I think I understand it. Just to check if the proposal is describing the intended check correctly, I would like to re-iterate the logic here. We would not roll the follower that has caught up to the leader while another one is not in sync yet. For example, if we have controller-1 (leader), controller-2 (follower in sync), controller-3 (follower is not in sync), we would only allow controller-3 to be rolled, since quorum would still stay healthy while it's down. When attempting to roll controller-1 or controller-2, the quorum check will return false because this would leave the quorum without any in sync follower. We will block rolling until controller-3 has caught up to the leader.

Also `controller.quorum.fetch.timeout.ms` is a configuration we allow users to change today so would it make sense for Strimzi to set this configuration?

056-kafka-roller-kraft.md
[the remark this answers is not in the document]
tinaselenge: @ppatierno KafkaRoller waits for pod's readiness until operational timeout is reached (300000ms by default). After this, if the pod is still in one of these states, it is considered stuck.

056-kafka-roller-kraft.md
fvaleri: > We would not roll the follower that has caught up to the leader while another one is not in sync yet.

Yes, that would avoid the new election in case `controller.quorum.fetch.timeout.ms` is exceeded. 

That said, in your example, what would happen if the leader fails while you are restarting a follower? The remaining follower would start an election, but won't be able to be elected as the new leader because 2 out of 3 nodes are down. This would be a service downtime. The new leader would be elected as soon as the controller you are restarting comes up again, restoring the service.

If you want to completely avoid any downtime, you need to check that a majority of controllers is in-sync (3/3, 3/5, 4/7, and so on).

> Also controller.quorum.fetch.timeout.ms is a configuration we allow users to change today so would it make sense for Strimzi to set this configuration?

Probably not, as we want to give users the flexibility to tune it, but only in very specific use cases (most users won't need to touch it).

tinaselenge: I think this is addressed now.

056-kafka-roller-kraft.md
scholzj: That is complicated. I guess the NodeRefs currently shows the desired state. However, right now it is the same as the current state given our lack of support for controller rolling etc. So there was no thought about this until now I guess.

056-kafka-roller-kraft.md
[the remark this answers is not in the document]
scholzj: This is tricky. It might be just a race condition during a pod startup and I think it happens from time to time under some circumstances. But in many other cases, the operator was already waiting for it for 5 minutes. So further delay might not be desired.

056-kafka-roller-kraft.md
scholzj: Actually, the Roller IIRC queries the pods to decide if they are ready or not. Maybe it can get the current state from their annotations that will have the current state until rolled.

katheris: In terms of ordering I wonder if it is actually better to restart the pods based on the desired roles, not the current roles. For example, if a node was previously a `broker` and now is `controller,broker` you probably want to restart it before you begin restarting the other broker nodes so it can join the controller quorum. I think in each process role change restarting the node based on the desired roles is preferable.

056-kafka-roller-kraft.md
tinaselenge: Yes, the controller gets retried. I have reworded this part.

056-kafka-roller-kraft.md
tinaselenge: This was actually incorrect , sorry! I have corrected this part.

README.md
tinaselenge: Oops, it was not intentional. Fixed it now.

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+- Pod or its StrimziPodSet is annotated for manual rolling update
+- Pod is unready
+- Broker's configuration has changed
+
+### Restart conditions
+ 1. Does not restart a broker performing log recovery
+ 2. Attempts to connect an admin client to a broker and if it can't connect, restarts the broker
+ 3. Restarts a broker if the pod is in stuck state
+ 4. Restarts a broker if the configuration has changed and cannot be updated dynamically
+ 5. Does not restart a broker if doing so would take the in-sync replicas count below `min.insync.replicas`
```
fvaleri, line 35:
```suggestion
 5. Does not restart a broker if doing so would take the in-sync replicas count below `min.insync.replicas` for any topic hosted on that broker
```

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+ 4. Restarts a controller if the controller configuration has changed
+ 5. Does not restart a controller if doing so would take the number of caught-up controllers (inc leader) to less than majority of the quorum
+
+For broker-only:
+ 1. Does not restart a broker performing log recovery
+ 2. Restarts a broker if the pod is in stuck state
+ 3. Attempts to connect an admin client to a broker and if it can't connect restarts the broker
+ > **NEW** Currently when pod is stuck and out of date, it gets restarted because admin client connection fails. With this change, we will not attempt admin client connection to the broker, if the pod is stuck. Instead the pod will be restarted without making the admin client connection. Therefore the resulting action that restarts the pod that is stuck stays the same for zookeeper mode. This resolves the issue mentioned at the beginning of the proposal that leaves some pods stuck in pending state.
+ 4. Restarts a broker if the broker configuration has changed and cannot be updated dynamically
+ 5. Does not restart a broker if doing so would take the in-sync replicas count below `min.insync.replicas`
```
fvaleri, line 118:
```suggestion
 5. Does not restart a broker if doing so would take the in-sync replicas count below `min.insync.replicas` for any topic hosted on that broker
```

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+
+
+#### Combined mode
+1. Does not restart a combined node performing log recovery
+2. Restarts a combined node if the pod is in stuck state
+3. Attempts to connect an admin client to a broker and if it can't connect restarts the broker
+4. Restarts a combined node if the controller configuration has changed OR if the broker configuration has changed and cannot be updated dynamically
+6. Does not restart a ready combined node if doing so would take the number of caught-up controllers (inc leader) to less than majority of the quorum
+6. Does not restart a ready combined node if doing so would take the in-sync replicas count below `min.insync.replicas`
```
fvaleri, line 127:
```suggestion
7. Does not restart a ready combined node if doing so would take the in-sync replicas count below `min.insync.replicas` for any topic hosted on that broker
```

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+
+#### Unready pod
+Remains the same as Zookeeper mode.
+
+
+#### Configuration changes
+As implemented in [PR #9125](https://github.com/strimzi/strimzi-kafka-operator/pull/9125) until [KIP 919](https://cwiki.apache.org/confluence/display/KAFKA/KIP-919%3A+Allow+AdminClient+to+Talk+Directly+with+the+KRaft+Controller+Quorum+and+add+Controller+Registration) is implemented:
+ - For broker only nodes, configuration changes are handled in the same way as brokers in the ZooKeeper case. If controller configurations changed, they will be ignored and brokers will not restart.
+ - For controller only nodes, a hash of the controller configs is calculated. A change to this hash causes a restart of the node.
+ - For combined nodes, both controller and broker configurations are checked and handled in the same was as brokers in the ZooKeeper case.
```
fvaleri, line 138:
```suggestion
 - For combined nodes, both controller and broker configurations are checked and handled in the same way as brokers in the ZooKeeper case.
```

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+
+
+#### Unready pod
+Remains the same as Zookeeper mode.
+
+
+#### Configuration changes
+As implemented in [PR #9125](https://github.com/strimzi/strimzi-kafka-operator/pull/9125) until [KIP 919](https://cwiki.apache.org/confluence/display/KAFKA/KIP-919%3A+Allow+AdminClient+to+Talk+Directly+with+the+KRaft+Controller+Quorum+and+add+Controller+Registration) is implemented:
+ - For broker only nodes, configuration changes are handled in the same way as brokers in the ZooKeeper case. If controller configurations changed, they will be ignored and brokers will not restart.
+ - For controller only nodes, a hash of the controller configs is calculated. A change to this hash causes a restart of the node.
```
fvaleri, line 137:
```suggestion
 - For controller only nodes, a hash of the controller configs is calculated. A change to this hash causes a restart of the node, regardless of whether this configuration can be updated dynamically or not.
```

056-kafka-roller-kraft.md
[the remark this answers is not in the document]
fvaleri: @tinaselenge this is not clear to me. What do you mean with operational timeout? Is it the reconciliation timeout or something else? Can you also add this part to the stuck pod description?

fvaleri (commented):
Thanks, added some more feedback, but looks mostly good to me.

056-kafka-roller-kraft.md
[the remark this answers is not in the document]
fvaleri: @tinaselenge Can you add the timeout part to the stuck pod description?

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+
+In KRaft mode the KafkaRoller currently skips controller only pods, but performs the above steps on any combined or broker only pods.
+This is causing a problem in combined mode because if the quorum has not formed due to some of the pods not being ready 
+the KafkaRoller will still try to contact the broker via the admin client.
+This call fails because the quorum is not formed, so in some cases this results in the cluster being stuck with some pods 
+in a pending state.
+
+## Motivation
+
+When running in the KRaft mode the controller pods need to be rolled if manually annoted, configuration changes occur or pod is unready. At the moment the
```
showuon, line 60: annoted -> annotated

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+
+
+#### Separate brokers and controllers
+For controller-only:
+ 1. Does not restart a controller performing log recovery
+ 2. Restarts a controller if the pod is in stuck state
+ 3. Attempts to connect an admin client to any of the brokers, if cannot connect, return `UnforceableProblem` which currently results in delay and retry for the pod until the maximum attempts is reached.
```
showuon, line 108: If the max retry attempt is reached, what will we do in this case? 
I know you mentioned it in the above section, but it'd be good to add more description here to make it complete. ex:
` Attempts to connect an admin client to any of the brokers, if cannot connect, return UnforceableProblem which currently results in delay and retry for the pod until the maximum attempts is reached. Then trying to reconcile the brokers to hope to restore admin client connections issue.`

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,158 @@
...
+- From the quorum info returned from the admin API, read `lastCaughtUpTimestamp` of each controller. `lastCaughtUpTimestamp` is the last millisecond timestamp at which a replica controller was known to be caught up with the quorum leader.
+- Check the quorum leader id using the quorum info and identify the `lastCaughtUpTimestamp` of the quorum leader
+- Retrieve value of the Kafka property `controller.quorum.fetch.timeout.ms` from the desired configurations specified in the Kafka CR. If this property does not exist in the desired configurations, then use the hard-coded default value for it which is `2000`. The reason for this is explained further in the **NOTE** below.
+- Mark a Kafka node as caught up if `leaderLastCaughtUpTimestamp - replicaLastCaughtUpTimestamp < controllerQuorumFetchTimeoutMs`. This will exclude the controller KafkaRoller is currently considering to restart, because we are trying to check whether the quorum would stay healthy when this controller is restarted.
+- Count each controller node that is caught up, including the leader (`numOfCaughtUpControllers`).
+- Can restart if: `numOfCaughtUpControllers >= ceil((double) totalNumOfControllers / 2)`.
```
showuon, line 99: The calculation is still not completely correct. 
In odds number's case, ex: 3 controllers case, `ceil((double) totalNumOfControllers / 2` -> 2, which is expected.
In even number's case, ex: 4 controllers case, `ceil((double) totalNumOfControllers / 2` -> 2, which is not right. the majority should be 3.

I think it should be `ceil((double) (totalNumOfControllers + 1) / 2`

showuon (commented):
Overall LGTM! Thanks again for the effort for this proposal! It's not an easy one.

056-kafka-roller-kraft.md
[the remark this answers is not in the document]
katheris: We've updated the proposal to be more explicit here on what happens when a pod is deemed "stuck". Is there anything else that needs addressing or can we view this thread as resolved?

056-kafka-roller-kraft.md
katheris: Yes you're right, I'll update it

056-kafka-roller-kraft.md
katheris: @fvaleri @scholzj are you happy with Tina's responses on this thread? Can it be marked as resolved?

katheris:
@scholzj @showuon @tombentley @ppatierno @fvaleri I think we've addressed all your comments. Is there anything outstanding we've missed? If not I'll rebase and fix the conflict

fvaleri (approved):
LGTM. Thanks.

showuon (approved):
LGTM!

056-kafka-roller-kraft.md
```diff
@@ -0,0 +1,159 @@
...
+- If failed to connect to the brokers, return `UnforceableProblem` which currently results in delay and retry for the pod until the maximum attempts is reached.
+- From the quorum info returned from the admin API, read `lastCaughtUpTimestamp` of each controller. `lastCaughtUpTimestamp` is the last millisecond timestamp at which a replica controller was known to be caught up with the quorum leader.
+- Check the quorum leader id using the quorum info and identify the `lastCaughtUpTimestamp` of the quorum leader
+- Retrieve value of the Kafka property `controller.quorum.fetch.timeout.ms` from the desired configurations specified in the Kafka CR. If this property does not exist in the desired configurations, then use the hard-coded default value for it which is `2000`. The reason for this is explained further in the **NOTE** below.
+- Mark a Kafka node as caught up if `leaderLastCaughtUpTimestamp - replicaLastCaughtUpTimestamp < controllerQuorumFetchTimeoutMs`. This will exclude the controller KafkaRoller is currently considering to restart, because we are trying to check whether the quorum would stay healthy when this controller is restarted.
```
ppatierno, line 97:
```suggestion
- Mark a KRaft controller node as caught up if `leaderLastCaughtUpTimestamp - replicaLastCaughtUpTimestamp < controllerQuorumFetchTimeoutMs`. This will exclude the controller KafkaRoller is currently considering to restart, because we are trying to check whether the quorum would stay healthy when this controller is restarted.
```

README.md
```diff
@@ -9,6 +9,7 @@ This repository list of proposals for the Strimzi project. A template for new pr
 
 |  #  | Title                                                                 |
 | :-: |:----------------------------------------------------------------------|
+| 56  | [Kafka Roller KRaft Support](./056-kafka-roller-kraft.md)
```
ppatierno, line 12: You should rebase, use a different updated index and renaming the MD file as well.

katheris: Yeah I will, I was holding off to see if there were any other comments as renaming the file sometimes makes it harder to see the previous discussions

ppatierno (approved):

056-kafka-roller-kraft.md
katheris: @showuon In this calculation, should "numOfCaughtUpControllers" only include the leader/active controller if we aren't trying to restart the current active leader?

i.e. if we have nodes 0,1,2 and 0 is the active leader, 1 is a caught up follower and 2 is a non-caught up follower, then we can restart 2 but not 0 or 1?

katheris: Based on discussion in the community call on 5th October I'm going to tweak the wording here

Frawless (approved):

scholzj:
This has no 3 binding +1 votes and 2 non-binding +1 votes. 

@tombentley you left some comments but did not approve it yet. Were they addressed? Do you have some more comments?

060-kafka-roller-kraft.md
```diff
@@ -0,0 +1,159 @@
...
+- Controller's configuration has changed
+
+The triggers for broker remain the same as Zookeeper mode.
+
+### Restart conditions
+The restart conditions that the KafkaRoller considers in different modes are described below.
+
+#### The new quorum check
+The restart conditions include a new check for controllers to verify that restarting the Kafka node does not affect the quorum health.
+The proposed check is:
```
PaulRMellor, line 91:
```suggestion
The proposed check ensures that a sufficient majority of controller nodes are caught up before allowing a restart:
```

060-kafka-roller-kraft.md
```diff
@@ -0,0 +1,159 @@
...
+- Count each controller node that is caught up (`numOfCaughtUpControllers`).
+- Can restart if: `numOfCaughtUpControllers >= ceil((double) (totalNumOfControllers + 1) / 2)`.
+
+> NOTE: Until [KIP-919](https://cwiki.apache.org/confluence/display/KAFKA/KIP-919%3A+Allow+AdminClient+to+Talk+Directly+with+the+KRaft+Controller+Quorum) is implemented, KafkaRoller cannot create an admin connection to the controller directly to describe its configuration or the quorum state. Therefore KafkaRoller checks the desired configurations to get the value of `controller.quorum.fetch.timeout.ms` and creates admin client connection to the brokers for the quorum check. If KafkaRoller cannot connect to any of the brokers after the maximum attempts to retry, the controllers will be marked as failed to reconcile because of not being able to determine the quorum health. KafkaRoller would then try to reconcile the brokers, which may help restoring admin client connections to them. In this scenario, the reconcilation will be completed with failure and reported to the operator. The controllers quorum check will be retried in the next round of reconcilation.
```
PaulRMellor, line 101: I get page not found on the link

060-kafka-roller-kraft.md
```diff
@@ -0,0 +1,159 @@
...
+
+
+#### Separate brokers and controllers
+For controller-only:
+ 1. Does not restart a controller performing log recovery
+ 2. Restarts a controller if the pod is in stuck state
+ 3. Attempts to connect an admin client to any of the brokers, if cannot connect, return `UnforceableProblem` which results in delay and retry for the pod until the maximum attempts is reached.
```
PaulRMellor, line 108:
```suggestion
 3. Attempts to connect an admin client to any of the brokers. If it cannot connect, returns `UnforceableProblem` which results in delay and retry for the pod until the maximum attempt is reached. 
```

060-kafka-roller-kraft.md
```diff
@@ -0,0 +1,159 @@
...
+When maximum attempts is reached the KafkaRoller will move on to restart broker pods incase this resolves connection issues but will still ultimately mark the reconciliation as failed.
+ 4. Restarts a controller if the controller configuration has changed
+ 5. Does not restart a controller if doing so would take the number of caught-up controllers (inc leader) to less than majority of the quorum
+
+For broker-only:
+ 1. Does not restart a broker performing log recovery
+ 2. Restarts a broker if the pod is in stuck state
+ 3. Attempts to connect an admin client to a broker and if it can't connect restarts the broker
+ > **NEW** Currently when pod is stuck and out of date, it gets restarted because admin client connection fails. With this change, we will not attempt admin client connection to the broker, if the pod is stuck. Instead the pod will be restarted without making the admin client connection. Therefore the resulting action that restarts the pod that is stuck stays the same for zookeeper mode. This resolves the issue mentioned at the beginning of the proposal that leaves some pods stuck in pending state.
```
PaulRMellor, line 117:
```suggestion
 > **NEW** Currently when pod is stuck and out of date, it is restarted because admin client connection fails. With this change, we will not attempt admin client connection to the broker if the pod is stuck. Instead the pod will be restarted without making the admin client connection. Therefore, the resulting action that restarts the pod that is stuck stays the same for ZooKeeper mode. This resolves the issue mentioned at the beginning of the proposal that leaves some pods stuck in pending state.
```

060-kafka-roller-kraft.md
```diff
@@ -0,0 +1,159 @@
...
+1. Does not restart a combined node performing log recovery
+2. Restarts a combined node if the pod is in stuck state
+3. Attempts to connect an admin client to a broker and if it can't connect restarts the broker
+4. Restarts a combined node if the controller configuration has changed OR if the broker configuration has changed and cannot be updated dynamically
+5. Does not restart a ready combined node if doing so would take the number of caught-up controllers (inc leader) to less than majority of the quorum
+6. Does not restart a ready combined node if doing so would take the in-sync replicas count below `min.insync.replicas` for any topic hosted on that broker
+
+
+#### Unready pod
+Remains the same as Zookeeper mode.
```
PaulRMellor, line 132:
```suggestion
Remains the same as ZooKeeper mode.
```

060-kafka-roller-kraft.md
```diff
@@ -0,0 +1,159 @@
...
+
+
+#### Configuration changes
+As implemented in [PR #9125](https://github.com/strimzi/strimzi-kafka-operator/pull/9125) until [KIP 919](https://cwiki.apache.org/confluence/display/KAFKA/KIP-919%3A+Allow+AdminClient+to+Talk+Directly+with+the+KRaft+Controller+Quorum+and+add+Controller+Registration) is implemented:
+ - For broker only nodes, configuration changes are handled in the same way as brokers in the ZooKeeper case. If controller configurations changed, they will be ignored and brokers will not restart.
```
PaulRMellor, line 137:
```suggestion
 - For broker-only nodes, configuration changes are handled in the same way as brokers in the ZooKeeper case. If controller configurations changed, they will be ignored and brokers will not restart.
```

PaulRMellor (approved):
The proposal clearly outlines the required behavior of KafkaRoller and restarts within a KRaft cluster. I left a few minor suggestions.

scholzj:
@tinaselenge @katheris Could you please incorporate the comments from Paul? Once done, I would say we give people one more day, and if nobody comes up with anything we will merge it as approved.

060-kafka-roller-kraft.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Restart conditions
+ 1. Does not restart a broker performing log recovery
+ 2. Attempts to connect an admin client to a broker and if it can't connect, restarts the broker
+ 3. Restarts a broker if the pod is in stuck state
+ 4. Restarts a broker if the configuration has changed and cannot be updated dynamically
+ 5. Does not restart a broker if doing so would take the in-sync replicas count below `min.insync.replicas` for any topic hosted on that broker
+
+#### Unready pod
+If pod is unready, KafkaRoller awaits its readiness before doing anything. If it hasn't become ready still, it checks whether it is stuck. A pod is considered stuck if it is in one of following states:
```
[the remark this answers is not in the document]
tinaselenge: I updated it to mention that the operational timeout is used for awaiting readiness.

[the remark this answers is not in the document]
scholzj: Operational timeout is 5 minutes -> that is way too long for stuck pods.

060-kafka-roller-kraft.md
```diff
@@ -0,0 +1,159 @@
...
+- Pod or its StrimziPodSet is annotated for manual rolling update
+- Pod is unready
+- Broker's configuration has changed
+
+### Restart conditions
+ 1. Does not restart a broker performing log recovery
+ 2. Attempts to connect an admin client to a broker and if it can't connect, restarts the broker
+ 3. Restarts a broker if the pod is in stuck state
+ 4. Restarts a broker if the configuration has changed and cannot be updated dynamically
+ 5. Does not restart a broker if doing so would take the in-sync replicas count below `min.insync.replicas` for any topic hosted on that broker
```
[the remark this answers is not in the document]
tinaselenge: KafkaRoller checks these conditions in order. Instead of splitting out into 2 lists, we added a sentence to clarify that the restart conditions are checked in the specified order.

scholzj:
This now has 4 binding +1 votes and 2 non-biding +1 votes. If nobody comes with any new comments until tomorrow EOB, we will merge it as approved.

060-kafka-roller-kraft.md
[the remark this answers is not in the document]
tinaselenge: Checking the current design again, the statement about waiting before checking if pod is stuck is actually incorrect. We only wait for pod readiness if pod is not stuck. I will correct the statement.

060-kafka-roller-kraft.md
```diff
@@ -0,0 +1,159 @@
...
+
+#### Unready pod
+If pod is unready, KafkaRoller awaits its readiness before doing anything. If it hasn't become ready still, it checks whether it is stuck. A pod is considered stuck if it is in one of following states:
+- `CrashLoopBackOff`
+- `ImagePullBackOff`
+- `ContainerCreating`
+- `Pending` and `Unschedulable`
+
+If pod is unready and stuck, KafkaRoller restarts it if pod is out of date. Otherwise, KafkaRoller fails the reconcilation and does not restart other pods. This is because a pod being stuck could mean restarting the other pods will result in then being in the same state.
+If pod is unready but not stuck, KafkaRoller considers the restart conditions and may restart it.
```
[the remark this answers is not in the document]
tinaselenge: It now refers to the specific restart conditions it considers.

060-kafka-roller-kraft.md
```diff
@@ -0,0 +1,159 @@
...
+
+
+#### Configuration changes
+ - Retrieves the current Kafka configurations of the broker via admin client and compares it with the desired configurations specified in the Kafka CR.
```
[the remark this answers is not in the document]
tinaselenge: Added a context introducing the list.

tinaselenge:
Thanks! I think we have addressed all the comments from @PaulRMellor . Please let us know if you see anything I have missed.

scholzj:
Approved with 4 binding +1 votes and 2 non-biding +1 votes (plus Kate and Tina as authors).

scholzj:
Thanks for the proposal @tinaselenge and @katheris

PR #93, opened 2023-09-20
Allow running ZooKeeper and KRaft based clusters in parallel

ppatierno:
This PR adds a proposal to allow running ZooKeeper and KRaft based clusters in parallel when the KRaft feature gate is enabled.

### DIFF
--- a/057-marking-zk-kraft-clusters.md
+++ b/057-marking-zk-kraft-clusters.md
@@ -31,7 +31,7 @@ The possible values would be:
 
 This way, during the reconciliation, the operator is able to detect a ZooKeeper-based cluster avoiding the warning, as described in the "Current" section, but allowing the user to operate it.
 On the KRaft side, the annotation would be needed to have the operator reconciling the corresponding `Kafka` custom resource.
-Without the annotation, but the `UseKRaft` feature gate enabled, the operator would ignore the cluster.
+Without the annotation, but the `UseKRaft` feature gate enabled, the operator would try to handle it as a ZooKeeper-based one.
 This approach is actually the same as for the node pools: enabling the `KafkaNodePools` feature gate on the operator is not enough and the user needs to apply the `strimzi.io/node-pools: enabled` annotation on the `Kafka` custom resource using node pools for brokers (and controllers, if KRaft enabled as well).
 
 The same annotation could be even used in the proposal [PR#90](https://github.com/strimzi/proposals/pull/90) for handling the ZooKeeper to KRaft migration steps.
--- a/057-marking-zk-kraft-clusters.md
+++ b/057-marking-zk-kraft-clusters.md
@@ -1,0 +1,53 @@
+# Marking ZooKeeper and KRaft based clusters for KRaft enabled operators
+
+This proposal is about providing to the Strimzi Cluster Operator a way to "detect" if an Apache Kafka cluster is ZooKeeper or KRaft based when the KRaft feature gate is enabled.
+
+## Current situation
+
+Currently, when the `UseKRaft` feature gate is enabled, the operator expects all the Apache Kafka clusters being KRaft-based.
+There is no way to differentiate between clusters running in ZooKeeper or KRaft mode.
+When a `Kafka` custom resource, configured to use ZooKeeper for metadata, is reconciled, the operator detects it with a missing KRaft controllers configuration logging the following warning:
+
+```shell
+io.strimzi.operator.common.model.InvalidResourceException: Tha Kafka cluster my-cluster is invalid: [At least one KafkaNodePool with the controller role and at least one replica is required when KRaft mode is enabled]
+```
+
+In this case, the reconciliation fails and the cluster is not operated at all.
+It means that the operator doesn't take any actions on changes applied by the user to the `spec.kafka` or `spec.zookeeper` in the `Kafka` custom resource.
+
+## Motivation
+
+Leaving a ZooKeeper-based cluster not operated when the `UseKRaft` feature gate is enabled looks like a bad thing for the user, until the ZooKeeper support will be completely removed.
+Before that, the user should be able to have both ZooKeeper and KRaft based clusters running and operated when the `UseKRaft` feature gate is enabled.
+Also, taking into account the future ZooKeeper to KRaft migration process, as described in the proposal [PR#90](https://github.com/strimzi/proposals/pull/90), it would help the operator migration component to detect which cluster is ZooKeeper-based so "migrate-able" and which one is already KRaft-based.
+
+## Proposal
+
+The proposal is about adding a new `strimzi.io/kraft` annotation to the `Kafka` custom resource, to be applied by the user, in order to identify a ZooKeeper or KRaft based cluster.
+The possible values would be:
+
+* `disabled` (or missing): identifies a ZooKeeper-based cluster.
+* `enabled`: identifies a KRaft-based cluster.
+
+This way, during the reconciliation, the operator is able to detect a ZooKeeper-based cluster avoiding the warning, as described in the "Current" section, but allowing the user to operate it.
+On the KRaft side, the annotation would be needed to have the operator reconciling the corresponding `Kafka` custom resource.
+Without the annotation, but the `UseKRaft` feature gate enabled, the operator would ignore the cluster.
+This approach is actually the same as for the node pools: enabling the `KafkaNodePools` feature gate on the operator is not enough and the user needs to apply the `strimzi.io/node-pools: enabled` annotation on the `Kafka` custom resource using node pools for brokers (and controllers, if KRaft enabled as well).
+
+The same annotation could be even used in the proposal [PR#90](https://github.com/strimzi/proposals/pull/90) for handling the ZooKeeper to KRaft migration steps.
+
+## Affected/not affected projects
+
+The Strimzi Cluster Operator is the only project affected by this proposal.
+Its logic needs to be updated in order to be able to handle the new `strimzi.io/kraft` annotation (or the missing of it).
+No other projects in the Strimzi ecosystem are impacted by this proposal.
+
+## Compatibility
+
+If the user is running the operator with `UseKRaft` feature gate enabled, it will detect ZooKeeper-based cluster allowing to operate them again.
+If the user has a KRaft-based cluster deployed, it would be ignored and not reconciled anymore, unless the user applies `strimzi.io/kraft: enabled` annotation on it.
+Breaking this backward compatibility is expected taking into account that the KRaft support in Strimzi is still behind a feature gate and should be considered just for development purposes. 
+
+## Rejected alternatives
+
+As first iteration of the proposal there are no rejected alternatives.
--- a/057-marking-zk-kraft-clusters.md
+++ b/057-marking-zk-kraft-clusters.md
@@ -1,12 +1,14 @@
 # Marking ZooKeeper and KRaft based clusters for KRaft enabled operators
 
-This proposal is about providing to the Strimzi Cluster Operator a way to "detect" if an Apache Kafka cluster is ZooKeeper or KRaft based when the KRaft feature gate is enabled.
+This proposal is about changing the way the current `UseKRaft` feature gate works.
+It changes the way the Strimzi Cluster Operator handles ZooKeeper and KRaft based clusters when such feature gate is enabled.
 
 ## Current situation
 
-Currently, when the `UseKRaft` feature gate is enabled, the operator expects all the Apache Kafka clusters being KRaft-based.
+Currently, when the `UseKRaft` feature gate is enabled (together with the required `KafkaNodePools` one), the operator expects all the already running Apache Kafka clusters being KRaft-based.
+It means that the `Kafka` custom resources are configured together with `KafkaNodePool`(s) with `broker` and `controller` roles.
 There is no way to differentiate between clusters running in ZooKeeper or KRaft mode.
-When a `Kafka` custom resource, configured to use ZooKeeper for metadata, is reconciled, the operator detects it with a missing KRaft controllers configuration logging the following warning:
+When a `Kafka` custom resource is configured to actually use ZooKeeper or it's just badly configured but supposed to be KRaft based, the operator detects it as having a missing KRaft controllers configuration and logs the following warning:
 
 ```shell
 io.strimzi.operator.common.model.InvalidResourceException: Tha Kafka cluster my-cluster is invalid: [At least one KafkaNodePool with the controller role and at least one replica is required when KRaft mode is enabled]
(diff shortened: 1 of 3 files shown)


ppatierno:
@scholzj I was thinking more on the proposal ... so my additional thoughts (it's even possible I got wrong our chat offline) before updating the proposal.
When the annotation is not there, we agreed on the cluster being considered as ZooKeeper.
My doubt is about should be it handled as actual ZooKeeper or just ignored during the reconciliation? (so not operated).
With KRaft feature gate enabled, we assume all cluster should be KRaft with annotation set on `enabled`.
It means that if the user, by upgrading the operator with this new version, has an already running KRaft cluster, it will be ignored and not reconciled unless he applies the `enabled` flag.
Because missing annotation on an already KRaft cluster would be handled as a ZooKeeper one ... but of course it's not possible in the operator. It would fail.
If the cluster is really ZooKeeper, only way user has is to apply the annotation to migrate as per corresponding proposal.

057-marking-zk-kraft-clusters.md
```diff
@@ -0,0 +1,59 @@
+# Marking ZooKeeper and KRaft based clusters for KRaft enabled operators
+
+This proposal is about changing the way the current `UseKRaft` feature gate works.
+It changes the way the Strimzi Cluster Operator handles ZooKeeper and KRaft based clusters when such feature gate is enabled.
```
scholzj, line 4:
```suggestion
It changes the way the Strimzi Cluster Operator handles ZooKeeper and KRaft based clusters when it is enabled.
```

057-marking-zk-kraft-clusters.md
```diff
@@ -0,0 +1,59 @@
...
+Leaving a ZooKeeper-based cluster not operated when the `UseKRaft` feature gate is enabled looks like a bad thing for the user, until the ZooKeeper support will be completely removed.
+Before that, the user should be able to have both ZooKeeper and KRaft based clusters running and operated when the `UseKRaft` feature gate is enabled.
+Also, taking into account the future ZooKeeper to KRaft migration process, as described in the proposal [PR#90](https://github.com/strimzi/proposals/pull/90), it would help the operator migration component to detect which cluster is ZooKeeper-based so "migrate-able" and which one is already KRaft-based.
+
+## Proposal
+
+The proposal is about adding a new `strimzi.io/kraft` annotation to the `Kafka` custom resource, to be applied by the user, in order to identify a ZooKeeper or KRaft based cluster.
+The possible values would be:
+
+* `disabled` (or missing): identifies a ZooKeeper-based cluster.
```
scholzj, line 31: I guess any invalid value should default to ZooKeeper as well. That way you can easily handle it with `if ("enabled".equals(annoValue)) { ... }`.

scholzj (commented):
I think the core of the proposal is good. Just some comments to the formulations in the text etc.

057-marking-zk-kraft-clusters.md
```diff
@@ -0,0 +1,59 @@
+# Marking ZooKeeper and KRaft based clusters for KRaft enabled operators
```
scholzj, line 1: I did not notice it initially ... but mAybe we can call it something like "Allow running ZooKeeper- and KRaft-based clusters in parallel"? I do not think I would understand what it is about from the current title if I didn't know about it upfront.

ppatierno: Yeah I was looking at a better title, I think your suggestion makes more sense.

fvaleri (approved):
LGTM. Thanks.

057-run-zk-kraft-clusters-parallel.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Current situation
+
+Currently, when the `UseKRaft` feature gate is enabled (together with the required `KafkaNodePools` one), the operator expects all the already running Apache Kafka clusters being KRaft-based.
+It means that the `Kafka` custom resources are configured together with `KafkaNodePool`(s) with `broker` and `controller` roles.
+There is no way to differentiate between clusters running in ZooKeeper or KRaft mode.
+When a `Kafka` custom resource is configured to actually use ZooKeeper or it's just badly configured but supposed to be KRaft based, the operator detects it as having a missing KRaft controllers configuration and logs the following warning:
+
+```shell
+io.strimzi.operator.common.model.InvalidResourceException: Tha Kafka cluster my-cluster is invalid: [At least one KafkaNodePool with the controller role and at least one replica is required when KRaft mode is enabled]
```
see-quick, line 14:
```suggestion
io.strimzi.operator.common.model.InvalidResourceException: The Kafka cluster my-cluster is invalid: [At least one KafkaNodePool with the controller role and at least one replica is required when KRaft mode is enabled]
```

057-run-zk-kraft-clusters-parallel.md
```diff
@@ -0,0 +1,65 @@
...
+* `enabled`: identifies a KRaft-based cluster.
+
+This way, during the reconciliation, the operator is able to "detect" a ZooKeeper-based cluster avoiding the warning and allowing the user to operate it.
+On the KRaft side, the annotation would be needed to have the operator reconciling the corresponding `Kafka` custom resource.
+Without the annotation, but the `UseKRaft` feature gate enabled, the operator would try to handle it as a ZooKeeper-based one.
+The `UseKRaft` feature gate does currently not support any upgrades of KRaft clusters.
+It is expected to be used only for short-lives clusters used for development and testing.
+So no clusters are expected to exist.
+Therefore adding the annotation now does not present any backwards compatibility issues.
+Only ZooKeeper-based clusters are expected and newly created KRaft-based clusters having the `strimzi.io/kraft: enabled`.
```
see-quick, line 44: One point with a consistency somewhere you use KRaft-based and somewhere Kraft based. I would stick with KRaft-based everywhere. But it's just a nit ).

057-run-zk-kraft-clusters-parallel.md
```diff
@@ -0,0 +1,65 @@
...
+The possible values would be:
+
+* `disabled`, missing or any invalid value: identifies a ZooKeeper-based cluster.
+* `enabled`: identifies a KRaft-based cluster.
+
+This way, during the reconciliation, the operator is able to "detect" a ZooKeeper-based cluster avoiding the warning and allowing the user to operate it.
+On the KRaft side, the annotation would be needed to have the operator reconciling the corresponding `Kafka` custom resource.
+Without the annotation, but the `UseKRaft` feature gate enabled, the operator would try to handle it as a ZooKeeper-based one.
+The `UseKRaft` feature gate does currently not support any upgrades of KRaft clusters.
+It is expected to be used only for short-lives clusters used for development and testing.
```
see-quick, line 41:
```suggestion
It is expected to be used only for short-lived clusters used for development and testing.
```

see-quick (approved):
Few nits. LGTM 👍

057-run-zk-kraft-clusters-parallel.md
scholzj: Maybe you should open a PR to fix it in the code instead? 😮

ppatierno: Yeah, I will update here and open the PR on code.

ppatierno: PR is here https://github.com/strimzi/strimzi-kafka-operator/pull/9149

057-run-zk-kraft-clusters-parallel.md
ppatierno: You are right but when it's just one I use "ZooKeeper-based cluster and KRaft-based cluster", when I want to mention both using plural I say "...ZooKeeper and KRaft based clusters...". Not sure if and where I should put `-based` because `based` is referring to both.

scholzj (approved):
LGTM. Thanks.

see-quick (approved):

im-konge (approved):

057-run-zk-kraft-clusters-parallel.md
```diff
@@ -0,0 +1,65 @@
...
+## Proposal
+
+The proposal is about adding a new `strimzi.io/kraft` annotation to the `Kafka` custom resource, to be applied by the user, in order to identify a ZooKeeper or KRaft based cluster.
+The possible values would be:
+
+* `disabled`, missing or any invalid value: identifies a ZooKeeper-based cluster.
+* `enabled`: identifies a KRaft-based cluster.
+
+This way, during the reconciliation, the operator is able to "detect" a ZooKeeper-based cluster avoiding the warning and allowing the user to operate it.
+On the KRaft side, the annotation would be needed to have the operator reconciling the corresponding `Kafka` custom resource.
```
tombentley, line 38: Maybe consider adding a table, to make the behaviour clear with fewer words?

| Operator feature gate | CR Annotation | Operator behaviour |
| --- | --- | --- |
| `-UseKraft` | `enabled` | ??log a warning??|
| `-UseKraft` | anything else | ZK reconciliation | 
| `+UseKraft` | `enabled` | KRaft reconciliation |
| `+UseKraft` | anything else |ZK reconciliation |

057-run-zk-kraft-clusters-parallel.md
```diff
@@ -0,0 +1,65 @@
...
+No other projects in the Strimzi ecosystem are impacted by this proposal.
+
+## Compatibility
+
+If the user is running the operator with `UseKRaft` feature gate enabled, it will detect ZooKeeper-based cluster allowing to operate them again.
+If the user has a KRaft-based cluster deployed, because of the missing annotation, the operator will try to handle it as it was ZooKeeper-based which would fail.
+In this case, if the user applies `strimzi.io/kraft: enabled` annotation on it, there could be unexpected results on the reconciliation. So doing this is not supported.
+In general, when this proposal is in place, the expectation is that there won't be KRaft-based clusters already running but there could be only ZooKeeper-based clusters running.
+Breaking this backward compatibility is expected by taking into account that the KRaft support in Strimzi is still behind a feature gate and should be considered just for development purposes. 
+
```
tombentley, line 62: I'd like to see mention of what happens to this annotation when

* KRaft is the only thing Kafka supports.
* An existing KRaft cluster has its annotation changed to `disabled`. 

The problem I see with it being a (user managed) annotation is that while that's fine for the user to express their intent, it cannot also be used by the operator to express how it interprets the resource.

057-run-zk-kraft-clusters-parallel.md
ppatierno: I can add the table if you like but regarding the first use case `-UseKRaft` with `strimzi.io/kraft: enabled` annotation, I made a comparison with what happens today with node pools (this proposal is aiming to do almost the same but for KRaft feature gate). When we have `-KafkaNodePools` (so no node pools feature gate enabled) but `strimzi.io/node-pools: enabled` on the Kafka CR (deployed together with some `KafkaNodePool`(s)), the operator just ignores that annotation (and node pools) and deploys the Kafka cluster just following the `spec.kafka` and `spec.zookeeper`. No warning printed, it's just ignored, which sounds fine to me imho. @scholzj what are your thoughts on this?

057-run-zk-kraft-clusters-parallel.md
ppatierno: > I'd like to see mention of what happens to this annotation when ...

I would expect the annotation handling just removed from the operator and the annotation ignored. It's going to happen the same way for the `node-pools` annotation as stated in the corresponding proposal: "So once the ZooKeeper support is removed from Strimzi, we will also remove the support for the virtual node pool and the strimzi.io/node-pools annotation" ... details [here](https://github.com/strimzi/proposals/blob/main/050-Kafka-Node-Pools.md).

> it cannot also be used by the operator to express how it interprets the resource.

I am not sure what do you mean by this.

057-run-zk-kraft-clusters-parallel.md
```diff
@@ -0,0 +1,73 @@
+# Allow running ZooKeeper and KRaft based clusters in parallel
+
+This proposal is about changing the way the current `UseKRaft` feature gate works.
+It changes the way the Strimzi Cluster Operator handles ZooKeeper and KRaft based clusters when it is enabled.
+
+## Current situation
+
+Currently, when the `UseKRaft` feature gate is enabled (together with the required `KafkaNodePools` one), the operator expects all the already running Apache Kafka clusters being KRaft-based.
```
PaulRMellor, line 8:
```suggestion
Currently, when the `UseKRaft` feature gate is enabled (together with the required `KafkaNodePools` one), the operator expects the Apache Kafka clusters already running to be KRaft-based.
```

057-run-zk-kraft-clusters-parallel.md
```diff
@@ -0,0 +1,73 @@
...
+Without the annotation, but the `UseKRaft` feature gate enabled, the operator would try to handle it as a ZooKeeper-based one.
+
+| Operator Feature Gate | `strimzi.io/kraft` annotation | Operator behaviour       |
+|-----------------------|-------------------------------|--------------------------|
+| `-UseKRaft`           | `enabled`                     | Ignore annotation        |
+| `-UseKRaft`           | missing or anything else      | ZooKeeper reconciliation |
+| `+UseKRaft`           | `enabled`                     | KRaft reconciliation     |
+| `+UseKRaft`           | missing or anything else      | ZooKeeper reconciliation |
+
+The `UseKRaft` feature gate does currently not support any upgrades of KRaft clusters.
```
PaulRMellor, line 48:
```suggestion
The `UseKRaft` feature gate does not currently support any upgrades of KRaft clusters.
```

057-run-zk-kraft-clusters-parallel.md
```diff
@@ -0,0 +1,73 @@
...
+The possible values would be:
+
+* `disabled`, missing or any invalid value: identifies a ZooKeeper-based cluster.
+* `enabled`: identifies a KRaft-based cluster.
+
+This way, during the reconciliation, the operator is able to "detect" a ZooKeeper-based cluster avoiding the warning and allowing the user to operate it.
+On the KRaft side, the annotation would be needed to have the operator reconciling the corresponding `Kafka` custom resource.
+Without the annotation, but the `UseKRaft` feature gate enabled, the operator would try to handle it as a ZooKeeper-based one.
+
+| Operator Feature Gate | `strimzi.io/kraft` annotation | Operator behaviour       |
```
PaulRMellor, line 41: would it help to have a column for KafkaNodePools settings too?

PaulRMellor (approved):
The proposal is clear and the suggested Kraft annotation makes sense

057-run-zk-kraft-clusters-parallel.md
ppatierno: KafkaNodePools feature gate is mandatory for using KRaft so it's implicit that it should be always enabled. Having it disabled doesn't make any KRaft feature gate value working. So I am not sure it makes sense to add a column for that.

057-run-zk-kraft-clusters-parallel.md
scholzj: I would probably just ignore it when the feature gate is disabled. But not a strong opinion on this one.

ppatierno:
@tombentley can you have another pass please, we replied your latest comments/suggestions. Thanks!

tombentley (approved):

ppatierno:
This has now 5 binding votes (+ my implicit one) and 1 non-binding vote. If there are no further comments till EOD, I will close this as approved.

katheris:
LGTM

katheris (approved):

ppatierno:
Approved with 5 binding +1 votes and 2 non-binding +1 vote.

PR #94, opened 2023-10-01
KRaft upgrades and downgrades

scholzj:
This proposal introduces the Kafka upgrade mechanism for KRaft clusters. It describes how will Strimzi upgrade or downgrade KRaft clusters from one Apache Kafka version to another.

This is currently a first draft.


fvaleri (commented):
Thanks Jakub. I left some initial comments, but overall it makes sense to me. Will have another pass when you remove the draft status.

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Motivation
+
+KRaft-to-Kraft upgrades and downgrades are one of the missing parts of Strimzi's KRaft support.
+Supporting upgrades and downgrades is required for calling KRaft production ready.
+This work can be done in parallel with the work on other proposals such as migration of ZooKeeper-based clusters to KRaft based clusters and others.
+
+## Proposal
+
+Strimzi should follow the Apache Kafka procedure for upgrading / downgrading Kafka.
```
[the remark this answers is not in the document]
scholzj: I linked to the docs.

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,177 @@
...
+
+Once all the Kafka pods are rolled, the next step will depend on whether `Kafka.spec.kafka.metadataVersion` is set or not.
+When it is not set, Strimzi will automatically update it to the default version corresponding to the new Kafka version.
+And the Kafka upgrade will be complete with this.
+
+In case the `Kafka.spec.kafka.metadataVersion` field is set, it will just check the Kafka cluster has the desired metadata version as in any other reconciliation.
+The user will be expected to verify the Kafka cluster and change the `Kafka.spec.kafka.metadataVersion` when everything seems to work fine.
+Only after changing the metadata version, the upgrade will be considered complete.
+Similarly to today's ZooKeeper based implementation, a warning will be issued (in log and in Kafka CR conditions) when the metadata version does not correspond to the Kafka version.
+
```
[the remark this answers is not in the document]
scholzj: I'm not sure this is relevant to the operator.

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,177 @@
...
+Apache Kafka [KRaft upgrade procedure](https://kafka.apache.org/documentation/#upgrade_350_kraft) consists of 2 different steps:
+1. Roll our the new Kafka version and verify everything is working fine.
+2. Update the `metadata.version` using the `kafka-reafutres.sh` command line tool (or using the Kafka Admin API).
+
+Downgrade is possibly as well, by:
+1. Downgrading the `metadata.version`
+2. Rolling out the older Kafka version
+
+The `metadata.version` downgrade might be subject to limitations as the older `metadata.version` might not support all the metadata used by the new version.
```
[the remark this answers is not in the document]
scholzj: This is a declarative mechanism driven by the user. There is no way to _ask for confirmation_. But yes, it is worth clarifying.

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,177 @@
...
+
+Strimzi currently supports upgrades of ZooKeeper-based Kafka clusters.
+The upgrades of ZooKeeper-based clusters are multi-stage process:
+1. First, ZooKeeper nodes are rolled to use the new container image from the new Kafka version and possibly also the new ZooKeeper version (if the new Kafka version uses new ZooKeeper version).
+2. Next, Kafka brokers are rolled to use the new container image with the new Apache Kafka version.
+3. Finally, the Kafka brokers are rolled one more time to update the `inter.broker.protocol.version`.
+   This step is done either when the user changes the `inter.broker.protocol.version` in `Kafka.spec.kafka.config`.
+   Or automatically in the next reconciliation when `inter.broker.protocol.version` is not set at all in `Kafka.spec.kafka.config`.
+
+Strimzi also supports downgrades of ZooKeeper-based Kafka clusters.
```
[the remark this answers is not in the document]
scholzj: I do not think it is needed for describing the existing implementation.

fvaleri:
Two considerations regarding metadata downgrades.

---

I was wondering if we should be a bit more opinionated about the downgrade versions we allow. Nowadays, it is not uncommon to see clusters running latest Kafka binaries with very old IBP versions (i.e. Kafka 3.5 with IBP 2.2). This could happen for various reasons: their upgrade documentation doesn't include the recommended IBP upgrade step, or they are simply afraid that this would break some of their client applications. In any case, this puts the user in an untested setup, which can trigger nasty bugs. I think the same misbehavior will happen with `metadata.version`, especially considering the Kafka 4.0 major release. What if we restrict the `metadata.version` downgrade to the supported Kafka versions? That way, we reduce the risks related to running new binaries with very old metadata records.

---

> When downgrading the metadata.version in Kafka, the operator will first try to do a safe downgrade and if it fails, it will try unsafe downgrade. This will allow us to log that unsafe upgrade was needed.

I'm not convinced that automating the unsafe (lossy) metadata downgrade by default is the right thing to do. Cluster metadata are critical, and there is a concrete risk of breaking the service. Information is lost when you force a downgrade, so you can't simply revert by upgrading. I think we need the user to opt in before unsafe downgrade is applied. If it's not already there, we could raise a warning (log and status) asking to enable unsafe downgrades by using the annotation `strimzi.io/unsafe-downgrade: "true"`. This would be disabled by default. Wdyt?

@mimaison @tombentley @showuon

scholzj:
> I was wondering if we should be a bit more opinionated about the downgrade versions we allow. Nowadays, it is not uncommon to see clusters running latest Kafka binaries with very old IBP versions (i.e. Kafka 3.5 with IBP 2.2). This could happen for various reasons: their upgrade documentation doesn't include the recommended IBP upgrade step, or they are simply afraid that this would break some of their client applications. In any case, this puts the user in an untested setup, which can trigger nasty bugs. I think the same misbehavior will happen with metadata.version, especially considering the Kafka 4.0 major release. What if we restrict the metadata.version downgrade to the supported Kafka versions? That way, we reduce the risks related to running new binaries with very old metadata records.

I think Strimzi (at the code level) does all they can do about this. We raise all kinds of warnings pushing the users to update the IBP and this proposal suggests the same for the metadata version.

I think you cannot expect users to understand how it works and what the consequences are if none of this is really documented in Kafka and without reading the code or all the KIPs you do not know what these things mean and do. Plus in the case of `metadata.version`, as we found out today, there is no real downgrade support. So if you actually know what it does, you would not want to upgrade it anyway because there is no real way back.

We could try to document it at least in our docs if Kafka is not able to do it. But I do not think we should impose any limitations of our own on the users.

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,213 @@
...
+Strimzi should follow the Apache Kafka [procedure for upgrading / downgrading](https://kafka.apache.org/documentation/#upgrade_350_kraft) Kafka.
+
+### KRaft upgrades according to Apache Kafka
+
+Apache Kafka [KRaft upgrade procedure](https://kafka.apache.org/documentation/#upgrade_350_kraft) consists of 2 different steps:
+1. Roll out the new Kafka version and verify everything is working fine.
+2. Update the `metadata.version` using the `kafka-reafutres.sh` command line tool (or using the Kafka Admin API).
+   Unlike updating the `inter.broker.protocol.version`, updating the `metadata.version` does not require a rolling update of all Kafka nodes.
+
+Downgrade is possibly as well, by:
```
tombentley, line 51:
```suggestion
Downgrade is possible as well:
```

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,213 @@
...
+1. Downgrading the `metadata.version`
+2. Rolling out the older Kafka version
+
+Changes to the metadata formats - that are defined by the `metadata.version` - might not be backwards compatible.
+As a result, the first step of the downgrade procedure - downgrading the `metadata.version` - can be done safely without any metadata loss only for selected metadata versions.
+For other versions, the downgrade will be possible only as _unsafe_ downgrade that might result in loosing some metadata information and might have negative impact on the cluster.
+
+#### Exiting downgrade limitations
```
tombentley, line 59:
```suggestion
#### Existing downgrade limitations
```

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,213 @@
...
+The operator will use this field to set the metadata version in the Kafka cluster.
+In the initial deployment of the Kafka cluster, it will be set using the `kafka-storage.sh` tool.
+In existing Kafka clusters, it will be queried and changed using the Kafka Admin API.
+
+The current metadata version will be tracked in the `.status` section of the Kafka CR in `.status.kafkaMetadataVersion`.
+This field will be used for validations and to track if upgrade / downgrade can be executed.
+
+##### Upgrading metadata version
+
+When upgrading the metadata version, the operator will validate the desired version and check if it is expected to work with the used Kafka version (e.g. is not higher than the Kafka version).
```
tombentley, line 96: I didn't quite follow this. Did you mean something like this?

```suggestion
When upgrading the metadata version, the operator will validate the desired version and check if it is expected to work with the used Kafka version (e.g. is not higher than `metadata.version` supported by the Kafka version).
```

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,213 @@
...
+* Unpause the reconciliation of the Kafka CR
+
+##### Configuring the initial metadata version
+
+When creating a new Kafka cluster, the initial metadata version can be configured using the `--release-version` option of the `kafka-storage.sh` script.
+To allow users to deploy Kafka clusters with older metadata version, Strimzi will also store the desired `metadata.version` in the per-broker configuration Config Map. 
+The value will be mounted into the Kafka pods and used when formatting the storage of a new Kafka cluster.
+That way - unlike when using environment variable - we will not need to roll the Pod every time the `metadata.version` changes (since in such cases we can update it dynamically without rolling the pods).
+
+#### Upgrades
```
tombentley, line 120: This section heading could be a bit more specific. I'm not sure if this is really much better...

```suggestion
#### Upgrading the Kafka container images
```

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,213 @@
...
+    * When the user changes `Kafka.spec.kafka.version` in parallel with the Strimzi Cluster Operator upgrade.
+* When user requests Kafka upgrade by changing `Kafka.spec.kafka.version`.
+
+In both cases, the Strimzi Cluster Operator will:
+1. Validate that the metadata version currently used is compatible with the new Kafka version (it will check that it is lower than the new Kafka version).
+   This check is expected to pass for upgrades since a Kafka cluster that would not pass it would be invalid already before the upgrade.
+2. Roll all Kafka pods to use the containers with the new Kafka version.
+
+Once all the Kafka pods are rolled, the next step will depend on whether `Kafka.spec.kafka.metadataVersion` is set or not.
+When it is not set, Strimzi will automatically update it to the default version corresponding to the new Kafka version.
```
tombentley, line 134: Does 'automatically update it' mean updating the ``Kafka.spec.kafka.metadataVersion`?

tombentley (commented):
Thanks @scholzj!

058-kraft-upgrades-and-downgrades.md
scholzj: No, this refers to updating the metadata version in the Kafka cluster itself. It will not update the Kafka CR, so `Kafka.spec.kafka.metadataVersion` will remain unset (and having it unset means the default metadata version for given Kafka version is used, so it will be in-sync).

058-kraft-upgrades-and-downgrades.md
scholzj: I made it a bit more specific. But this chapter really describes the whole process and how it should work. Not just the rolling update for the new containers.

058-kraft-upgrades-and-downgrades.md
scholzj: (I also clarified this in the proposal to make it clear there)

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,221 @@
+# KRaft upgrades and downgrades
+
+This proposal introduces the Kafka upgrade mechanism for KRaft clusters.
+It describes how will Strimzi upgrade or downgrade KRaft clusters from one Apache Kafka version to another.
```
PaulRMellor, line 4:
```suggestion
It describes how Strimzi will upgrade or downgrade KRaft clusters from one Apache Kafka version to another.
```

PaulRMellor (commented):
Clear proposal. I left a few suggestions, but also a couple of comments for sections I can revisit if changes are made

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,221 @@
...
+The upgrades of ZooKeeper-based clusters are multi-stage process:
+1. First, ZooKeeper nodes are rolled to use the new container image from the new Kafka version and possibly also the new ZooKeeper version (if the new Kafka version uses new ZooKeeper version).
+2. Next, Kafka brokers are rolled to use the new container image with the new Apache Kafka version.
+3. Finally, the Kafka brokers are rolled one more time to update the `inter.broker.protocol.version` (in case the `inter.broker.protocol.version` changed).
+   This step is done either when the user changes the `inter.broker.protocol.version` in `Kafka.spec.kafka.config`.
+   Or automatically in the next reconciliation when `inter.broker.protocol.version` is not set at all in `Kafka.spec.kafka.config`.
+
+Strimzi also supports downgrades of ZooKeeper-based Kafka clusters.
+The downgrades of ZooKeeper-based clusters are multi-stage process:
+1. If the `inter.broker.protocol.version` is set to a higher version then supported by the Kafka version being downgraded to, it has to be set to a version supported by the older Kafka version.
```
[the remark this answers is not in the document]
scholzj: I don't think I understand the differences in detail. But it is possible to do that with `inter.broker.protocol.version`.

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,221 @@
...
+
+When downgrading the `metadata.version` in Kafka, the operator will always try to do a _safe_ downgrade only.
+And if the safe downgrade fails, it will report an error (in log and in `.status` section of the Kafka CR) and continue reconciliation without changing the metadata version.
+
+In case the safe downgrade is not supported, an _unsafe_ downgrade can be attempted (as explained in one of the earlier sections, the _unsafe_ downgrade is not implemented in current Kafka versions, but it is planned to be supported in the future).
+Unsafe upgrade might result in metadata loss that might cause various problems to the Kafka cluster.
+Users who decide they want to do the unsafe downgrade can do so manually:
+* Pause the reconciliation of the Kafka CR
+* Update the Kafka CR to the updated metadata version (to avoid the operator upgrading it back immediately after the reconciliation is unpaused)
+* Use the Kafka Admin API or CLI tools to do the unsafe downgrade
```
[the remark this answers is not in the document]
Frawless: I think the `safe` part means, that the metadata version between versions used during downgrade are compatible. If not user can still do it, but it could break something, correct @scholzj ?

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,221 @@
...
+When it is not set, Strimzi will automatically update the metadata version in the Kafka cluster using Kafka Admin API to the default version corresponding to the new Kafka version.
+The operator will not change the `Kafka.spec.kafka.metadataVersion` field - it will remain unset.
+And the Kafka upgrade will be complete with this.
+
+In case the `Kafka.spec.kafka.metadataVersion` field is set, it will just check the Kafka cluster has the desired metadata version as in any other reconciliation.
+The user will be expected to verify the Kafka cluster and change the `Kafka.spec.kafka.metadataVersion` when everything seems to work fine.
+Only after changing the metadata version, the upgrade will be considered complete.
+Similarly to today's ZooKeeper-based implementation, a warning will be issued (in log and in Kafka CR `.status` section) when the metadata version does not correspond to the Kafka version and tell the user to complete the upgrade by updating the metadata version.
+
+The upgrades will be supported to skip multiple Strimzi and Kafka versions in single step.
```
Frawless, line 143: It means that Kafka will support metadata upgrade to skip multiple versions as well?

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,221 @@
...
+#### Risks
+
+Kafka provides APIs to manage the `metadata.version`.
+Strimzi has no way to block the users from using these APIs.
+If the users decide to manipulate the `matadata.version` them self, this can lead to unexpected issues.
+This risk is deemed acceptable, because we expect that:
+* Users would prefer to have the `metadata.version` managed through Strimzi.
+* Users would have no reason to manipulate the `metadata.version` on their own.
+
+The risk can be also at least partially mitigated through documentation.
```
Frawless, line 203: Would it make sense to raise a warning in Kafka CR when Strimzi detects manual change of `metadata.version` ?

058-kraft-upgrades-and-downgrades.md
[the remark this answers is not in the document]
scholzj: Yes. The safe downgrade will be done by the operator as it should guarantee that the stuff will keep working. The unsafe downgrade has to be done by the users and might cause issues and metadata loss.

058-kraft-upgrades-and-downgrades.md
scholzj: Yes, that is correct.

058-kraft-upgrades-and-downgrades.md
scholzj: I'm not sure we can track it reliably. So I do not think we should have any special handling for it.

058-kraft-upgrades-and-downgrades.md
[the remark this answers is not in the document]
PaulRMellor: Okay. Got it.

PaulRMellor (approved):

Frawless (approved):

see-quick (approved):
LGTM

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Proposal
+
+Strimzi should follow the Apache Kafka [procedure for upgrading / downgrading](https://kafka.apache.org/documentation/#upgrade_350_kraft) Kafka.
+
+### KRaft upgrades according to Apache Kafka
+
+The Apache Kafka [KRaft upgrade procedure](https://kafka.apache.org/documentation/#upgrade_350_kraft) consists of 2 different steps:
+1. Roll out the new Kafka version and verify everything is working fine.
+2. Update the `metadata.version` using the `kafka-reafutres.sh` command line tool (or using the Kafka Admin API).
```
tinaselenge, line 48:
```suggestion
2. Update the `metadata.version` using the `kafka-features.sh` command line tool (or using the Kafka Admin API).
```

fvaleri (approved):
LGTM now. Thanks.

tombentley (approved):
I left a few nits about clarity in a few places. But I think the process being described here is reasonable. Thanks!

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,221 @@
...
+Users who decide they want to do the unsafe downgrade can do so manually:
+* Pause the reconciliation of the Kafka CR
+* Update the Kafka CR to the updated metadata version (to avoid the operator upgrading it back immediately after the reconciliation is unpaused)
+* Use the Kafka Admin API or CLI tools to do the unsafe downgrade
+* Unpause the reconciliation of the Kafka CR
+
+##### Configuring the initial metadata version
+
+When creating a new Kafka cluster, the initial metadata version can be configured using the `--release-version` option of the `kafka-storage.sh` script.
+To allow users to deploy Kafka clusters with an older metadata version, Strimzi will also store the desired `metadata.version` in the per-broker configuration Config Map.
```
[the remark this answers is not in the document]
scholzj: Yes, users will specify the metadata version only in `Kafka.spec.kafka.metadataVersion`. This section mostly describes how it will be handled internally, because unlike the `inter.broker.protocol.version`, the way Kafka hnaldes the metadata version requires special implementation. I have rewritten the section to make it more clear what is done by user and what does the operator do internally.

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,221 @@
...
+
+Users can use this new field to configure the metadata version.
+The default value of this field when not set will be the metadata version corresponding to the current Kafka version.
+The default version corresponding to a given Kafka version will be stored in `kafka-versions.yaml` in the same way we today store the default `inter.broker.protocol.version` and `log.message.format.version`.
+
+The operator will use this field to set the metadata version in the Kafka cluster.
+In the initial deployment of the Kafka cluster, it will be set using the `kafka-storage.sh` tool.
+In existing Kafka clusters, it will be queried and changed using the Kafka Admin API.
+
+The current metadata version will be tracked in the `.status` section of the Kafka CR in `.status.kafkaMetadataVersion`.
```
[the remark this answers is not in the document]
fvaleri: Good point. I already see new features coming in from recent KIPs.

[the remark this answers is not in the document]
scholzj: Kafka does not make it clear what the intentions are. There is no clear documentation of how the features should be used or even a list of them beyond the `metadata.version`. It is also hard to predict what impact might the other features have for upgrades / downgrades and how will they integrate with the `metadata.version`. 

This proposal covers only the `metadata.version` as currently the only feature relevant to upgrades / downgrades and managed by Strimzi. It also designs the `.spec` and `.status` sections accordingly. There is currently no reason to use some complicated maps for them instead of a simple field.

The expectation is, that after ZooKeeper support is dropped, we will migrate to a new API version that will give us the chance to re-evaluate this decision if needed.

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,228 @@
...
```
ppatierno, line 93: We are going to have `status.metadataState` for the ZooKeeper to KRaft migration FSM.
I was wondering if ... we should have `status.metadataVersion` here or the other way around changing on the migration side by having `status.kafkaMetadataState`. I have no strong opinion about it tbh.
Of course, these names make the two status fields related somehow. If I was a user I could think "my metadata state is .... " and "its version is". I think that's correct, because when metadata state is ZooKeeper I don't expect to be a metadata version there. When the cluster is KRaft both fields will be set. What about during a migration? I guess the metadata version should be set at least at the end of the migration process or even before during the dual write? Maybe the latter, because in dual write we are effectively using KRaft as metadata store (while copying to ZooKeeper as well).

058-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,228 @@
...
+To allow users to configure it in our Kafka custom resource, a new String-type field named `metadataVersion` will be added to `Kafka.spec.kafka`.
+This field will be validated by the operator to contain a valid Kafka metadata version.
+
+Users can use this new field to configure the metadata version.
+The default value of this field when not set will be the metadata version corresponding to the current Kafka version.
+The default version corresponding to a given Kafka version will be stored in `kafka-versions.yaml` in the same way we today store the default `inter.broker.protocol.version` and `log.message.format.version`.
+
+The operator will use this field to set the metadata version in the Kafka cluster.
+In the initial deployment of the Kafka cluster, it will be set using the `kafka-storage.sh` tool.
+In existing Kafka clusters, it will be queried and changed using the Kafka Admin API.
```
ppatierno, line 91: Out of curiosity, why two different ways are needed? I mean tool vs Admin API if cluster is new or existing?

058-kraft-upgrades-and-downgrades.md
scholzj: I opted for `kafkaMetadataVersion` because we already have `kafkaVersion`. I'm fairly flexible. I can rename it `metadataVersion` if you prefer. But we can't rename `kafkaVersion` anymore, so that is set.

058-kraft-upgrades-and-downgrades.md
scholzj: It is a bit more explained in the `Configuring the initial metadata version` section. Essentially, If you do not set the metadata version first with `kafka-storage.sh`, the cluster will start with the latest metadata version and you will have to downgrade it afterwards (and that is not supported today anyway, so you would be essentially stuck). Does that explain? I can add a mention of the section covering this later if it helps.

058-kraft-upgrades-and-downgrades.md
ppatierno: Well it comes to try thinking if "metadata" could be related to something else and not just "Kafka metadata".
I would say, let me change my `status.metadataState` to `status.kafkaMetadataState` so we are also aligned better with the `kafkaVersion` field.

What about my last question on setting the metadata version during migration?

058-kraft-upgrades-and-downgrades.md
ppatierno: I read that but then I missed during reading these lines. Yeah adding a link could be better.

058-kraft-upgrades-and-downgrades.md
scholzj: > What about during a migration? I guess the metadata version should be set at least at the end of the migration process or even before during the dual write? Maybe the latter, because in dual write we are effectively using KRaft as metadata store (while copying to ZooKeeper as well).

In the status, it should be IMHO set from the point when the controllers are enabled. For the documentation, we should also consider suggesting the users to set the `Kafka.spec.kafka.metadataVersion` field explicitely in the Kafka CR.

But I do not expect any major impact on the code. Depending on which Pr is merged first, the other would need to adjust the `isKraftEnabled` boolean flag versus the migration state. But apart from that, you should not need to write anything special in the migration code to set the metadata version or anything like that.

scholzj: I'm not sure if you want to include this in the proposal or not. I think it is a good point, b ut I would probably prefer to not include the details in it since they are a bit unclear and depend on the exact code (that is I guess still WiP on your side and does not exist for this proposal).

ppatierno (approved):
LGTM

scholzj:
This has now 5 binding +1 votes and 2 non-binding +1. If you have any comments or need more time, please let me know till tomorrow EOB. Otherwise, I will merge it as approved.

@tinaselenge You seem to be the only one who commented but did not approve. Do you have something more to this? Or is it ok?

061-kraft-upgrades-and-downgrades.md
```diff
@@ -0,0 +1,229 @@
...
+If it passes the validation, the Kafka Admin API will be used to upgrade the metadata.
+If the validation fails, the existing metadata version will be used and an error will be raised in the log and in the `.status` section of the Kafka CR.
+
+##### Downgrading metadata version
+
+When downgrading the `metadata.version` in Kafka, the operator will always try to do a _safe_ downgrade only.
+And if the safe downgrade fails, it will report an error (in the log and in the `.status` section of the Kafka CR) and continue reconciliation without changing the metadata version.
+
+In case the safe downgrade is not supported, an _unsafe_ downgrade can be attempted (as explained in one of the earlier sections, the _unsafe_ downgrade is not implemented in current Kafka versions, but it is planned to be supported in the future).
+Unsafe upgrade might result in metadata loss that might cause various problems to the Kafka cluster.
```
im-konge, line 108: Maybe I miss-read it, but should it be `Unsafe downgrade` ?

im-konge (approved):
Sorry for the late review, LGTM

scholzj:
Approved with 6 binding +1 votes and 2 non-binding +1 votes.

PR #97, opened 2023-10-27
Add API for Quotas Management

scholzj:
Thi PR opens a proposal for adding a new API for Quotas management to the `Kafka` custom resource.

### DIFF
--- a/062-quotas-management.md
+++ b/062-quotas-management.md
@@ -139,6 +139,7 @@ Based on this configuration, the operator will:
 The quota plugin configuration will not be set a forbidden configuration.
 When the new `.spec.kafka.quotas` API would be not set at all, Strimzi will keep the default Kafka configuration (which is the default built-in quotas plugin).
 Or - in case the user customized the quota plugin configuration in the `.spec.kafka.config` section - Strimzi will pass this configuration to Kafka.
+That means that the Apache Kafka `client.quota.callback.class` will not be marked as forbidden option.
 
 That helps to ensure backwards compatibility.
 Any users who current use the default plugin or any custom plugin would not be required to start using the `.spec.kafka.quotas` section immediately.
@@ -155,7 +156,7 @@ There is no change to the Strimzi User Operator or the `KafkaUser` custom resour
 ## Backwards compatibility
 
 This proposal is fully backwards compatible.
-When the new `.spec.kafka.quotas` APi is not set, the operator handles the quotas configuration exactly as in previous Strimzi versions.
+When the new `.spec.kafka.quotas` API is not set, the operator handles the quotas configuration exactly as in previous Strimzi versions.
 
 ## Rejected alternatives
 
--- a/062-quotas-management.md
+++ b/062-quotas-management.md
@@ -47,7 +47,7 @@ The absence of a dedicated API for quota management in the `Kafka` CR currently 
    While it can be used by users, it does not follow Strimzi's basic idea of declarative management.
    Having a dedicated API for quota configuration as part of the `Kafka` CR would make it much easier to configure the default quotas.
 2. The initial releases of the Strimzi Quota Plugin were simple, preventing users from producing messages directly to a broker running out of disk space. 
-   However, these versions did not address the issue of replicated messages produced on other brokers in the Kafka cluster, which could still fill up the disk space of the destination broker."
+   However, these versions did not address the issue of replicated messages produced on other brokers in the Kafka cluster, which could still fill up the disk space of the destination broker.
    They also didn't support many useful features, such as JBOD storage.
    Because the initial releases were so simple, it was possible to configure them in the `.spec.kafka.config` section of the `Kafka` CR.
    The next - not yet released - version of the Strimzi Quota plugin brings many improvements.
--- a/062-quotas-management.md
+++ b/062-quotas-management.md
@@ -1,0 +1,163 @@
+# Quotas management
+
+This proposal introduces a new mechanism for managing the Apache Kafka quota plugins.
+It supports both the Apache Kafka's built-in Quota plugin as well as the new (not yet released) version of the [Strimzi Quota plugin](https://github.com/strimzi/kafka-quotas-plugin).
+
+## Current situation
+
+### Quotas in Apache Kafka
+
+Apache Kafka supports a pluggable quota mechanism that allows to limit the load Kafka users and/or clients can put on the Kafka brokers.
+
+Apache Kafka provides a built-in plugin that is enabled by default and can apply quotas to:
+* Maximum bytes per-second that each client group can publish to a broker
+* Maximum bytes per-second that each client group can fetch from a broker
+* Maximum CPU utilization as a percentage of network and I/O threads
+* Rate at which mutations are accepted for the create topics request, the create partitions request and the delete topics request
+
+These quotas can be set at the user or client level and are useful to:
+* Prevent a single client from overloading the broker
+* Prevent noisy neighbor situations, where one very busy user or client negatively impacts others
+* Ensure fairness in allocating resources to different users and clients
+
+The built-in quota plugin is enabled when `client.quota.callback.class` is set to `null` (i.e. not set) in Apache KAfka configuration.
+
+Instead of using the default plugin, users can also write and use their own quota plugins by implementing the `ClientQuotaCallback` Java interface and configuring it with the `client.quota.callback.class` in Kafka configuration.
+The custom implementations can have different goals.
+For example the [Strimzi Quota plugin](https://github.com/strimzi/kafka-quotas-plugin) focus on preventing the Kafka brokers from running out of disk space instead of limiting the individual clients or users.
+
+### Quotas in Strimzi
+
+Currently, Strimzi does not have any API support for quotas in the `Kafka` custom resource.
+All quotas are configured directly by the users in `.spec.kafka.config` section of the `Kafka` custom resource.
+That gives users a lot of freedom and flexibility to configure the quotas.
+But it also does not provide them any support when the quota plugin requires some more complicated configuration.
+
+The only place where Strimzi directly supports configuring quotas is in the `KafkaUser` resource.
+It can be used to configure a specific quota for a given user using Kafka's built-in quota plugin.
+
+## Motivation
+
+The absence of an dedicated API for quota management in the `Kafka` CR currently creates two major limitations.
+1. The built-in Apache Kafka quota plugin has an ability to configure _default quotas_.
+   Default quotas apply to any user that has no specific quota set.
+   However, the default quotas cannot be configured declaratively in the `.spec.kafka.config` section.
+   And they cannot be configured in a `KafkaUser` resource either because they do not belong to any particular user.
+   The only way to configure them right now is imperatively use the Kafka Admin API.
+   While it can be used by users, it does not follow Strimzi's basic idea of declarative management.
+   Having a dedicated API for quota configuration as part of the `Kafka` CR would make it much easier to configure the default quotas.
+2. The initial releases of the Strimzi Quotas Plugin were very simple.
+   It helps to prevent users producing directly to a broker that was running out of disk space.
+   But it fails to prevent replication of messages produced to other brokers in the Kafka cluster from filling in the disk.
+   It also doesn't support many additional features such as JBOD storage.
+   Because the initial releases were very simple, they had also very simple configuration.
+   And it was possible to configure them in the `.spec.kafka.config` section of the `Kafka` CR.
+   The next - not yet released - version of the Strimzi Quotas plugin brings many improvements.
+   More details about how the new version works can be found in the [Cluster Wide Volume Usage Quota Management](./047-cluster-wide-volume-usage-quota-management.md) proposal.
+   It operates at the level of a whole Kafka cluster and it has improved ability to prevent replicated messages to fill in the broker storage.
+   It also supports JBOD storage.
+   But it has also much more complicated configuration.
+   The plugin instances running in the different Kafka brokers communicate with each other using a Kafka topic.
+   And need therefore to connect to the Kafka brokers as clients as well.
+   Having the user configure this directly in the `.spec.kafka.config` section would be much more complicated as the user would need to configure the keystores and truststores, their passswords, bootstrap address etc.
+   Having a dedicated API for quota configuration would make it much easier to have the operator configure the client automatically.
+
+## Proposal
+
+This proposal suggests to add a new `quotas` API to the `Kafka` custom resource.
+It will be inside the `.spec.kafka` section.
+And it will allow configuring different types of quotas.
+As many existing Strimzi APIs, it will use the `type` field to differentiate the quota plugin that will be used.
+
+This proposal suggests to initially add support for two Quota plugins.
+1. The build-in Kafka plugin
+2. The Strimzi Quotas plugin
+
+The first supported plugin - the Kafka's built-in plugin - will be configured as `type: kafka`.
+The API will allow to configure the default user quotas for this plugin in the same format as the `KafkaUser` API allows it today:
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    quotas:
+      type: kafka
+      producerByteRate: 1048576
+      consumerByteRate: 2097152
+      requestPercentage: 55
+      controllerMutationRate: 50
+    # ... 
+  # ...
+```
+
+Based on this configuration, the operator will:
+* Configure the default quota plugin (make sure the `client.quota.callback.class` option is not set)
+* Set the default user quotas
+* When the `type: kafka` quotas APi will be set, but one or more of fields with the individual quotas will be missing, the operator will make sure that given default quota is not set.
+* The other configuration options for the built-in Kafka plugin can be configured by the user in the `.spec.kafka.config` section in the `Kafka` CR.
+
+The second supported plugin - the Strimzi Quotas plugin - will be configured as `type: strimzi`.
+The API will allow to configure the storage limits at which the plugin will start limiting the producers in order to prevent the disk from getting full.
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    quotas:
+      type: strimzi
+      produceQuota: 1000000
+      fetchQuota: 1000000
+      minAvailableBytesPerVolume: 5368709120
+      excludedPrincipals:
+        - my-special-user
+    # ... 
+  # ...
+```
+
+The `type: strimzi` quotas API will support the following fields:
+* `produceQuota` for setting the default produce quota
+* `fetchQuota` for setting the default fetch quota
+* `minAvailableBytesPerVolume` for setting the minimal required free space per volume as number of bytes (this field can be configured only when `minAvailableRatioPerVolume`is not set)
+* `minAvailableRatioPerVolume` for setting the minimal required free space per volume as percentage of the overall volume size (this field can be configured only when `minAvailableBytesPerVolume`is not set)
+* `excludedPrincipals` for setting s list of users to be excluded from the applied quotas
+
+Other options might be configured directly by the user in `.spec.kafka.config` section of the `Kafka` CR.
+
+Based on this configuration, the operator will:
+* Set the `client.quota.callback.class` field in Kafka configuration to `io.strimzi.kafka.quotas.StaticQuotaCallback`
+* Set the configuration on how to connect to the Kafka broker as a client using the `client.quota.callback.static.kafka.admin.` options
+* Configures the default produce and fetch quota (`client.quota.callback.static.produce` and `client.quota.callback.static.fetch`)
+* Configure the minimal available disk space per-volume using the `client.quota.callback.static.storage.per.volume.limit.min.available.bytes` and `client.quota.callback.static.storage.per.volume.limit.min.available.ratio` options
+* Set the excluded users for that the quotas will not be applied using the `client.quota.callback.static.excluded.principal.name.list` option
+
+The quota plugin configuration will not be set a forbidden configuration.
+When the new `.spec.kafka.quotas` API would be not set at all, Strimzi will keep the default Kafka configuration (which is the default built-in quotas plugin).
+Or - in case the user customized the quota plugin configuration in the `.spec.kafka.config` section - Strimzi will pass this configuration to Kafka.
+
+That helps to ensure backwards compatibility.
+Any users who current use the default plugin or any custom plugin would not be required to start using the `.spec.kafka.quotas` section immediately.
+They will be able to keep the configuration as is.
+
+When the `.spec.kafka.quotas` API is used and the `.spec.kafka.config` section contains some custom quota plugin configuration, Strimzi will configure the quotas according to the `.spec.kafka.quotas` API.
+But it will set a warning condition and log a warning about this conflict.
+
+## Affected projects
+
+This proposal affects only the Strimzi Cluster Operator and the `Kafka` custom resource.
+There is no change to the Strimzi User Operator or the `KafkaUser` custom resource API or any other projects.
+
+## Backwards compatibility
+
+This proposal is fully backwards compatible.
+When the new `.spec.kafka.quotas` APi is not set, the operator handles the quotas configuration exactly as in previous Strimzi versions.
+
+## Rejected alternatives
+
+Currently, there are no rejected alternatives.
+
--- a/062-quotas-management.md
+++ b/062-quotas-management.md
@@ -7,7 +7,7 @@ It supports both the Apache Kafka's built-in Quota plugin as well as the new (no
 
 ### Quotas in Apache Kafka
 
-Apache Kafka supports a pluggable quota mechanism that allows to limit the load Kafka users and/or clients can put on the Kafka brokers.
+Apache Kafka supports a pluggable quota mechanism that allows limiting the load that Kafka users and/or clients can place on the Kafka brokers.
 
 Apache Kafka provides a built-in plugin that is enabled by default and can apply quotas to:
 * Maximum bytes per-second that each client group can publish to a broker
@@ -20,25 +20,25 @@ These quotas can be set at the user or client level and are useful to:
 * Prevent noisy neighbor situations, where one very busy user or client negatively impacts others
 * Ensure fairness in allocating resources to different users and clients
 
-The built-in quota plugin is enabled when `client.quota.callback.class` is set to `null` (i.e. not set) in Apache KAfka configuration.
+The built-in quota plugin is enabled when `client.quota.callback.class` is set to `null` (i.e. not set) in Apache Kafka configuration.
 
 Instead of using the default plugin, users can also write and use their own quota plugins by implementing the `ClientQuotaCallback` Java interface and configuring it with the `client.quota.callback.class` in Kafka configuration.
 The custom implementations can have different goals.
-For example the [Strimzi Quota plugin](https://github.com/strimzi/kafka-quotas-plugin) focus on preventing the Kafka brokers from running out of disk space instead of limiting the individual clients or users.
+For example, the [Strimzi Quota plugin](https://github.com/strimzi/kafka-quotas-plugin) focuses on preventing the Kafka brokers from running out of disk space instead of limiting the individual clients or users.
 
 ### Quotas in Strimzi
 
 Currently, Strimzi does not have any API support for quotas in the `Kafka` custom resource.
 All quotas are configured directly by the users in `.spec.kafka.config` section of the `Kafka` custom resource.
 That gives users a lot of freedom and flexibility to configure the quotas.
-But it also does not provide them any support when the quota plugin requires some more complicated configuration.
+But it also does not provide them any support when the quota plugin requires a more complicated configuration.
 
 The only place where Strimzi directly supports configuring quotas is in the `KafkaUser` resource.
 It can be used to configure a specific quota for a given user using Kafka's built-in quota plugin.
 
 ## Motivation
 
-The absence of an dedicated API for quota management in the `Kafka` CR currently creates two major limitations.
+The absence of a dedicated API for quota management in the `Kafka` CR currently creates two major limitations.
 1. The built-in Apache Kafka quota plugin has an ability to configure _default quotas_.
    Default quotas apply to any user that has no specific quota set.
    However, the default quotas cannot be configured declaratively in the `.spec.kafka.config` section.
@@ -57,23 +57,23 @@ The absence of an dedicated API for quota management in the `Kafka` CR currently
    It operates at the level of a whole Kafka cluster and it has improved ability to prevent replicated messages to fill in the broker storage.
    It also supports JBOD storage.
    But it has also much more complicated configuration.
-   The plugin instances running in the different Kafka brokers communicate with each other using a Kafka topic.
-   And need therefore to connect to the Kafka brokers as clients as well.
-   Having the user configure this directly in the `.spec.kafka.config` section would be much more complicated as the user would need to configure the keystores and truststores, their passswords, bootstrap address etc.
+   The plugin is using the Kafka Admin API to gather the information about the disk capacity from the Kafka brokers.
+   And needs therefore to connect to the Kafka cluster as a client.
+   Having the user configure this directly in the `.spec.kafka.config` section would be much more complicated, as the user would need to configure the keystores, truststores, their passwords, bootstrap address, etc.
    Having a dedicated API for quota configuration would make it much easier to have the operator configure the client automatically.
 
 ## Proposal
 
 This proposal suggests to add a new `quotas` API to the `Kafka` custom resource.
 It will be inside the `.spec.kafka` section.
-And it will allow configuring different types of quotas.
+And it will allow users to configure different types of quotas.
 As many existing Strimzi APIs, it will use the `type` field to differentiate the quota plugin that will be used.
 
 This proposal suggests to initially add support for two Quota plugins.
-1. The build-in Kafka plugin
+1. The built-in Kafka plugin
 2. The Strimzi Quotas plugin
 
-The first supported plugin - the Kafka's built-in plugin - will be configured as `type: kafka`.
+The first supported plugin - Kafka's built-in plugin - will be configured as `type: kafka`.
 The API will allow to configure the default user quotas for this plugin in the same format as the `KafkaUser` API allows it today:
 
 ```yaml
@@ -96,7 +96,7 @@ spec:
 Based on this configuration, the operator will:
 * Configure the default quota plugin (make sure the `client.quota.callback.class` option is not set)
 * Set the default user quotas
-* When the `type: kafka` quotas APi will be set, but one or more of fields with the individual quotas will be missing, the operator will make sure that given default quota is not set.
+* When the `type: kafka` quotas API will be set, but one or more of the fields with the individual quotas are missing, the operator will make sure that the given default quota is not set (set to null).
 * The other configuration options for the built-in Kafka plugin can be configured by the user in the `.spec.kafka.config` section in the `Kafka` CR.
 
 The second supported plugin - the Strimzi Quotas plugin - will be configured as `type: strimzi`.
@@ -125,7 +125,7 @@ The `type: strimzi` quotas API will support the following fields:
 * `fetchQuota` for setting the default fetch quota
 * `minAvailableBytesPerVolume` for setting the minimal required free space per volume as number of bytes (this field can be configured only when `minAvailableRatioPerVolume`is not set)
 * `minAvailableRatioPerVolume` for setting the minimal required free space per volume as percentage of the overall volume size (this field can be configured only when `minAvailableBytesPerVolume`is not set)
-* `excludedPrincipals` for setting s list of users to be excluded from the applied quotas
+* `excludedPrincipals` for setting a list of users to be excluded from the applied quotas
 
 Other options might be configured directly by the user in `.spec.kafka.config` section of the `Kafka` CR.
 
@@ -145,7 +145,7 @@ Any users who current use the default plugin or any custom plugin would not be r
 They will be able to keep the configuration as is.
 
 When the `.spec.kafka.quotas` API is used and the `.spec.kafka.config` section contains some custom quota plugin configuration, Strimzi will configure the quotas according to the `.spec.kafka.quotas` API.
-But it will set a warning condition and log a warning about this conflict.
+It will also set a warning condition and log a message about this conflict.
 
 ## Affected projects
 
--- a/062-quotas-management.md
+++ b/062-quotas-management.md
@@ -9,11 +9,11 @@ It supports both the Apache Kafka's built-in Quota plugin as well as the new (no
 
 Apache Kafka supports a pluggable quota mechanism that allows limiting the load that Kafka users and/or clients can place on the Kafka brokers.
 
-Apache Kafka provides a built-in plugin that is enabled by default and can apply quotas to:
+Apache Kafka provides a built-in plugin that is enabled by default and can apply the following quotas:
 * Maximum bytes per-second that each client group can publish to a broker
 * Maximum bytes per-second that each client group can fetch from a broker
 * Maximum CPU utilization as a percentage of network and I/O threads
-* Rate at which mutations are accepted for the create topics request, the create partitions request and the delete topics request
+* Rate per-second at which concurrent mutations are accepted from requests to create topics, create partitions, and delete topics.
 
 These quotas can be set at the user or client level and are useful to:
 * Prevent a single client from overloading the broker
@@ -34,7 +34,7 @@ That gives users a lot of freedom and flexibility to configure the quotas.
 But it also does not provide them any support when the quota plugin requires a more complicated configuration.
 
 The only place where Strimzi directly supports configuring quotas is in the `KafkaUser` resource.
-It can be used to configure a specific quota for a given user using Kafka's built-in quota plugin.
+It can be used to configure specific quotas for a given user using Kafka's built-in quota plugin.
 
 ## Motivation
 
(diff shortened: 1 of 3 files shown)


scholzj:
FYI: @robobario @SamBarker

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+* Maximum bytes per-second that each client group can fetch from a broker
+* Maximum CPU utilization as a percentage of network and I/O threads
+* Rate at which mutations are accepted for the create topics request, the create partitions request and the delete topics request
+
+These quotas can be set at the user or client level and are useful to:
+* Prevent a single client from overloading the broker
+* Prevent noisy neighbor situations, where one very busy user or client negatively impacts others
+* Ensure fairness in allocating resources to different users and clients
+
+The built-in quota plugin is enabled when `client.quota.callback.class` is set to `null` (i.e. not set) in Apache KAfka configuration.
```
tombentley, line 23:
```suggestion
The built-in quota plugin is enabled when `client.quota.callback.class` is set to `null` (i.e. not set) in Apache Kafka configuration.
```

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+  # ...
+```
+
+Based on this configuration, the operator will:
+* Configure the default quota plugin (make sure the `client.quota.callback.class` option is not set)
+* Set the default user quotas
+* When the `type: kafka` quotas APi will be set, but one or more of fields with the individual quotas will be missing, the operator will make sure that given default quota is not set.
+* The other configuration options for the built-in Kafka plugin can be configured by the user in the `.spec.kafka.config` section in the `Kafka` CR.
+
+The second supported plugin - the Strimzi Quotas plugin - will be configured as `type: strimzi`.
```
tombentley, line 102: I don't think `strimzi` is the right name. If we wrote a new kind of quota plugin in the future what would we call it?

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    quotas:
+      type: strimzi
+      produceQuota: 1000000
+      fetchQuota: 1000000
```
tombentley, line 115: What are the units of these?

tombentley (commented):
Thanks @scholzj!

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+   It also doesn't support many additional features such as JBOD storage.
+   Because the initial releases were very simple, they had also very simple configuration.
+   And it was possible to configure them in the `.spec.kafka.config` section of the `Kafka` CR.
+   The next - not yet released - version of the Strimzi Quotas plugin brings many improvements.
+   More details about how the new version works can be found in the [Cluster Wide Volume Usage Quota Management](./047-cluster-wide-volume-usage-quota-management.md) proposal.
+   It operates at the level of a whole Kafka cluster and it has improved ability to prevent replicated messages to fill in the broker storage.
+   It also supports JBOD storage.
+   But it has also much more complicated configuration.
+   The plugin instances running in the different Kafka brokers communicate with each other using a Kafka topic.
+   And need therefore to connect to the Kafka brokers as clients as well.
```
SamBarker, line 61: The new version of the quota plug-in doesn't store any state in Kafka, in fact there is no inter node co-ordination. It does however still connect to the admin API to call `describeCluster` and `describeLogDirs`. 

That doesn't really change the motivation/goals, but is probably relevant for context.

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Proposal
+
+This proposal suggests to add a new `quotas` API to the `Kafka` custom resource.
+It will be inside the `.spec.kafka` section.
+And it will allow configuring different types of quotas.
+As many existing Strimzi APIs, it will use the `type` field to differentiate the quota plugin that will be used.
+
+This proposal suggests to initially add support for two Quota plugins.
+1. The build-in Kafka plugin
```
SamBarker, line 73: nit typo: 
```suggestion
1. The built-in Kafka plugin
```

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+      requestPercentage: 55
+      controllerMutationRate: 50
+    # ... 
+  # ...
+```
+
+Based on this configuration, the operator will:
+* Configure the default quota plugin (make sure the `client.quota.callback.class` option is not set)
+* Set the default user quotas
+* When the `type: kafka` quotas APi will be set, but one or more of fields with the individual quotas will be missing, the operator will make sure that given default quota is not set.
```
SamBarker, line 99: By this do you mean the operator will explicitly set empty values for the missing quota types? (and thus the broker inherits from the default configuration) Or jus that it will not set them to some default value?

062-quotas-management.md
SamBarker, line 102: Hmm does the project really want to commit to the proposed quota plug-in being type strimzi? Or would it be better to describe it as `ClusterWideLimits` or `ClusterWideVolumeLimits`?

To me using the proposed implementation as `strimzi` (which I admit seems like the natural name) implicitly binds `type: strimzi` to behaviour of outlined in proposal 47 and thus makes it difficult in practice to evolve that behaviour.

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+    # ... 
+  # ...
+```
+
+The `type: strimzi` quotas API will support the following fields:
+* `produceQuota` for setting the default produce quota
+* `fetchQuota` for setting the default fetch quota
+* `minAvailableBytesPerVolume` for setting the minimal required free space per volume as number of bytes (this field can be configured only when `minAvailableRatioPerVolume`is not set)
+* `minAvailableRatioPerVolume` for setting the minimal required free space per volume as percentage of the overall volume size (this field can be configured only when `minAvailableBytesPerVolume`is not set)
+* `excludedPrincipals` for setting s list of users to be excluded from the applied quotas
```
SamBarker, line 128: Is this an exclusive list or does Strimzi implicitly know of other super user type user which should not be blocked and can thus add them automagically?

nit typo: 
```suggestion
* `excludedPrincipals` for setting a list of users to be excluded from the applied quotas
```

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+
+Other options might be configured directly by the user in `.spec.kafka.config` section of the `Kafka` CR.
+
+Based on this configuration, the operator will:
+* Set the `client.quota.callback.class` field in Kafka configuration to `io.strimzi.kafka.quotas.StaticQuotaCallback`
```
SamBarker, line 133: It seem inherent in the proposal but I think it would help to state it explicitly. Defining `spec.kafka.quotas` implicitly overrides any value specified for `client.quota.callback.class` in `kafka.config`

062-quotas-management.md
SamBarker, line 128: Obviously this is equivalent to what the proposed version of the quota plug-in offers, but shouldn't it be driven by the plug-in rather than explicitly by this (or some other subsequent proposal)? Otherwise how would a user pass their own custom configuration through to their chosen plug-in?

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+* Set the configuration on how to connect to the Kafka broker as a client using the `client.quota.callback.static.kafka.admin.` options
+* Configures the default produce and fetch quota (`client.quota.callback.static.produce` and `client.quota.callback.static.fetch`)
+* Configure the minimal available disk space per-volume using the `client.quota.callback.static.storage.per.volume.limit.min.available.bytes` and `client.quota.callback.static.storage.per.volume.limit.min.available.ratio` options
+* Set the excluded users for that the quotas will not be applied using the `client.quota.callback.static.excluded.principal.name.list` option
+
+The quota plugin configuration will not be set a forbidden configuration.
```
SamBarker, line 139: How does the operator know what the rules governing what is acceptable configuration really are? To me that requires an interaction with the plug-in it self which I don't see how the operator can do.

SamBarker (commented):
I think the proposal makes sense.

My two concerns: 

1. are binding proposal 47 to the type name `strimzi`
2. How to express custom configuration through the quota API.
2.1. Arbitrary property keys
2.2. Custom validation rules.

I wonder how much effort the cluster operator should go to in validating the custom configuration? Given the broker will fail to start with a fairly explicit error message when validating the plug-in configuraiton.

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
+# Quotas management
+
+This proposal introduces a new mechanism for managing the Apache Kafka quota plugins.
+It supports both the Apache Kafka's built-in Quota plugin as well as the new (not yet released) version of the [Strimzi Quota plugin](https://github.com/strimzi/kafka-quotas-plugin).
+
+## Current situation
+
+### Quotas in Apache Kafka
+
+Apache Kafka supports a pluggable quota mechanism that allows to limit the load Kafka users and/or clients can put on the Kafka brokers.
```
see-quick, line 10:
```suggestion
Apache Kafka supports a pluggable quota mechanism that allows limiting the load that Kafka users and/or clients can place on the Kafka brokers.
```

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+These quotas can be set at the user or client level and are useful to:
+* Prevent a single client from overloading the broker
+* Prevent noisy neighbor situations, where one very busy user or client negatively impacts others
+* Ensure fairness in allocating resources to different users and clients
+
+The built-in quota plugin is enabled when `client.quota.callback.class` is set to `null` (i.e. not set) in Apache KAfka configuration.
+
+Instead of using the default plugin, users can also write and use their own quota plugins by implementing the `ClientQuotaCallback` Java interface and configuring it with the `client.quota.callback.class` in Kafka configuration.
+The custom implementations can have different goals.
+For example the [Strimzi Quota plugin](https://github.com/strimzi/kafka-quotas-plugin) focus on preventing the Kafka brokers from running out of disk space instead of limiting the individual clients or users.
```
see-quick, line 27:
```suggestion
For example, the [Strimzi Quota plugin](https://github.com/strimzi/kafka-quotas-plugin) focuses on preventing the Kafka brokers from running out of disk space instead of limiting the individual clients or users.
```

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Quotas in Strimzi
+
+Currently, Strimzi does not have any API support for quotas in the `Kafka` custom resource.
+All quotas are configured directly by the users in `.spec.kafka.config` section of the `Kafka` custom resource.
+That gives users a lot of freedom and flexibility to configure the quotas.
+But it also does not provide them any support when the quota plugin requires some more complicated configuration.
```
see-quick, line 34:
```suggestion
But it also does not provide them any support when the quota plugin requires a more complicated configuration.
```

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+
+The only place where Strimzi directly supports configuring quotas is in the `KafkaUser` resource.
+It can be used to configure a specific quota for a given user using Kafka's built-in quota plugin.
+
+## Motivation
+
+The absence of an dedicated API for quota management in the `Kafka` CR currently creates two major limitations.
```
see-quick, line 41:
```suggestion
The absence of a dedicated API for quota management in the `Kafka` CR currently creates two major limitations.
```

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+   Because the initial releases were very simple, they had also very simple configuration.
+   And it was possible to configure them in the `.spec.kafka.config` section of the `Kafka` CR.
+   The next - not yet released - version of the Strimzi Quotas plugin brings many improvements.
+   More details about how the new version works can be found in the [Cluster Wide Volume Usage Quota Management](./047-cluster-wide-volume-usage-quota-management.md) proposal.
+   It operates at the level of a whole Kafka cluster and it has improved ability to prevent replicated messages to fill in the broker storage.
+   It also supports JBOD storage.
+   But it has also much more complicated configuration.
+   The plugin instances running in the different Kafka brokers communicate with each other using a Kafka topic.
+   And need therefore to connect to the Kafka brokers as clients as well.
+   Having the user configure this directly in the `.spec.kafka.config` section would be much more complicated as the user would need to configure the keystores and truststores, their passswords, bootstrap address etc.
```
see-quick, line 62:
```suggestion
   Having the user configure this directly in the `.spec.kafka.config` section would be much more complicated, as the user would need to configure the keystores, truststores, their passwords, bootstrap address, etc.
```

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+   Having a dedicated API for quota configuration would make it much easier to have the operator configure the client automatically.
+
+## Proposal
+
+This proposal suggests to add a new `quotas` API to the `Kafka` custom resource.
+It will be inside the `.spec.kafka` section.
+And it will allow configuring different types of quotas.
```
see-quick, line 69:
```suggestion
It will allow users to configure different types of quotas.
```

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+As many existing Strimzi APIs, it will use the `type` field to differentiate the quota plugin that will be used.
+
+This proposal suggests to initially add support for two Quota plugins.
+1. The build-in Kafka plugin
+2. The Strimzi Quotas plugin
+
+The first supported plugin - the Kafka's built-in plugin - will be configured as `type: kafka`.
```
see-quick, line 76:
```suggestion
The first supported plugin - Kafka's built-in plugin  - will be configured as `type: kafka`.
```

062-quotas-management.md
see-quick, line 99:
```suggestion
* When the `type: kafka` quotas API will be set, but one or more of the fields with the individual quotas are missing, the operator will make sure that the given default quota is not set.
```

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+The quota plugin configuration will not be set a forbidden configuration.
+When the new `.spec.kafka.quotas` API would be not set at all, Strimzi will keep the default Kafka configuration (which is the default built-in quotas plugin).
+Or - in case the user customized the quota plugin configuration in the `.spec.kafka.config` section - Strimzi will pass this configuration to Kafka.
+
+That helps to ensure backwards compatibility.
+Any users who current use the default plugin or any custom plugin would not be required to start using the `.spec.kafka.quotas` section immediately.
+They will be able to keep the configuration as is.
+
+When the `.spec.kafka.quotas` API is used and the `.spec.kafka.config` section contains some custom quota plugin configuration, Strimzi will configure the quotas according to the `.spec.kafka.quotas` API.
+But it will set a warning condition and log a warning about this conflict.
```
see-quick, line 148:
```suggestion
However, it will set a warning condition and log a message about this conflict.
```

062-quotas-management.md
see-quick, line 102: I was thinking about considering naming conventions based on their specific functionality e.g., `type: 
strimziDiskSpace` or `strimziRateLimit`? Then if in the future we introduce a new quota plug-in we can easily differentiate by their names.

see-quick (commented):
The content is clear and well-composed—just a couple of minor points to consider.👍

062-quotas-management.md
scholzj: Yes, it is declarative. So not being set in the Kafka CR means it should not be set in the Kafka cluster either. (in this case, the default value for the default quota is null / not set).

062-quotas-management.md
scholzj: The configuration mostly copies the plugin configuration as described in the README. So this expects the user to specify any users for that it would not apply (not entirely sure what was the motivation was for this).

If some internal users need to be excluded as well, please let me know which and why and I can add it to the proposal. We should probably also add it to the plugin README?

062-quotas-management.md
scholzj: I think this is explained on  the current lines 147-148?

062-quotas-management.md
scholzj: I don't follow this. Strimzi controls the configuration. Something like _own custom configuration_ is something what is really available only in _seelcted places_ because there was no way how to reasonably validate it. But this does not seem to be the case.

062-quotas-management.md
scholzj: Again, I don't follow this. The plugin runs inside Kafka. There is no interaction between the operator and the plugin. And I do not follow what logic are you talking about here. Again, the README of the quotas plugin does not seem to mention anything like that.

scholzj:
@tombentley @see-quick @SamBarker I do not have a strong preference for the `type: strimzi` name and I realize it has its issues. But TBH, I think it is still better than all your suggestions. So I think we should come up with something better. We should also consider it in the context of `type: kafka` which suffers from the same issues as the `type: strimzi`. So I think we need a better name for that one as well if we do not want to name it after the project it comes from.

062-quotas-management.md
SamBarker: No, I'm not aware of any specific user which need to be excluded. I was just wanting to clarify if the values in here were expected to be an exclusive list or merged with a Strimzi internal one. From your answer its clear it's exclusive. 

The motivation is to ensure that configured admin clients could still access the cluster e.g. to persist more up to date readings for disk usage in the cluster (from the earlier iteration of the plug-in).

062-quotas-management.md
SamBarker: Yes, I'd agree. I didn't pick it up and remove this when reading sequentially initially.

062-quotas-management.md
SamBarker: I interpreted this line as saying the Operator **will not** present invalid configuration to the Kafka Broker (and by extension the plug-in). Thus the operator would need to understand what does and doesn't represent valid configuration for the plug-in. Which is simple enough for something like the callback class but seems difficult to do for mutually exclusive properties. Thus the operator would need some mechanism to understand what the plug-in validation.

062-quotas-management.md
```diff
@@ -0,0 +1,163 @@
...
+      producerByteRate: 1048576
+      consumerByteRate: 2097152
+      requestPercentage: 55
+      controllerMutationRate: 50
+    # ... 
+  # ...
+```
+
+Based on this configuration, the operator will:
+* Configure the default quota plugin (make sure the `client.quota.callback.class` option is not set)
```
ppatierno, line 97: will `client.quota.callback.class` be forbidden or still usable but the operator checks that both `quotas` and `client.quota.callback.class` are not set?

062-quotas-management.md
ppatierno: I would agree with you Tom but at this point, if you are mentioning a "new kind of quota plugin", even the current quota plugin name `kafka-quotas-plugin` is not that good, or? Because we could have in Strimzi a new kind of Kafka-related quota plugin.

062-quotas-management.md
scholzj: > The motivation is to ensure that configured admin clients could still access the cluster e.g. to persist more up to date readings for disk usage in the cluster (from the earlier iteration of the plug-in).

Well, Strimzi has _admin clients_. But they usually do not produce or fetch messages to/from topics. So they should be mostly unaffected?

That said, there are some exceptions ...
* Replication => I assume that does not have to be excluded, as the plugin should not affect it?
* Cruise Control => Do we have any reasons to have it excluded? For doing a rebalance with Cruise Control if the disk space issue caused by an imbalance in disk usage?
* Bidirectional Topic Operator => I would not expect it to produce enough data to fill in the disk. But not sure we want it excluded. Maybe to allow users to reconfigure retention using it? This problem will go away once BTO is removed since UTO does not have any topic.

TBH, I do not know the plugin details, I'm just trying to bring it to some end, but really rely on your input here 😉

062-quotas-management.md
scholzj: Well, I would definitely not say `will not present invalid configuration to the Kafka Broker` in bold. It can do some validation. It often cannot do everything and it often relies on the users not being complete idiots (that assumption is a sometimes pushing it a lot, but not easy to avoid it).

It can do some basic validation such as that `minAvailableBytesPerVolume` and `minAvailableRatioPerVolume` should not be set at the same time. But it will probably not have a complex understanding of the whole configuration. So it is definitely not doing everything. 

I think that normally works reasonably well. But if you are aware of any things we should validate, please share them and I can consider them in the proposal.

> Which is simple enough for something like the callback class but seems difficult to do for mutually exclusive properties. Thus the operator would need some mechanism to understand what the plug-in validation.

So, are you saying the plugin has a validation callback I can just call, pass it the 6 or ho many string values and it tells me it is valid?

062-quotas-management.md
scholzj: So, what if we use the `producerByteRate` and `consumerByteRate` already used by the Kafka's own plugin?

062-quotas-management.md
scholzj: No. That would make this backward incompatible. It should be implicit from the backward compatibility part of the proposal. But I added it explicitly as well.

062-quotas-management.md
SamBarker: The Strimzi plug-in only applies limits to produce requests to topics so all other operations are un-affected.

So:

- `Replication` is un-affected. 
- `Cruise Control` might be worth so that it can refresh its metrics even when external users are blocked. (If I am right in saying it has a topic to store its metrics)
- `BTO` Would need to be included in the allow list so that it could operate while the cluster is blocked.
- `Canary` would be an other example of something to potentially exclude by default (though that might be better handled in the canary documentation).

scholzj: Canary is not deployed through the Kafka CR. So that is something users will need to exclude them self using the option.

SamBarker: I suspect I'm over thinking it, but I was imagining users being able to configure their own quota plug-ins not just the Kafka and Strimzi provided ones, but they can still do that directly without operator support so we don't need to support arbitrary config blobs here.

062-quotas-management.md
SamBarker: Ok, that makes sense to me. 

I consider this resolved but don't see a button to mark it as resolved 🤷‍♂️

062-quotas-management.md
SamBarker: The plug-in does that validation during startup but doesn't currently expose that in any other way. We could expose a validation API but I don't think its really worth the effort for a single plug-in.

Working on the assumption the project only wants to support the plug-ins it packages then I think encoding the validation in the operator is fine.

062-quotas-management.md
scholzj: We could in theory add some third type `type: custom` where users can freely configure everything. But as we have to let them also do it directly in `.spec.kafka.config`, I'm not sure it would have any value since they can do it anyway there. I guess it would make sense only if we bundle this inside a feature gate to somehow disable the quotas configuration in `.spec.kafka.config` after some time. But I'm not sure that is really useful and easily implementable.

scholzj: I updated the proposal accordingly.

062-quotas-management.md
PaulRMellor: It might be useful to indicate in the name what the plugin's main focus is. Something like the suggestions mentioned or `storageLimits` or `storageQuotas` and so on

062-quotas-management.md
```diff
@@ -0,0 +1,168 @@
...
+* Set the default user quotas
+* When the `type: kafka` quotas API will be set, but one or more of the fields with the individual quotas are missing, the operator will make sure that the given default quota is not set (set to null).
+* The other configuration options for the built-in Kafka plugin can be configured by the user in the `.spec.kafka.config` section in the `Kafka` CR.
+
+The second supported plugin - the Strimzi Quotas plugin - will be configured as `type: strimzi`.
+The API will allow to configure the storage limits at which the plugin will start limiting the producers in order to prevent the disk from getting full.
```
PaulRMellor, line 103:
```suggestion
The API will allow you to configure the storage limits at which the plugin will start limiting the producers in order to prevent the disk from getting full.
```

062-quotas-management.md
```diff
@@ -0,0 +1,168 @@
...
+* `minAvailableBytesPerVolume` for setting the minimal required free space per volume as number of bytes (this field can be configured only when `minAvailableRatioPerVolume`is not set)
+* `minAvailableRatioPerVolume` for setting the minimal required free space per volume as percentage of the overall volume size (this field can be configured only when `minAvailableBytesPerVolume`is not set)
+* `excludedPrincipals` for setting a list of users to be excluded from the applied quotas
+
+Other options might be configured directly by the user in `.spec.kafka.config` section of the `Kafka` CR.
+
+Based on this configuration, the operator will:
+* Set the `client.quota.callback.class` field in Kafka configuration to `io.strimzi.kafka.quotas.StaticQuotaCallback`.
+* Set the configuration on how to connect to the Kafka broker as a client using the `client.quota.callback.static.kafka.admin.` options.
+* Configures the default produce and fetch quota (`client.quota.callback.static.produce` and `client.quota.callback.static.fetch`).
```
PaulRMellor, line 135:
```suggestion
* Configure the default produce and fetch quota (`client.quota.callback.static.produce` and `client.quota.callback.static.fetch`).
```

062-quotas-management.md
```diff
@@ -0,0 +1,168 @@
...
+* Configure the minimal available disk space per-volume using the `client.quota.callback.static.storage.per.volume.limit.min.available.bytes` and `client.quota.callback.static.storage.per.volume.limit.min.available.ratio` options.
```
PaulRMellor, line 136:
```suggestion
* Configure the minimal available disk space per-volume using the `client.quota.callback.static.storage.per.volume.limit.min.available.bytes` or `client.quota.callback.static.storage.per.volume.limit.min.available.ratio` options.
```

062-quotas-management.md
```diff
@@ -0,0 +1,168 @@
...
+* Set the excluded users for that the quotas will not be applied using the `client.quota.callback.static.excluded.principal.name.list` option.
```
PaulRMellor, line 137:
```suggestion
* Set excluded users for whom quotas will not be applied using the `client.quota.callback.static.excluded.principal.name.list` option.
```

062-quotas-management.md
```diff
@@ -0,0 +1,168 @@
...
+* Configures the default produce and fetch quota (`client.quota.callback.static.produce` and `client.quota.callback.static.fetch`).
+* Configure the minimal available disk space per-volume using the `client.quota.callback.static.storage.per.volume.limit.min.available.bytes` and `client.quota.callback.static.storage.per.volume.limit.min.available.ratio` options.
+* Set the excluded users for that the quotas will not be applied using the `client.quota.callback.static.excluded.principal.name.list` option. 
+  The list of excluded users will also automatically (regardless whether user specified some custom excluded users or not) include the internal users used by Topic Operator and Cruise Control to not block them as they might be useful to resolve the disk space issue:
+  * Topic Operator by changing the retention of the existing topics. 
+    (Producing / fetching messages is required only by the Bidirectional Topic Operator. So the TO user can be removed only BTO is not supported anymore.)
+  * Cruise Control by rebalancing the topics between the brokers to address the disk space shortage.
```
PaulRMellor, line 141:
```suggestion
Regardless of whether custom exclusions are specified, the list of excluded users will automatically include internal users employed by the Topic Operator and Cruise Control. This ensures these internal users are not obstructed, as they might be required for resolving disk space issues:
  * The Topic Operator modifies the retention settings of existing topics. Message production and fetching activities are required only for the Bidirectional Topic Operator. Therefore, the Topic Operator user can be removed if support for the Bidirectional Topic Operator is discontinued.
  * Cruise Control rebalances topics among brokers to address disk space shortages.
```

PaulRMellor: I'm not entirely clear about what we're trying to say with this: _(Producing / fetching messages is required only by the Bidirectional Topic Operator._ I rewrote a bit in the suggestion above, but it feels like it's missing something.

062-quotas-management.md
```diff
@@ -0,0 +1,168 @@
...
+
+The quota plugin configuration will not be set a forbidden configuration.
```
PaulRMellor, line 143:
```suggestion
It will not be possible to set a forbidden configuration for the Strimzi Quota plugin.
```

062-quotas-management.md
```diff
@@ -0,0 +1,168 @@
...
+   The only way to configure them right now is imperatively use the Kafka Admin API.
+   While it can be used by users, it does not follow Strimzi's basic idea of declarative management.
+   Having a dedicated API for quota configuration as part of the `Kafka` CR would make it much easier to configure the default quotas.
+2. The initial releases of the Strimzi Quotas Plugin were very simple.
+   It helps to prevent users producing directly to a broker that was running out of disk space.
+   But it fails to prevent replication of messages produced to other brokers in the Kafka cluster from filling in the disk.
+   It also doesn't support many additional features such as JBOD storage.
+   Because the initial releases were very simple, they had also very simple configuration.
+   And it was possible to configure them in the `.spec.kafka.config` section of the `Kafka` CR.
+   The next - not yet released - version of the Strimzi Quotas plugin brings many improvements.
```
PaulRMellor, line 55:
```suggestion
   The next - not yet released - version of the Strimzi Quota plugin brings many improvements.
```

062-quotas-management.md
```diff
@@ -0,0 +1,168 @@
...
+## Proposal
+
+This proposal suggests to add a new `quotas` API to the `Kafka` custom resource.
+It will be inside the `.spec.kafka` section.
+And it will allow users to configure different types of quotas.
+As many existing Strimzi APIs, it will use the `type` field to differentiate the quota plugin that will be used.
+
+This proposal suggests to initially add support for two Quota plugins.
+1. The built-in Kafka plugin
+2. The Strimzi Quotas plugin
```
PaulRMellor, line 74:
```suggestion
2. The Strimzi Quota plugin
```

062-quotas-management.md
```diff
@@ -0,0 +1,168 @@
...
+  # ...
+```
+
+Based on this configuration, the operator will:
+* Configure the default quota plugin (make sure the `client.quota.callback.class` option is not set)
+* Set the default user quotas
+* When the `type: kafka` quotas API will be set, but one or more of the fields with the individual quotas are missing, the operator will make sure that the given default quota is not set (set to null).
+* The other configuration options for the built-in Kafka plugin can be configured by the user in the `.spec.kafka.config` section in the `Kafka` CR.
+
+The second supported plugin - the Strimzi Quotas plugin - will be configured as `type: strimzi`.
```
PaulRMellor, line 102:
```suggestion
The second supported plugin - the Strimzi Quota plugin - will be configured as `type: strimzi`.
```

062-quotas-management.md
```diff
@@ -0,0 +1,168 @@
...
+* Set the excluded users for that the quotas will not be applied using the `client.quota.callback.static.excluded.principal.name.list` option. 
+  The list of excluded users will also automatically (regardless whether user specified some custom excluded users or not) include the internal users used by Topic Operator and Cruise Control to not block them as they might be useful to resolve the disk space issue:
+  * Topic Operator by changing the retention of the existing topics. 
+    (Producing / fetching messages is required only by the Bidirectional Topic Operator. So the TO user can be removed only BTO is not supported anymore.)
+  * Cruise Control by rebalancing the topics between the brokers to address the disk space shortage.
+
+The quota plugin configuration will not be set a forbidden configuration.
+When the new `.spec.kafka.quotas` API would be not set at all, Strimzi will keep the default Kafka configuration (which is the default built-in quotas plugin).
```
PaulRMellor, line 144:
```suggestion
If the new `.spec.kafka.quotas` API is not set at all, Strimzi will keep the default Kafka configuration (which is Apache Kafka's built-in Quota plugin).
```

062-quotas-management.md
PaulRMellor: I may have missed the point here. If so, maybe this _forbidden_ aspect needs some clarification.

062-quotas-management.md
```diff
@@ -0,0 +1,168 @@
...
+Or - in case the user customized the quota plugin configuration in the `.spec.kafka.config` section - Strimzi will pass this configuration to Kafka.
+That means that the Apache Kafka `client.quota.callback.class` will not be marked as forbidden option.
```
PaulRMellor, line 146: Same for the forbidden comment above

062-quotas-management.md
```diff
@@ -0,0 +1,168 @@
...
+
+That helps to ensure backwards compatibility.
+Any users who current use the default plugin or any custom plugin would not be required to start using the `.spec.kafka.quotas` section immediately.
```
PaulRMellor, line 149:
```suggestion
Any users who currently use the default plugin or any custom plugin would not be required to start using the `.spec.kafka.quotas` section immediately.
```

PaulRMellor (commented):
Nice proposal. The need for the change is clearly presented, as too the approach to introduce the feature.

PaulRMellor (approved):

scholzj:
I was thinking a bit more about the naming as that seemed to be the sticky point. I still think my original choice is the best ...

Kafka supports only one quota plugin at a time
* Do you really expect that if we add we need a new type of quota, we create a second Quota plugin and force the users to choose one or the other? (leaving aside that we are struggling to finish the current plugin)
* Same for Kafka, do you really see it adding a new separate plugin?

I think that giving it a generic name based on the origin of the plugin allows us to extend the plugin in the future. If we call it today based on storage, what if we add some new feature not related to it? Will we rename it? Will we apologetically explain why is it called `storageQuota` while it has 4 other features?

An additional argument is that already today the quota plugin does two things:
* It has a generic per-cluster throughput quota that is valuable on its own
* It has a storage-based part that depends on the storage utilization

So calling it `type: storage` is already a bit confusing. Also, if the focus of the quota is storage, should its class name be `StorageQuota` instead of `StaticQuotaCallback` with pretty much no meaning for the user? Doesn't that also suggest that it does not necessarily have a focus on one particular feature?

CC @PaulRMellor @tombentley @see-quick @ppatierno @SamBarker

PaulRMellor:
> I was thinking a bit more about the naming as that seemed to be the sticky point. I still think my original choice is the best

I guess leaving room for future extensions and for the plugin to evolve is quite an important consideration.  And we don't want to be stuck with a specific name that's then not fit for purpose. If the generic name is less self-explanatory, we can make sure the docs explain it well ;-)

scholzj:
As this seems to be dead now, I'm probably going to close this proposal.

What do we expect to do around the quotas? Should we have a proposal suggesting the removal of the quota plugin from the container images and archival of the quota plugin project?

SamBarker (approved):
Sorry I missed the December ping. 

> Do you really expect that if we add we need a new type of quota, we create a second Quota plugin and force the users to choose one or the other? (leaving aside that we are struggling to finish the current plugin)

Yes, I think thats the lesser evil. As the alternative is we call it `Strimzi` and then in some future version we want to add quota support based on some other property we then have to make build logic to ensure the different mechanisms interact sensibly and document that logic. We also then have to expose a bunch more configuration for users to understand even when they only care about storage based quotas. 

> So calling it type: storage is already a bit confusing. Also, if the focus of the quota is storage, should its class name be StorageQuota instead of StaticQuotaCallback with pretty much no meaning for the user? Doesn't that also suggest that it does not necessarily have a focus on one particular feature?

Yes, your right the name is strange but I didn't change as it made for a similar migration story from single node limits to cluster wide limits. 

I'd don't feel that strongly about the naming that I think the proposal should be blocked over it calling the plug-ins `strimzi` or `kafka` is definitely better than the status quo

062-quotas-management.md
```diff
@@ -0,0 +1,167 @@
...
+* `minAvailableRatioPerVolume` for setting the minimal required free space per volume as percentage of the overall volume size (this field can be configured only when `minAvailableBytesPerVolume`is not set)
+* `excludedPrincipals` for setting a list of users to be excluded from the applied quotas
+
+Other options might be configured directly by the user in `.spec.kafka.config` section of the `Kafka` CR.
+
+Based on this configuration, the operator will:
+* Set the `client.quota.callback.class` field in Kafka configuration to `io.strimzi.kafka.quotas.StaticQuotaCallback`.
+* Set the configuration on how to connect to the Kafka broker as a client using the `client.quota.callback.static.kafka.admin.` options.
+* Configure the default produce and fetch quota (`client.quota.callback.static.produce` and `client.quota.callback.static.fetch`).
+* Configure the minimal available disk space per-volume using the `client.quota.callback.static.storage.per.volume.limit.min.available.bytes` or `client.quota.callback.static.storage.per.volume.limit.min.available.ratio` options.
```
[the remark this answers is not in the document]
scholzj: I guess they will not be set and the defaults from the Quotas plugin will be used. But I think it is a valid use-case that you want to just set the throughput quota and not care about the storage.

scholzj:
> Yes, I think thats the lesser evil. As the alternative is we call it Strimzi and then in some future version we want to add quota support based on some other property we then have to make build logic to ensure the different mechanisms interact sensibly and document that logic. We also then have to expose a bunch more configuration for users to understand even when they only care about storage based quotas.

@SamBarker So, how would you explain it to some users asking for some new kind of quota?
```
Yeah, we will add that as a new quota type. You have to choose whether to use the new quota and risk your cluster dying from a full disk or not to use the new quota type.
```

As long as Kafka does not handle the full disks gracefully, it seems unlikely to me that we would add new separate _quota plugins_. The full disk protection will be likely more important and not being able to have both of them would significantly reduce the possible users of the new plugin and make it not worth the effort. And if one day Kafka handles the full disks gracefully, we can anyway deprecate most of the plugins as not needed anymore and start with a new types for separate features. So yes, adding new quotas into the same plugin sounds like something that would not be nice and might never happen. But I think it is still more likely than creating separate quota plugins for other features.

062-quotas-management.md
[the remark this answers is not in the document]
fvaleri: > I guess they will not be set and the defaults from the Quotas plugin will be used.

Looks like there is no default, which I guess it means no check: https://github.com/strimzi/proposals/blob/20a4043c5a2f86e35995eb12b1ac50aac8929154/047-cluster-wide-volume-usage-quota-management.md#limit-configuration

I think it would be better to be opinionated on that an set a default value.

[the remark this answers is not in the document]
scholzj: I don't think so. As I said, it might be a valid situation that you just want to disable it.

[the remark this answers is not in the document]
scholzj: Also, if we would decide to have a default, it should be done in the Qoutas plugin and not in the operator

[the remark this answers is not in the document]
fvaleri: > if we would decide to have a default, it should be done in the Qoutas plugin and not in the operator

I agree on this. @SamBarker wdyt?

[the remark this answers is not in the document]
SamBarker: I sort of agree. If we are to have an opinionated default it should be done in the plug-in.

There was a lot of hesitation during the proposal review about defaulting to blocking access so we default `client.quota.callback.static.throttle.factor.fallback` to `1.0` so I'm not sure we have a different calculus here (personally I think we should default to blocking but thats another conversation). 

The bigger question is how to determine a sensible default? Setting `min.available.bytes=1GB` allows almost maximum usage but potentially makes for un-recoverable brokers. If they crash for other reasons there still isn't room on the volume to do log recovery. Using `min.available.ratio=0.05` is more forgiving but leads to questions about wasted disk space and potentially the same flaw (I don't have a picture of how much space is required for log recovery).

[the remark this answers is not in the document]
fvaleri: Can't we figure out a way to compute this default based on the number of IO threads and the segment size (Tom's suggestion)? We may need to dig into the code to figure this out.

I've seen many users blindingly trusting Kafka and Strimzi default values in production. For this reason, I also think it's important to have a default value, even if it's not perfect.

@showuon @mimaison @tombentley what's your opinion on this?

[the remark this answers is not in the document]
scholzj: * We haven't been able to finish the quotas stuff and bring it to working state for several years now. Do you really want to now start creating some complex schemas how to define storage limits based on number of threads?
* I think you are still missing the main point. The quotas will not be enabled by default and they cannot be enabled by default because of Kafka's native quotas. So the default affects only people who intentionally enable the quotas but without defining the storage limit. We can discuss how big the group of people who would enable a plugin for quotas based on storage but not set the limit by mistake is. But it is certainly not equal to `users blindingly trusting Kafka and Strimzi default values in production`.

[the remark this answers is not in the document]
fvaleri: > the default affects only people who intentionally enable the quotas but without defining the storage limit.

I'm aware, see my first comment on this thread.

> We can discuss how big the group of people who would enable a plugin for quotas based on storage but not set the limit by mistake is.

The real question is: Can you trust them setting a sensible value that would avoid the potential log recovery issue that Sam was mentioning?

Not saying we should do it now, but maybe spending some time thinking about it would be good.

fvaleri (approved):
LGTM. Thanks.

062-quotas-management.md
```diff
@@ -0,0 +1,167 @@
...
+The absence of a dedicated API for quota management in the `Kafka` CR currently creates two major limitations.
+1. The built-in Apache Kafka quota plugin has an ability to configure _default quotas_.
+   Default quotas apply to any user that has no specific quota set.
+   However, the default quotas cannot be configured declaratively in the `.spec.kafka.config` section.
+   And they cannot be configured in a `KafkaUser` resource either because they do not belong to any particular user.
+   The only way to configure them right now is imperatively use the Kafka Admin API.
+   While it can be used by users, it does not follow Strimzi's basic idea of declarative management.
+   Having a dedicated API for quota configuration as part of the `Kafka` CR would make it much easier to configure the default quotas.
+2. The initial releases of the Strimzi Quota Plugin were simple, preventing users from producing messages directly to a broker running out of disk space. 
+   However, these versions did not address the issue of replicated messages produced on other brokers in the Kafka cluster, which could still fill up the disk space of the destination broker."
```
ppatierno, line 50:
```suggestion
   However, these versions did not address the issue of replicated messages produced on other brokers in the Kafka cluster, which could still fill up the disk space of the destination broker.
```

062-quotas-management.md
```diff
@@ -0,0 +1,167 @@
...
+metadata:
+  name: my-cluster
+spec:
+  kafka:
+    quotas:
+      type: kafka
+      producerByteRate: 1048576
+      consumerByteRate: 2097152
+      requestPercentage: 55
+      controllerMutationRate: 50
```
ppatierno, line 89: I was wondering if an opaque list isn't better than specific fields (I was thinking at the tiered storage proposal doing this way).

scholzj: It is just 4 fields -> We have it defined like this also in the `KafkaUser` CRD, it gives us better validation etc. I do not have a strong opinion if more people want it as a Map, but I would probably prefer to keep the dedicated fields.

ppatierno: Well they are just for filed for this type of plugin. They could be more for another type. Maybe I am overthinking?

scholzj: I don't know. We use the non-validated `<String, String>` maps normally in places where we cannot really use proper fields:
* Because there are too many options to model each one in the API (e.g. `.spec.kafka.config`)
* Because the list of options is not finite and new options can be part of plugins or new releases (e.g. `.spec.kafka.config`)
* Because we add support for a custom plugin and thus do not know the configuration options (this would be the tiered storage for example)

We do not use it in places like this where we have few clearly defined options.

You are right that there might be another quota plugin somewhere in the future. But given how this proposal goes, I would not see it happen anytime soon 😉. Also, such a plugin would have its own type and configurations anyway.

ppatierno: Ok let's move on with this for now.

ppatierno (approved):

scholzj:
Technically, this now has the required number of binding votes: Paolo, Paul + me as an author.

@tombentley @see-quick and anyone else interested: Please comment with any new ideas or suggestions or place your votes by Monday 11th EOB or I will close it as approved.

062-quotas-management.md
Frawless: Sounds reasonable to me to use the same naming

Frawless (approved):

scholzj:
Approved with 3 binding +1 votes and 2 non-binding +1 votes.

PR #98, opened 2023-11-06
Proposal to add tiered storage configuration

gaobosince1987:


ppatierno:
Tbh I don't think that this proposal is covering what @scholzj raised in https://github.com/strimzi/strimzi-kafka-operator/issues/9259. 
This seems to be just an explanation of how updating only Helm charts to have such a support. And Helm charts are only just one way to deploy a Kafka cluster. I think you are just solving your own problem, not providing a real tiered storage support to the project.

scholzj (commented):
I agree with Paolo. Thi is more a guide of how you are doing it. Rather then a proposal how strimzi should address it. It does not really address the security for example for providing the AWS credentials or configuring the Kafka client connection for the tired storage plugin. It also talks about some non-existent Strimzi Helm Chart. The only Helm Chart Strimzi provides is to install the Cluster Operator. So you are likely refering to some Helm Chart that is your own. It should also cover wider range of possible plugins and how they would work and not just a single one.

gaobosince1987:
> I agree with Paolo. Thi is more a guide of how you are doing it. Rather then a proposal how strimzi should address it. It does not really address the security for example for providing the AWS credentials or configuring the Kafka client connection for the tired storage plugin. It also talks about some non-existent Strimzi Helm Chart. The only Helm Chart Strimzi provides is to install the Cluster Operator. So you are likely refering to some Helm Chart that is your own. It should also cover wider range of possible plugins and how they would work and not just a single one.

Line 68 is talking about how tier storage plugin configure client to talk to secured internal topic. 
Line 85 mentions credentials can be deferred to custom plugins. AWS/GCP have preferred ways to manage secrets, custom plugins can handle it better.
Line 64 mentions a way to include plugin jars in a new image and specify configuration path for tier storage to use, the same mechanism can be used in the other plugins.

gaobosince1987:
> Tbh I don't think that this proposal is covering what @scholzj raised in [strimzi/strimzi-kafka-operator#9259](https://github.com/strimzi/strimzi-kafka-operator/issues/9259). This seems to be just an explanation of how updating only Helm charts to have such a support. And Helm charts are only just one way to deploy a Kafka cluster. I think you are just solving your own problem, not providing a real tiered storage support to the project.

Just for my understanding, what are the other ways you are talking about to deploy Kafka cluster is needed here?

scholzj:
I appreciate the effort you put into this. And I assume everything you wrote in the proposal would work for you fine today. But for the official support, we need to make sure that the solution is:
* User friendly
* Extensible
* Can provide backward compatibility and maintainability in the long term
* Can handle all of the various environments and infrastructures where Strimzi is used.

The two points I raised in particular do not provide something what can be supported and maintained in the long term and that respects the diversity of environments Strimzi is used in.

> Line 68 is talking about how tier storage plugin configure client to talk to secured internal topic. 

That relies heavily on internal implementation details. That is not stable and might not work in the next release. So that is not something that can be used as an official solution. This is likely to change in the future.

> Line 85 mentions credentials can be deferred to custom plugins. AWS/GCP have preferred ways to manage secrets, custom plugins can handle it better. 

Strimzi is used in a variety of environments. Not just in AWS/GCP. So I do not think we can rely on their credential injection. It is good to be compatible with it, but not sufficient. Also, how would it work for other technologies than just S3? For example with some data lake etc.

> Just for my understanding, what are the other ways you are talking about to deploy Kafka cluster is needed here?

Strimzi operands are deployed using a custom resource. That can be done from a command line using `kubectl apply`, it can be done using various GitOps tooling, or it can be done dynamically from someone's application. As I said, Strimzi actually does not have any Helm Chart as you refer to - I know there are users who wrote such Helm Charts and you are likely one of them. But none of that is part of Strimzi. So the proposal should deal at the level of things whcih are part of Strimzi.

scholzj:
Replaced by #99

PR #106, opened 2024-01-19
Proposal for moving data between two JBOD disks using Cruise Control

ShubhamRwt:
This proposal shows how we can implement the feature to move data between two JBOD disks using Cruise Control


063-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,71 @@
...
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaRebalance
+metadata:
+  name: my-rebalance
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  mode: remove-disks
+  brokerandlogdirs: [brokerid1-logdir1,brokerid2-logdir2...] # for eg. 0-/var/lib/kafka/data-0/kafka-log0
```
scholzj, line 51: `brokerandlogdirs` seems strange as a name. If nothing else, you should use proper case: `brokerAndLogDirs`. For the users, it is quite complicated to understand what the paths are etc. as they are an internal implementation. So I also wonder if you should create more elegant APIs. E.g. something like:
```
spec:
  mode: remove-disks
  brokerandlogdirs:
    - brokerId: 0
      volumeId: 0
    - brokerId: 1
      volumeId: 0
    - brokerId: 2
      volumeId: 0
```

Or something similar that would be based on what the user configures int he Kafka / KafkaNodePool resources and nto on the internal paths.

063-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,71 @@
...
+    strimzi.io/cluster: my-cluster
+spec:
+  mode: remove-disks
+  brokerandlogdirs: [brokerid1-logdir1,brokerid2-logdir2...] # for eg. 0-/var/lib/kafka/data-0/kafka-log0  
+# ...
+```
+
+### Flow
+
+- The user should be using the `Kafka` resource with JBOD configured.
```
scholzj, line 57: What if the user uses KafkaNodePool resources? What if the user does not use JBOD type storage?

063-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Flow
+
+- The user should be using the `Kafka` resource with JBOD configured. 
+- When the Kafka cluster is ready, the user creates a `KafkaRebalance` custom resource with the `spec.mode` field as `remove-disks` and the list of the broker and the corresponding logdirs to move in the `spec.brokerandlogdirs` field.
+- The `KafkaRebalanceAssemblyOperator` starts the interaction with Cruise Control via the `/remove_disks` endpoint for getting an optimization proposal (by using the dryrun feature). 
+- The user accepts the proposal by applying the `strimzi.io/rebalance=approve` annotation on it. 
+- The `KafkaRebalanceAssemblyOperator` starts the interaction with Cruise Control via the `/remove_disks` endpoint for running the actual rebalancing.
+
+Note: The movement of data between the JBOD disks doesn't affect the broker load, therefore there will be no changes in the before/after broker load.
```
scholzj, line 63: Why does it not impact it? The moving of data between disks has to impact performance and load of the broker.

scholzj (commented):
Thanks for the proposal. I left some quick comments.

063-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,71 @@
+# Moving data between two JBOD disks using Cruise Control
+
+This proposal is about integrating the `remove_disks` endpoint from Cruise Control into Strimzi cluster operator. 
+This endpoint will allow us to move the data between two JBOD disks. 
+
+## Current situation
+
+Currently, we get a multiple requests to add the ability for moving all Kafka logs between two disks on the JBOD storage array. This feature can be useful in following scenarios:
```
ppatierno, line 8: multiple requests ... from community users I guess? Let's make it clear.

063-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,71 @@
...
+- The current disk is too big and the user wants to use smaller one
+- When we want to use different Storage Class with different parameters or different storage types.
+For now, we can do this using the Kafka CLI tools but it is not very user-friendly.
```
ppatierno, line 11: which Kafka CLI tools?

063-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,71 @@
+# Moving data between two JBOD disks using Cruise Control
+
+This proposal is about integrating the `remove_disks` endpoint from Cruise Control into Strimzi cluster operator.
```
ppatierno, line 3: I can't find any documentation about this endpoint in Cruise Control, for example it's missing in the wiki.
Is there an official place where we can find how it works and put a link here?

063-moving-data-between-two-jbod-disks-using-cruise-control.md
ppatierno: I agree with Jakub. In our current Strimzi implementation, the users have no knowledge of the internal paths on the volume where we configure the brokers to store logs. The only thing the user knows is the volume on a specific broker as pointed out  in the above example.
The user doesn't know about `/var/lib/kafka/data-0/kafka-log0` at all but we are using a very well defined pattern to make it depending on the volume for a broker.

063-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,71 @@
...
+  brokerandlogdirs: [brokerid1-logdir1,brokerid2-logdir2...] # for eg. 0-/var/lib/kafka/data-0/kafka-log0  
+# ...
+```
+
+### Flow
+
+- The user should be using the `Kafka` resource with JBOD configured. 
+- When the Kafka cluster is ready, the user creates a `KafkaRebalance` custom resource with the `spec.mode` field as `remove-disks` and the list of the broker and the corresponding logdirs to move in the `spec.brokerandlogdirs` field.
+- The `KafkaRebalanceAssemblyOperator` starts the interaction with Cruise Control via the `/remove_disks` endpoint for getting an optimization proposal (by using the dryrun feature). 
+- The user accepts the proposal by applying the `strimzi.io/rebalance=approve` annotation on it.
```
ppatierno, line 60: I guess the user can still use the auto approval feature we have. It's better to highlight it here.

063-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,71 @@
...
+  mode: remove-disks
+  brokerandlogdirs: [brokerid1-logdir1,brokerid2-logdir2...] # for eg. 0-/var/lib/kafka/data-0/kafka-log0  
+# ...
+```
+
+### Flow
+
+- The user should be using the `Kafka` resource with JBOD configured. 
+- When the Kafka cluster is ready, the user creates a `KafkaRebalance` custom resource with the `spec.mode` field as `remove-disks` and the list of the broker and the corresponding logdirs to move in the `spec.brokerandlogdirs` field.
+- The `KafkaRebalanceAssemblyOperator` starts the interaction with Cruise Control via the `/remove_disks` endpoint for getting an optimization proposal (by using the dryrun feature).
```
PaulRMellor, line 59:
```suggestion
- The `KafkaRebalanceAssemblyOperator` interacts with Cruise Control via the `/remove_disks` endpoint to generate an optimization proposal (by using the dryrun feature). 
```

063-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,71 @@
...
+# ...
+```
+
+### Flow
+
+- The user should be using the `Kafka` resource with JBOD configured. 
+- When the Kafka cluster is ready, the user creates a `KafkaRebalance` custom resource with the `spec.mode` field as `remove-disks` and the list of the broker and the corresponding logdirs to move in the `spec.brokerandlogdirs` field.
+- The `KafkaRebalanceAssemblyOperator` starts the interaction with Cruise Control via the `/remove_disks` endpoint for getting an optimization proposal (by using the dryrun feature). 
+- The user accepts the proposal by applying the `strimzi.io/rebalance=approve` annotation on it. 
+- The `KafkaRebalanceAssemblyOperator` starts the interaction with Cruise Control via the `/remove_disks` endpoint for running the actual rebalancing.
```
PaulRMellor, line 61:
```suggestion
- The `KafkaRebalanceAssemblyOperator` interacts with Cruise Control via the `/remove_disks` endpoint to perform the actual rebalancing.
```

063-moving-data-between-two-jbod-disks-using-cruise-control.md
PaulRMellor: I guess we want to mention `kafka-reassign-partitions.sh` tool

063-moving-data-between-two-jbod-disks-using-cruise-control.md
PaulRMellor: Do we want to link to the [yaml](https://github.com/ilievladiulian/cruise-control/blob/ad275079f271b28704a1d5ff00fff82891aebc89/cruise-control/src/main/resources/yaml/endpoints/removeDisks.yaml)

PaulRMellor (approved):
useful sought-after feature. :+1: 
I made a few suggestions as I read through

063-moving-data-between-two-jbod-disks-using-cruise-control.md
kyguy, line 8: Would this be better in the  `Motivation`section?

063-moving-data-between-two-jbod-disks-using-cruise-control.md
kyguy, line 11: This is a reference to the `kafka-reassign-partitions` script right? If so, you could emphasize how it is devising the partition mappings to migrate data between disks in this scenario is a manual, time-consuming process. Solutions exist for moving data of brokers but not off specific disks of brokers which have JBOD config

063-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,71 @@
+# Moving data between two JBOD disks using Cruise Control
+
+This proposal is about integrating the `remove_disks` endpoint from Cruise Control into Strimzi cluster operator. 
+This endpoint will allow us to move the data between two JBOD disks.
```
kyguy, line 4: Does it only allow moving data between two disks?

063-moving-data-between-two-jbod-disks-using-cruise-control.md
kyguy: I mislead Shubham on this one, I thought it wouldn't affect the broker load since the data going in and out of the broker would be the same regardless of how the data was distributed amongst the disks. But I understand that if one disk was faster than another, moving the data between them within a broker certainly would impact the performance 

@ShubhamRwt did we ever find out why the loadBefore/loadAfter information was not provided by this `remove_disk` endpoint? Is it just an oversight by upstream Cruise Control?

fvaleri (commented):
Hi @ShubhamRwt. I left just few comments. I will have another pass when out of draft state.

063-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,71 @@
...
+spec:
+  mode: remove-disks
+  brokerandlogdirs: [brokerid1-logdir1,brokerid2-logdir2...] # for eg. 0-/var/lib/kafka/data-0/kafka-log0  
+# ...
+```
+
+### Flow
+
+- The user should be using the `Kafka` resource with JBOD configured. 
+- When the Kafka cluster is ready, the user creates a `KafkaRebalance` custom resource with the `spec.mode` field as `remove-disks` and the list of the broker and the corresponding logdirs to move in the `spec.brokerandlogdirs` field.
```
im-konge, line 58: You have `remove_disks` on one place and `remove-disks` one another - it should be the same across whole proposal

063-moving-data-between-two-jbod-disks-using-cruise-control.md
ShubhamRwt: I will create a documentation bug and create a PR for the same on upsteam repo regarding this. The only piece of information for using this is only on the PR

tombentley:
Thanks @ShubhamRwt, what you describe here could certainly be made to work. But I'd like there to be some discussion about whether this is the right approach.

Before I get on to that, let's talk about terminology for a moment:
* Kafka provides an API to move replicas (logs) between log dirs.
The main (only?) use case for configuring a broker to use multiple log dirs is so that those directories are on separate disks/devices/volumes.
* Cruise Control has decided to use the term "disks", rather than log dirs.
*Kubernetes uses the term "volume" for something mounted into the container vfs. 
* Strimzi uses the Kubernetes terminology, so the Jbod storage API lists a number of `volumes`, which are each identified by integer ids.

I think this feature should stick with the established Strimzi terminology, so if we're going to add a new mode to the `KafkaRebalance` resource it should be `mode: remove-volumes`, rather than `mode: remove-disks`. 

### User experience

Kafka clusters need rebalancing because of uneven load on brokers imposed by the replicas they're hosting. 
"Load" is composed of various resources that brokers have in limited supply (CPU cycles, disk space, network bandwidth)
Rebalancing estimates the load imposed by each replica on a broker and tries to rearrange partition leader- and followership to get a better balance.
That rearrangement involves moving the data between brokers.

But moving data between log dirs is quite different. It's not a global rebalance at all. What CC is providing is much less involved, because the space of possible changes is much smaller: It can only move replicas to the remaining disks on the same broker, and all it's using for that assignment (currently) is knowledge of the disk space required by each replica.

It's possible that these things are done by different people, and they're always done for different reasons. Removing disks is usually going to be done because it's related to some operational necessity. Rebalancing is usually going to be done to get better resource usage. 

The schema of the `KafkaRebalance` CRD has all sorts of things which apply to the global rebalance case which do not apply for the volume removal case. Things like goals to do with CPU usage and network bandwidth are irrelevant for the volume removal case, because they'll be the same however the replicas are spread over the disks.

All this is to say that I'm not convinced by the approach of repurposing `KafkaRebalance` for this use case. I can see that it's _convenient_ to lean on `KafkaRebalance`, because a lot of the underlying machinery will be the same. 
But I don't think it fully addresses what the user wants to do.
With the `KafkaRebalance` approach the user will have to perform a multi-step change:

1. Do the `KafkaRebalance` with `mode: remove-volumes` to move the data. This can take quite some time if there are many TB to move.
2. Update the `Kafka`, or `KafkaNodePool`, to actually remove the volumes.
3. Possibly do a global rebalance.

There's a few things that can go wrong between steps 1 and 2:

* There's nothing in Kafka which will prevent replicas of new topics getting assigned to the to-be-removed volumes. We have some check in place to prevent disaster, but the fact remains that step 1 may need to be repeated if new topics get created.
* The user can screw up and name different volumes in step 1 than in step 2. There are two cases:
    - The don't remove data off volumes in step which which they try to remove from brokers in step 2. The safety check prevents disaster, but they have to re-do step 1.
    - They move data off volumes in step 1 which aren't ultimately removed from the brokers in step 2. 
      They then need to do step 3 to make use of those volumes again.
* The user can do step 1 and then forget about step 2 entirely.

Further more we need to consider the complexity we're imposing on users with this approach. They need to operationally coordinate steps 1 and 2. If they're using gitops this would involve applying a change for step 1 and at some later point one for step 2. This might involve different people (e.g. on different shifts). This all adds to the burden we're placing on users.

I think we should consider if this is the experience we want users to have. 

### Possible alternative

One alternative would be for this to be driven from the `Kafka` (or `KafkaNodePool`) CRs:
1. The user removes a volume from the JBOD `storage`
2. The CO checks whether this volume contains any replicas
3. If so, the CO:
    1. Checks for existing Cruise Control tasks: If any exist then we update the status with some new condition to say they volume removal is still pending, and proceed with the current reconciliation without taking further action. The user has to choice of cancelling those tasks themselves or waiting till they finish and a later CO reconciliation is able to proceed with the volume removal (at the next step). We could finesse this at some point so that, for example, `KafkaRebalance` itself declares whether it's auto-cancellable, and if so the CO could cancel it automatically.
    2. Eventually there are no on-going Cruise Control tasks, so the CO uses the `remove_disks` endpoint to orchestrate the data movement, updating the `Kafka` or `KafkaNodePool` CR status
    3. Eventually the `remove_disks` fails or succeeds: If it fails we give up and fail the reconciliation with a suitable message. If it succeeds then the to-be-removed volumes no longer have and replicas and we can proceed to the next step.
3. The CO proceeds with deleting the PVCs and restarting those brokers.

This is certainly more complicated for us to implement. (It happens over a number of reconciliations, for a start). But I think it's much closer to what users actually need: In the normal case they have a single CR to interact with (the one which already declares the `volumes` for their brokers) and they have to interact with it only once, to express their intent.

With this alternative the `KafkaRebalance` remains a CR used for optimizing the cluster. The API it expresses is more "type safe", in the sense that it doesn't have a load of options to do with global rebalancing which get ignored by a special mode which isn't really a rebalance.

ShubhamRwt:
Hi @scholzj @ppatierno , I do think that the above suggestion by @tombentley  looks good so do we want to put the `remove-disk` endpoint usage in `Kafka` resource?

scholzj:
I do not think it looks good.
* The original design Follows the existing logic for adding or removing brokers. The proposal from Tom creates yet another way to do what is basically the same thing.
* As I explained in a different discussion the other day, we should avoid tight couplings in the KafkaAssemblyOperator. It is already today overloaded and tightly coupling it with Cruise Control would make it even worse. It will make it hard to test things.

It might make sense to one day automate the removal of the disks. But I think that should be left to another proposal. It should be also done through the `KafkaRebalance` resources to decouple the logic of the different operations.

ShubhamRwt:
@ppatierno @tombentley Hi,  It would be great to have some views from you too , I will be starting to  re-write this.

ShubhamRwt:
Another thing to consider is that upstream CC is not so happy to enable the `verbose` parameter for this `remove-disk` endpoint while all the other  CC endpoints that we currently have in strimzi have the `verbose` parameter enabled. So the issue we will get with it is that we won't be able to get any `loadBeforeOptimization` part and that can cause error since we need both `loadBeforeOptimization` and `loadAfterOptimization` for proposal generation in Strimzi. I am currently trying to seek a work around for that. I am not sure why upstream CC has disabled verbose for this endpoint but on testing my self I was able to get this by enabling `verbose` -> https://gist.github.com/ShubhamRwt/f09f7c1a856eaa452b2696a86f1b7896

ShubhamRwt:
@tombentley Hi, actually I wanted your views on which path to move forward with. I see one way you mentioned above which is more towards adding this mechanism in the `Kafka` CR and at the same time after reading Jakub's comment, I think we want keep it in the `KafkaRebalance` CR itself. So i wanted to have discussion around it and make sure that we all agree on one approach

tombentley:
I'm not quite sure how to parse @ppatierno's comment, but I _think_ he's saying that he agrees with @scholzj. In which case I'm happy to respect the consensus.

ShubhamRwt:
Thanks for the views @tombentley @ppatierno @scholzj, I will continue with this way then, using the `kafkaRebalance` resource. Going to apply the other suggestions now

ShubhamRwt:
Hi, I was working on this on trying to get the POC ready, there is an issue which Iam having. Due to the `verbose` parameter not being enabled, I can't get the load before optimation part. The way we use CC in our code, we have the `verbose` parameter available with all the endpoints we are leveraging in Strimzi but with the `remove_disks` endpoint, it is not possible. I have asked the upstream community the reason why it is not there and haven't yet got any response. There is one  downside of this feature 
```After all replicas are moved from the specified disk, the disk may still be used by CC during rebalances and Kafka can still use it when creating topics.
```
[](Maybe this can be a reason why verbose was not enables since there is not much change in the after/before load. I also tried to enable verbose myself on this endpoint and see the changes but there were no big changes  -> https://gist.github.com/ShubhamRwt/c12fa47d36d53a0cba3473667020319f)

There are some alternatives to move forward:

1. Not showing any before load in the `optimization proposal` when using this endpoint and maybe prompting the reason that
2. Showing the same load in the before/after load i.e. using `loadAfterOptimization` in both the fields
3. Trying again to push the CC community to enable the verbose flag w.r.t endpoint. (This is something we have already tried by no good)

@ppatierno @kyguy @scholzj @tombentley wdyt, what should be the best approach?

scholzj:
If the Cruise Control doesn't provide the information, we just don't have it and don't show it. I do not think this is a blocker in any way. If the users do not like that they do not have the information, they don't have to use it. But I suspect most of them will not care.

ppatierno:
Yeah I tend to agree with Jakub. We don't have such information and the CC community seems not helping on this. We cannot block just for it but we should move forward by providing the users what we have.

ShubhamRwt:
Updated the proposal with improvements and moved it to `Ready` state now

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,85 @@
...
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaRebalance
+metadata:
+  name: my-rebalance
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  mode: remove_disks
+  brokerAndVolumeIds:
```
scholzj, line 53: I wonder if we should utilize the existing `brokers` field and just add new `volumes` field and combine them. It would not allow to remove different volumes form different brokers in one go. But I'm not sure that would matter for the use cases we have. If we don't do it, would calling it `brokersAndVolumes` make more sense given the `brokers` field name?

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Motivation
+
+We should introduce the logic to Strimzi to leverage Cruise Control integration and make it possible to move the data between two JBOD disks.
+This feature will also allow us to remove the disks without the loss of data.
+
+## Proposal
+
+Cruise Control provides `remove_disks` HTTP REST endpoint to move replicas from a specified disk to other disks of the same broker.
+This endpoint triggers a rebalancing operation by moving replicas in a size-based manner to the remaining disks, from the largest to the smallest, while checking the following constraint:
```
scholzj, line 23: So, this is always intra-broker rebalancing not moving any data between brokers?

scholzj (commented):
Some nits, but it seems mostly good to me. Thanks.

076-moving-data-between-two-jbod-disks-using-cruise-control.md
scholzj: PS: Maybe we should see what for example @ppatierno thinks about this before you change it.

fvaleri: I also like `brokerAndVolumes`. Also, we use dash a separator for the other modes, so it shuold be `remove-disks`.

> It would not allow to remove different volumes form different brokers in one go. 

Why should we limit to only one broker? In a Kafka cluster is common to have the same storage configuration and sizing in all brokers. If the cluster is well balanced, you may need to remove disks on multiple brokers, and doing it in one go is great.

fvaleri (approved):
Hi @ShubhamRwt, thanks for this useful proposal. I just let a couple of minor suggestions and considerations.

076-moving-data-between-two-jbod-disks-using-cruise-control.md
ShubhamRwt: Yes, its always intra-broker rebalancing.

scholzj: Ok, thanks. Maybe its worth clarifying it in the proposal and in the future docs?

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,85 @@
...
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaRebalance
+metadata:
+  name: my-rebalance
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  mode: remove_disks
```
ppatierno, line 52: let's use `remove-disks` as for the other modes.

076-moving-data-between-two-jbod-disks-using-cruise-control.md
ppatierno: I would use something similar to the CC API (it has the `brokerid_and_logdirs` param) so maybe `brokersAndVolumes` as Jakub suggested is better than `brokerAndVolumeIds`.
Using already existing `brokers` list would add a limitation which CC doesn't have.
If we leave this API with, I guess the `volumeId` should be `volumeIds` instead being an array. AFAICS from the CC endpoint, you can list more log dirs on the same broker (i.e. `101-/tmp/kafka-logs-1,101-/tmp/kafka-logs-2`) so you could have something like this:

```yaml
brokersAndVolumes
- brokerId:101
  volumeIds: [1, 2]
- brokerId:102
  volumeIds: [0]
```

ShubhamRwt: Yeah, that sounds good. I will update POC and the proposal with this change

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,85 @@
...
+      volumeIds: [1, 2]
+    - brokerId: 2
+      volumeIds: [1]
+# ...
+```
+
+### Flow
+
+- The user should be using the `Kafka` resource with JBOD configured. Make sure that you have more than one disk configured on the brokers.
+- When the Kafka cluster is ready, the user creates a `KafkaRebalance` custom resource with the `spec.mode` field as `remove_disks` and the list of the broker and the corresponding logdirs to move in the `spec.brokerandlogdirs` field.
```
ppatierno, line 64:
```suggestion
- When the Kafka cluster is ready, the user creates a `KafkaRebalance` custom resource with the `spec.mode` field as `remove-disks` and the list of the broker and the corresponding logdirs to move in the `spec.brokersAndVolumes` field.
```

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,85 @@
...
+remainingCapacity = sum of capacities of the remaining disks
+errorMargin = configurable property (default 0.1); it makes sure that a disk percentage is always free when moving replicas
+```
+
+In order to use the `remove_disks` endpoint in the Strimzi cluster operator, it would be added to the [`CruiseControlApi`](https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApi.java) interface and developing the corresponding implementation.
+
+### Implementation
+
+To implement this feature, we will be adding a new mode to the `KafkaRebalanceMode` class.
+* `remove_disks`: It moves replicas from a specified disk to other disks of the same broker. It always uses intra-broker re-balancing.
```
ppatierno, line 39:
```suggestion
* `remove-disks`: It moves replicas from a specified disk to other disks of the same broker. It always uses intra-broker re-balancing.
```

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,85 @@
...
+You can use this mode by changing the `spec.mode` to `remove_disks` in the `KafkaRebalance` resource.
```
ppatierno, line 40:
```suggestion
You can use this mode by changing the `spec.mode` to `remove-disks` in the `KafkaRebalance` resource.
```

ppatierno (changes requested):

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,85 @@
...
+
+We should introduce the logic to Strimzi to leverage Cruise Control integration and make it possible to move the data between two JBOD disks.
+This feature will also allow us to remove the disks without the loss of data.
+
+## Proposal
+
+Cruise Control provides `remove_disks` HTTP REST endpoint to move replicas from a specified disk to other disks of the same broker. It always uses intra-broker re-balancing.
+This endpoint triggers a rebalancing operation by moving replicas in a size-based manner to the remaining disks, from the largest to the smallest, while checking the following constraint:
+```sh
+1 - (remainingUsageAfterRemoval / remainingCapacity) > errorMargin
```
scholzj, line 25: Out of curiosity ... Does Cruise Control take into account the sizes of individual partitions and the free space on individual disks? E.g. what happens in a situation when:
* I have broker with 3 disks, each 100Gi
* The disks 1 and 2 have each 1Gi of free space
* The 3rd disk hosts only one partition with 1,5Gi in size
* User requests to remove the 3rd disk
* => The overall free capacity is 2Gi, so the partition fits to that. But because each of the remaining two disks has only 1Gi of free space, it cannot be moved safely. So what happens? Does Cruise Control handle it and fail?

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,85 @@
...
+
+Note: The optimization proposal will not show the load before optimization, it will only show the load after optimization. This is because in upstream Cruise Control, we don't have the verbose tag enabled with the `remove_disks` endpoint.
+
+### Other Scenarios
+
+- In case the user is not using JBOD storage and tries to generate the optimization proposal, the `KafkaRebalance` resource will move to `NotReady` state prompting invalid log dirs provided for the broker.
+- If you are using JBOD with single disk configured on the brokers, in that case `KafkaRebalance` will move to `NotReady` state prompting that you don't have enough log dirs to move the repicas to for that broker.
+- If the disk capacity has exceeded for the broker, in that case `KafkaRebalance` will move to `NotReady` prompting that enough capacity is not remaining to move replicas for that broker.
+- This feature works fine with KafkaNodePools. 
+- This feature work with Kraft only if Kafka version is greater than 3.7.0, as that version supports multiple JBOD disks on brokers.
```
scholzj, line 77: Will these errors be based on an error reported by Cruise Control? Or will our code need to handle this?

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,85 @@
...
+  name: my-rebalance
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  mode: remove-disks
+  brokersAndVolumes:
+    - brokerId: 0
+      volumeIds: [1, 2]
+    - brokerId: 2
+      volumeIds: [1]
```
scholzj, line 57: Are all these combinations packed into a single Cruise Control request? Is there any limit for the number of items this list can have?

ppatierno: Yes they will go all together in the query parameter, so something like `/remove_disks?brokerid_and_logdirs=0-/tmp/kafka-logs-1,0-/tmp/kafka-logs-2,2=-/tmp/kafka-logs-1`. From the CC code I didn't find any limit. It just parses a string using `,` as separator across brokers/volumes and `-` between broker id and log dir.

scholzj: I gues the URL size will have some limit?

ppatierno: Good point. I didn't find anything useful in the official HTTP 1.1 RFC. I just found different length for different browsers which of course we don't use. From 2000 up to more and more.
Are you wondering if we should put a limit on how many brokers and volumes could be involved in order to limit that query parameter length?

scholzj: Anyway, I guess if it has a separate item for each brokerID-logDir pair it doesn't matter as we cannot change that. I was mostly wondering if it would make sense to have a list of brokers or range for the `brokerId`. But it does not seem so. Thanks for the explanation @ppatierno

076-moving-data-between-two-jbod-disks-using-cruise-control.md
ShubhamRwt: These errors are based on errors reported by Cruise Control

scholzj: You should make it clear in the proposal then who is responsible for the validation.

076-moving-data-between-two-jbod-disks-using-cruise-control.md
ShubhamRwt: Yes, looking at the code 
```
            double futureUsage = 0.0;
            double remainingCapacity = 0.0;
            for (Disk disk : broker.disks()) {
                futureUsage += disk.utilization();
                if (!logDirsToRemove.contains(disk.logDir())) {
                    remainingCapacity += disk.capacity();
                }
            }
            if (futureUsage / remainingCapacity > capacityThreshold) {
                LOG.error("Not enough remaining capacity to move replicas to for broker {}.", brokerId);
                throw new IllegalArgumentException(String.format("Not enough remaining capacity to move replicas to for broker %d.", brokerId));
            }
        }
```
If we exceed the capacity threshold then in that case, Cruise control will report not enough capacity

scholzj: But having multiple disks cannot be only dealt with through the threshold, or? Or what if the threshold is not big enough to detect the issue? In an extreme case, you can have something like this:
* Three 100Gi disks
* Three partitions, each 51Gi big, one on each disk
* What happens when you try to remove one of the disks? 
* You have 98Gi of free disk space on the remaining two disks. So the partitions from the third disk should seemingly fit in easily. But it will not.
 
Does Cruise Control give you a reasonable error and doesn't free the disk?

im-konge (approved):

076-moving-data-between-two-jbod-disks-using-cruise-control.md
ShubhamRwt: I do see a check which makes sure that we are violating the disk capacity 
```

Comment (ShubhamRwt) [reply 076-moving-data-between-two-jbod-disks-using-cruise-control.md:25]: I do see a check which makes sure that we are not violating the disk capacity while doing intra broker rebalancing
```
  /**
   * Check whether the movement of source replica to destination disk is acceptable for this goal.
   *
   * @param sourceReplica   Source replica.
   * @param destinationDisk Destination disk.
   * @return {@code true} if movement is acceptable for this goal, {@code false} otherwise.
   */
  private boolean isMovementAcceptableForCapacity(Replica sourceReplica, Disk destinationDisk) {
    double replicaUtilization = sourceReplica.load().expectedUtilizationFor(RESOURCE);
    return isUtilizationUnderLimitAfterAddingLoad(destinationDisk, replicaUtilization);
  }
```

Comment (ShubhamRwt) [reply 076-moving-data-between-two-jbod-disks-using-cruise-control.md:25]: I n case the disk utilization is over limit we have this code handling the errors -> 
```
      for (Disk disk : broker.disks()) {
        if (disk.isAlive() && isUtilizationOverLimit(disk)) {
          // The utilization of the host for the resource is over the capacity limit.
          double requiredCapacity = disk.utilization() / _balancingConstraint.capacityThreshold(RESOURCE);
          ProvisionRecommendation recommendation = new ProvisionRecommendation.Builder(ProvisionStatus.UNDER_PROVISIONED)
              .numDisks(1).totalCapacity(requiredCapacity).build();
          throw new OptimizationFailureException(String.format("[%s] Utilization (%.2f) for disk %s on broker %d is above capacity limit.",
                                                               name(), disk.utilization(), disk, broker.id()), recommendation);
        }
        if (_shouldEmptyZeroCapacityDisks && disk.capacity() == 0 && !disk.replicas().isEmpty()) {
          throw new OptimizationFailureException(String.format("[%s] Could not move all replicas from disk %s on broker %d",
                                                               name(), disk.logDir(), broker.id()));
        }
      }
 ```

scholzj: Is that a yes or no for my question?

ShubhamRwt: From what I understood from the code, Yes

scholzj: Great, thanks.

076-moving-data-between-two-jbod-disks-using-cruise-control.md
scholzj: I think this still needs to be updated?

ShubhamRwt: Yes, on it.

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -76,6 +76,8 @@ Note: The optimization proposal will not show the load before optimization, it w
 - This feature works fine with KafkaNodePools. 
 - This feature work with Kraft only if Kafka version is greater than 3.7.0, as that version supports multiple JBOD disks on brokers.
 
+The errors in the above scenarios are reported by Cruise Control. Based on those we move the `KafkaRebalance` resource to `NotReady` state and update the `KafkaRebalance` status with the corresponding error message.
```
ShubhamRwt, line 79: Does this sound good @scholzj?

scholzj (approved):
Thanks.

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,87 @@
...
+kind: KafkaRebalance
+metadata:
+  name: my-rebalance
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  mode: remove-disks
+  brokersAndVolumes:
+    - brokerId: 0
+      volumeIds: [1, 2]
```
tombentley, line 55: In conjunction with the `mode: remove-disks` it's not clear whether this should be interpreted as 
* "I want broker 0 to end up with just volumes 1 and 2; remove any others". Resources in Kubernetes are normally declarative, so this its a reasonable interpretation for someone who doesn't know the semantics of the resource.
* "I want to remove volumes 1 and 2 from broker 0". `KafkaRebalance` is intrinsically imperative/specifying an action, so this its a reasonable interpretation for someone who know that.

And actually both those interpretations are wrong! The 2nd is closer, but it won't actually remove the volumes (in the kube sense), it just moves the data.

Perhaps renaming `brokersAndVolumes` to something like `moveReplicasOffVolumes` (or `movingReplicasOffVolumes` or some alternative) would make that clearer. I'm not sure we need to include the word broker or even volume, since the items in the array make clear we're talking about volumes within brokers.

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,87 @@
...
+    - brokerId: 2
+      volumeIds: [1]
+# ...
+```
+
+### Flow
+
+- The user should be using the `Kafka` resource with JBOD configured. Make sure that you have more than one disk configured on the brokers.
+- When the Kafka cluster is ready, the user creates a `KafkaRebalance` custom resource with the `spec.mode` field as `remove-disks` and the list of the brokers and the corresponding volumes to move in the `spec.brokersAndVolumes` field.
```
tombentley, line 64: "volumes to move" is really confusing. The volumes don't move! The replicas on them do.

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,87 @@
...
+
+Note: The optimization proposal will not show the load before optimization, it will only show the load after optimization. This is because in upstream Cruise Control, we don't have the verbose tag enabled with the `remove_disks` endpoint.
+
+### Other Scenarios
+
+- In case the user is not using JBOD storage and tries to generate the optimization proposal, the `KafkaRebalance` resource will move to `NotReady` state prompting invalid log dirs provided for the broker.
+- If you are using JBOD with single disk configured on the brokers, in that case `KafkaRebalance` will move to `NotReady` state prompting that you don't have enough log dirs to move the repicas to for that broker.
+- If the disk capacity has exceeded for the broker, in that case `KafkaRebalance` will move to `NotReady` prompting that enough capacity is not remaining to move replicas for that broker.
+- This feature works fine with KafkaNodePools. 
+- This feature work with Kraft only if Kafka version is greater than 3.7.0, as that version supports multiple JBOD disks on brokers.
```
tombentley, line 77: In the above description you said " When the Kafka cluster is ready", so I need closure on the scenario where the Kafka cluster is not initially ready.

tombentley (commented):
Thanks @ShubhamRwt, this looks pretty good, but I had a few questions about some minor points.

076-moving-data-between-two-jbod-disks-using-cruise-control.md
scholzj: I don't think this is completely wrong. But I think we also need to keep in mind the existing parts of the API that already exist in this way for a long time. So I wonder if we should prefer inconsistency over something what might not be completely clear. I personally would probably prefer consistency.

tombentley: I agree consistency is a good thing. But good APIs are also a good thing. I wonder if we should take a moment to consider what other CC features we might want to integrate into this CR and how well they they'd fit with `brokersAndVolumes`. Otherwise I can see that the consistency argument might be used to justify an even more distasteful compromise in the future.

Frawless (approved):

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,87 @@
...
+- You can use `strimzi.io/rebalance-auto-approval:true` annotation on the `KafkaRebalance` resource for auto-approval of proposal. In case you want to do it manually you can do it by applying the `strimzi.io/rebalance=approve` annotation on it.
+- The `KafkaRebalanceAssemblyOperator` interacts with Cruise Control via the `/remove_disks` endpoint to perform the actual rebalancing.
+
+Note: The optimization proposal will not show the load before optimization, it will only show the load after optimization. This is because in upstream Cruise Control, we don't have the verbose tag enabled with the `remove_disks` endpoint.
+
+### Other Scenarios
+
+- In case the user is not using JBOD storage and tries to generate the optimization proposal, the `KafkaRebalance` resource will move to `NotReady` state prompting invalid log dirs provided for the broker.
+- If you are using JBOD with single disk configured on the brokers, in that case `KafkaRebalance` will move to `NotReady` state prompting that you don't have enough log dirs to move the repicas to for that broker.
```
kyguy, line 74:
```suggestion
- If you are using JBOD with single disk configured on the brokers, in that case `KafkaRebalance` will move to `NotReady` state prompting that you don't have enough log dirs to move the replicas for that broker.
```

076-moving-data-between-two-jbod-disks-using-cruise-control.md
kyguy, line 77:
```suggestion
- This feature works with Kraft only if Kafka version is greater than 3.7.0, as that version supports multiple JBOD disks on brokers.
```

kyguy (approved):
Looks great Shubham, awesome work

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,87 @@
...
+A `KafkaRebalance` custom resource would look like this.
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaRebalance
+metadata:
+  name: my-rebalance
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
```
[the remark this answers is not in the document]
ShubhamRwt: Sure

076-moving-data-between-two-jbod-disks-using-cruise-control.md
ShubhamRwt: When the Kafka Cluster is not ready, we have a check in place in the `KafkaRebalanceAssemblyOperator` which will move the `KafkaRebalance` resource to `NotReady` prompting the reason  as `Kafka Cluster is not ready` in the `status` field of the rebalance resource. So basically its a pre-check which needs to be met.

076-moving-data-between-two-jbod-disks-using-cruise-control.md
ShubhamRwt: I will fix this

076-moving-data-between-two-jbod-disks-using-cruise-control.md
ShubhamRwt: Okay, I will change the name to `moveReplicasOffVolumes`, that sound good to me

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,89 @@
...
+      volumeIds: [1, 2]
+    - brokerId: 2
+      volumeIds: [1]
+# ...
+```
+
+### Flow
+
+- The user should be using the `Kafka` resource with JBOD configured. Make sure that you have more than one disk configured on the brokers.
+- When the Kafka cluster is ready, the user creates a `KafkaRebalance` custom resource with the `spec.mode` field as `remove-disks` and provides list of the brokers, and the corresponding volumes from which you want to move the replicas in the `spec.moveReplicasOffVolumes` field.
```
ppatierno, line 66: what happens if the user set the mode but forgets to set the `moveReplicasOffVolumes` field? I guess the operator should check it and put the resource in `NotReady` with a proper condition warning. Can we make it clear here?

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,89 @@
...
+- The `KafkaRebalanceAssemblyOperator` interacts with Cruise Control via the `/remove_disks` endpoint to perform the actual rebalancing.
+
+Note: The optimization proposal will not show the load before optimization, it will only show the load after optimization. This is because in upstream Cruise Control, we don't have the verbose tag enabled with the `remove_disks` endpoint.
+
+### Other Scenarios
+
+- In case the user is not using JBOD storage and tries to generate the optimization proposal, the `KafkaRebalance` resource will move to `NotReady` state prompting invalid log dirs provided for the broker.
+- If you are using JBOD with single disk configured on the brokers, in that case `KafkaRebalance` will move to `NotReady` state prompting that you don't have enough log dirs to move the replicas for that broker.
+- If the disk capacity has exceeded for the broker, in that case `KafkaRebalance` will move to `NotReady` prompting that enough capacity is not remaining to move replicas for that broker.
+- This feature works fine with KafkaNodePools.
```
ppatierno, line 78: Do we need this sentence? What does it add/mean? We have everything now moving to node pools (mostly because KRaft) so I would expect that any new features will work fine with node pools as explicit.

ShubhamRwt: There was earlier a commet by @scholzj, if it will work fine with nodepools and KRaft, therefore I thought it would be good to mention about all of them. I agree, any new feature should work with nodepool but just for clarity maybe we can still have this pointer.

076-moving-data-between-two-jbod-disks-using-cruise-control.md
ShubhamRwt: Sure, I can add an extra sentence saying
```
In case, the user forgets to set the `spec.moveReplicasOffVolumes` field, then the `KafkaRebalance` resource will move to `NotReady` state prompting that `spec.moveReplicasOffVolumes` field is missing`
```
Does this sound good?

ppatierno: I would not say "user forgets", just that the field is missing :-)

ShubhamRwt:
@PaulRMellor @im-konge @scholzj @tombentley @Frawless Hi, It would be great to get your reviews again after my last update on the proposal

Frawless (approved):
I am not SME, but LGTM.

im-konge (approved):
I'm not SME, but LGTM :) thanks

076-moving-data-between-two-jbod-disks-using-cruise-control.md
ShubhamRwt: I added the extra sentence

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,89 @@
...
+1 - (remainingUsageAfterRemoval / remainingCapacity) > errorMargin
+```
+where:
+```sh
+remainingUsageAfterRemoval = current usage for remaining disks + additional usage from removed disks
+remainingCapacity = sum of capacities of the remaining disks
+errorMargin = configurable property (default 0.1); it makes sure that a disk percentage is always free when moving replicas
+```
+
+In order to use the `remove_disks` endpoint in the Strimzi cluster operator, it would be added to the [`CruiseControlApi`](https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApi.java) interface and developing the corresponding implementation.
```
PaulRMellor, line 34:
```suggestion
To use the `remove_disks` endpoint in the Strimzi cluster operator, it should be added to the [`CruiseControlApi`](https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/resource/cruisecontrol/CruiseControlApi.java) interface, and the corresponding implementation developed.
```

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,89 @@
...
+    - brokerId: 0
+      volumeIds: [1, 2]
+    - brokerId: 2
+      volumeIds: [1]
+# ...
+```
+
+### Flow
+
+- The user should be using the `Kafka` resource with JBOD configured. Make sure that you have more than one disk configured on the brokers.
```
PaulRMellor, line 65:
```suggestion
- The user should be using the `Kafka` resource with JBOD configured, making sure that they have more than one disk configured on the brokers.
```

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,89 @@
...
+- When the Kafka cluster is ready, the user creates a `KafkaRebalance` custom resource with the `spec.mode` field as `remove-disks` and provides list of the brokers, and the corresponding volumes from which you want to move the replicas in the `spec.moveReplicasOffVolumes` field. In case, the `spec.moveReplicasOffVolumes` field is not set, then the `KafkaRebalance` resource will move to `NotReady` state prompting that `spec.moveReplicasOffVolumes` field is missing.
```
PaulRMellor, line 66:
```suggestion
- When the Kafka cluster is ready, the user creates a `KafkaRebalance` custom resource with the `spec.mode` field as `remove-disks` and provides a list of the brokers, and the corresponding volumes from which they want to move the replicas in the `spec.moveReplicasOffVolumes` field. In case, the `spec.moveReplicasOffVolumes` field is not set, then the `KafkaRebalance` resource will move to `NotReady` state prompting that `spec.moveReplicasOffVolumes` field is missing.
```

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,89 @@
...
+- The `KafkaRebalanceAssemblyOperator` interacts with Cruise Control via the `/remove_disks` endpoint to generate an optimization proposal (by using the dryrun feature).
+- You can use `strimzi.io/rebalance-auto-approval:true` annotation on the `KafkaRebalance` resource for auto-approval of proposal. In case you want to do it manually you can do it by applying the `strimzi.io/rebalance=approve` annotation on it.
+- The `KafkaRebalanceAssemblyOperator` interacts with Cruise Control via the `/remove_disks` endpoint to perform the actual rebalancing.
+
+Note: The optimization proposal will not show the load before optimization, it will only show the load after optimization. This is because in upstream Cruise Control, we don't have the verbose tag enabled with the `remove_disks` endpoint.
```
PaulRMellor, line 71:
```suggestion
> **NOTE** The optimization proposal will not show the load before optimization, it will only show the load after optimization. This is because in upstream Cruise Control we don't have the verbose tag enabled with the `remove_disks` endpoint.
```

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Other Scenarios
+
+- In case the user is not using JBOD storage and tries to generate the optimization proposal, the `KafkaRebalance` resource will move to `NotReady` state prompting invalid log dirs provided for the broker.
+- If you are using JBOD with single disk configured on the brokers, in that case `KafkaRebalance` will move to `NotReady` state prompting that you don't have enough log dirs to move the replicas for that broker.
+- If the disk capacity has exceeded for the broker, in that case `KafkaRebalance` will move to `NotReady` prompting that enough capacity is not remaining to move replicas for that broker.
+- This feature works fine with KafkaNodePools.
```
PaulRMellor, line 78:
```suggestion
- This feature works fine with `KafkaNodePool` resources. 
```

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,89 @@
...
+
+Note: The optimization proposal will not show the load before optimization, it will only show the load after optimization. This is because in upstream Cruise Control, we don't have the verbose tag enabled with the `remove_disks` endpoint.
+
+### Other Scenarios
+
+- In case the user is not using JBOD storage and tries to generate the optimization proposal, the `KafkaRebalance` resource will move to `NotReady` state prompting invalid log dirs provided for the broker.
+- If you are using JBOD with single disk configured on the brokers, in that case `KafkaRebalance` will move to `NotReady` state prompting that you don't have enough log dirs to move the replicas for that broker.
+- If the disk capacity has exceeded for the broker, in that case `KafkaRebalance` will move to `NotReady` prompting that enough capacity is not remaining to move replicas for that broker.
+- This feature works fine with KafkaNodePools. 
+- This feature works with Kraft only if Kafka version is greater than 3.7.0, as that version supports multiple JBOD disks on brokers.
```
PaulRMellor, line 79:
```suggestion
- This feature works with KRaft only if Kafka version is greater than 3.7.0, as that version supports multiple JBOD disks on brokers.
```

076-moving-data-between-two-jbod-disks-using-cruise-control.md
```diff
@@ -0,0 +1,89 @@
...
+
+The errors in the above scenarios are reported by Cruise Control. Based on those we move the `KafkaRebalance` resource to `NotReady` state and update the `KafkaRebalance` status with the corresponding error message.
```
PaulRMellor, line 81:
```suggestion
Errors for these scenarios are reported by Cruise Control. 
Based on these errors, we transition the `KafkaRebalance` resource to the `NotReady` state and update its status with the corresponding error message.
```

tombentley (approved):
LGTM, thanks @ShubhamRwt

scholzj (approved):
+1. Thanks @ShubhamRwt

scholzj:
This proposal now has 6 binding and 2 non-binding +1 votes. If there are no new comments until the end of the week, we will merge it as approved.

see-quick (approved):
LGTM, thanks! 👍

scholzj:
Approved with 7 binding and 2 non-biding +1 votes.

PR #110, opened 2024-03-05
Add proposal for passing ACLs in JWT

iblutrifork:
Proposal for an alternative Authorizer, which reads ACLs from the JWT.
Discussion on the topic: https://github.com/strimzi/strimzi-kafka-oauth/issues/223

### DIFF
--- a/068-access-policies-in-jwt.md
+++ b/068-access-policies-in-jwt.md
@@ -17,8 +17,12 @@ and Keycloak authorizer.
 It also supports custom authorizers, which can be used to point to the implementation of a custom authorizer class.
 There is currently no inbuilt support for JWT-based authorization.
 If a user wants to use JWT-based authorization, they have to implement their own authorizer and use the custom authorizer feature.
+As the documentation states:
+
+> In addition to the Kafka custom resource configuration, the JAR file containing the custom authorizer class along with its dependencies must be available on the classpath of the Kafka broker.
+
 This proposal aims to add the JWT-based authorizer to Strimzi, so it can be selected in the configuration as
-another option apart from the simple and keycloak ones.
+another option apart from the simple and Keycloak ones, without having to implement a custom authorizer.
 
 ## Motivation
 
--- a/068-access-policies-in-jwt.md
+++ b/068-access-policies-in-jwt.md
@@ -72,7 +72,7 @@ e.g. for a list of strings:
 and for a string:
 ```json
 {
-  "acls": "cluster_x:topic_1:read,topic2:write"
+  "acls": "cluster_x:topic_1:read,:topic2:write"
 }
 ```
 
--- a/068-access-policies-in-jwt.md
+++ b/068-access-policies-in-jwt.md
@@ -169,7 +169,7 @@ permissions.
 
 10s of permissions in the JWT would be very little overhead in terms of computation.
 The main overhead is the size of the JWT.
-If the average topic length is 20 characters, with 10 ACLs the overhead would be 200-300 bytes.
+If the average topic name length is 20 characters, with 10 ACLs the overhead would be 200-300 bytes.
 
 Passing 100s of permissions would be too much of an overhead, so it is advised against.
 Passing 1000s of permissions is not feasible.
--- a/068-access-policies-in-jwt.md
+++ b/068-access-policies-in-jwt.md
@@ -110,9 +110,8 @@ If the ACLs are passed as a string, they are separated by `,`.
 i.e., `CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS,CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS`
 
 - `RESOURCE_TYPE` can be `topic` or `group`, or the shortened versions `t` and `g`.
-- `PERMITTED_ACTIONS` is a list of a subset of `read`, `write`, `create`, `delete`, `alter`, `alter_configs`, `describe`, `describe_configs`, `cluster_action`, `all`, or the shortened versions `r`, `w`, `c`, `d`, `a`, `ac`, `de`, `dc`, `ca`. The short version for `all` is `*`.
-The list items are separated by `|`.
-The choice of the separator `|` is based on the typical pattern of OR-ing boolean permissions.
+- `PERMITTED_ACTIONS` is a list of a subset of `read`, `write`, `create`, `delete`, `alter`, `describe`, `cluster_action`, `describe_configs`, `alter_configs`, `idempotent_write`, `create_tokens`, `describe_tokens`, `all`, or the shortened versions `r`, `w`, `c`, `d`, `a`, `de`, `ca`, `dc`, `ac`, `iw`, `ct`, `dt`. The short version for `all` is `*`.
+The list items are separated by a `+`.
 - `CLUSTER_NAME` is the name of the cluster, and `RESOURCE_SPEC` is the name of the resource (topic/group).
 These fields can start or end with `*` to match any prefix/suffix.
 
@@ -128,11 +127,11 @@ The default for `PERMITTED_ACTIONS` is none to prevent accidental access to all 
 
 ### Examples of ACLs
 
-- `my_cluster:t:topic1:r|w` - allows reading and writing to `topic1` in the cluster `my_cluster`.
+- `my_cluster:t:topic1:r+w` - allows reading and writing to `topic1` in the cluster `my_cluster`.
 - `:::` - denies all actions on all topics in all clusters.
 - `:::*` - allows all actions on all topics in all clusters.
 - `my_cluster:group:*_app2:read` - allows reading from all groups ending with `_app2` in the cluster `my_cluster`.
-- `::edge_*:write|r` - allows writing and reading topics starting with `edge_` in all clusters.
+- `::edge_*:write+r` - allows writing and reading topics starting with `edge_` in all clusters.
 
 ### Implementation details
 
--- a/068-access-policies-in-jwt.md
+++ b/068-access-policies-in-jwt.md
@@ -1,0 +1,153 @@
+# Pass Access Policies (ACLs) in JWT
+
+This proposal addresses the possibility of passing access policies in the JWT tokens.
+
+## Current situation
+
+Currently, authorization is supported via [Keycloak](https://www.keycloak.org/).
+Keycloak must be deployed, and the identities must be managed there.
+This makes using Strimzi in larger systems more complicated.
+Most users use their own identity management systems.
+In order to integrate it with Strimzi, they would have to duplicate the identities in Keycloak.
+This solution doesn't follow the single source of truth architecture.
+It also requires to run a redundant service - Keycloak.
+
+## Motivation
+
+Having a generic solution is always preferable as it allows for less effort to be maintained.
+While we cannot achieve a perfect generic solution due to the nature of OAuth2, and how different
+identity providers are implemented, we can strive to create a solution as generic as possible.
+
+Passing additional data in the JWT seems to be supported by the majority of identity providers.
+Therefore, we can use this to pass the access policies in the JWT.
+This solution will work in most cases and with most identity providers, however it will not
+scale to infinity.
+It will, however, be decoupled from Keycloak and therefore be more generic.
+
+## Proposal
+
+Identity providers support passing custom fields in the JWT token.
+Some providers allow the data type to be string, while others allow a list of strings.
+There are not many (if any) identity providers that allow a list of objects.
+
+Therefore, in this proposal we allow the ACLs to be passed as either.
+e.g. for a list of strings:
+```json
+{
+  "<acls>": [
+    "cluster_x:topic_1:read",
+    ":topic2:write"
+  ]
+}
+```
+
+and for a string:
+```json
+{
+  "<acls>": "cluster_x:topic_1:read,topic2:write"
+}
+```
+
+The field name is configurable with default value `acls`.
+
+The new authorizer will be implemented as a new class, extending the
+[AclAuthorizer](https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/security/authorizer/AclAuthorizer.scala).
+
+```java
+public class JWTKafkaAuthorizer extends AclAuthorizer {
+    @Override
+    public void configure(Map<String, ?> configs) {
+        // ... parse acls
+    }
+
+    @Override public List<AuthorizationResult> authorize(AuthorizableRequestContext requestContext, List<Action> actions) {
+        // ... get JWT from requestContext
+        // for each access request
+        //   add to the result whether the topic or cluster claims allow the access request
+        // return the result
+    }
+}
+```
+
+Operations are described in the [Confluent docs](https://docs.confluent.io/platform/current/kafka/authorization.html#operations).
+Allowing an operation will follow the same logic as in [AclEntry.supportedOperations](https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/security/authorizer/AclEntry.scala#L99).
+
+The above code will be supplied with unit tests and documentation explaining how to set up Strimzi with
+the new authorizer using Azure Active Directory B2C as an example.
+The setup will be similar for other identity providers, with the main difference being the configuration of the identity provider itself and the `<acls>` field.
+
+## Analysis of more popular identity providers
+
+Adding custom attributes/roles/groups to the token works as follows in the different auth solutions:
+
+In Azure Active Directory B2C:
+
+The attribute can be a string/list of strings.
+
+- https://learn.microsoft.com/en-us/azure/active-directory-b2c/user-flow-custom-attributes?pivots=b2c-user-flow
+- https://learn.microsoft.com/en-us/azure/active-directory-b2c/client-credentials-grant-flow?pivots=b2c-user-flow
+
+In AWS Cognito:
+
+The attribute type can be string/number/bool so probably not an object.
+Also, the roles might be fixed to have the aws prefixes.
+Supports a list of objects and strings.
+Might need more configurable field name/parsing nested structures.
+
+- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AddCustomAttributes.html
+- https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_SchemaAttributeType.html
+- https://docs.aws.amazon.com/cognito/latest/developerguide/role-based-access-control.html
+
+In OneLogin:
+
+Does not currently support the field "values", which is an array. Supports adding roles as a semicolon separated list.
+
+- https://developers.onelogin.com/api-docs/2/api-authorization/list-claims
+- https://developers.onelogin.com/api-docs/1/users/set-custom-attribute
+- https://developers.onelogin.com/api-docs/2/api-authorization/add-claim
+
+In Google Firebase:
+
+Custom claims are of the form "claim": true, so not just a list of strings, but a list of attributes. I guess this could be used to model the allow/deny kafka rules, or use the same custom string format as for every other auth solution and ignore the true/false value.
+
+- https://firebase.google.com/docs/auth/admin/create-custom-tokens
+- https://firebase.google.com/docs/auth/admin/custom-claims
+
+In Auth0:
+
+Can use custom claims as roles/groups.
+event.authorization has a field roles which is a list of permissions for the user.
+It supports an array of strings.
+
+- https://auth0.com/docs/get-started/authentication-and-authorization-flow/client-credentials-flow/call-your-api-using-the-client-credentials-flow#sample-use-cases
+- https://auth0.com/docs/customize/actions/flows-and-triggers/machine-to-machine-flow#m2m-client-credentials
+- https://auth0.com/docs/get-started/apis/scopes/sample-use-cases-scopes-and-claims
+- https://auth0.com/docs/customize/actions/flows-and-triggers/login-flow/event-object
+
+In Okta:
+
+Group claims could represent roles.
+Supports a list of strings
+
+- https://developer.okta.com/docs/guides/customize-tokens-returned-from-okta/main/
+- https://developer.okta.com/docs/guides/customize-tokens-groups-claim/main/
+
+## Affected/not affected projects
+
+This proposal will not affect any other projects.
+It will be added as a new feature, an alternative implementation of the Keycloak Authorizer.
+
+## Compatibility
+
+This proposal is backward compatible and will not affect any existing functionality.
+
+## Rejected alternatives
+
+There do not currently exist alternatives apart from the Keycloak Authorizer.
+Using another OAuth2 provider is not possible without implementing a new authorizer,
+which is tightly coupled with said provider.
+While that is an alternative, we either have to let every user implement their own authorizer,
+or we have to implement a new authorizer for every OAuth2 provider.
+This is not feasible.
+Using more generic solutions, such as OpenFGA, overcomplicates the approach to the problem.
+Such solutions would need to be implemented on top of OAuth2, while OAuth2 works by default with JWT.
--- a/068-access-policies-in-jwt.md
+++ b/068-access-policies-in-jwt.md
@@ -78,6 +78,29 @@ and for a string:
 
 The field name is configurable with default value `acls`.
 
+### Using roles to pass ACLs
+
+Sometimes the identity provider supports very fine customization of the JWT,
+but this is not always the case.
+However, pretty much all identity providers support the concept of roles or claims.
+Roles are often used to define the access level of a user or service.
+For example, the database service is given the `admin` role, while the frontend service is given the `read` role. 
+In the application layer, the role is checked to see if the user has the necessary permissions when the user tries to perform an action.
+
+We can use fine-grained roles to represent ACLs.
+Instead of naming the role `admin`, which implicitly implies all permissions, we can name the role `*:topic:*:*` or if we want to limit the permissions - `*:topic:my_topic:read`.
+This way we are using the name of the role to encode the permissions the user has,
+instead of using a generic name such as `admin` and having the application layer understand what `admin` means.
+
+As an example we can define the roles for a service A in the IDP to be:
+```json
+"roles": [
+  "*:topic:input_topic:read",
+  "*:topic:output_topic:write",
+],
+```
+Now when these "roles" are passed in the JWT, we can use them as Kafka ACLs in order to authorize service A.
+
 ### ACL Syntax
 
 A single ACL has the syntax `CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS`.
--- a/068-access-policies-in-jwt.md
+++ b/068-access-policies-in-jwt.md
@@ -91,6 +91,21 @@ The above code will be supplied with unit tests and documentation explaining how
 the new authorizer using Azure Active Directory B2C as an example.
 The setup will be similar for other identity providers, with the main difference being the configuration of the identity provider itself and the `acls` field.
 
+## Scalability
+
+This solution is intended for small to medium-sized clients.
+In practice the permissions that are granted to a client are usually at most a few dozen.
+In most cases in microservices, even less than ten as the client is usually designed to perform a single task,
+i.e., read from a single topic (or a few) and write to a single topic.
+If a client has hundreds of permissions it is likely that architecture is flawed or
+the client performs too many tasks, which is again a sign of a flawed architecture.
+
+10s of permissions in the JWT would be very little overhead.
+If the average topic length is 20 characters, with 10 ACLs the overhead would be 200-300 bytes.
+
+Passing 100s of permissions would be too much of an overhead, so it is advised against.
+Passing 1000s of permissions is not feasible.
+
 ## Analysis of more popular identity providers
 
 Adding custom attributes/roles/groups to the token works as follows in the different auth solutions:
--- a/068-access-policies-in-jwt.md
+++ b/068-access-policies-in-jwt.md
@@ -78,6 +78,41 @@ and for a string:
 
 The field name is configurable with default value `acls`.
 
+### ACL Syntax
+
+A single ACL has the syntax `CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS`.
+The separator `:` is chosen because it is a common separator in the Unix world.
+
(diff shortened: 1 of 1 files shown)


068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,153 @@
...
+    @Override public List<AuthorizationResult> authorize(AuthorizableRequestContext requestContext, List<Action> actions) {
+        // ... get JWT from requestContext
+        // for each access request
+        //   add to the result whether the topic or cluster claims allow the access request
+        // return the result
+    }
+}
+```
+
+Operations are described in the [Confluent docs](https://docs.confluent.io/platform/current/kafka/authorization.html#operations).
```
scholzj, line 72: You should link to Kafka docs as Confluent Platform is not the same as Apache Kafka.

scholzj (commented):
Thanks for the proposal. I think it would be great to include the following:
* Why can't you use the `type: custom` authorization with your own authorizer plugin for this?
* How will the API look like in the `Kafka` custom resource?
* What is the performance? What are the scalability limits?
* How will the code evaluating these rights look like / work?
* Are there any similar projects using this pattern?

You should also try to fix the DCO signoff -> the instructions should be inder the _Details_ link in the PR status section.

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,153 @@
...
+
+Identity providers support passing custom fields in the JWT token.
+Some providers allow the data type to be string, while others allow a list of strings.
+There are not many (if any) identity providers that allow a list of objects.
+
+Therefore, in this proposal we allow the ACLs to be passed as either.
+e.g. for a list of strings:
+```json
+{
+  "<acls>": [
```
mstruk, line 37: Wouldn't that make more sense as `"acls"` rather than `"<acls>"`?

iblutrifork: Yes, I originally had it `<acls>` to indicate it's a configurable parameter, but I mention this in the text, so I can change this to `acls`.

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,153 @@
...
+
+```java
+public class JWTKafkaAuthorizer extends AclAuthorizer {
+    @Override
+    public void configure(Map<String, ?> configs) {
+        // ... parse acls
+    }
+
+    @Override public List<AuthorizationResult> authorize(AuthorizableRequestContext requestContext, List<Action> actions) {
+        // ... get JWT from requestContext
```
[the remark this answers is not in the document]
iblutrifork: In the case where we want to pass the ACLs in the JWT, the ACLs should always be accessible via `getClaimsJSON()`.
I think by passing the ACLs in the JWT always, we don't need to take into account the two different flows.

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,183 @@
...
+The setup will be similar for other identity providers, with the main difference being the configuration of the identity provider itself and the `acls` field.
+
+## Scalability
+
+This solution is intended for small to medium-sized clients.
+In practice the permissions that are granted to a client are usually at most a few dozen.
+In most cases in microservices, even less than ten as the client is usually designed to perform a single task,
+i.e., read from a single topic (or a few) and write to a single topic.
+If a client has hundreds of permissions it is likely that architecture is flawed or
+the client performs too many tasks, which is again a sign of a flawed architecture.
```
scholzj, line 101: I don't think this is true. It is from my experience quite common to consume from many topics. You have also Kafka Streams applications that consume and produce many topics if they do some complicated stateful operations. So I do not think that is a flawed architecture.

iblutrifork:
Thanks for the comments!
I'm back working on this.

> Are there any similar projects using this pattern?

Unfortunately, I cannot link to projects that I know of, as they are not public. I could look through the public strimzi forks if that is a point that is necessary to be in the proposal.

> Why can't you use the type: custom authorization with your own authorizer plugin for this?

I am not sure if I understand this question and the next one fully, can you help me out?

> How will the API look like in the Kafka custom resource?

Do you mean I have to add something similar to:

```yaml
kind: Kafka
metadata:
  name: my-cluster
  namespace: myproject
spec:
  kafka:
    authorization:
      type: custom
      authorizerClass: io.mycompany.CustomAuthorizer
```

scholzj:
> Unfortunately, I cannot link to projects that I know of, as they are not public. I could look through the public strimzi forks if that is a point that is necessary to be in the proposal.

This is not about Strimzi forks or about any private projects. It is about other products or open-source projects that are using this kind of authorization pattern. To show that this is a generally recognized pattern used out there by others.

> I am not sure if I understand this question and the next one fully, can you help me out?

Strimzi supports users to bring and use custom authorizers using the `type: custom` in `.spec.kafka.authorization`. So you should explain why that is not an option here.

068-access-policies-in-jwt.md
iblutrifork: I have only once worked with consuming data from more than 20 topics, and it was not a nice experience.
Calling it flawed was my bad, I will remove the comment.

scholzj: To be honest, I think you should maybe remove the whole subjective section on how you think Kafka should be used and instead focus on the hard data of the actual scalability and limitations. I.e. things such as:
* With how many permissions is it expected to work
* With how many permissions does it break
* What is the performance and memory impact for each

Once the limits are clearly listed, people can decide on their own of how useful they are.

For me personally, I don't think 100s of permissions would be common. But 10s of them are definitely very common.

iblutrifork: I'll rework it, leaving only the numbers.

iblutrifork:
> Strimzi supports users to bring and use custom authorizers using the type: custom in .spec.kafka.authorization. So you should explain why that is not an option here.

I've added a couple of sentences on this.

scholzj:
> > Strimzi supports users to bring and use custom authorizers using the type: custom in .spec.kafka.authorization. So you should explain why that is not an option here.
> 
> I've added a couple of sentences on this.

I see it mentioned on line 17. But I do not see any explanation why you cannot use it.

iblutrifork:
> I see it mentioned on line 17. But I do not see any explanation why you cannot use it.

I have added a quote from the documentation, which repeats what the paragraph around line 17 says:

Using a custom authorizer is always an option, but is more work for users, instead of having the JWT authorizer built-in. Everyone who needs to interact with a different OAuth identity provider has to implement a custom authorizer.
This is the pain point this proposal is aiming to solve.

scholzj:
> > I see it mentioned on line 17. But I do not see any explanation why you cannot use it.
> 
> I have added a quote from the documentation, which repeats what the paragraph around line 17 says:
> 
> Using a custom authorizer is always an option, but is more work for users, instead of having the JWT authorizer built-in. 

Ok, so I take it as that nothing technical prevents users from using it as a custom authorizer.

> Everyone who needs to interact with a different OAuth identity provider has to implement a custom authorizer. This is the pain point this proposal is aiming to solve.

The proposal suggests that this will work with all kind of OAuth providers. So why do you now suggest that everyone would suddenly need to implement their own?

I don't really understand you. You are trying to convince us how great idea this is, how useful it is for everyone, how little maintenance cost it will have, etc. But you absolutely don't want to consider maintaining this yourself and insist that adding a JAR to the container image is too _cumbersome_ for anyone to use it. That does not really add up.

tombentley (commented):
@iblutrifork thanks for the proposal. I think what you're aiming for would be of benefit for people wanting to use their own existing IDP without having to resort to writing their own. 

You give an example policy of `cluster_x:topic_1:read,topic2:write`. I think it would be  good to describe in more detail the proposed syntax. There's a balance to be struck there between verbosity and clarity. Obviously it needs to cover all the resource types and ACL operations. There's some variation in what's supported for attributes and roles by the different IDPs mentioned, so it would be good to confirm that whatever is proposed is actually going to be practical with those, for example by showing how an example policy maps to the format/structure supported by each IDP.

I also didn't entirely understand how the IDP's support for roles was (or was not) intended to map to Kafka's model, which is not explicitly role-based. I think a statement and maybe an example there might go a long way.

Thanks again!

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,251 @@
...
+We have listed the most popular ones in the [Support for roles or claims in popular identity providers](#support-for-roles-or-claims-in-popular-identity-providers) section with links to the documentation
+on how custom claims or roles are supported.
+
+The concept of roles can be directly mapped to the concept of access policies in Kafka.
+
+## Proposal
+
+Identity providers support passing custom fields in the JWT token.
+Some providers allow the data type to be string, while others allow a list of strings.
+There are not many (if any) identity providers that allow a list of objects.
```
fvaleri, line 59: Are we going to support regex expressions for the topic name? I think this would help to support use cases where you have many topics with a common prefix (e.g. the app name).

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,251 @@
...
+## Scalability
+
+This solution is intended for small to medium-sized clients,
+that have a couple 10s of permissions.
+Since permissions would be passed as strings, the overhead would be proportional to the number and kind of
+permissions.
+
+10s of permissions in the JWT would be very little overhead in terms of computation.
+The main overhead is the size of the JWT.
+If the average topic length is 20 characters, with 10 ACLs the overhead would be 200-300 bytes.
```
fvaleri, line 172:
```suggestion
If the average topic name length is 20 characters, with 10 ACLs the overhead would be 200-300 bytes.
```

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,251 @@
...
+
+Passing 100s of permissions would be too much of an overhead, so it is advised against.
+Passing 1000s of permissions is not feasible.
```
fvaleri, line 175: Are we going to log warnings in these cases?

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,251 @@
...
+
+    // ...
+}
+```
+
+Operations are described in the [Kafka docs](https://kafka.apache.org/documentation/#operations_in_kafka).
+Allowing an operation will follow the same logic as in [AclEntry.supportedOperations](https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/security/authorizer/AclEntry.scala#L99).
+
+The above code will be supplied with unit tests and documentation explaining how to set up Strimzi with
+the new authorizer using Azure Active Directory B2C as an example.
```
fvaleri, line 160: Can we also add a Keycloak example? Some users may want to switch to this new authorizer, and that would make their life easier, and increase adoption.

fvaleri (commented):
Hi, thanks for this interesting proposal.
Left some comments.

I also think that it would be good to provide some more details on the proposed syntax for the custom ACL field.

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,251 @@
...
+    ":topic2:write"
+  ]
+}
+```
+
+and for a string:
+```json
+{
+  "acls": "cluster_x:topic_1:read,topic2:write"
+}
```
ppatierno, line 76: why we need two different ways? Isn't it having `acls` as an array enough?

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,251 @@
...
+Some providers allow the data type to be string, while others allow a list of strings.
+There are not many (if any) identity providers that allow a list of objects.
+
+Therefore, in this proposal we allow the ACLs to be passed as either.
+e.g. for a list of strings:
+```json
+{
+  "acls": [
+    "cluster_x:topic_1:read",
+    ":topic2:write"
```
ppatierno, line 67: is this syntax a kind of standard? My understanding is that omitting the cluster name, could mean all clusters like using a regex like `.*`. Is it right? Maybe it's related to the @fvaleri question if we are going to support regex and what's the syntax for this acls.

068-access-policies-in-jwt.md
iblutrifork: That is an interesting idea. We are currently using an implementation that is a subset of glob patterns - handling wildcard matches to allow matching a prefix or a suffix. Considering that we want to compress the ACLs as much as possible, regex could be a good fit, but my concern is that it might overcomplicate it. I cannot think of another situation apart from common prefix/suffix where you'd need the power of regex.
One more concern I have with regex is that it would be hard to debug. Glob patterns are a bit more expressive and have more consistent performance impact.

068-access-policies-in-jwt.md
iblutrifork: That makes sense to me

068-access-policies-in-jwt.md
iblutrifork: I propose two ways in order to cover as many identity providers as possible. While passing `acls` as an array should be preferable, some identity providers support only passing strings, from what I could tell. Looking at [Google Firebase](https://firebase.google.com/docs/auth/admin/custom-claims), I cannot tell if they support passing arrays or just bools/strings/numbers. [OneLogin](https://developers.onelogin.com/api-docs/2/api-authorization/add-claim) is similar - they definitely support strings, but I cannot find information on supporting arrays.

068-access-policies-in-jwt.md
fvaleri: I agree on the points you made in favor of glob patterns for name matching. Can you add them to the proposal with some examples? Also, I would add regex to rejected alternatives with the motivation.

068-access-policies-in-jwt.md
fvaleri: Right, documentation is the best place to raise these warnings.

068-access-policies-in-jwt.md
mstruk: It should probably be something like: `CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS`

Specifically:
* `my-cluster:Topic:topic_1:Read,Describe,DescribeConfigs`
ACL specifically for Kafka cluster named `my-cluster`, grants Read, Describe and DescribeConfigs permissions to topic called `topic_1`

* `:Group:prod_consumers_*:Describe,Delete,Read`
ACL grants Describe,Delete,Read to consumer groups that start with `prod_consumers_` on any Kafka cluster

iblutrifork: I've updated the proposal with more details and examples. This should cover most of the other questions on the ACL syntax in previous threads as well.

I'm using mostly the same format as @mstruk proposes.
I have opted to have the `RESOURCE_TYPE` and `PERMITTED_ACTIONS` be all lowercase.
I'd also change the separator for the `PERMITTED_ACTIONS` to `|` and keep the `,` for separating different ACLs when they are passed in a single string.
However, I have no strong opinion about the exact format.

068-access-policies-in-jwt.md
iblutrifork: done

068-access-policies-in-jwt.md
```diff
@@ -78,6 +78,41 @@ and for a string:
...
+### ACL Syntax
+
+A single ACL has the syntax `CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS`.
+The separator `:` is chosen because it is a common separator in the Unix world.
+
+If the ACLs are passed as a string, they are separated by `,`.
+i.e., `CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS,CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS`
+
+- `RESOURCE_TYPE` can be `topic` or `group`, or the shortened versions `t` and `g`.
+- `PERMITTED_ACTIONS` is a list of a subset of `read`, `write`, `create`, `delete`, `alter`, `alter_configs`, `describe`, `describe_configs`, `cluster_action`, `all`, or the shortened versions `r`, `w`, `c`, `d`, `a`, `ac`, `de`, `dc`, `ca`. The short version for `all` is `*`.
```
fvaleri, line 90: Looks like we miss some operations here.
https://kafka.apache.org/documentation/#operations_in_kafka

068-access-policies-in-jwt.md
```diff
@@ -78,6 +78,41 @@ and for a string:
...
+If `RESOURCE_TYPE` is empty, it takes the default value `topic`.
+
+If `PERMITTED_ACTIONS` is empty, no actions are allowed.
+
+The defaults for `CLUSTER_NAME`, `RESOURCE_SPEC`, and `RESOURCE_TYPE` are chosen for convenience.
+The default for `PERMITTED_ACTIONS` is none to prevent accidental access to all resources.
+
+### Examples of ACLs
+
+- `my_cluster:t:topic1:r|w` - allows reading and writing to `topic1` in the cluster `my_cluster`.
```
fvaleri, line 108: Why did you change from `,` to `|` for permitted actions? Now the symbol reads "or", but the description says "and", which is kind of weird.

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,293 @@
...
+
+### ACL Syntax
+
+A single ACL has the syntax `CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS`.
+The separator `:` is chosen because it is a common separator in the Unix world.
+
+If the ACLs are passed as a string, they are separated by `,`.
+i.e., `CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS,CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS`
+
+- `RESOURCE_TYPE` can be `topic` or `group`, or the shortened versions `t` and `g`.
```
fvaleri, line 89: Looks like we miss some resources here.
https://kafka.apache.org/documentation/#resources_in_kafka

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,316 @@
...
+and for a string:
+```json
+{
+  "acls": "cluster_x:topic_1:read,topic2:write"
+}
+```
+
+The field name is configurable with default value `acls`.
+
+### Using roles to pass ACLs
```
iblutrifork, line 81: @tombentley Adding a section to cover how roles are mapped to kafka ACLs. I am using roles as these are supported by all IDPs. If an IDP supports more advanced mechanisms of authorization or JWT manipulation, then it could also be used.

068-access-policies-in-jwt.md
iblutrifork: Thanks for noticing, added the missing ones.

068-access-policies-in-jwt.md
mstruk: I think the idea is to piggyback on the roles by naming them using this custom ACL syntax. However, that may look quite hackish as the UI used to configure that may offer a drop down list of all the different ACLs and it may be quite confusing.

068-access-policies-in-jwt.md
iblutrifork: My intuition for `|` is the same as when talking about read/write/execute file permissions in Unix, where permissions are represented as 3 bits, so if you want to enable read and write you turn on the read and write permission, which is done by OR-ing them. `100 | 010 = 110`
Again, I have no strong preference, this is the first thing that came to my mind. I would prefer not to use a `,` for AND and instead use it to separate the different ACLs. I think the `&` sign is also a bad idea as it's not as easy to read alongside words.

iblutrifork: I like the `+`

068-access-policies-in-jwt.md
iblutrifork: I'm adding the other resources to the proposal. I do not fully understand if they are all applicable, but I can clear that up later.

068-access-policies-in-jwt.md
iblutrifork: I can see why you would see this as hack-ish. Some IDPs support other ways of enriching the JWT, but they are specific to the IDP.  I chose to give the example with roles as it is what most people would know and what would work under any IDP (albeit with the drawbacks that you mention). For the same reason, I'd give the same example in the documentation.
I have used fine-grained roles like this before and can say it works fine for people.
For the UI concerns, I can talk to a team member who has more experience with this and come back to you.

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,315 @@
...
+    "cluster_x:topic_1:read",
+    ":topic2:write"
+  ]
+}
+```
+
+and for a string:
+```json
+{
+  "acls": "cluster_x:topic_1:read,topic2:write"
```
fvaleri, line 75:
```suggestion
  "acls": "cluster_x:topic_1:read,:topic2:write"
```

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,315 @@
...
+A single ACL has the syntax `CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS`.
+The separator `:` is chosen because it is a common separator in the Unix world.
+
+If the ACLs are passed as a string, they are separated by `,`.
+i.e., `CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS,CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS`
+
+- `RESOURCE_TYPE` can be `topic`, `group`, `transactional_id`, `delegation_token`, `user`, or the shortened versions `t`, `g`, `tid`, `dt`, `u`.
+- `PERMITTED_ACTIONS` is a list of a subset of `read`, `write`, `create`, `delete`, `alter`, `describe`, `cluster_action`, `describe_configs`, `alter_configs`, `idempotent_write`, `create_tokens`, `describe_tokens`, `all`, or the shortened versions `r`, `w`, `c`, `d`, `a`, `de`, `ca`, `dc`, `ac`, `iw`, `ct`, `dt`. The short version for `all` is `*`.
+The list items are separated by a `+`.
+- `CLUSTER_NAME` is the name of the cluster, and `RESOURCE_SPEC` is the name of the resource (topic/group).
```
fvaleri, line 115: I would use the same suffix for these two fields. My preference is on `_NAME`.

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,315 @@
...
+- `:::` - denies all actions on all topics in all clusters.
+- `:::*` - allows all actions on all topics in all clusters.
+- `my_cluster:group:*_app2:read` - allows reading from all groups ending with `_app2` in the cluster `my_cluster`.
+- `::edge_*:write+r` - allows writing and reading topics starting with `edge_` in all clusters.
+
+### Implementation details
+
+The new authorizer will be implemented as a new class, similarly to
+[KeycloakAuthorizer](https://github.com/strimzi/strimzi-kafka-oauth/blob/229daee85b096804d16e2904c8c0f1add599cc99/oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakAuthorizer.java),
+making use of the [KeycloakRBACAuthorizer](https://github.com/strimzi/strimzi-kafka-oauth/blob/229daee85b096804d16e2904c8c0f1add599cc99/oauth-keycloak-authorizer/src/main/java/io/strimzi/kafka/oauth/server/authorizer/KeycloakRBACAuthorizer.java).
```
fvaleri, line 140: Please use a tag instead of main branch for links.

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,315 @@
...
+        // ...
+      }
+    }
+
+    // ...
+}
+```
+
+Operations are described in the [Kafka docs](https://kafka.apache.org/documentation/#operations_in_kafka).
+Allowing an operation will follow the same logic as in [AclEntry.supportedOperations](https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/security/authorizer/AclEntry.scala#L99).
```
fvaleri, line 214: Broken link. Please use a tag instead of main branch.

fvaleri (approved):
LGTM. Thanks.

scholzj:
I'm -0.9 on this proposal. I'm not going to -1 this proposal, so I'm not going to block this if other maintainers want to proceed with this. But I think this should be rejected for the following reasons:
* The OAuth authentication and authorization proved to be very expensive in terms of maintenance - especially compared to how many users it seems to have. This seems to be because of the high cost of maintenance of the tests as well of the Oauth libraries (updating versions, releases etc.). But also because of the bad standardization in the OAuth space where everyone seems to implement things differently and requires more and more options to handle their Oauth setup.
* The proposal does not convince me that this will be the thing used by all OAuth users out of the box as it is proposed here - without each of them requiring a special customization to make it work for them. And the proposal was not able to provide any examples where this is used in similar projects to Strimzi.
* There also seems to be a clear alternative path. The authors of the proposal can create the authorizer in their own repository and keep maintaining it. Strimzi users can then easily add their authorizer to the container images and use it as the `type: custom` authorizer. While this might not have the same level of comfort, there is nothing complicated with that - especially not if this is really the thing everyone wants to use as presented in the proposal. If needed, we can also consider adding some _directory of plugins_ to our website or docs to make it easier to advertise plugins like this to the Strimzi users. And if this proves to be successful and used by many users, this discussion can be re-opened. This seems to be also much better for the Strimzi users than accepting it and then again removing it after a few releases if it proves that the concerns are valid.
* I do not think we have enough resources to take on and maintain things like these. Especially when there is a clear path how it can live outside of Strimzi as a plugin.

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,315 @@
...
+If `PERMITTED_ACTIONS` is empty, no actions are allowed.
+
+The defaults for `CLUSTER_NAME`, `RESOURCE_SPEC`, and `RESOURCE_TYPE` are chosen for convenience.
+The default for `PERMITTED_ACTIONS` is none to prevent accidental access to all resources.
+
+### Examples of ACLs
+
+- `my_cluster:t:topic1:r+w` - allows reading and writing to `topic1` in the cluster `my_cluster`.
+- `:::` - denies all actions on all topics in all clusters.
+- `:::*` - allows all actions on all topics in all clusters.
```
ppatierno, line 132: tbh I think the last two example are not readable. I think the empty defaults (for cluster and resource) or using `topic` if no resource type is specified don't work really well imho. I would use `*` or making the resource type mandatory.

068-access-policies-in-jwt.md
```diff
@@ -0,0 +1,315 @@
...
+
+A single ACL has the syntax `CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS`.
+The separator `:` is chosen because it is a common separator in the Unix world.
+
+If the ACLs are passed as a string, they are separated by `,`.
+i.e., `CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS,CLUSTER_NAME:RESOURCE_TYPE:RESOURCE_SPEC:PERMITTED_ACTIONS`
+
+- `RESOURCE_TYPE` can be `topic`, `group`, `transactional_id`, `delegation_token`, `user`, or the shortened versions `t`, `g`, `tid`, `dt`, `u`.
+- `PERMITTED_ACTIONS` is a list of a subset of `read`, `write`, `create`, `delete`, `alter`, `describe`, `cluster_action`, `describe_configs`, `alter_configs`, `idempotent_write`, `create_tokens`, `describe_tokens`, `all`, or the shortened versions `r`, `w`, `c`, `d`, `a`, `de`, `ca`, `dc`, `ac`, `iw`, `ct`, `dt`. The short version for `all` is `*`.
+The list items are separated by a `+`.
```
ppatierno, line 114: shortened versions are really complicated to make any sense when reading a rule imho.
Also you say using `+` as separated while in one of previous comments you were suggesting the `|`.

ppatierno (commented):
I left a couple of more comments but my feeling is the same as Jakub. Why can't we just allow users to have their own plugins and using `custom` type?
What I see here is a lot of effort to maintain the code working for different IDPs but mostly testing them. Are you going to write tests for all of them as well? I am not sure it would scale well.

068-access-policies-in-jwt.md
fvaleri: I suggested to use +, as | usually is interpreted as "or", but here actually means "and".

fvaleri:
IMO this feature has the potential to simplify the IdP configuration (mapping between users/roles and Kafka ACLs), which would allow more users to benefit from it. Maybe we need more evidence that maintenance wouldn't be a burden, and that this approach really works fine with the most popular IdPs.

068-access-policies-in-jwt.md
iblutrifork: I'm ok with this change. These names wouldn't be used anywhere except maybe in code and in the documentation.

068-access-policies-in-jwt.md
iblutrifork: I somewhat agree. I think best practice would be to be explicit about all the fields. The only reasonable time to use such shorthand syntax would be when developing locally or debugging.

068-access-policies-in-jwt.md
iblutrifork: I decided to propose the shorthand variants only to allow users with very complex and lengthy ACL lists to be able to use this feature with as little overhead as possible. I would personally advise against using them.

iblutrifork:
Thank you for your feedback on this proposal.
Our conversation has been really helpful, and I've learned a lot from it.

I will leave a short summary to help others quickly gauge the proposal's goals.

We use a custom authorizer with Strimzi, which reads ACLs from the JWT.
We created this solution to avoid being bound to a specific IDP, such as Keycloak.
In our IDP we enrich the JWT with a list of ACLs, which the authorizer uses to authorize requests.

We thought this solution could be useful for others and thought the best way to share it would be to contribute it to the Strimzi project.

I believe we should put this proposal on hold until there is more interest from the community for something similar.

scholzj:
Discussed on the community call on 18.4.: This can be closed. If needed, it can be reopened later to get back to it.

yvessavoy:
Hey @iblutrifork 

We have a very similar situation in our company and before I start implementing the same custom authorizer, I wanted to ask if your code is available anywhere to take a look at?

Thanks & cheers!

cthtrifork:
@yvessavoy  we have made a much simpler and less generic version, which lead to this proposal. We do keep it updated.
It is available here:
https://github.com/trifork/cheetah-infrastructure-utils-kafka

Do consider forking it and keeping it public, so we can inspire each other.

PR #111, opened 2024-03-13
Raising proposal based on PR discussion for: Enhance KafkaBridge resurce with consumer inactivity timeout and HTTP consumer/producer parts enablement

steffen-karlsson:


068-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,87 @@
+<!-- This template is provided as an example with sections you may wish to comment on with respect to your proposal. Add or remove sections as required to best articulate the proposal. -->
+
```
scholzj, line 2: You can delete this I guess.

068-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,87 @@
...
+            description: Whether the HTTP producer should be enabled or disabled.
+        description: Configurations for the HTTP Producer.
+      consumer:
+        type: object
+        properties:
+          enabled:
+            type: boolean
+            description: Whether the HTTP consumer should be enabled or disabled.
+        description: Configurations for the HTTP Consumer.
+```
```
scholzj, line 47: The CRD structure is not really much useful for the review as it is hard to read. It would be much more useful to include examples of a KafkaBridge resource with the new API fields to show how it will look like.

068-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,87 @@
...
+
+Properties are not yet supported in the Strimzi Kubernetes Operator.
+
+## Motivation
+
+Raised in the discussion here: https://github.com/strimzi/strimzi-kafka-operator/issues/8732 and triaged on 29.6.2023: We should enable the configuration of these fields. But how should the API look like? Should have a proposal to clarify the API changes.
+
+## Proposal
+
+Proposal is implemented here: https://github.com/strimzi/strimzi-kafka-operator/pull/9820, both Paolo and Jakub suggested to raise a proposal for the implementation of `http.consumer.enabled` and `http.producer.enabled` specifically, as it can be done in two ways, motivated here: https://github.com/strimzi/strimzi-kafka-operator/pull/9820#discussion_r1523500115.
```
scholzj, line 20: Ideally, the proposal should be readable as a single document -> can you maybe summarize the things from the comments instead of linking to them?

scholzj (commented):
Thanks for the proposal. I left few initial comments.

068-kafkabrige-consumer-producer.md
steffen-karlsson: Will work on that and update

steffen-karlsson:
> Thanks for the proposal. I left few initial comments.

Comments resolved @scholzj.

068-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,83 @@
+# Enhance KafkaBridge resource with consumer inactivity timeout and HTTP consumer/producer parts enablement
+
+Providing support in the Strimzi Kubernetes Operator for following properties supported in the Strimzi HTTP Kafka Bridge:
+* `http.timeoutSeconds`   - For deleting inactive consumers after a timeout (disabled by default).
+* `http.consumer.enabled` - To enable/disable the HTTP consumer part (enabled by default).
+* `http.producer.enabled` - To enable/disable the HTTP producer part (enabled by default.
```
ppatierno, line 6:
```suggestion
* `http.producer.enabled` - To enable/disable the HTTP producer part (enabled by default).
```

068-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,83 @@
...
+  http:
+    timeoutSeconds: 60
+  consumer:
+    enabled: true
+    config:
+      foo: "bar"
+  producer:
+    enabled: false
+    config:
+      foo: "buz"
```
ppatierno, line 69: I prefer Version 2 and it also lets me think that we made a mistake in the past to put enablement of producer and consumer under the `http.` section. In the end, producer and consumer are an intrinsic part of the bridge or anyway when enabling/disabling one of them you are actually modifying the overall behaviour for both. I mean, you are enabling/disabling both HTTP and Kafka part of each. There is no distinction.

steffen-karlsson: Maybe it should be called `httpEnabled` instead of just `enabled` then?

If you'd like, I can fix this in the operator API and then we can look at moving it inside the Bridge after?

ppatierno: No I think that `enabled` is the right word to use because I don't see it totally related to HTTP protocol.
Also there are no changes to make on the PR. It has to be parked. We need to reach an agreement on the API through the proposal with all interested maintainers.
Even because just because I like Version 2, it doesn't mean we'll go with Version 2 :-) ... we need to wait what the others think.

068-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,83 @@
...
+  enableMetrics: false
+```
+
+## Affected/not affected projects
+
+PR raised for the HTTP Bridge to use the new ENV parameters introduced from the operator: https://github.com/strimzi/strimzi-kafka-bridge/pull/882
```
scholzj, line 75: So, I guess the affected projects are the HTTP Bridge (changes to the start-up script) and the operators project where you extend the API and the Cluster Operator.

068-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Compatibility
+
+None.
```
scholzj, line 79: You can add that there is no impact on backward compatibility or something like that.

scholzj (commented):
Maybe I should try to explain the proposal process a bit ... the proposal is used to discuss, design, and decide how some more complicated issues will be addressed. Thsi includes typically things that:
* Change the API (the CRDs)
* Might have some techncial consequences, cause future limitations etc.

In this case, the change of the API / CRDs would be the main reason for the proposal. While it is sometimes useful to have a prototype PR written in is not necessarily required. I personally would not link it in the proposal but for example in the comments / PR description as an example of how it might be implemented.

If there are multiple variants -> you would normally choose one and propose it and list the others in the rejected alternatives for others to consider. This could be based on your liking but also on the follow-up discussion. So in this case we should pick one version and use it and move the other to the rejected alternatives. It seems the version two is prefereed by Paolo and probbaly slighly by me as well. So maybe we can pick that one?

Having the one variant selected is important, because at the end, the proposal is voted upon and needs to get at least 3 maintainer votes (approvals) (see more info [here](https://github.com/strimzi/governance/blob/main/GOVERNANCE.md#proposals)). So it needs to be clear what the plan is.

So we should pick the right variant, get the votes which I think should be pretty straigth forward here and then we can proceed with the PRs.

steffen-karlsson:
@scholzj, thanks for explaining :)
Though it was merged and then discussed after, but that makes sense - my bad.

068-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,80 @@
...
+metadata:
+  name: "test-kafka-bridge"
+spec:
+  replicas: 1
+  image: "my-test-image"
+  bootstrapServers: "my-cluster-kafka:9092"
+  http:
+    timeoutSeconds: 60
+  consumer:
+    enabled: true
```
scholzj, line 34: I guess this (and the producer one) defaults to true for backward compatibility? It would be worth mentioning it in the text either here or in the compatibility section.

steffen-karlsson: Sure, that makes sense :)

scholzj:
@k-wall I would try to avoid doing the deep nesting and having `consumer` sections everywhere. So I would not put it under `.soec.http.consumer` and should it be related only to consumer, I would go for `.spec.consumer.timeoutSeconds`. WDYT @ppatierno?

steffen-karlsson:
@scholzj if `timeoutSeconds` should be available for both `consumer` and `producer`, I think thats a better solution, more in line with version 2 presented.

scholzj:
> @scholzj if `timeoutSeconds` should be available for both `consumer` and `producer`, I think thats a better solution, more in line with version 2 presented.

If it applies to both, it makes more sense under `spec.http` as you have it I guess. That also coresponds to `http.timeoutSeconds` as it exists in the bridge itself. But I would defer to @ppatierno as to what he thinks.

steffen-karlsson:
> > @scholzj if `timeoutSeconds` should be available for both `consumer` and `producer`, I think thats a better solution, more in line with version 2 presented.
> 
> If it applies to both, it makes more sense under `spec.http` as you have it I guess. That also coresponds to `http.timeoutSeconds` as it exists in the bridge itself. But I would defer to @ppatierno as to what he thinks.

Sure thing. Sorry for the potential misunderstanding, what I meant was if its relevant for those `timeoutSeconds` to differ from `consumer` to `producer`.

k-wall:
https://github.com/strimzi/strimzi-kafka-bridge/blob/main/src/main/java/io/strimzi/kafka/bridge/http/HttpConfig.java#L37 shows this property applies only to the consumer.  I there should be a change considered against the Bridge itself that would rename the property `http.consumer.timeoutSeconds` (retaining but deprecating the old name). The API being proposed by this change should be `.spec.consumer.timeoutSeconds`.

ppatierno:
The choice is not so obvious :-( and I think that the initial choice of `http.timeoutSeconds` was bad (sorry for that), because it didn't make it clear that the configuration is for the consumer only (as pointed by @k-wall).
The thing is that the bridge has two parts: http and kafka. But they are not that clear in the `KafkaBridge` CRD (i.e. `consumer.config` is about Kafka related parameter so everything like `kafka.*`, but it's not `consumer.kafka.config` which could have been better).
Another idea ... I know it's going to use a different order from the application.properties file but what about using `consumer.http.timeoutSeconds`, `consumer.http.enabled`, `producer.http.enabled` ? I know they will be the opposite of what we have in the application properties starting with `http.*` but we already have the approach of `consumer` and `producer` sections in the `KafkBridge` custom resource.
At the same time it means an additional inner level, which we could avoid.
@scholzj @k-wall @steffen-karlsson wdyt?

scholzj:
@ppatierno I would try to avoid the deep nesting and the confusing use of `http` in multiple places. So maybe you could do something like:

```
spec:
  http:
    consumerEnabled: true
    producerEnabled: true
    consumerTimeoutSeconds: 60
```

if you think these things belong to HTTP?

ppatierno:
@scholzj also your suggestion is fine but I am really struggling to take a strong position. I am really regretting the choices made in the past for the current names in the application.properties file.
Regarding the `enabled` field for example, while it's enabling/disabling the HTTP part of consumer/producer, in reality it's enabling/disabling the consumer/producer as whole (disabling HTTP, no Kafka client is created) so the best should have been just `consumer.enabled` and `producer.enabled` (without the `http` prefix). So the timeout seconds should have been just consumer (if we assume that any other Kafka related timeout should go in the config part).
At this point I would be for having these params directly under consumer (enabled and timeoutSeconds) and producer (enabled) sections.

scholzj:
APIs are hard :-(. When it comes to bad decisions in the `KafkaBridge` custom resource, I guess the `v1` api will be an opportunity. To change the `application.properties` is I guess less complicated technically, but would break compatibility. But in any case I don't think the `KafkaBridge` schema absolutely has to follow the `application.properties` if it doesn't make sense.

k-wall:
Isn't calling it a http timeout is actually misleading?   It could be confused with the timeouts applied to the HTTP protocol itself.   The timeout we are talking about is intrinsic to the consumer bridge.  So that train of thought leads me to:

```
consumer:
    enabled: true
    timeoutSeconds: 60
    config:
      auto.offset.reset: earliest
      enable.auto.commit: true
```

steffen-karlsson:
@ppatierno Sure, let me do that :)

071-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,84 @@
...
+
+Properties are not yet supported in the Strimzi Kubernetes Operator.
+
+## Motivation
+
+Raised in the discussion here: https://github.com/strimzi/strimzi-kafka-operator/issues/8732 and triaged on 29.6.2023: We should enable the configuration of these fields. But how should the API look like? Should have a proposal to clarify the API changes.
+
+## Proposal
+
+This proposal suggests based on discussions here: https://github.com/strimzi/strimzi-kafka-operator/pull/9820 to add the `http enablement` for the `consumer` and `producer` in their respective sessions, i.e. `spec.consumer.enabled` and `spec.producer.enabled`, as well as adding `timeoutSeconds` to the `http` properties.
```
k-wall, line 18: This sentence is now inconsistent with the updated API proposal.

071-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,84 @@
...
+  image: "my-test-image"
+  bootstrapServers: "my-cluster-kafka:9092"
+  consumer:
+    enabled: true
+    timeoutSeconds: 60
+    config:
+      foo: "bar"
+  producer:
+    enabled: false
+    timeoutSeconds: 60
```
k-wall, line 38: I don't think producer actually has the concept of timeout.  It is only the consumer that does.

071-kafkabrige-consumer-producer.md
ppatierno: +1 for @k-wall 
Also I would avoid the code formatting on "http enablement", "consumer" and "producer" but using it for actual code and fields.

071-kafkabrige-consumer-producer.md
ppatierno: +1 It has not, you are right.

071-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,84 @@
...
+    config:
+      foo: "bar"
+  producer:
+    config:
+      foo: "buz"
+  enableMetrics: false
+```
+
+### 2.
+`timeoutSeconds` has been moved from `http` to `consumer` and `producer` section.
```
ppatierno, line 84: again `timeoutSeconds` is valid for consumer only.

ppatierno (changes requested):

steffen-karlsson:
Not sure exactly what you mean by "Also I would avoid the code formatting on "http enablement", "consumer" and "producer" but using it for actual code and fields." @ppatierno

Everything else is updated @ppatierno and @k-wall.

ppatierno:
> Not sure exactly what you mean by "Also I would avoid the code formatting on "http enablement", "consumer" and "producer" but using it for actual code and fields." @ppatierno

The usage of the single apex which provides the formatting like this `formatting with single apex` is typically for code and yaml fields while you are just mentioning "http enablement", "consumer" and "producer". So I would avoid having something like `http enablement`, `consumer` and `producer` unless you refer to fields in the spec or code.

071-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+## Affected/not affected projects
+
+Strimzi Kafka Bridge - PR raised for the HTTP Bridge to use the new ENV parameters introduced from the operator: https://github.com/strimzi/strimzi-kafka-bridge/pull/882
+
+## Compatibility
+
+`http.consumer.enabled` and `http.producer.enabled` both defaults to `true`, thus there are no impacts on backward compatibility.
+
```
k-wall, line 50: for thoroughness, we should talk here about compatibility aspects of `timeoutSeconds`.  If field is omitted from CR, I would expect the effect on the bridge would be as if `http.consumer.timeoutSeconds` had the value [`-1`](https://github.com/strimzi/strimzi-kafka-bridge/blob/25b4106bd5cf3b9b4c8f128b6c284159bb736149/src/main/java/io/strimzi/kafka/bridge/http/HttpConfig.java#L50).

ppatierno: Yes when you omit `http.consumer.timeoutSeconds` from the application.properties, it takes `-1` as default value (no timeout). Missing `timeoutSeconds` in the CR would mean not setting it in the application.properties so providing the same effect. Of course, I agree it should be clarified in this section.

ppatierno:
@steffen-karlsson maybe I was not clear but now your removed single apex from everywhere even where it makes sense :-D ... for example `http.consumer.timeoutSeconds` makes sense having the single apex and formatted as code, but you removed from it having just http.consumer.timeoutSeconds.
As I said, YAML fields should have such format.

071-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,85 @@
+# Enhance KafkaBridge resource with consumer inactivity timeout and HTTP consumer/producer parts enablement
+
+Providing support in the Strimzi Kubernetes Operator for following properties supported in the Strimzi HTTP Kafka Bridge:
+* http.consumer.timeoutSeconds   - For deleting inactive consumers after a timeout (disabled by default).
```
ppatierno, line 4:
```suggestion
* `http.consumer.timeoutSeconds`   - For deleting inactive consumers after a timeout (disabled by default).
```

071-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,85 @@
...
+* http.consumer.enabled          - To enable/disable the HTTP consumer part (enabled by default).
```
ppatierno, line 5:
```suggestion
* `http.consumer.enabled`          - To enable/disable the HTTP consumer part (enabled by default).
```

071-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,85 @@
...
+* http.producer.enabled          - To enable/disable the HTTP producer part (enabled by default).
```
ppatierno, line 6:
```suggestion
* `http.producer.enabled`          - To enable/disable the HTTP producer part (enabled by default).
```

071-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,85 @@
...
+
+Properties are not yet supported in the Strimzi Kubernetes Operator.
+
+## Motivation
+
+Raised in the discussion here: https://github.com/strimzi/strimzi-kafka-operator/issues/8732 and triaged on 29.6.2023: We should enable the configuration of these fields. But how should the API look like? Should have a proposal to clarify the API changes.
+
+## Proposal
+
+This proposal suggests based on discussions here: https://github.com/strimzi/strimzi-kafka-operator/pull/9820 to add the http enablement for the consumer and producer in their respective sessions, i.e. spec.consumer.enabled and spec.producer.enabled, as well as adding timeoutSeconds to the spec.consumer properties.
```
ppatierno, line 18:
```suggestion
This proposal suggests based on discussions here: https://github.com/strimzi/strimzi-kafka-operator/pull/9820 to add the http enablement for the consumer and producer in their respective sessions, i.e. `spec.consumer.enabled` and `spec.producer.enabled`, as well as `spec.consumer.timeoutSeconds` property.
```

071-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,85 @@
...
+  enableMetrics: false
+```
+
+## Affected/not affected projects
+
+Strimzi Kafka Bridge - PR raised for the HTTP Bridge to use the new ENV parameters introduced from the operator: https://github.com/strimzi/strimzi-kafka-bridge/pull/882
+
+## Compatibility
+
+http.consumer.enabled and http.producer.enabled both defaults to true, thus there are no impacts on backward compatibility.
```
ppatierno, line 49:
```suggestion
Not specifying `spec.consumer.enabled` or `spec.producer.enabled` implies not configuring the corresponding bridge parameters `http.consumer.enabled` and `http.producer.enabled` which default to true, thus there are no impacts on backward compatibility.
```

071-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,85 @@
...
+
+If http.consumer.timeoutSeconds is not specified, it should default to -1 to have the same effect, as when it is omitted from the application.properties.
```
ppatierno, line 51: It seems you are referring to the property you would have in the `KafkaBridge` which is omitted but it's not `http.consumer.timeoutSeconds` but it's `spec.consumer.timeoutSeconds`.

```suggestion
Not specifying `spec.consumer.timeoutSeconds` implies not configuring the corresponding bridge parameter `http.consumer.timeoutSeconds` which defaults to -1 to have the same effect (no timeout), thus there are no impacts on backward compatibility.
```

071-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Rejected alternatives
+
+### 1.
+Initially in the implementation suggestion: https://github.com/strimzi/strimzi-kafka-operator/pull/9820, the enabled properties for producer and consumer were added to as part of the http properties, i.e. spec.http.consumer.enabled and spec.http.producer.enabled.
```
ppatierno, line 56:
```suggestion
Initially in the implementation suggestion: https://github.com/strimzi/strimzi-kafka-operator/pull/9820, the enabled properties for producer and consumer were added to as part of the http properties, i.e. `spec.http.consumer.enabled` and `spec.http.producer.enabled`.
```

071-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,85 @@
...
+    config:
+      foo: "bar"
+  producer:
+    config:
+      foo: "buz"
+  enableMetrics: false
+```
+
+### 2.
+timeoutSeconds has been moved from http to consumer section.
```
ppatierno, line 85: I think we can remove this sentece. It's just part of the proposed solution.

ppatierno (changes requested):

scholzj (approved):
I think Paolo's comments make sense and should be applied. But otherwise it looks good to me.

ppatierno:
@steffen-karlsson just adding one more thing. The `071` number was just taken by another merged proposal.
You should rename the file using the next available number and also update the README file with this proposal (number and title).

ppatierno (approved):
LGTM. Thanks for the proposal!

im-konge (approved):

072-kafkabrige-consumer-producer.md
```diff
@@ -0,0 +1,82 @@
...
+  producer:
+    enabled: false
+    config:
+      foo: "buz"
+  enableMetrics: false
+```
+
+## Affected/not affected projects
+
+Strimzi Kafka Bridge - PR raised for the HTTP Bridge to use the new ENV parameters introduced from the operator: https://github.com/strimzi/strimzi-kafka-bridge/pull/882
```
see-quick, line 45:
```suggestion
Strimzi Kafka Bridge - A PR has been raised for the HTTP Bridge to use the new ENV parameters introduced by the operator:https://github.com/strimzi/strimzi-kafka-bridge/pull/882
```

see-quick (approved):
LGTM

k-wall (approved):

ppatierno:
@steffen-karlsson as for the other PR, I see a lot of commits which are not related to the proposal. What's happened? Can you try to clean it please? Closing and opening is not good because we'll lost all the approvals we already have so you should fix on this PR imho.

steffen-karlsson:
> @steffen-karlsson as for the other PR, I see a lot of commits which are not related to the proposal. What's happened? Can you try to clean it please? Closing and opening is not good because we'll lost all the approvals we already have so you should fix on this PR imho.

Git happend, tried to update from main due to conflict and then my fork was out of sync.
Let me squash it all into one

ppatierno:
This proposal has enough votes to be approved and merged (4 binding, 1 non-binding).
Because it's has been opened for long time, I will wait the end of tomorrow (April 16th) and then approve/merged if no further feedback are provided.

ppatierno:
This proposal is approved with 4 binding and 1 non-binding votes. Going to merge.
@steffen-karlsson you can even back to work on the implementation PR changing it accordingly. Thanks!

PR #112, opened 2024-03-19
Enhancements to the Strimzi Kafka Operator for managing shared volume mounts,

rao2100:
Proposal to introduce enhancements to the Strimzi Kafka Operator for managing shared volume mounts, improving task resiliency across nodes.

The changes for the proposal in the PR (https://github.com/strimzi/strimzi-kafka-operator/pull/9761)

### DIFF
--- a/069-shared-volume-mount.md
+++ b/069-shared-volume-mount.md
@@ -1,0 +1,108 @@
+# Shared Volume Mount Changes
+
+The proposed update to the Strimzi Kafka Operator introduces management of shared volume mounts for kafka connect cluster, 
+enhancing task resilience across pods. It allows custom specification of the container directory for storage volume mounting 
+and defines storage access modes.
+
+## Current situation
+
+In the existing setup, Strimzi Operator allows mounting of storage volumes for Apache Kafka and Zookeeper pods.  
+However, it currently faces a limitation in terms of flexibility when managing shared volume mounts across kafka connect cluster.
+It only allows mounting of configmaps and secrets. So if there are any files written to disk (by sink file connectors) then will  be lost
+if the pod restarts. This limits task resiliency in distributed environments, which is a critical aspect of maintaining high availability and fault tolerance in such systems.
+
+## Motivation
+
+The primary motivation for this proposed enhancement to the Strimzi Kafka Operator is to ensure uninterrupted access to data across all kafka connect pods in a distributed cluster. In the current setup, if a pod goes down, the data stored in its mount path becomes inaccessible, potentially disrupting operations. By enabling shared volume mounts across all pods, we can ensure that even if one pod is down, the data remains accessible from other pods. This feature is crucial for maintaining high availability and fault tolerance, thereby significantly improving the resilience and efficiency of tasks of connectors. Hence, this proposed change is not just an improvement, but a necessity for robust and seamless data management in distributed kafka connect clusters.
+ 
+Support for resilliency for sink file connnectors: Mounting volumes will allow a task that gets reallocted to a different pod to continue processing the file. Also allows for exactly once  sink file connectors to store their offsets to file without them being lost in case  of pod restart.
+Support for Pod Running Logs: Ability to dump kafka connect logs to storage volume instead of just stdout.
+GC Logs Management: Implement a JVM option to output GC logs to a file instead of stdout when GC logging is enabled. This will help in identifying JVM hangs and keep stdout uncluttered.
+Additional diagnostics: Explore the possibility of mounting additional volumes into the Kafka Connect pods beyond ConfigMaps or Secrets. This could facilitate writing event data to parquet files for later analysis.
+
+## Proposal
+
+This change introduces enhancements to the Strimzi Kafka Operator for managing shared volume mounts, improving task resiliency across nodes. Key changes include:
+
+1) **mountPath:** A string property to specify the container directory for mounting the storage volume. While the default is /var/lib/kafka/data, this allows for custom path specification. By default, it will try to mount read write many storage if accessMode is not specified.
+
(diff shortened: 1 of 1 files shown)


069-shared-volume-mount.md
```diff
@@ -0,0 +1,108 @@
...
+
+1) **mountPath:** A string property to specify the container directory for mounting the storage volume. While the default is /var/lib/kafka/data, this allows for custom path specification. By default, it will try to mount read write many storage if accessMode is not specified.
+
+2) **accessMode:** An optional string property defining the storage access mode, determining permissions like read-only or read-write.
+
+These changes, tested successfully with Read-Write-Many on Ceph-block storage and NFS, provide greater control over volume configuration and usage, improving operator performance in distributed environments. 
+
+**Sample Test**
+
+ - This test case is to test the strimzi operator customization that allows mounting of shared volume across kafka-connect pods
```
scholzj, line 35: You should also explain how will this be cover in the various test environments -> how will you provide the shared storage for example?

069-shared-volume-mount.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Compatibility
+
+The proposed enhancement to the Strimzi Kafka Operator has been designed with both future and backward compatibility in mind. For future compatibility, the feature is built to be flexible and adaptable to accommodate potential changes or additions to the storage and access modes. This ensures that the feature remains relevant and useful as the system evolves. For backward compatibility, the feature maintains the existing functionality and behavior of the operator when the new options are not specified. This means that existing projects can continue to operate without any changes, ensuring a smooth transition and minimizing disruption to current operations..
+
+## Rejected alternatives
+
+While formulating this proposal, several alternatives were considered but ultimately rejected for various reasons:
+
+- **Using Local Storage**: Initially, the idea of using local storage for each pod was considered. However, this was rejected because it would not allow data sharing across pods, which is a crucial requirement for ensuring high availability and fault tolerance in distributed systems.
```
scholzj, line 104: I think it is fine to _reject_ local storage. But I do not think shared persistent volumes are a good basis for distributed systems as you suggest as they cause many limitations as to how _distributed_ they can be.

069-shared-volume-mount.md
```diff
@@ -0,0 +1,108 @@
...
+- **Static Volume Mounts**: Another alternative was to use static volume mounts with predefined paths. This was rejected due to its lack of flexibility. It would not allow custom specification of the container directory for storage volume mounting, which is necessary for accommodating different system configurations and use cases.
```
scholzj, line 105: I don't think it is really clear what `Static Volume Mounts` are.

069-shared-volume-mount.md
```diff
@@ -0,0 +1,108 @@
+# Shared Volume Mount Changes
+
+The proposed update to the Strimzi Kafka Operator introduces management of shared volume mounts for kafka connect cluster,
```
scholzj, line 3: We need to have a reusable solution that covers all the different use cases for mounting some volumes. So you should not focus just on your immediate need. We have to make sure this change can survive the _next proposal_ from someone else with different requirement.

scholzj (commented):
Thanks for the proposal. As I already mentioned in the PR, we have to make sure it matches all the various usecases mentioned in the existing Strimzi issues such as https://github.com/strimzi/strimzi-kafka-operator/issues/3693. And not just your single use-case. So you should cover all of them in this proposal and explain how they will be fulfilled.

My (personal) view is that we do not want to extend the existing `.storage` APi which is designed in sync withj Kafka's requirement. Instead, I think the right way might be to extend the Pod and Container template to allow adding volumes and mount points?

069-shared-volume-mount.md
```diff
@@ -0,0 +1,108 @@
...
+Support for Pod Running Logs: Ability to dump kafka connect logs to storage volume instead of just stdout.
+GC Logs Management: Implement a JVM option to output GC logs to a file instead of stdout when GC logging is enabled. This will help in identifying JVM hangs and keep stdout uncluttered.
+Additional diagnostics: Explore the possibility of mounting additional volumes into the Kafka Connect pods beyond ConfigMaps or Secrets. This could facilitate writing event data to parquet files for later analysis.
+
+## Proposal
+
+This change introduces enhancements to the Strimzi Kafka Operator for managing shared volume mounts, improving task resiliency across nodes. Key changes include:
+
+1) **mountPath:** A string property to specify the container directory for mounting the storage volume. While the default is /var/lib/kafka/data, this allows for custom path specification. By default, it will try to mount read write many storage if accessMode is not specified.
```
ppatierno, line 27: It's not just about where the volume is mounted but this path is also related to the Kafka broker configuration where putting the logs via `log.dirs` param. 
Also how it's going to work in terms of backward compatibility if for an already running cluster you change this path. What happens to the log segments?

069-shared-volume-mount.md
```diff
@@ -0,0 +1,108 @@
...
+
+2) **accessMode:** An optional string property defining the storage access mode, determining permissions like read-only or read-write.
```
ppatierno, line 29: can you elaborate what's the rationale behind this need?

069-shared-volume-mount.md
ppatierno: I am with Jakub on this.

ppatierno (commented):
I left a few comments but one thing you should do is splitting all the sentences, one per line. It's really hard to make comments on a specific sentence when there are more and how GitHub works.

scholzj:
@rao2100 Do you plan to incorporate the comments into the proposal? Or should this be closed?

scholzj:
Discussed on the community call on 2.5.2024: Should be closed as there does not seem to be any progress. Feel free to reopen it / open new PR with the comments incorporated.

PR #113, opened 2024-03-20
Introducing Performance Testing

see-quick:
This PR adds a proposal for performance testing within Strimzi.


fvaleri (commented):
Thanks for this proposal. 
I like the approach you describe and the implementation details.
Left some comments.

IMO it is critical to base these performance tests on real world use cases, as they can be really helpful to boost confidence in case of major changes. Maybe we could mention this.

068-performance-testing.md
```diff
@@ -0,0 +1,311 @@
...
+            Arguments.of("500", "100", false),    // Increased batch size with longer linger time
+            Arguments.of("1000", "100", false)    // Large batch size with longer linger time
+    );
+}
+```
+
+```java
+@ParameterizedTest
+@MethodSource("provideConfigurationsForAliceBulkBatchUseCase")
+public void testAliceBulkBatchUseCase(String maxBatchSize, String maxBatchLingerMs, boolean withClientsEnabled) throws IOException {
```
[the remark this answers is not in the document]
see-quick: Yeah, good idea. Added.

scholzj (approved):
I think having performance tests would be definitely useful. Although I'm not sure the suggested change is really significant enough to need a proposal. But since you already wrote it, I guess we can approve it 😉.

ppatierno (approved):
I agree with Jakub. Not sure a proposal was needed for that. Maybe I missed something but not clear if it's going to be just performance tests for TO or even other components like bridge for example.

068-performance-testing.md
```diff
@@ -0,0 +1,319 @@
...
+@Override
+public void run() {
+    LOGGER.info("Thread started with selector: {}", this.selector);
+
+    while (!Thread.currentThread().isInterrupted()) {
+        this.topicOperatorMetricsCollector.collectMetricsFromPods();
+        // record specific time when metrics were collected
+        Long timeWhenMetricsWereCollected = System.nanoTime();
+
+        Map<String, List<Double>> currentMetrics = new HashMap<>();
```
tombentley, line 228: Shouldn't do something with `currentMetrics` once you're populated it?

tombentley (approved):
Thanks @see-quick

fvaleri (approved):

im-konge (approved):

068-performance-testing.md
see-quick: We do...this is just for brevity and illustration that we would create a separate Thread within test execution where we would gather metrics. 
For instance:
1. We calculate derived metrics in the loop (I know this could be calculated within that loop)
2. (main one) We create a file for each metric so we loop through them in the `Reporter` class.

```java
metricsHistory.forEach((time, metrics) -> {
            // Including milliseconds in the time format
            DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd-HH-mm-ss-SSS");
            dateTimeFormatter = dateTimeFormatter.withZone(ZoneId.systemDefault());
            String formattedTime = dateTimeFormatter.format(Instant.ofEpochMilli(time));

            metrics.forEach((metricName, values) -> {
                String content = String.format("Timestamp: %s, Values: %s%s", formattedTime, values, newLine);

                // Using just the metric name for the file name
                String fileName = metricName + ".txt";
                Path metricFilePath = alicePathDir.resolve(fileName);

                // Appending to the file if it exists, or creating a new one if it doesn't
                try (BufferedWriter writer = Files.newBufferedWriter(metricFilePath,
                        StandardCharsets.UTF_8,
                        Files.exists(metricFilePath) ? StandardOpenOption.APPEND : StandardOpenOption.CREATE)) {
                    writer.write(content);
                } catch (IOException e) {
                    LOGGER.error("Failed to write metrics to file", e);
                }
            });
        });
```

see-quick:
> I agree with Jakub. Not sure a proposal was needed for that. Maybe I missed something but not clear if it's going to be just performance tests for TO or even other components like bridge for example.

Initial PR will cover just TO, but the plan is to extend it and test our other Strimzi components. The reason, why I opened a proposal is that I wanted to be sure we all decide on place where such tests should live (i.e., `systemtest` or `performance` module) and using totally different repository.

ppatierno:
> Initial PR will cover just TO, but the plan is to extend it and test our other Strimzi components.

Is it stated in the proposal? Maybe I missed that, can you point me to it. It should be clear somehow to set expectations and commitment.

see-quick:
> > Initial PR will cover just TO, but the plan is to extend it and test our other Strimzi components.
> 
> Is it stated in the proposal? Maybe I missed that, can you point me to it. It should be clear somehow to set expectations and commitment.

Well, I wrote this proposal in a general way and then provided an example of a test scenario (i.e., TO). But yeah sure I can add a sentence to explicitly say that. Thanks

scholzj:
> > Initial PR will cover just TO, but the plan is to extend it and test our other Strimzi components.
> 
> Is it stated in the proposal? Maybe I missed that, can you point me to it. It should be clear somehow to set expectations and commitment.

To be honest, I understood the proposal as a generic proposal with a single example based on the TO. So it seemed to be quite clear to me.

ppatierno:
I think that without this addition by Maros ...

> The initial PR will cover the Topic Operator (TO) component, but the plan is to extend it and test our other Strimzi components.

It was not clear at all to me. Thanks @see-quick !

Frawless (approved):

068-performance-testing.md
```diff
@@ -0,0 +1,321 @@
...
+        }
+    }
+}
+```
+
+And in the test case we would simply interrupt such `Thread` to finish with scraping.
+Also, we could make an interval for scraping such metrics configurable via environment variable or system property.
+
+Moreover, we propose that all metrics would be gathered in the `systemtest/target/performance` directory, with same pattern as we do it with the `logs` directory. 
+For instance from performance directory we could:
```
PaulRMellor, line 258:
```suggestion
For instance, the performance directory could be structured as follows:
```

068-performance-testing.md
```diff
@@ -0,0 +1,321 @@
...
+            ...
+            ├── strimzi_update_status_duration_seconds_sum.txt
+            ├── system_cpu_count.txt
+            ├── system_cpu_usage.txt
+            ├── system_load_average_1m.txt
+            ├── system_load_average_per_core_percent.txt
+            └── test-performance-metrics.txt
+... # more dates
+```
+where each of these files contains values, scraped by `Thread` in the test.
```
PaulRMellor, line 278:
```suggestion
where each of these files contains values scraped by `Thread` in the test. 
```

068-performance-testing.md
```diff
@@ -0,0 +1,321 @@
...
+... # other records skipped for brevity
+Timestamp: 2024-03-16-13-57-59-797, Values: [6.12]
+Timestamp: 2024-03-16-13-58-05-067, Values: [6.19]
+```
+
+And these values/data could be used in the next phase (i.e., analyzing and reporting).
+
+### Analyze and report 
+
+We propose, that a `Reporter.class`, which is responsible for retrieving metrics from `<module-name>/target/performance` directory.
```
PaulRMellor, line 291:
```suggestion
We propose a `Reporter.class` that is responsible for retrieving metrics from `<module-name>/target/performance` directory. 
```

068-performance-testing.md
```diff
@@ -0,0 +1,321 @@
...
+Timestamp: 2024-03-16-13-58-05-067, Values: [6.19]
+```
+
+And these values/data could be used in the next phase (i.e., analyzing and reporting).
+
+### Analyze and report 
+
+We propose, that a `Reporter.class`, which is responsible for retrieving metrics from `<module-name>/target/performance` directory. 
+Reporter will generate a table of all related metrics to that specific test case.
+For instance for Topic Operator performance test case such report could look like this:
```
PaulRMellor, line 293:
```suggestion
For instance, for the Topic Operator performance test case such a report could look like this:
```

068-performance-testing.md
```diff
@@ -0,0 +1,321 @@
...
+| 2          | 500              | 84.553            | 169.537           | 254.090             | 1000                   | 30000               | 7.281                           | 385.000        | 324.000        | 3091.712                 | 0.237088658                       | 6.530933468                    | 7.281358907                      | 0.307568988                    | 22.8%                                   | 172.74                    |
+| ...        | ...              | ...               | ...               | ...                 | ...                    | ...                 | ...                             | ...            | ...            | ...                      | ...                               | ...                            | ...                              | ...                            | ...                                     | ...                       |
+
+We could use such report as a `post-processing` phase after test execution and all needed metrics were gathered from Pods.
+
+### CI: Testing farm
+
+We copy pattern from the system tests and just adapt specific label to it.
```
PaulRMellor, line 305: Maybe people unfamiliar with Testing Farm would benefit from a little context here? The purpose and setup.

PaulRMellor (commented):
The proposal seems like a great addition to the testing effort.
I left some suggestions as I read through.

068-performance-testing.md
```diff
@@ -0,0 +1,321 @@
...
+
+We propose integrating performance testing directly alongside existing system tests, leveraging the established infrastructure and testing utilities. 
+This method ensures that performance and system tests can share resources efficiently, streamlines the development process, facilitates immediate feedback on performance impacts, and reduces overhead.
+
+**Advantages**:
+- Streamlines the development process by using existing test infrastructure.
+- Facilitates immediate feedback on performance impacts within the same testing framework used for system tests.
+- Reduces the overhead of maintaining separate modules for system and performance tests.
+
+The initial PR will cover the Topic Operator (TO) component, but the plan is to extend it and test our other Strimzi components.
```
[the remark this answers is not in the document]
see-quick: No, I don't think it will be necessary. The main reason is decide where such tests should be placed :)

068-performance-testing.md
```diff
@@ -0,0 +1,321 @@
...
+3. **Metrics Collection:** 
+   1. Base our metrics collection on the system test module's `MetricsCollector`, with plans to possibly extend this functionality in the future from an external library like [Test Frame](https://github.com/skodjob/test-frame). 
+   2. Additionally, introduce specialized collectors such as `TopicOperatorMetricsCollector` to gather detailed metrics relevant to the Strimzi Topic Operator, tailoring our data gathering to the nuances of Kafka and Strimzi.
+4. **Performance Reporting:** 
+   1. Develop a performance test reporting mechanism. This component will be essential for documenting and analyzing the outcomes of performance tests. 
+   2. The reporter should be capable of generating detailed summaries and insights, 
+   ideally supporting formats conducive to both human reading and automated parsing, such as Markdown or structured JSON.
+
+**Including within the `systemtest` module:** Integrate performance tests to ensure they coexist seamlessly with system tests, potentially categorizing tests to maintain clarity and prevent any confusion between system and performance testing efforts (i.e., using performance profile).
```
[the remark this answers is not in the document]
see-quick: No.

068-performance-testing.md
```diff
@@ -0,0 +1,321 @@
...
+ * subsequently consume from them in parallel, and finally delete all the topics. This use case is particularly focused on minimizing
+ * the latency associated with the bulk creation and deletion of topics. In such scenarios, Kafka's performance and its ability to handle
+ * batching operations efficiently become crucial factors.
+ *
+ * The test leverages Kafka's batching capabilities for both creation and deletion of topics to aid in this process. By using a larger batch size
+ * and a longer linger time, the test aims to optimize the throughput and minimize the operational latency.
+ *
+ * This test scenario is implemented to help users in understanding the performance characteristics of Kafka when dealing with
+ * high volumes of topics and to assist in configuring their systems for optimal performance during bulk operations.
+ */
```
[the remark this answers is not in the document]
see-quick: Hmmm, I think having this as a `JavaDoc` is enough or do you think it would make more sense removing it here and adding two steps above? :)

068-performance-testing.md
```diff
@@ -0,0 +1,321 @@
...
+
+### CI: Azure Pipelines
+
+We assume Azure machines may not suffice for all performance tests, meaning we anticipate that lower-complexity performance tests, 
+not requiring as much memory/CPU, could run on such infrastructure. 
+However, for tests needing more memory in the context of performance, we prefer another approach (i.e., using a Testing farm) to execute these tests.
+
+### CI: Possible integration with Testing Farm 
+
+Special consideration will be given to the potential integration of performance tests with Testing farm, utilizing higher-capacity machines than those available in Azure, to ensure comprehensive performance evaluation.
```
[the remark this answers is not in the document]
see-quick: > possible integration, potential integration: does this mean integrating when necessary or integrating if using a testing farm is found to be suitable?

integrating if using a testing farm is found to be suitable. But I am sure we will integrate it with testing farm :)

068-performance-testing.md
see-quick: Updated :)

PaulRMellor (approved):

see-quick:
This proposal now has 6 binding and 1 non-binding vote. I will close the vote and merge it tomorrow morning if there is no objection.

PR #118, opened 2024-05-06
Extend Feature Gates to all Strimzi operators

scholzj:
This proposal suggests extending the support for Feature gates to all Strimzi operators. The main motivation for it is the work on the Server-Side-Apply support that would apply to all operators.


see-quick (approved):
LGTM, well written. Thanks

fvaleri (approved):
LGTM. Thanks.

073-extend-feature-gates-to-all-operators.md
```diff
@@ -0,0 +1,114 @@
...
+## Proposal
+
+The feature gates and the `STRIMZI_FEATURE_GATES` environment variable will keep its current format.
+The support for the `STRIMZI_FEATURE_GATES` environment variable will be added to the User and Topic Operators.
+The same feature gates will be defined and handled by every operator.
+However, if the operator doesn't have any use for the feature gate, it can simply ignore it in its code.
+
+When the User and Topic Operators are deployed through the Cluster Operator as part of the Kafka cluster, the user will configure the feature gates as today using the `STRIMZI_FEATURE_GATES` environment variable.
+And the Cluster Operator will then automatically pass the environment variable to the User and Topic Operators when deploying them.
+When the User and Topic Operators are deployed as standalone, used can configure the `STRIMZI_FEATURE_GATES` environment variable directly in their deployments.
```
Frawless, line 37:
```suggestion
When the User and Topic Operators are deployed as standalone, user can configure the `STRIMZI_FEATURE_GATES` environment variable directly in their deployments.
```

Frawless (approved):

073-extend-feature-gates-to-all-operators.md
```diff
@@ -0,0 +1,114 @@
...
+
+Additional feature gates that apply to the other operators might be added in the future.
+To have a unified way how to manage such feature gates, this proposal provides the plan for extending the feature gates to the User and Topic Operators as well.
+
+## Proposal
+
+The feature gates and the `STRIMZI_FEATURE_GATES` environment variable will keep its current format.
+The support for the `STRIMZI_FEATURE_GATES` environment variable will be added to the User and Topic Operators.
+The same feature gates will be defined and handled by every operator.
+However, if the operator doesn't have any use for the feature gate, it can simply ignore it in its code.
```
PaulRMellor, line 33: Would it be beneficial to provide more details on how this would be implemented?

073-extend-feature-gates-to-all-operators.md
```diff
@@ -0,0 +1,114 @@
...
+
+When the User and Topic Operators are deployed through the Cluster Operator as part of the Kafka cluster, the user will configure the feature gates as today using the `STRIMZI_FEATURE_GATES` environment variable.
+And the Cluster Operator will then automatically pass the environment variable to the User and Topic Operators when deploying them.
+When the User and Topic Operators are deployed as standalone, user can configure the `STRIMZI_FEATURE_GATES` environment variable directly in their deployments.
```
PaulRMellor, line 37: To be clear, typically all the feature gate configuration is done in the Cluster Operator config. If standalone UO and TO are used, you configure in Cluster Operator and in the standalone operators?

073-extend-feature-gates-to-all-operators.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Trade-offs and limitations
+
+This design has several trade-offs and limitations: 
+* When a feature gate is enabled or disabled in the Cluster Operator, it will cause rolling update of all the User and Topic Operators managed by it.
+  This will happen even if they in reality don't use given feature gate in their code and simply ignore it.
+  That said, changing the feature gates configuration is not something what is expected to be done very often.
```
PaulRMellor, line 44:
```suggestion
  That said, changing the feature gates configuration is not something that is expected to be done very often.
```

073-extend-feature-gates-to-all-operators.md
```diff
@@ -0,0 +1,114 @@
...
+* The centrally managed feature gate configuration will also apply to all Kafka clusters managed by given Cluster Operator.
+  It cannot be applied to individual Kafka clusters which can be seen as a limitation.
+  But it does not differ significantly from the current state.
+
+### Advantages
+
+It has also aspects that can be considered as advantages:
```
PaulRMellor, line 51:
```suggestion
The proposed change also has aspects that can be considered advantages:
```

073-extend-feature-gates-to-all-operators.md
```diff
@@ -0,0 +1,114 @@
...
+* The configuration of the feature gates is unchanged compared to how it works today.
+  It also maintains the simplicity of the configuration.
+  The user needs to set only one environment variable in the Cluster Operator to configure all feature gates.
```
PaulRMellor, line 54: unless using standalone operators?

073-extend-feature-gates-to-all-operators.md
```diff
@@ -0,0 +1,114 @@
...
+  This also means that we do not need any changes to our system tests as they can keep using the Feature Gates in the same way as they do today.
+* Having a consistency and clarity how the operators are configured simplifies support as there are less variables that need to be understood before helping our users.
+* While the implementation is different, the proposed behavior mirrors the current behavior:
+    * The feature gates apply to all operands (i.e. to all User and Topic Operators managed by the same Cluster Operator)
+    * It follows the behavior of the `UnidirectionalTopicOperator` feature gate that was also configured centrally and changing it caused rolling update of the Topic Operator(s).
+* Given that one of the goals of the feature gates is to allow users easily test new features before introducing them to everyone when they are enabled by default, the central configuration makes sure the feature gate is properly tested.
```
PaulRMellor, line 60:
```suggestion
* Given that one of the goals of the feature gates is to allow users to easily test new features before they are enabled by default for all users, the central configuration makes sure the feature gate is properly tested.
```

073-extend-feature-gates-to-all-operators.md
```diff
@@ -0,0 +1,114 @@
...
+  It for example helps to avoid situation when a user would by mistake test the `UseServerSideApply` feature gate only in the Cluster Operator, but not in the other operators.
```
PaulRMellor, line 61:
```suggestion
  For example, it helps to avoid situation when a user would test the `UseServerSideApply` feature gate only in the Cluster Operator by mistake, but not in the other operators.
```

073-extend-feature-gates-to-all-operators.md
```diff
@@ -0,0 +1,114 @@
...
+* If there is a significant demand for configuring the feature gates per-operand or more dynamically (i.e. without restarts), it can be probably better addressed as part if the [OpenFeature support](https://github.com/strimzi/strimzi-kafka-operator/issues/7520) building on top of this proposal.
```
PaulRMellor, line 62:
```suggestion
* If there is a significant demand for configuring the feature gates per-operand or more dynamically (i.e. without restarts), it can be probably better addressed as part of the [OpenFeature support](https://github.com/strimzi/strimzi-kafka-operator/issues/7520) building on top of this proposal.
```

073-extend-feature-gates-to-all-operators.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Expected source code changes
+
+To implement the proposed changes, the feature gates source code and the feature gates definitions will move to the `operator-common` module and be shared between the 3 different operators that will support the feature gates.
+The feature gates configuration will be added to the configuration classes of the User and Topic Operators.
+
+As of today, there are no (merged) feature gates that would apply to the User or Topic Operator.
+So not other changes to their source code are needed.
```
PaulRMellor, line 70:
```suggestion
So no other changes to their source code are needed.
```

073-extend-feature-gates-to-all-operators.md
scholzj: Yes, correct. I added it there. But when using the standalone operators, you have completely standalone installations, so you cannot really expect it to be done somehow centrally.

073-extend-feature-gates-to-all-operators.md
scholzj: When using standalone User and/or Topic operators, they are completely independent and it is not really expected that you use the Cluster Operator at all. I tried to make it more clear in the proposl.

073-extend-feature-gates-to-all-operators.md
```diff
@@ -0,0 +1,116 @@
...
+The feature gates and the `STRIMZI_FEATURE_GATES` environment variable will keep its current format.
+The support for the `STRIMZI_FEATURE_GATES` environment variable will be added to the User and Topic Operators.
+The same feature gates will be defined and handled by every operator.
+However, if the operator doesn't have any use for the feature gate, it can simply ignore it in its code.
+
+When the User and Topic Operators are deployed through the Cluster Operator as part of the Kafka cluster, the user will configure the feature gates as today using the `STRIMZI_FEATURE_GATES` environment variable.
+And the Cluster Operator will then automatically pass the environment variable to the User and Topic Operators when deploying them.
+
+When the User and Topic Operators are deployed as standalone, there is no Cluster Operator to configure the feature gates through it.
+But the users can configure the `STRIMZI_FEATURE_GATES` environment variable directly in the User and Topic operator deployments.
```
PaulRMellor, line 39:
```suggestion
When the User and Topic Operators are deployed as standalone, there is no Cluster Operator available to configure the feature gates.
Instead, users can configure the `STRIMZI_FEATURE_GATES` environment variable directly in the User and Topic Operator deployments.
```

073-extend-feature-gates-to-all-operators.md
PaulRMellor: Right. Makes sense now. Thanks

PaulRMellor (approved):
A nit, but looks good to me

073-extend-feature-gates-to-all-operators.md
```diff
@@ -0,0 +1,116 @@
...
+
+Additional feature gates that apply to the other operators might be added in the future.
+To have a unified way how to manage such feature gates, this proposal provides the plan for extending the feature gates to the User and Topic Operators as well.
+
+## Proposal
+
+The feature gates and the `STRIMZI_FEATURE_GATES` environment variable will keep its current format.
+The support for the `STRIMZI_FEATURE_GATES` environment variable will be added to the User and Topic Operators.
+The same feature gates will be defined and handled by every operator.
+However, if the operator doesn't have any use for the feature gate, it can simply ignore it in its code.
```
tombentley, line 33: In the implementation can we log (e.g. at INFO) the feature gates which are, and are not, understood/being consumed by an operator?

tombentley (approved):

im-konge (approved):

073-extend-feature-gates-to-all-operators.md
scholzj: To be honest, I did not plan for that. I don't think it is completely trivial ... I think it leads to a small _spiral of complexity_ ...
* I do not think you would want to have a dedicated code in each operator with hardcoded log message as that might fail to get updated etc.
* So you probably need to have some flag in the Feature Gate itself ... `supportsTopicOperator`, `supportsClusterOperator`, etc.
* But if you have that flag, should you really use it just for logging? Maybe you can use it earlier, not pass the irrelevant feature gates down the stream (e.g. not update the running UO when a feature gate that is not used by UO is not enabled), and save an unnecessary rolling update.
* But if you do that, what about the standalone UO and TO ... should it reject the feature gates it does not use as if it was an invalid feature gate? Or what should it do?

So I wonder ... is it really worth the effort and the complexity? Does it make sense to stop only with the log message if you already have the information? 

I think it is not worth it. We can for example use the docs to explain which components are affected by the feature gate and I think that would be sufficient. But obviously, it is not like the implementation would take weeks ... so if people think this would be worth it, I can update the proposal.

katheris (approved):
LGTM

tinaselenge (approved):

073-extend-feature-gates-to-all-operators.md
scholzj: @tombentley So, what should we do about this? Everyone seems to have approved it without this being incorporated. But I can still try to incorporate it and restart the reviews if you want to have it in.

tombentley: My concern was it just opens the door to users thinking they've configured something, but not actually, because the operator they're configuring it on doesn't honour it. These are the sorts of dumb mistakes that are among the most frustrating.

I wasn't thinking of a flag-based mechanism that you propose with `supportsTopicOperator` etc. Instead I was thinking about encapsulating the flags so that an operator can only get the enabled flags by passing in the set of flats it supports. So you have some class which does the initial parsing of the env var into some `FeatureFlags` object (which can reject completely unknown flags), but that doesn't directly expose methods for getting the enabled flags. Instead it exposes a method like `operatorScopedFlags(Set<String> operatorFlags)` which accepts the set of flags understood by that operator, and returns the subset of enabled flags which the operator can then use. That allows the `operatorScopedFlags()` to log about flags which we passed by the user, but not requested/consumed by the operator, but without having to model that within `common`. 

Obviously this doesn't allow the rolling avoidance feature you mention, but I guess no one thinks that's a big enough problem (or else they wouldn't have voted to accept this proposal).

All that said, if no one else thinks this is a problem worth solving, or if that just a lot of refactoring, then I'm happy if you proceed with out it. It's something we could revisit if it turned out to be a problem in practice.

scholzj: Ok, I guess given the number of approvals, we should keep it as it is and we can extend it if we see any issues. Thanks.

scholzj:
This proposal now has 5 binding and 3 non-binding approvals. Unless there are any new comments, I will close it tomorrow. Thanks.

scholzj:
Approved with 5 binding and 3 non-binding votes.

PR #116, opened 2024-04-24
Decouple CA and end-entity certificate management

katheris:
This proposal describes how to create better separation between managing the self-signed CA that Strimzi owns and end-entity certificates. It is based on PR #46 which was closed. It proposes a smaller change but still working towards the wider goal of making it possible to use a tool like Cert Manager in Strimzi

PR #119, opened 2024-05-07
 Issue #9372 Enhance Kafka Spec with cert algorithm management

steffen-karlsson:


scholzj:
Thanks for the proposal. I think we need to be careful how does this fit with https://github.com/strimzi/proposals/pull/116.

scholzj (commented):
I think the API looks reasonable. But I also thing that this is a bit more tricky than just adding the API. I think it would be worth elaborating on the following details:
* What will be the suported values?
* How will they be validated?
* How do you ensure that they are supported by both the OpenSSL (where the certificates are issues) as well as by Java (where the certificates are used)
* How will this handle FIPS where different algs might be available than in the non-FIPS mode?
* How will this information be passed to the User Operator?
* It is also worth thinking about how this would relate to when we (possibly) use tools such as Cert Manager in the future

steffen-karlsson:
> I think the API looks reasonable. But I also thing that this is a bit more tricky than just adding the API. I think it would be worth elaborating on the following details:
> 
> * What will be the suported values?
> * How will they be validated?
> * How do you ensure that they are supported by both the OpenSSL (where the certificates are issues) as well as by Java (where the certificates are used)
> * How will this handle FIPS where different algs might be available than in the non-FIPS mode?
> * How will this information be passed to the User Operator?
> * It is also worth thinking about how this would relate to when we (possibly) use tools such as Cert Manager in the future

Sure that makes sense, let me do an FAQ and answer some of these concerns in the proposal it self.

katheris:
I agree with Jakub that some more details in the PR are needed. The original issue links [here](https://github.com/strimzi/strimzi-kafka-operator/blob/main/systemtest/src/main/java/io/strimzi/systemtest/security/SystemTestCertAndKeyBuilder.java#L65) for where it is "hardcoded", but that is actually just system tests. Implementing this will be more involved than just changing those few lines of code, as the use of RSA is assumed in a few places. E.g. in several methods in the OpenSslCertManager class. As a result I suspect it might be challenging to let the user specify whatever they like, so having a set list of supported values makes more sense to me.

Looking at how this fits with future plans to make use of Cert Manager, you can specify the key algorithm and size when requesting certificates, so I think that's fine. The tricky part is the signature hashing algorithm because as far as I can tell it cannot be specified when using Cert Manager, instead it looks like it gets determined by the key size and algorithm https://github.com/cert-manager/cert-manager/issues/3040

katheris:
Hey @steffen-karlsson did you have any time to think about what the supported values would be and Jakub's other questions?

073-kafka-crd-certalg-option.md
```diff
@@ -0,0 +1,51 @@
...
+
+Requested by the community here: https://github.com/strimzi/strimzi-kafka-operator/issues/9372
+
+## Proposal
+
+The proposal will introduce cert configs for the cluster and client CA, this is so that it is possible to use other algorithms than the currently hardcoded RSA. The proposal will add three new attributes to the `clusterCa` and `clientsCa` specs:
+
+* `keyAlgorithm`: The algorithm for generating the private key.
+* `keySize`: The size of the private key generated.
+* `signatureAlgorithm`: The hashing algorithm used for signing.
```
tombentley, line 19: We need more detail about these:

*Are they validated via the CRD Schema, or the Cluster Operator itself? Or do we just pass them to `openssl` and treat algorithms that are unknown to it as any other openssl failure? 
* Are we going to document some key algorithms which we guarantee will be available? How might we, in the future, go about deprecating and removing support for some combination of these parameters?
* Documenting allowed key sizes is tricky because they depend on the key algorithm. 
* Do the same settings apply for both CA and end-entity certificates?
* Are these settings only use for _new_ certificates, or does changing them imply that certificates need to be immediately regenerated?

tombentley (commented):
Thanks for the proposal. It would be nice to give users this level of control, but I think we need to define the API and behaviour in a little more detail.

im-konge:
Triaged on 27.6.2024: @steffen-karlsson are you waiting for Kate's proposal to proceed? Did you have time to look at the comments from Kate and Jakub?

katheris:
@steffen-karlsson did you get a chance to see how this might fit with the cert-manager proposal? In case you missed it my original proposal PR was closed in favour of this new one https://github.com/strimzi/proposals/pull/135

ppatierno:
@steffen-karlsson any news? I don't see any interaction on this since months. Did you also take a look at last Kate comments. Otherwise I am going to close this one if you are not willing to help with this anymore.

PR #120, opened 2024-05-14
proposal: support of additional volumes in pod

cthtrifork:


074-additional-volumes-support.md
```diff
@@ -0,0 +1,95 @@
...
+        driver: csi-driver-name
+        readOnly: true
+        volumeAttributes:
+          secretProviderClass: example-secret-provider-class
+    - name: example-projected-volume
+      path: /path/to/mount/volume
+      projected:
+        sources:
+          serviceAccountToken
+```
```
scholzj, line 62: The template part of the API in general mirrors the Kubernetes API. So, why not do the same here as well:
* Add a `volumes` array in `PodTemplate` that mirrors the volume objects?
* The `Volume` API seems quite big and I have concerns about the CRD size especially during the transition to `v1` API in the future when multiple versions would need to be specified. Could you please add some considerations and analysis for that? Maybe it should be just defined as unstructured `Map<Stirng, Object>` to deal with the API size? 🤔
* Add `volumeMounts` to the `ContainerTemplate` to configure the mount points separately. I think this is important as there are other enhancement requests for things such as sidecars and those might need to share the volumes in some cases.

074-additional-volumes-support.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Supported volumes
+
+The following types of volumes are proposed as initial support. More types could be added later.
+
+- Secret
+- ConfigMap
+- EmptyDir
+- CSI
+- Projected volume
```
scholzj, line 72: It would be good to add some considerations about the security implications this has. Mounting Secrets or projected volumes is possibly dangerous as it allows you to mount things you otherwise might not have access to and use them. I'm not saying this is a blocker, but:
* It should be raised here for consideration while reviewing the proposal
* It should be documentad in the docs as part of the implementation if this is approved

074-additional-volumes-support.md
```diff
@@ -0,0 +1,95 @@
...
+ - Use additional configurations or secret management tools that require mounting extra volumes.
+
+Improve flexibility in managing Kafka deployments by allowing users to attach custom storage solutions based on their specific requirements.
+
+## Proposal
+
+### Considerations
+
+- CRD Changes: Update the Kafka CRDs to include an `additionalVolumes` field.
+- Validation: Ensure that the additional volumes are validated correctly and do not conflict with existing volume definitions such as `/var/lib/kafka/` and `/tmp`.
```
scholzj, line 25: I think this needs more detail as there might be more to it than just duplicate mount paths.
* Duplicate volume names?
* Duplicate or conflicting paths as already mentioned?
* What if the same volume is mounted multiple times? Does it have any implications?

074-additional-volumes-support.md
```diff
@@ -0,0 +1,95 @@
...
+ - Access log files for enhanced monitoring and debugging.
+ - Use additional configurations or secret management tools that require mounting extra volumes.
+
+Improve flexibility in managing Kafka deployments by allowing users to attach custom storage solutions based on their specific requirements.
+
+## Proposal
+
+### Considerations
+
+- CRD Changes: Update the Kafka CRDs to include an `additionalVolumes` field.
```
scholzj, line 24: Please be more exact what CRDs and which parts of them will be affected and which operands will be implemented as `Kafka CRDs` is a bit confusing. I would expect it to be implemented for all CRDs, all operands ... including all individual components of them? E.g. the Kafka CR has not just the Kafka node pods. It has also ZooKeeper pods, Cruise Control pods, Entity OPerator Pods, Kafka Exporter etc. Then there are the others such as Connect, Bridge, MirrorMaker etc. So it would be good to understand hat will be changed, where and how.

The simplest answer is all should change. But I can imagine for example saying that ZooKeeper pods or Mirror Maker 1 pods will be ignored as they will be removed soon etc.

074-additional-volumes-support.md
```diff
@@ -0,0 +1,95 @@
...
+- EmptyDir
+- CSI
+- Projected volume
+
+See initial work in this PR draft:
+<https://github.com/strimzi/strimzi-kafka-operator/pull/10099>
+
+## Affected/not affected projects
+
+<https://github.com/strimzi/strimzi-kafka-operator>
```
scholzj, line 79: Similar to one of the points I raised above -> it would be good to go deeper here and explain what exactly are the affected operands / components.

074-additional-volumes-support.md
```diff
@@ -0,0 +1,95 @@
...
+Improve flexibility in managing Kafka deployments by allowing users to attach custom storage solutions based on their specific requirements.
+
+## Proposal
+
+### Considerations
+
+- CRD Changes: Update the Kafka CRDs to include an `additionalVolumes` field.
+- Validation: Ensure that the additional volumes are validated correctly and do not conflict with existing volume definitions such as `/var/lib/kafka/` and `/tmp`.
+- Documentation: Update the Strimzi documentation to include examples and guidelines on how to use the new `additionalVolumes` field.
+- Security: Ensure that the additional volumes do not introduce security vulnerabilities by validating the configurations.
```
scholzj, line 27: i added some points around security in a comment below. I'm not sure if this can be validated or just have to be accepted as a risk. But it would be good if you can elaborate on what this means.

scholzj (commented):
Thanks for the proposal. I left some initial comments.

074-additional-volumes-support.md
cthtrifork: This is quite an interesting discussion. Looking at other operators and how the kubeapi works, I would propose not to add too much validation here. If the platform engineer defines the custom resource in a invalid way, then kubernetes will do validation and throw errors when reconciling. E.g telling the user that the volume with same name already exists etc.

I do not think we need to reproduce this validation logic from kubeapi into this operator. However, I do see an argument for validating cases such as adding a volume that already conflicts with the *existing* volumes that the operator adds such as `/tmp`. I hope the distinction makes sense.

074-additional-volumes-support.md
cthtrifork: I will investigate the codebase further. When making changes to the PodTemplate.java the CRDGenerator only modified the CRD for KafkaCluster/brokers. I think we should also consider what would requiress less implementation effort, but is made in way so we can extend it over time to support more types.

But again, I think we need to understand the codebase a bit better before scoping this.

074-additional-volumes-support.md
scholzj: I think validating the name is fairly simple and easy to do. So that should be done by the operator. You are right that Kubernetes will validate these things. But it will do so only once the original Pod is deleted which weakens the availability. So I think it needs to be evaluated case by case what does or doesn't make sense -> for names and mount path I think the validation is reasonably simple and should be done.

074-additional-volumes-support.md
cthtrifork: I am not sure why this is a concern from the operator point of view? It would be the ones deploying and operating kafka which has that responsbility and the RBAC setup in kubernetes.

074-additional-volumes-support.md
scholzj: I guess it is up for discussion. The first thing is to at least clarify the intention of the proposal - I do not think it is clear right now. There are also various options. For example having a generic proposal covering everything but implement it only somewhere.

My personal view (might differ from other maintainers) is that I think for example the demand for using custom volumes with Connect is equal if not bigger to the Kafka brokers. Plus I think many of the usecases (including yours) more or less apply to all of the components. So I would probably tend to go with doing all of them.

I did not check your PR. But the [PodTemplate](https://strimzi.io/docs/operators/latest/full/configuring.html#type-PodTemplate-reference) API is used in many places. So is the ContainerTemplate. So any changes would impact all the component APIs and make it strange if only one actually uses it.

074-additional-volumes-support.md
cthtrifork: > The template part of the API in general mirrors the Kubernetes API. So, why not do the same here as well:
> 
> * Add a `volumes` array in `PodTemplate` that mirrors the volume objects?
> * The `Volume` API seems quite big and I have concerns about the CRD size especially during the transition to `v1` API in the future when multiple versions would need to be specified. Could you please add some considerations and analysis for that? Maybe it should be just defined as unstructured `Map<Stirng, Object>` to deal with the API size? 🤔

A proposal would to not have an array of the direct `volumes`but something like this for the PodTemplate:
```
    private List<AdditionalVolume> additionalVolumes;
```

The new datatype would contain specific types we've selected:

```
    private SecretVolumeSource secret;
    private ConfigMapVolumeSource configMap;
    private EmptyDirVolumeSource emptyDir;
    private CSIVolumeSource csi;
```

This would also make the API size much smaller. I would like to avoid `Map<String, Object>` if possible

> * Add `volumeMounts` to the `ContainerTemplate` to configure the mount points separately. I think this is important as there are other enhancement requests for things such as sidecars and those might need to share the volumes in some cases.

Yeah that is fine. I have also talked to some other colleagues and I see operators such as Prometheus uses this approach.

scholzj: Not using directly the Kubernetes Volume API but having out own might be reasonable. I think that we will need way more than just the 4 you mentioned. But it might still save something and keep some control. And we might avoid supporting some of the deprecated in-tree driver volume types. But I think I personally would still call it `volumes`.

scholzj: > But I think I personally would still call it volumes.

I mean the field in the PodTemplate. The class name `AdditionalVolume` is of course fine.

cthtrifork: > `volumes` vs `additionalVolumes`
I think it is a bit misleading, and there will already be several other volumes present that is created automatically by the operator, but not visible in the CRD schema.

However, I have seen both naming convention used in other operators and I do not have a strong opinion here, so I am fine with getting inline with recommendations from the maintenance 😄

074-additional-volumes-support.md
tombentley: Would this allow a route to running code that's not part of the container image, and therefore isn't immutable?

I guess I'm thinking of what's proposed here in conjunction with changing the `$CLASSPATH` to include a mounted directory and loading code/plugins from there? TBH I don't recall if we try to prevent users setting the `$CLASSPATH`, but even if we do, this would still be possible for Kafka Connect, because it uses a configurable class loading mechanism. 

I'm sure there are users who would like to be able to do this (and avoid the steps needed to include plugin jars in their own container images), but I'm not so sure that the project wants to support users who did this.

tombentley (commented):
Thanks for the proposal!

074-additional-volumes-support.md
AndersBennedsgaard: I agree that this isn't an RBAC issue, since Kubernetes RBAC can't be used to disallow a Pod from mounting a Secret. Regardless, I don't really see an issue here. It is a design choice by the Kubernetes developers that Pods are able to mount Secrets in the same namespace, so Kubernetes users should never place Secrets in a namespace where a Pod shouldn't have access to it.
Adding a note about a possible security issue also seems weird, since it is already possible to access arbitrary secrets in the namespace using several mechanisms in the CRD. For example, it is possible to use something like:

```yaml
template:
  kafkaContainer:
    env:
    - name: VERY_SECRET_SECRET
      valueFrom:
        secretKeyRef:
          name: my-secret
          key: some-key
    envFrom:
    - secretRef:
        name: my-secret
```
 
The `ContainterTemplate` documentation does not contain a security notice either: https://strimzi.io/docs/operators/latest/configuring#type-ContainerTemplate-reference

AndersBennedsgaard: I agree that this isn't an RBAC issue, since Kubernetes RBAC can't be used to disallow a Pod from mounting a Secret.
Regardless, I don't really see an issue here. It is a design choice by the Kubernetes developers that Pods are able to mount Secrets in the same namespace, so cluster admins should never place Secrets in a namespace where a Pod shouldn't have access to it.
Adding a note about a possible security issue also seems weird, since it is already possible to access arbitrary secrets in the namespace using several mechanisms in the different CRDs, and none of these ways have a warning about it

074-additional-volumes-support.md
AndersBennedsgaard: @tombentley I agree that users shouldn't mount arbitrary code such as plugins using these custom volumes. So we should probably add a note in the documentation about this not being recommended

074-additional-volumes-support.md
```diff
@@ -0,0 +1,95 @@
...
+See initial work in this PR draft:
+<https://github.com/strimzi/strimzi-kafka-operator/pull/10099>
+
+## Affected/not affected projects
+
+<https://github.com/strimzi/strimzi-kafka-operator>
+
+## Compatibility
+
+The proposed changes are designed to be forward-compatible. Future updates to the Strimzi Kafka Operator can build upon this feature without breaking existing configurations.
```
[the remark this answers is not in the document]
AndersBennedsgaard: That is true. As a workaround, we could either enforce users to always use `/mnt` or similar as a root mounting point, or add a warning about not overriding anything managed by Strimzi.

Personally, I would prefer to have the power to mount anything, anywhere, since that would allow users to configure the instances to very specific use-cases.

cthtrifork:
@tombentley @scholzj Hi maintainers. Thanks for all the feedback. We can continue investigating and updating the proposal + PR, but how can we reach a verdict if it make sense to go forward the with the PR?

I can see on the feedback that we need to ensure some basic validation, ensure the API schema does not become too large and we need some security statements/recommendations. All these seem to be something that can be managed with collaboration.

Any blockers or major design details we still need to cover before putting effort into implementation details?

074-additional-volumes-support.md
scholzj: > Would this allow a route to running code that's not part of the container image, and therefore isn't immutable?
> 
> I guess I'm thinking of what's proposed here in conjunction with changing the `$CLASSPATH` to include a mounted directory and loading code/plugins from there? TBH I don't recall if we try to prevent users setting the `$CLASSPATH`, but even if we do, this would still be possible for Kafka Connect, because it uses a configurable class loading mechanism.
> 
> I'm sure there are users who would like to be able to do this (and avoid the steps needed to include plugin jars in their own container images), but I'm not so sure that the project wants to support users who did this.

I share the concern about this being used to add runtime dependencies. But I'm not entirely convinced we should try to prevent it beyond discouraging it in the docs. It seems some users want to do it and the idea of playing some game where we try to plug the way to change the classpath with them seems like a waste of effort.

So I guess I would stick to docs and possibly enforce some mount path to avoid them remounting some existing binary directories directly etc.

scholzj:
> @tombentley @scholzj Hi maintainers. Thanks for all the feedback. We can continue investigating and updating the proposal + PR, but how can we reach a verdict if it make sense to go forward the with the PR?
> 
> I can see on the feedback that we need to ensure some basic validation, ensure the API schema does not become too large and we need some security statements/recommendations. All these seem to be something that can be managed with collaboration.
> 
> Any blockers or major design details we still need to cover before putting effort into implementation details?

In general, it is an iterative process. So you should go through the comments you got so far and:
* Try to incorporate them if you agree with them
* Discuss them if you disagree (and essentially try to offer a better alternative or convince us why is your way better)

And once done, there might be another round of comments. It would be great to have comments from more @strimzi/maintainers to make sure you will not get some pushback against the changes requested by me or Tom. But it is often the reality that others might see that there are already open comments and wait for them to be incorporated before doing their reviews. If you want, we could also use the [community calls](https://github.com/strimzi/strimzi-kafka-operator?tab=readme-ov-file#strimzi-community-meetings) for some more interactive discussions (or if needed we can have some dedicated call to make sure the timezones work for the relevant people). 

As for the PR with the PoC -> that is for sure useful to point out any issues or test things. But there is no need to keep it up-to-date with every single change in the proposal right now. That can be all updated once the proposal is approved.

Hope this helps.

MichaelMorrisEst:
I would be interested in seeing this implemented more generally than just the Kafka pods and am happy to help with the implementation effort to achieve it

cthtrifork:
> I would be interested in seeing this implemented more generally than just the Kafka pods and am happy to help with the implementation effort to achieve it

Sounds great. Can you help update the proposal? I am helping a collague to understand the codebase, so he can assist with proposal and later PR.

MichaelMorrisEst:
> > I would be interested in seeing this implemented more generally than just the Kafka pods and am happy to help with the implementation effort to achieve it
> 
> Sounds great. Can you help update the proposal? I am helping a collague to understand the codebase, so he can assist with proposal and later PR.

Sure, no problem

MichaelMorrisEst:
> > I would be interested in seeing this implemented more generally than just the Kafka pods and am happy to help with the implementation effort to achieve it
> 
> Sounds great. Can you help update the proposal? I am helping a collague to understand the codebase, so he can assist with proposal and later PR.

I have some updates ready to push. I think in order to push them directly to the proposal you need to add me as a collaborator in the settings on your https://github.com/trifork/strimzi-proposals fork. If you would rather not do that, I can add as comments, or I am open to any other suggestions

cthtrifork:
> I have some updates ready to push. I think in order to push them directly to the proposal you need to add me as a collaborator in the settings on your https://github.com/trifork/strimzi-proposals fork. If you would rather not do that, I can add as comments, or I am open to any other suggestions

Ah yes, of cause. I have invited you as an external collaborator!

MichaelMorrisEst:
This recent change https://github.com/strimzi/strimzi-kafka-operator/issues/2779 is causing an issue for supporting EmptyDir - the io.fabric8.kubernetes.api.model.EmptyDirVolumeSource class contains a field of type io.fabric8.kubernetes.api.model.Quantity and that class does not have the @JsonPropertyOrder annotation or the @JsonInclude annotation. This code is from a different project so it may be difficult to add the annotation without a good reason to do so. This could also potentially be an issue for other classes in the future.

Perhaps the new checks introduced by the above change should  be limited to classes in the strimzi project (i.e. io.strimzi package)? Or does anybody have any other suggestions on how to overcome this?

074-additional-volumes-support.md
scholzj: I think this still should be covered here and in the future docs. As I said, I do not think it is a blocker and it obviously cannot be easily avoided unless we significantly limit what can be mounted. But it is a notable thing that should be documented in the docs in the future as well as here in the proposal.

> Adding a note about a possible security issue also seems weird, since it is already possible to access arbitrary secrets in the namespace using several mechanisms in the different CRDs, and none of these ways have a warning about it

Connect supports mounting custom Secrets or Config Maps through the CRD. Secrets are allowed also for TLS purposes. We do not support things such as Downward API. We also do not support it for all kind of containers. It is also less obvious and more hidden so it might not be clear on a quick look during a security review. So I think this is different.

TBH, I'm not sure I understand why it seems to be such a problem - I'm not blocking anything, I'm not asking to rewrite Kubernetes to prohibit it ... just asking for it to be properly documented in a paragraph or two.

074-additional-volumes-support.md
scholzj: I still think this should go into more details what will be validated, what will be the possible check for the volume path etc.

074-additional-volumes-support.md
```diff
@@ -0,0 +1,127 @@
...
+            - name: temp
+              path: /tmp/logs
+            - name: example-csi-volume
+              path: /path/to/mount/csi-volume
+            - name: example-projected-volume
+              path: /path/to/mount/proj-volume
+
+```
+
+### Supported volumes
```
scholzj, line 94: Some stuff I'm wondering about ...
* Should [PVC](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#claims-as-volumes) be added to the list as well? Wouldn't PVCs be the most common way to create the volume?
* Should Projected volumes be really on the list? What do you need it for? Do you have any use-case? It does not seem to match any of the motivations for this I'm aware of. (It does not seem to be part of the PR anyway)

074-additional-volumes-support.md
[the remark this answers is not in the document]
scholzj: I'm not sure I would want to limit it only to `/mnt`, But I do not think we want to allow anything. We definitely want to prevent using the paths Strimzi is using => for the data, for tmp, for the binaries.

scholzj (commented):
I took another pass through it. I do not think there are any real blockers. But we should get all the previous comments incorporated into the proposal. We can discuss more on the community call.

074-additional-volumes-support.md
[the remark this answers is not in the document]
tombentley: @scholzj do you have something specifically in mind that would be prevented by limiting users to `/mnt`? Because it would be simple to implement and simple to explain. If someone later found it wasn't enough we could always allow more in addition.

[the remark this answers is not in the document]
scholzj: I don't. I guess you are right. If we limit it to something like `/mnt` now, it is easy to open it more later if needed.

cthtrifork:
> This recent change [strimzi/strimzi-kafka-operator#2779](https://github.com/strimzi/strimzi-kafka-operator/issues/2779) is causing an issue for supporting EmptyDir - the io.fabric8.kubernetes.api.model.EmptyDirVolumeSource class contains a field of type io.fabric8.kubernetes.api.model.Quantity and that class does not have the @JsonPropertyOrder annotation or the @JsonInclude annotation. This code is from a different project so it may be difficult to add the annotation without a good reason to do so. This could also potentially be an issue for other classes in the future.
> 
> Perhaps the new checks introduced by the above change should be limited to classes in the strimzi project (i.e. io.strimzi package)? Or does anybody have any other suggestions on how to overcome this?

As discussed in the community call, we should limit it to classes in the strimzi project. We will make this change in our upcoming PR:

074-additional-volumes-support.md
```diff
@@ -0,0 +1,127 @@
...
+
+
+It is proposed to not add support in the following locations - if `volumes` or `volumeMounts` are specified in the PodTemplate or ContainerTemplate in any of these locations it will be ignored:
+
+|Pod             |CRD             |Schema Location                                          |Reason                                 |
+|----------------|----------------|---------------------------------------------------------|---------------------------------------|
+|KafkaMirrorMaker|KafkaMirrorMaker|spec -> template -> pod/mirrorMakerContainer             |KafkaMirrorMaker has been deprecated   |
+|Kafka           |Kafka           |spec -> jmxTrans -> template -> pod/container            |JmxTrans no longer supported           |
+|EntityOperator  |Kafka           |spec -> entityOperator -> template -> tlsSidecarContainer|tlsSidecarContainer no longer supported|
+|CruiseControl   |Kafka           |spec -> cruiseControl -> template -> tlsSidecarContainer |tlsSidecarContainer deprecated         |
```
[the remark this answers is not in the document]
cthtrifork: @MichaelMorrisEst Will you adjust this in the proposal?

074-additional-volumes-support.md
cthtrifork: We will rework this. We will remove projected volumes and look more into CSI volumes vs. PVC

074-additional-volumes-support.md
```diff
@@ -0,0 +1,128 @@
...
+## Current situation
+
+The operator handles many configurations and operational aspects but lacks support for specifying additional volumes in the Custom Resource Definition (CRD). This limitation can restrict users from accessing log files or integrating with external storage solutions directly.
+
+## Motivation
+
+Adding support for additional volumes in the CRD will allow users to:
+
+ - Access log files for enhanced monitoring and debugging.
+ - Use additional configurations or secret management tools that require mounting extra volumes.
```
cthtrifork, line 16: - Using for analyzing JVM issues (e.g. heap dumps)

And prepare support for future use-cases such as:

- Using Kafka Connect connectors to read/write files to/from the disk
- Using it for tiered storage
- CSI Secrets
- Adding plugins to the container image

cthtrifork, line 16:
```suggestion
 - Use additional configurations or secret management tools that require mounting extra volumes.
 - Using for analyzing JVM issues (e.g. heap dumps)

And prepare support for future use-cases such as:

- Using Kafka Connect connectors to read/write files to/from the disk
- Using it for tiered storage
- CSI Secrets
- Adding plugins to the container image
```

074-additional-volumes-support.md
```diff
@@ -0,0 +1,136 @@
+# Support of additional volumes in pod
+
+We propose to enhance the Strimzi Kafka Operator by adding support for specifying additional volumes in the CRDs. This will involve modifying the Kafka CRDs (see details below) to include a `volumes` field which will allow users to define extra volumes, and a `volumeMounts` field which will allow users to define extra volume mounts.
+
+The specific usecase of ours, is that we want to set custom log4j logging for audit logs, to ensure that audit logs are *not* present in stdout, but can be picked up by a sidecar which uses OpenTelemtry or logstash to forward to the audit logs to a secure location.
```
cthtrifork, line 5:
```suggestion
The specific usecase of ours, is that we want to set custom log4j logging for audit logs, to ensure that audit logs are *not* present in stdout, but can be picked up by a sidecar which uses OpenTelemtry or logstash to forward to the audit logs to a secure location.
See motivation for more use-cases and see also original discussion in <https://github.com/strimzi/strimzi-kafka-operator/issues/3693>
```

074-additional-volumes-support.md
cthtrifork: As discussed in the community call, we will limit the path to `/mnt` to ensure backwards compatibility further

074-additional-volumes-support.md
```diff
@@ -0,0 +1,137 @@
...
+- Adding plugins to the container image
+
+Improve flexibility in managing Kafka deployments by allowing users to attach custom storage solutions based on their specific requirements.
+
+## Proposal
+
+### Considerations
+
+- CRD Changes: Update the Kafka CRDs to include a `volumes` and a `volumeMounts` field.
+- Validation: Ensure that the additional volumes are validated correctly and do not conflict with existing volume definitions such as `/var/lib/kafka/` and `/tmp`.
```
cthtrifork, line 34:
```suggestion
- Validation: All additional mounted paths should be inside `/mnt` to ensure backwards compatibility for the evolution of kafka and this operator.

```

074-additional-volumes-support.md
cthtrifork: We will rework this. We will remove projected volumes and look more into CSI volumes vs. PVC.
As discussed in the community call, this should remove most security concerns.

074-additional-volumes-support.md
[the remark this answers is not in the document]
cthtrifork: We will limit it to `/mnt` for now, as discussed in the community call

074-additional-volumes-support.md
scholzj: The CSI volumes actually might have had some value for the CSI secrets. But its fine, we can limit the scope and add it later when needed 👍

074-additional-volumes-support.md
```diff
@@ -0,0 +1,138 @@
...
+|----------------|----------------|---------------------------------------------------------|---------------------------------------|
+|KafkaMirrorMaker|KafkaMirrorMaker|spec -> template -> pod/mirrorMakerContainer             |KafkaMirrorMaker has been deprecated   |
+|Kafka           |Kafka           |spec -> jmxTrans -> template -> pod/container            |JmxTrans no longer supported           |
+|EntityOperator  |Kafka           |spec -> entityOperator -> template -> tlsSidecarContainer|tlsSidecarContainer no longer supported|
+|CruiseControl   |Kafka           |spec -> cruiseControl -> template -> tlsSidecarContainer |tlsSidecarContainer deprecated         |
+
+
+An example configuration could look as follows:
+
+```yaml
```
scholzj, line 68: Can you update the example to match the other changes? E.g. the `/mnt/...` muont paths, the volume types supported etc.

ppatierno:
I had another pass. No more comments from me, but waiting for addressing the example YAML to be fixed as requested by Jakub.

074-additional-volumes-support.md
```diff
@@ -0,0 +1,138 @@
...
+- Adding plugins to the container image
+
+Improve flexibility in managing Kafka deployments by allowing users to attach custom storage solutions based on their specific requirements.
+
+## Proposal
+
+### Considerations
+
+- CRD Changes: Update the Kafka CRDs to include a `volumes` and a `volumeMounts` field.
+- Validation: All additional mounted paths should be inside `/mnt` to ensure backwards compatibility for the evolution of kafka and this operator.
```
MichaelMorrisEst, line 34: I thought we agreed on a blacklist rather than /mnt?

cthtrifork: I am quite sure there was a pushback from maintainers to use `/mnt` for now, as Strimzi would like to have a lot of responsibility regarding backwards compatible with users of the operators. 

EDIT: Just double checked: https://youtu.be/_8xd_NsM2Fo?si=4qmgq-q7pI-6XvDk&t=1122

074-additional-volumes-support.md
```diff
@@ -0,0 +1,141 @@
...
+                volumeAttributes:
+                  secretProviderClass: example-secret-provider-class
+            - name: example-projected-volume
+              projected:
+                sources:
+                  serviceAccountToken
+          kafkaContainer:
+            volumeMounts:
+            - name: example-configmap
+              path: /path/to/mount/cm-volume
```
cthtrifork, line 96:
```suggestion
              path: /mnt/cm-volume

```

074-additional-volumes-support.md
```diff
@@ -0,0 +1,141 @@
...
+            - name: temp
```
cthtrifork, line 97:
```suggestion


```

074-additional-volumes-support.md
```diff
@@ -0,0 +1,141 @@
...
+              path: /tmp/logs
```
cthtrifork, line 98:
```suggestion
```

074-additional-volumes-support.md
```diff
@@ -0,0 +1,141 @@
...
+            - name: example-csi-volume
+              path: /path/to/mount/csi-volume
```
cthtrifork, line 100:
```suggestion
```

074-additional-volumes-support.md
```diff
@@ -0,0 +1,141 @@
...
+            - name: example-projected-volume
```
cthtrifork, line 101:
```suggestion
            - name: example-pvc-volume

```

074-additional-volumes-support.md
```diff
@@ -0,0 +1,141 @@
...
+              path: /path/to/mount/proj-volume
```
cthtrifork, line 102:
```suggestion
              path: /mnt/pvc-volume

```

074-additional-volumes-support.md
ppatierno: Actually I supported @tombentley to use `/mnt`

cthtrifork:
> I had another pass. No more comments from me, but waiting for addressing the example YAML to be fixed as requested by Jakub.

What do you think? is the proposal ready for a vote or is there final feedback?

scholzj (approved):
LGTM. Thanks.

074-additional-volumes-support.md
```diff
@@ -14,51 +15,85 @@ Adding support for additional volumes in the CRD will allow users to:
...
 ## Proposal
 
 ### Considerations
 
-- CRD Changes: Update the Kafka CRDs to include an `additionalVolumes` field.
-- Validation: Ensure that the additional volumes are validated correctly and do not conflict with existing volume definitions such as `/var/lib/kafka/` and `/tmp`.
-- Documentation: Update the Strimzi documentation to include examples and guidelines on how to use the new `additionalVolumes` field.
+- CRD Changes: Update the Kafka CRDs to include a `volumes` and a `volumeMounts` field.
+- Validation: All additional mounted paths should be inside `/mnt` to ensure backwards compatibility for the evolution of kafka and this operator.
+- Documentation: Update the Strimzi documentation to include examples and guidelines on how to use the new `volumes` and `volumeMounst` fields.
```
tombentley, line 35:
```suggestion
- Documentation: Update the Strimzi documentation to include examples and guidelines on how to use the new `volumes` and `volumeMounts` fields.

```

tombentley (approved):
LGTM, thank you!

im-konge (approved):
LGTM, thanks for the proposal :)

ppatierno (approved):
LGTM. Left just a nit. Thanks!

074-additional-volumes-support.md
```diff
@@ -0,0 +1,135 @@
...
+
+| Pod              | CRD              | Schema Location                              | Reason                               |
+| ---------------- | ---------------- | -------------------------------------------- | ------------------------------------ |
+| KafkaMirrorMaker | KafkaMirrorMaker | spec -> template -> pod/mirrorMakerContainer | KafkaMirrorMaker has been deprecated |
+
+Support will also not be added to the following locations which are no longer supported:
+
+|Kafka           |Kafka           |spec -> jmxTrans -> template -> pod/container            |^M
+|EntityOperator  |Kafka           |spec -> entityOperator -> template -> tlsSidecarContainer|^M
+|CruiseControl   |Kafka           |spec -> cruiseControl -> template -> tlsSidecarContainer |^M
```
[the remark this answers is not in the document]
cthtrifork: @MichaelMorrisEst  Can you help here?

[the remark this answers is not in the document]
MichaelMorrisEst: windows CR character was duplicated somehow, corrected now

scholzj:
@strimzi/maintainers This now has the required votes to be approved. Please have a look at this. If there are no more comments, I will close it at Thursday.

scholzj:
@cthtrifork In order to merge it, can you please:
* Rename the proposal to `75` which is the next available number
* Add a link to it in https://github.com/strimzi/proposals/blob/main/README.md?

Thanks. Once done, I will close the vote and merge the PR.

cthtrifork:
> @cthtrifork In order to merge it, can you please:
> 
> * Rename the proposal to `75` which is the next available number
> * Add a link to it in https://github.com/strimzi/proposals/blob/main/README.md?
> 
> Thanks. Once done, I will close the vote and merge the PR.

It will have to be slightly postponed. During test on our live cluster, we found an issue when having multiple (3) brokers and using persistent volume claims. The current setup would require the volume to be `ReadWriteMany` as we are sharing the volume. This was not the intended design and each pod should have its own volume with appended broker name `kafka-0, kafka-1` etc.
I think we will need to redesign to support `Volume Claim Templates` like in ´statefulsets´ but for strimzypodset`. See <https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#volume-claim-templates>.

This changes the implementation design a bit and I think we need to add `Volume Claim Templates` details in this proposal.

Any objections @MichaelMorrisEst  ?

cthtrifork:
> Well, I thought that was expected - at least it was expected by me from the design ... and as far as I'm concerned it is also desired feature.

So you prefer to finish this proposal and design and open a new proposal for supporting `PersistentVolumeClaimSet` ?
I think that is a good idea, so we get some progress.

> * I'm skeptical that you can actually implement it in a reasonable way from this proposal. If you wanna do per-pod stuff, you might want to consider using some Mutating webhook to modify the pods?

I think we can add an additional type, so it is

- Secret
- ConfigMap
- EmptyDir
- PersistentVolumeClaims
- **PersistentVolumeClaimsSet** 

which then would create one `PersistentVolumeClaimSource` PER pod when creating containers. This would archive the same as using native statefulsets and Persistent Volume Templates. I think we can do it in the existing operator logic without too many additions of changes.

cthtrifork:
If you think the current design is the desired feature, I think we should merge this and we will create a new proposal in the future for Persistent Volume sets - for a new discussion.

I have renamed the proposal to `75`

@scholzj

scholzj:

Thanks for the proposal.

PR #121, opened 2024-05-21
Add support for managing offsets in Connect

katheris:


074-connector-offsets-support.md
```diff
@@ -0,0 +1,180 @@
...
+* PATCH /connectors/{connector}/offsets
+* DELETE /connectors/{connector}/offsets
+* PUT /connectors/{connector}/stop
+
+Support for the "stop" endpoint was added via [PR 9095](https://github.com/strimzi/strimzi-kafka-operator/pull/9095).
+Currently users that are utilising the KafkaConnector resource cannot make use of the API endpoints for managing offsets.
+
+## Motivation
+
+Previously if users wanted to list, update or delete offsets for a connector they had to interact with the relevant Kafka topics directly.
```
mimaison, line 19:
```suggestion
Previously if users wanted to list, update or delete offsets for a connector they had to interact with the relevant consumer groups for sink connectors or directly write to the Kafka Connect internal offsets topic for source connectors.
```

074-connector-offsets-support.md
```diff
@@ -0,0 +1,180 @@
...
+      }
+    }
+  ]
+}
+```
+
+### Listing offsets
+The KafkaConnector resource will be updated with a new property to enable the current offsets to be listed in the resource status.
+
+The new property will be a boolean property called `showOffsets` and default to `false`. It should default to `false` because enabling it requires the operator to do an additional API call every time it reconciles the connector.
```
mimaison, line 78: `every time it reconciles the connector`, How often is that?

074-connector-offsets-support.md
```diff
@@ -0,0 +1,180 @@
...
+  This is to simplify the logic and means we don't have to account for what happens if the KafkaConnector spec has been updated but the connector isn't stopped yet.
+
+### Resetting offsets
+
+To reset the offsets for a particular connector the user will annotate the KafkaConnector resource with a new annotation: `strimzi.io/offsets-reset=true`.
+When the reset has been completed the operator will remove this annotation from the resource.
+
+If the reset fails the operator will add a condition to the KafkaConnector resource called `strimzi.io/offsets-reset` to communicate the failure and remove the annotation.
+
+Strimzi will shortcut and automatically fail to do the reset if the KafkaConnector resource is not stopped.
```
mimaison, line 154: Could Strimzi stop the connector if needed?

074-connector-offsets-support.md
```diff
@@ -0,0 +1,180 @@
...
+* Skipping records that cause issues with a connector and cannot be addressed using existing error-handling features
+
+## Proposal
+
+Strimzi should include mechanisms to let users manage offsets when utilising the KafkaConnector resources.
+This proposal suggests different approaches for each action, since they are all slightly different.
+In summary:
+ - listing offsets via the KafkaConnector status and enabled/disabled through a property
+ - altering offsets via a new resource called KafkaConnectorOffsets
+ - deleting offsets via an annotation on the KafkaConnector resource
```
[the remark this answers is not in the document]
katheris: I hadn't, but presumably it wouldn't be hard to implement, it would just require additional steps for the operator to follow.

For deleting offsets when deleting a connector, I guess the operator would have to do the following steps:
1. stop the connector
2. reset offsets
3. delete the connector

I think we would want a way to disable it, but I wonder if it should be the default behaviour since you can now stop connectors, I'm not sure why you would delete it if you wanted the offsets to be maintained.

For specifying offsets when creating a connector the steps would be:
1. create the connector in a stopped state (we don't currently support this as far as I know but Connect has made it possible)
2. alter the offsets
3. resume the connector
With the current proposal I'm not sure how this would fit in, it would perhaps work better to have the user do it. Although it could be an argument for not having the separate resource, then it would be easier for the operator to do everything.

What do you think?

074-connector-offsets-support.md
katheris: I believe every 2 minutes (similar to the cluster operator), and then if something changes in the KafkaConnector or KafkaConnect CR.

I realise that isn't very often if users are wanting live updates of the offsets. However it's enough to check whether a connector is progressing or not. And if the user is wanting to list offsets to see where a connector got stuck then they only need it to be updated once, since the connector is likely in a failed state.

074-connector-offsets-support.md
katheris: It could, however the reason I chose not to do that is because I don't think it's best practice for the Strimzi operator to update the spec in the KafkaConnector resource. And it would need to change the `state` to `stopped`

scholzj (commented):
Thanks for the proposal. I left some comments ... but apart from them, I have a bit mixed feelings about it. 

I think it is well designed in terms of the API and I'm not sure it could be done in some significantly better way. But it seems very _expensive_ for something what seems like a pretty niche feature:
* The initial implementation will be non-trivial
* The maintenance effort will be very significant because you have a new custom resource watches, CRDs, you will need to have some metrics => historically, all of these are very expensive to maintain.

So I wonder if it is really worth the effort. And I guess for me it falls to whether we still expect that we can disable the Connect REST API access completely to users using the KafkaConnector resources. If we still think we can cut off the REST API access, why haven't we done it yet? Or if we think it is not something we can really do and that users will always want the REST API access, isn't it better to leave this on the REST API and focus on securing the REST API?

Or did I somehow missed the importance of this feature and it is now (despite being missing for many years) expected that users will be reseting the offsets twice per day?

074-connector-offsets-support.md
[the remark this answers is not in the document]
katheris: Once you delete the connector, you can't delete, change or list them via the `/offsets` endpoint anymore. You can still do it by directly interacting with the either the consumer group for sink connectors, or the topic for source connectors.

074-connector-offsets-support.md
katheris: Is it possible to set up the watch in some way so that it doesn't trigger a reconciliation when the offset part of the status is updated? Or could we perhaps have some configurable timeout value and we only list offsets once the timeout is hit?

074-connector-offsets-support.md
```diff
@@ -0,0 +1,180 @@
...
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaConnector
+#...
+status:
+  #...
+  connectorOffsets:
+    lastUpdateTime: "2024-05-21T09:08:58.467712198Z"
+    source:
+      - partition: "{\"connector-determined-partition\": \"1\"}" # value is a String containing JSON as received from the API
+        offset: "{\"connector-determined-offset\": \"1\"}" # value is a String containing JSON as received from the API
```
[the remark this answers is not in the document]
katheris: The structure of the partition and offset is determined by the connector you are running, so yeah it's a completely random structure from the POV of Strimzi. There's no way to predetermine what it would be. In terms of size I guess it's limited to at least the size of a Kafka record, since it is stored in Kafka, but otherwise there is no limitation. I think the CM idea is a good alternative, I'll have a look at incorporating that into the proposal.

074-connector-offsets-support.md
```diff
@@ -0,0 +1,180 @@
...
+      - partition:
+          kafkaTopic: "my-topic"
+          kafkaPartition: 1
+        offset:
+          kafkaOffset: 0
+```
+
+Notes:
+* Only one of the `source` and `sink` properties will be present.
+* The operator will update the `lastUpdateTime` whenever it updates the offsets section of the status.
```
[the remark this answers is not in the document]
katheris: When you say "to not update it just because of the timestamp" do you mean to not trigger a never-ending loop liked discussed earlier in the proposal?

074-connector-offsets-support.md
```diff
@@ -0,0 +1,180 @@
...
+    - partition:
+        kafkaTopic: "my-topic"
+        kafkaPartition: 1
+      offset:
+        kafkaOffset: 0
+```
+
+Notes:
+* Only one of `sourceOffsets` and `sinkOffsets` should be specified.
+* The format matches the format shown in the status when listing offsets so it can be copy pasted here.
```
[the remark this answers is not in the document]
katheris: I don't follow this question, I meant that I have chosen to use the same format in the offsets CR as the status section of the KafkaConnector CR, just to make it easier for people interacting with the CRs

README.md
```diff
@@ -8,7 +8,8 @@
 This repository list of proposals for the Strimzi project. A template for new proposals can be found [here](./000-template.md).
 
 |  #  | Title                                                                 |
-| :-: |:----------------------------------------------------------------------|
+|:---:|:----------------------------------------------------------------------|
+| 74  | [Connector Offsets Support](./074-connector-offsets-support.md) |
```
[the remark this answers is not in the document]
katheris: Thanks, but don't you still have the problem of having to renumber the proposal md file either way?

074-connector-offsets-support.md
```diff
@@ -0,0 +1,180 @@
...
+* Both the `offset` properties are optional since the API allows these to be passed as `null` (this resets the offsets for that partition).
+* When the offset patch has been completed the status will have a `Ready` condition of `true`.
+* If the request fails the status will include the response from the API.
+* Strimzi will not attempt to verify that the connector is already stopped, it will rely on the API failing if that is the case.
+  This is to simplify the logic and means we don't have to account for what happens if the KafkaConnector spec has been updated but the connector isn't stopped yet.
+
+### Resetting offsets
```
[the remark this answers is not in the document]
katheris: The operator will use the DELETE API, so all partitions and offsets for the connector are reset to 0. So yes if they wanted to reset a single partition they would have to do offset modification

074-connector-offsets-support.md
scholzj: No, the watch fires at every update. The onUpdate method of the watch gives you the old and new resource, so I guess you could do some filtering there and ignore the watch if only the offsets differ. Btu we do not do anything like that anywhere else, so not sure we want to go that way.

You could probably have some timestamp check and for example update the offsets only if they are over 1 minute old ... we don't do things like that anywhere, but technically it should not be a problem I guess.

That said, none of this would solve the size issues related to the resource status. So we should first figure out if that is a problem or not.

074-connector-offsets-support.md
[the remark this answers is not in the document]
scholzj: Yeah. With updating the timestamp even without the offset change just to indicate that at this time these offsets still applied it will be just the timestamp change triggering the loop. So you will have a busy loop even for connectors that are almost idle. That is what I meant I think.

074-connector-offsets-support.md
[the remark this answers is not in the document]
scholzj: Just keep in mind that CMs still have size limits. their main advantage would be that:
* If needed you can probably compress the JSON
* If you are too big, you would get a nicer error to share in the actual resource status and tell the user to disable the offset collection

But the size limit will still be there, so if there is a concern that it will be a common issue, it might be still not sufficient (I have no idea how much of a niche situation that would be to be honest).

074-connector-offsets-support.md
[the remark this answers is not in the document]
scholzj: So, if I got it right, the offset format is defined by the connector. So what happens if I tell Strimzi to set the offsets to a JSON that is not valid for given connector? Are there any validation facilities in Connect for this? Or will Connect just store it in the topic and break the connector?

The logic that it can be copy-pasted is fine. But you still expect the user to change the current status, so there is a pretty big space to create mistakes.

katheris:
> Thanks for the proposal. I left some comments ... but apart from them, I have a bit mixed feelings about it.
> 
> I think it is well designed in terms of the API and I'm not sure it could be done in some significantly better way. But it seems very _expensive_ for something what seems like a pretty niche feature:
> 
> * The initial implementation will be non-trivial
> * The maintenance effort will be very significant because you have a new custom resource watches, CRDs, you will need to have some metrics => historically, all of these are very expensive to maintain.
> 
> So I wonder if it is really worth the effort. And I guess for me it falls to whether we still expect that we can disable the Connect REST API access completely to users using the KafkaConnector resources. If we still think we can cut off the REST API access, why haven't we done it yet? Or if we think it is not something we can really do and that users will always want the REST API access, isn't it better to leave this on the REST API and focus on securing the REST API?
> 
> Or did I somehow missed the importance of this feature and it is now (despite being missing for many years) expected that users will be reseting the offsets twice per day?

I wouldn't say it is a niche feature, but I do think it is something that users won't do especially frequently. I think for me the most valuable feature is being able to delete offsets before deleting a connector. So at a minimum I think we should support the delete offset path, which also happens to be the simplest and least additional maintenance.

As a bonus we could also look at providing an option that allows users to request we clean up the offsets before deleting a connector, although I don't think we'd be able to do that in a way that guarantees it was successful (e.g. if the operator goes down half way through stopping, deleting offsets and deleting the connector).

In terms of supporting securing the Connect REST API that's still stuck behind the other certificate work which is obviously progressing, but isn't a small task.

In terms of listing and altering offsets, I do think those are less likely to be used often. If we switched to using a configmap as you suggested for the altering of offsets then perhaps we could do it without a separate CR to reduce the additional maintenance? Then for listing offsets we could again reduce the support, so e.g allow a user to request the offsets using an annotation or something, but make it a one time thing, so to get the updated offsets you have to add the annotation again.

@mimaison what are your thoughts on how frequently these endpoints would be used?

mimaison:
These update/delete endpoints are not expected to be used a lot apart during development. You typically only need to update offsets when setting up a new pipeline or when handling issues like an unprocessable record.

The list offsets endpoint is useful to monitor a connector and validate it is making progress. This is especially the case with MirrorMaker at least until [KIP-971](https://cwiki.apache.org/confluence/display/KAFKA/KIP-971%3A+Expose+replication-record-lag+MirrorMaker2+metric) gets implemented.

Still being able to fully operate connectors via CRDs would be nice.

Securing the REST API is in my opinion more fundamental but it's an orthogonal issue and seems to be "in progress".

katheris:
@mimaison do you think it will be a common usecase to want to continually fetch offsets, or would having the user annotate the CR every time they want the offsets to be listed be sufficient for most usecases?

mimaison:
I'm not sure continually fetching offsets is very important. You can track consumer offsets for sink connectors and for source connectors the offsets are specific for the connector and may not may a lot of users. In my opinion the main values of the API are 1) retrieving offsets to update them (or apply them to another Connect cluster), 2) see they are changing to validate the connector is making progress.

Today the main reason for continuous fetching would be MirrorMaker as in this case source offsets are Kafka offsets and understandable by users. But as I said, this should be addressed by KIP-971.

074-connector-offsets-support.md
```diff
@@ -0,0 +1,226 @@
...
+To list the current offsets the user will annotate the KafkaConnector resource with the new annotation `strimzi.io/connectorOffsets` set to `list`.
+After the annotation is added, on the next reconciliation the operator will fetch the current offsets and create a Kubernetes ConfigMap containing the JSON response.
+If the ConfigMap already exists the operator will replace the data with the updated data.
+The operator will then remove the annotation from the KafkaConnector CR.
+
+If the user wants to see an updated set of offsets they will need to re-annotate the resource.
+
+The ConfigMap the operator creates will be called `{connector}-offsets`, where `{connector}` is the name of the KafkaConnector CR that was annotated.
+The ConfigMap will contain a single file called `offsets.json` that contains the response received from the GET /connectors/{connector}/offsets endpoint.
+The KafkaConnector CRD will also be updated to contain a new property called `listOffsetsCM` that when set will override the name of the ConfigMap the operator creates/updates with the endpoint response.
```
ppatierno, line 92: Not sure I would use the `CM` to specify it's a ConfigMap (it would come from the CRD spec). Should it be `offsetsList` as well?

074-connector-offsets-support.md
```diff
@@ -0,0 +1,226 @@
...
+It is possible for the operator to help the user in cleaning up resources.
+The KafkaConnector CR will have a new property added called `resetOffsetsOnDelete`.
+This will be optional and when not present it will default to `false`.
+When `resetOffsetsOnDelete` is set to `true` the operator will perform the following steps when it sees a connector needs to be deleted:
+1. Stop the connector using the PUT /connectors/{connector}/stop endpoint 
+2. Reset offsets using the DELETE /connectors/{connector}/offsets endpoint 
+3. Delete the connector the DELETE /connectors/{connector} endpoint
+
+Notes:
+* Steps 1 & 2 are new compared to the existing steps the operator takes when deleting connectors
```
scholzj, line 188: Can we really guarantee this? What if the Connect is not working anymore or is deleted together with the connector etc.?

074-connector-offsets-support.md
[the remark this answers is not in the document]
katheris: Resolving conversation, since I've added this as an option now

074-connector-offsets-support.md
```diff
@@ -0,0 +1,226 @@
...
+The possible values will be `list`, `alter` and `reset`.
+
+### Listing offsets
+
+To list the current offsets the user will annotate the KafkaConnector resource with the new annotation `strimzi.io/connectorOffsets` set to `list`.
+After the annotation is added, on the next reconciliation the operator will fetch the current offsets and create a Kubernetes ConfigMap containing the JSON response.
+If the ConfigMap already exists the operator will replace the data with the updated data.
+The operator will then remove the annotation from the KafkaConnector CR.
+
+If the user wants to see an updated set of offsets they will need to re-annotate the resource.
```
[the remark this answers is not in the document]
katheris: I state that on line 86 already

074-connector-offsets-support.md
```diff
@@ -0,0 +1,226 @@
...
+Notes:
+* The format of the request body will be validated to make sure it is valid JSON, but the operator won't perform any further validation, instead relying on the API endpoint to return a reasonable error message.
+* If the request fails the KafkaConnector CR status will be updated to include a warning message with the response from the API, e.g.
+  ```yaml
+  - lastTransitionTime: "2024-06-04T08:44:15.913138115Z"
+    message: Failed to alter the connector offsets due to "message from endpoint".
+    reason: AlterOffsets
+    status: "True"
+    type: Warning
+  ```
```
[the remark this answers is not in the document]
katheris: I don't think we should retry unless the user explicitly re-annotates the resource, since there aren't any expected failures that should resolve automatically, however I worry that expecting the user to read the log to spot a failure will result in people not noticing it hasn't worked.

Is it feasible to leave the condition there for a period of time? Or could we use another annotation to communicate that a failure ocurred?

074-connector-offsets-support.md
katheris: I think we can give the same level of guarantees as we do with deleting the connector.

For example if Connect isn't working or has been deleted the call to the delete the connector would also fail, so if in future a new Connect cluster was created that re-used the same topics the connector would be started on that Connect cluster automatically, since it wouldn't have been removed from the Connect configs topic.

074-connector-offsets-support.md
```diff
@@ -0,0 +1,226 @@
...
+
+To list the current offsets the user will annotate the KafkaConnector resource with the new annotation `strimzi.io/connectorOffsets` set to `list`.
+After the annotation is added, on the next reconciliation the operator will fetch the current offsets and create a Kubernetes ConfigMap containing the JSON response.
+If the ConfigMap already exists the operator will replace the data with the updated data.
+The operator will then remove the annotation from the KafkaConnector CR.
+
+If the user wants to see an updated set of offsets they will need to re-annotate the resource.
+
+The ConfigMap the operator creates will be called `{connector}-offsets`, where `{connector}` is the name of the KafkaConnector CR that was annotated.
+The ConfigMap will contain a single file called `offsets.json` that contains the response received from the GET /connectors/{connector}/offsets endpoint.
```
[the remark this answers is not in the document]
katheris: The user can see when the ConfigMap was last updated by viewing the `time` in the `managedFields` of the ConfigMap. Do you think that is good enough?

074-connector-offsets-support.md
```diff
@@ -0,0 +1,226 @@
...
+
+### Resetting offsets
+
+To reset the offsets for a particular connector the user will annotate the KafkaConnector resource with the new annotation `strimzi.io/connectorOffsets` set to `reset`.
+When the operator observes this annotation it will use the DELETE /connectors/{connector}/offsets endpoint to reset all offsets for the connector.
+When the reset has been completed the operator will remove this annotation from the resource.
+
+If the reset fails the operator will add a condition to the KafkaConnector resource called `ResetOffsets` to communicate the failure and remove the annotation.
+
+Strimzi will shortcut and automatically fail to do the reset if the KafkaConnector resource is not stopped.
```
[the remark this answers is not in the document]
katheris: Good spot, that is because previously the alter was being done in a different resource, not in KafkaConnector CR. We can check ourselves in both cases now, since the state should be correct in the KafkaConnector, so I'll update it to say that in both cases.

074-connector-offsets-support.md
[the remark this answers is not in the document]
katheris: I've changed it to say "entry" instead, but I've left the `.json` file extension to make it clear the format that is expected in the entry.

074-connector-offsets-support.md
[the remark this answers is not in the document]
scholzj: That is not how the logic works in that case:
1) You remove the annotation or update the status to add the condition
2) That triggers reconciliation
3) This reconciliation doesn't do the operation anymore as the annotation is not there
4) The condition is removed as it is not valid anymore

Doing some time test would IMHO add a lot of complexity to the operator. I think we need either some other way to do this or we should stick with log messages only.

Or we should keep the annotation and retry it. I agree that it seems a bit silly and the retrying alone might not solve much. But maybe it helps the user to realize it doesn't work?

074-connector-offsets-support.md
scholzj: I'm not sure this is so simple:
* The connector will stay in the Connect topics and be recovered -> but deleted right away if the KafkaConnector CR is missing. So you can probably call it a _very eventual consistency_ :-).
* In your case, the connector will be gone. But the offset will not be reset and later when the Connect cluster is started again, you will not know you were supposed to delete some offsets.

074-connector-offsets-support.md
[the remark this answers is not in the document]
scholzj: I would think it would be more user friendly to have it somewhere in the data section?

074-connector-offsets-support.md
```diff
@@ -0,0 +1,226 @@
...
+    }
+```
+
+### Altering offsets
+
+To alter offsets the user will annotate the KafkaConnector resource with the new annotation `strimzi.io/connectorOffsets` set to `alter`.
+After the annotation is added, on the next reconciliation the operator will call the PATCH /connectors/{connector}/offsets endpoint to alter the offsets.
+Once the patch is complete the operator will then remove the annotation from the KafkaConnector CR.
+
+By default, the operator will construct the request body from a ConfigMap called `{connector}-offsets`, where `{connector}` is the name of the KafkaConnector CR that was annotated.
```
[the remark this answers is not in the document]
scholzj: You browse through your bash history to find the right annotation and use the wrong one? Maybe that is just me, but it happens to me quite often.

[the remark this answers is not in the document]
scholzj: In any case, making it optional later ould be easier than making it required later.

074-connector-offsets-support.md
katheris: In the proposal I'm suggesting that when we delete the connector we first stop the connector and reset the offsets, so effectively adding those two steps before this line https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectAssemblyOperator.java#L488 I'm not following why that would result in the connector being deleted before we can reset the offsets.

However, I have spotted a different problem, which is since we only know to delete the connector because the KafkaConnector CR is missing, we can't use a property on the KafkaConnector CR to tell us whether to reset the offsets first can we 🤔 I think it only works if we always reset the offsets, which as previously discussed we probably don't want

scholzj: > In the proposal I'm suggesting that when we delete the connector we first stop the connector and reset the offsets, so effectively adding those two steps before this line https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/KafkaConnectAssemblyOperator.java#L488 I'm not following why that would result in the connector being deleted before we can reset the offsets.

This is the normal flow. Sure, you can do it there. But in many of the edge cases, you will not.

> However, I have spotted a different problem, which is since we only know to delete the connector because the KafkaConnector CR is missing, we can't use a property on the KafkaConnector CR to tell us whether to reset the offsets first can we 🤔 I think it only works if we always reset the offsets, which as previously discussed we probably don't want

If I get right what you mean, then check the `connectorEventHandler` in KafkaConnectAssemblyOperator. You should (IIRC) get there the event handled inclusing the deleted CR. So you can see what the settings were. But of course, if the operator misses the watch because it was for example rolling at that time, it will not have it. 

That is the point I'm making here. You cannot easily make it 100% reliable. So is it worth it with the caveat it might not work? Or is it better to leave it to users to do it as combination of steps and take the respnsibility?

074-connector-offsets-support.md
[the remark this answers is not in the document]
katheris: Would it work to add the condition but also keep the annotation? That way the first time the annotation is added a new reconciliation would be triggered, however since the error message returned should be identical the second attempt would not change the status of the KafkaConnector. So then after that the reconciliation would just happen every 2 minutes, giving the user time to notice the condition. What do you think @scholzj?

074-connector-offsets-support.md
```diff
@@ -0,0 +1,264 @@
...
+If the user wants to see an updated set of offsets they will need to re-annotate the resource.
+
+The name of the ConfigMap the operator creates/updates will be set by the user in the KafkaConnector CR.
+The KafkaConnector CRD will be updated to contain a new property called `listOffsets` that must be set for the `strimzi.io/connectorOffsets=list` annotation to take effect.
+The structure of the `listOffsets` property will be as below, to allow it to be extended in future if required:
+
+```yaml
+listOffsets:
+  toConfigMap:
+    name: my-connector-offsets
```
scholzj, line 100: * What if the config map already exists?
* Also, will it have ownerReference attached to the KafkaConnector resource?

074-connector-offsets-support.md
```diff
@@ -0,0 +1,264 @@
...
+The user can update the KafkaConnector to stop the connector and reset the offsets at the same time.
+In that case the operator will first stop the connector, and once that API call returns, then it will make the call to reset the offsets.
+
+## Affected/not affected projects
+
+This only affects the Connect parts of the cluster-operator.
+
+## Compatibility
+
+If a `KafkaConnectorOffsets` resource is created against a Kafka Connect version that does not support the APIs, the API calls will fail and the operator will put a suitable message in the status to indicate what has happened.
```
scholzj, line 228: I guess this is not valid anymore because the resource is not used?

Also, the KIP JIRA suggests it is supported from Kafka 3.6.0, in which case it should be supported in all versions we support. Or is there some catch?

scholzj (commented):
SOme nits. But LGTM otherwise.

074-connector-offsets-support.md
mimaison: Yes I think Kafka 3.6.0 has all the APIs this relies on. 
Kafka 3.5.0 introduced `PUT /connectors/{connector}/pause` and `GET /connectors/{connector}/offsets`. The `PATCH /connectors/{connector}/offsets` and `DELETE /connectors/{connector}/offsets` were added in 3.6.0.

074-connector-offsets-support.md
katheris: I stated higher up that we would overwrite the contents if it existed already. For owner references, I would suggest if the operator creates it we add one, but if it's there already we don't add it.

074-connector-offsets-support.md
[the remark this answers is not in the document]
scholzj: That sounds like a great design :-D. Anyway, not much one can do - maybe we should clarify it in the proposal at least for future reference?

074-connector-offsets-support.md
scholzj: Well, at least until server-side-apply is the main way to do things, you will likely overwrite the CM if it already exists. So keeping any owner references is an additional effort. Also, you can have more of them IIRC.

katheris: Do you think we shouldn't add any owner references then?

scholzj: TBH, I do not have a strong opinion ... that is why I was curious about what the plan is :-D. I guess it has both pros and cons. The positive thing is that it will be cleaned up when you delete the connector and clean up all related things. The negative is that it will be deleted with the connector even if you would want to keep it. 🤔 So I'm really not sure what is the best.

075-connector-offsets-support.md
```diff
@@ -0,0 +1,265 @@
...
+  ```yaml
+  - lastTransitionTime: "2024-06-04T08:44:15.913138115Z"
+    message: Failed to alter the connector offsets due to "message from endpoint".
+    reason: AlterOffsets
+    status: "True"
+    type: Warning
+  ```
+  The operator will leave the annotation on the KafkaConnector resource until either the patch operation succeeds, or the user removes the annotation.
+  This means the operator will retry the patch on every reconciliation, allowing the condition to remain present for the user to see.
+* Strimzi will shortcut and automatically fail to do the reset if the KafkaConnector resource does not have it's `state` set as `stopped`.
```
scholzj, line 196: Typo?

```suggestion
* Strimzi will shortcut and automatically fail to alter the offsets if the KafkaConnector resource does not have it's `state` set as `stopped`.
```

scholzj (approved):
LGTM. Thanks.

You do not mention anything about MM2 -> Is it because the MM2 source connectors store the offsets as regular commits in the source cluster? Or will we possibly need a separate proposal to cver MM2 in the future? (In both cases, it might be worth explaining  it in the affected projects section)?

075-connector-offsets-support.md
katheris: Yeah, good spot

katheris:
> You do not mention anything about MM2 -> Is it because the MM2 source connectors store the offsets as regular commits in the source cluster? Or will we possibly need a separate proposal to cver MM2 in the future? (In both cases, it might be worth explaining it in the affected projects section)?

That's a good point, if the users are using the KafkaConnector CR they can manage the offsets for the MM2 source connectors like any other source connector. However, if they are using the KafkaMirrorMaker2 CR I think we should also provide support. Since the format of the offsets is known for these I think we can give a more guided experience, rather than allowing a free-form ConfigMap, so I've added a note in the Affected/not affected projects that this should be covered in a separate proposal.

katheris:
@mimaison @ppatierno when you have some time please take another look at the proposal with the recent updates

mimaison (approved):
LGTM. I think this simpler approach works well and it should still enable users to perform all the offsets operations on their connectors.

075-connector-offsets-support.md
```diff
@@ -0,0 +1,270 @@
...
+        "kafka_offset": // Kafka offset
+      }
+    }
+  ]
+}
+```
+
+### New annotation
+
+All three actions will make use of a new annotation called `strimzi.io/connectorOffsets`.
```
ppatierno, line 81: most of our annotations are using a format like this `strimzi.io/connector-offsets`. I would change this way (here and of course everywhere it's mentioned again).

075-connector-offsets-support.md
```diff
@@ -0,0 +1,270 @@
...
+            "kafka_topic": "my-topic"
+              "kafka_partition": 2
+          },
+          "offset": {
+            "kafka_offset": 4
+          }
+        }
+      ]
+    }
+```
```
ppatierno, line 151: both the above ConfigMap examples are missing the "owner reference to the Kubernetes ConfigMap pointing to the KafkaConnector resource that was annotated" you mentioned before. Maybe you can add it?

ppatierno (approved):
LGTM. I added two more comments/nits.

075-connector-offsets-support.md
tombentley: +1 camel case looks very weird here

075-connector-offsets-support.md
```diff
@@ -0,0 +1,270 @@
...
+The structure of the `alterOffsets` property will be as below, to allow it to be extended in future if required:
+
+```yaml
+alterOffsets:
+  fromConfigMap:
+    name: my-connector-offsets
+```
+
+Notes:
+* The format of the request body will be validated to make sure it is valid JSON, but the operator won't perform any further validation, instead relying on the API endpoint to return a reasonable error message.
```
tombentley, line 185: 1. What does "valid JSON" mean here? I guess it must mean "syntactically valid JSON", because source connectors have an unknown schema for their offset JSON. But it would be good to be explicit in the doc. 
2. Let's also be explicit that `offsets.json` being absent from the ConfigMap is also treated as an error. 
3. I suppose there's also no harm in saying that it's not an error if the ConfigMap contains keys others than `offsets.json`

075-connector-offsets-support.md
```diff
@@ -0,0 +1,270 @@
...
+
+All three actions will make use of a new annotation called `strimzi.io/connectorOffsets`.
+The possible values will be `list`, `alter` and `reset`.
+
+### Listing offsets
+
+To list the current offsets the user will annotate the KafkaConnector resource with the new annotation `strimzi.io/connectorOffsets` set to `list`.
+After the annotation is added, on the next reconciliation the operator will fetch the current offsets and create a Kubernetes ConfigMap containing the JSON response.
+The operator will add an owner reference to the Kubernetes ConfigMap pointing to the KafkaConnector resource that was annotated.
+If the ConfigMap already exists the operator will replace the data with the updated data and not add an owner reference.
```
tombentley, line 89: Assuming by "the data" you mean "the `data`", then we're saying we'll obliterate any other keys in the ConfigMap. It might be worth calling this out, even if only by monospacing the `data`.

tombentley (approved):
Thank you, I think this will make a nice improvement. I left a few minor comments, but over all this LGTM.

fvaleri (approved):
LGTM. Thanks.

075-connector-offsets-support.md
katheris: Yes, I've adjusted the wording

075-connector-offsets-support.md
katheris: 1. Correct, I've added the word "syntactically" to be more explicit.
2. Added
3. Added

075-connector-offsets-support.md
katheris: Good point, I've updated it

scholzj:
This has now 3 binding and 2 non-binding +1 votes. If you have any comments or plan to review it, please do so until Friday. Otherwise, we will close it as approved.

076-connector-offsets-support.md
```diff
@@ -0,0 +1,286 @@
...
+      kind: KafkaConnector
+      name: my-sink-connector
+      uid: 5678
+data:
+  offsets.json: |
+    {
+      "offsets": [
+        {
+          "partition": {
+            "kafka_topic": "my-topic"
```
PaulRMellor, line 156:
```suggestion
            "kafka_topic": "my-topic",
```

076-connector-offsets-support.md
```diff
@@ -0,0 +1,286 @@
...
+  This means the operator will retry the list on every reconciliation, allowing the condition to remain present for the user to see.
+
+### Altering offsets
+
+To alter offsets the user will annotate the KafkaConnector resource with the new annotation `strimzi.io/connector-offsets` set to `alter`.
+After the annotation is added, on the next reconciliation the operator will call the PATCH /connectors/{connector}/offsets endpoint to alter the offsets.
+Once the patch is complete the operator will then remove the annotation from the KafkaConnector CR.
+
+The operator will read in the entry called `offsets.json` from the ConfigMap.
+The entry name matches the entry name used in the list operation, so that a user can list offsets, then edit the ConfigMap the operator created, then request the offsets be altered using that same ConfigMap.
```
PaulRMellor, line 186:
```suggestion
The entry name matches the entry name used in the list operation, so that a user can list offsets, then edit the ConfigMap the operator created, then request the offsets to be altered using that same ConfigMap.
```

076-connector-offsets-support.md
```diff
@@ -0,0 +1,286 @@
...
+  ```yaml
+  - lastTransitionTime: "2024-06-04T08:44:15.913138115Z"
+    message: Failed to alter the connector offsets due to "message from endpoint".
+    reason: AlterOffsets
+    status: "True"
+    type: Warning
+  ```
+  The operator will leave the annotation on the KafkaConnector resource until either the patch operation succeeds, or the user removes the annotation.
+  This means the operator will retry the patch on every reconciliation, allowing the condition to remain present for the user to see.
+* Strimzi will shortcut and automatically fail to alter the offsets if the KafkaConnector resource does not have it's `state` set as `stopped`.
```
PaulRMellor, line 210: shortcut meaning ending the operation early?

076-connector-offsets-support.md
```diff
@@ -0,0 +1,286 @@
...
+### Resetting offsets
+
+To reset the offsets for a particular connector the user will annotate the KafkaConnector resource with the new annotation `strimzi.io/connector-offsets` set to `reset`.
+When the operator observes this annotation it will use the DELETE /connectors/{connector}/offsets endpoint to reset all offsets for the connector.
+When the reset has been completed the operator will remove this annotation from the resource.
+
+If the request to the Connect API fails the operator will add a condition to the KafkaConnector CR status with a warning message that includes the response from the API, e.g.
+  ```yaml
+  - lastTransitionTime: "2024-06-04T08:44:15.913138115Z"
+    message: Failed to alter the connector offsets due to "message from endpoint".
```
PaulRMellor, line 226: Should this be Failed to reset?

076-connector-offsets-support.md
PaulRMellor: `paused` state not sufficient?

076-connector-offsets-support.md
```diff
@@ -0,0 +1,286 @@
...
+Strimzi will shortcut and automatically fail to do the reset if the KafkaConnector resource does not have it's `state` set as `stopped`.
+Similar to if Connect returns on error, the operator will add a condition stating that the operation has failed because the connector is not stopped and therefore offsets cannot be reset, and leave the annotation on the resource.
+The user can update the KafkaConnector to stop the connector and reset the offsets at the same time.
+In that case the operator will first stop the connector, and once that API call returns, then it will make the call to reset the offsets.
+
+## Affected/not affected projects
+
+This proposal only affects the Connect parts of the cluster-operator.
+This proposal does not apply to the use of the KafkaMirrorMaker2 Custom Resource.
+If users are using a KafkaConnector Custom Resource to manage the Mirror Maker connectors, they can manage offsets using the method described in this proposal.
```
PaulRMellor, line 243:
```suggestion
If users are using a KafkaConnector Custom Resource to manage the MirrorMaker connectors, they can manage offsets using the method described in this proposal.
```

076-connector-offsets-support.md
```diff
@@ -0,0 +1,286 @@
...
+We should have a different proposal to cover how we might support managing offsets for Mirror Maker connectors that are managed via the KafkaMirrorMaker2 Custom Resource.
```
PaulRMellor, line 244:
```suggestion
We should have a different proposal to cover how we might support managing offsets for MirrorMaker connectors that are managed via the KafkaMirrorMaker2 Custom Resource.
```

076-connector-offsets-support.md
```diff
@@ -0,0 +1,286 @@
...
+
+### Listing offsets continually
+A previous version of this proposal suggested to have a property that prompted the operator to add the current offsets to the KafkaConnector status on every reconciliation.
+This was rejected because it will be tricky to implement this in a way that doesn't result in cyclical reconciliations.
+In addition, it seems likely that listing offsets will be used most often as a one-off action, rather than users needing to continually track offsets.
+
+### Listing offsets in the KafkaConnector status
+
+A previous version of this proposal suggested using the KafkaConnector status field to list the returned offsets.
+This was rejected because updating the status would trigger an addition reconcile and the response from the endpoint may be a very large size if there are many partitions.
```
PaulRMellor, line 262:
```suggestion
This was rejected because updating the status would trigger an additional reconcile and the response from the endpoint may be a very large size if there are many partitions.
```

076-connector-offsets-support.md
```diff
@@ -0,0 +1,286 @@
...
+
+Using an annotation on its own wouldn't really be feasible since the body of the request might be very complex.
+
+We could use the spec described for `KafkaConnectorOffsets`, combined with an annotation to trigger the reset.
+This has the benefit of not needing a new resource type, however, it means polluting the `KafkaConnector` resource with offset properties.
+Having separate resources means users can also provide different permissions for the two kinds of resource.
+
+### Allowing users to reset offsets when the connector is removed
+
+A previous version of this proposal suggested introducing a new property on the KafkaConnector resource to allow users to request the offsets be reset when the connector is removed.
```
PaulRMellor, line 282:
```suggestion
A previous version of this proposal suggested introducing a new property on the KafkaConnector resource to allow users to request the offsets to be reset when the connector is removed.
```

076-connector-offsets-support.md
katheris: > shortcut meaning ending the operation early?

Yes so rather than making the request to Kafka Connect and dealing with the error response back, we don't make the request at all and just inform the user of the error.

> `paused` state not sufficient?

My understanding is that `paused` is not sufficient, is that yours as well @mimaison?

scholzj:
Approved with 5 binding and 2 non-binding +1 votes.

PR #123, opened 2024-06-20
Added proposal to improve `KafkaRebalance` deletion handling

ppatierno:
As per issue https://github.com/strimzi/strimzi-kafka-operator/issues/10245, this proposal is about improving the way how the deletion of a `KafkaRebalance` custom resource is handled to provide the user a better experience and still getting feedback on the ongoing operations on Cruise Control.

### DIFF
--- a/075-kafkarebalance-deletion-handling.md
+++ b/075-kafkarebalance-deletion-handling.md
@@ -1,0 +1,55 @@
+# `KafkaRebalance` custom resource deletion handling
+
+This proposal introduces a better mechanism of handling the deletion attempt of a `KafkaRebalance` custom resource by the user.
+It's about providing a consistent user experience on getting feedback on the status of an ongoing operation on Cruise Control.
+
+## Current situation
+
+When the user creates a `KafkaRebalance` custom resource for getting an optimization proposal from Cruise Control and then running the corresponding actual rebalancing, such custom resource is also used to provide feedback on what's the status of the ongoing operation.
+The `KafkaRebalance.status` includes a condition which uses the `type` field to describe the status of the request such as `PendingProposal`, `ProposalReady`, `Rebalancing`, `Ready` (rebalancing done), `NotReady` (error) and `Stopped`.
+When the operation is done (proposal ready or rebalancing done), the `KafkaRebalance.status` also includes details about the optimization itself.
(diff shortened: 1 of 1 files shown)


ppatierno:
>  Also, you say that the CO will remove the finilaizer when it gets to ProposalReady or to Ready => so at that point it will be deleted and the user will not get the feedback anyway.

It's not true because while the resource deletion is blocked it's still updated by the operator and the user has the feedback. They are also still on time on aborting the deletion if done by mistake.

> The only situation when the finalizer like this might make any sense is if the operator is going to immediately stop any ongoing proposal preparation or rebalances when the resource is deleted. But otherwise they do not seem to add any value.

Are you saying the if manual (or future autorebalancing), if the user deletes the custom resource, the operator should trigger a stop operation? So deletion implies stopping? This could be another approach. But I don't think the first one doesn't make any sense.

scholzj:
> > Also, you say that the CO will remove the finilaizer when it gets to ProposalReady or to Ready => so at that point it will be deleted and the user will not get the feedback anyway.
> 
> It's not true because while the resource deletion is blocked it's still updated by the operator and the user has the feedback. They are also still on time on aborting the deletion if done by mistake.

The deletion will be pending until you remove the finalizer. So the user will likely have no chance to read the "feedback". No idea what this `They are also still on time on aborting the deletion if done by mistake.` means.

> > The only situation when the finalizer like this might make any sense is if the operator is going to immediately stop any ongoing proposal preparation or rebalances when the resource is deleted. But otherwise they do not seem to add any value.
> 
> Are you saying the if manual (or future autorebalancing), if the user deletes the custom resource, the operator should trigger a stop operation? So deletion implies stopping? This could be another approach. But I don't think the first one doesn't make any sense.

I don't think it matters whether is manual or automated. And I'm not saying this should be done -> but I'm saying that this is the only situation where a finalizer used as describer here would make any sense. That would also correspond to how the finalizer is used in the TO for example.

> We have a specific `stop` annotation for stopping execution. Not sure we want the deletion to trigger it.

As I said, all I'm saying is that without stopping it at deletion, the finalizers like this add negative value (no benefits, but more code, more testing, more confused users, more maintenance).

ppatierno:
Going to close it because not relevant to the autorebalancing stuff on which we have to focus.

PR #126, opened 2024-07-22
Deprecation and removal of Storage overrides

scholzj:
This PR proposes deprecation and later removal of the Storage overrides: https://strimzi.io/docs/operators/latest/full/deploying.html#storage_class_overrides


im-konge (approved):
I think that the timeline is fine, the users can do it together with the migration to NodePools, so this LGTM. Thanks :)

079-deprecation-and-removal-of-storage-overrides.md
```diff
@@ -0,0 +1,95 @@
...
+
+The persistent-volume-claim storage overrides will be deprecated immediately in Strimzi 0.43.
+The will remain supported and used, but users would be encouraged to move away from them and use Kafka Node Pools instead.
+This also falls into the time period when we expect most users to migrate to node pools and to KRaft which provides a good opportunity to move to per-node-pool storage class instead of using the overrides.
+
+Later, in the Strimzi version where support for ZooKeeper-based Kafka clusters is removed, the support for the storage overrides will be dropped and the overrides will be ignored.
+The overrides will remain in the CRD API for backwards compatibility but will not be used by the operator code.
+Finally, when moving the new `v1` CRD API in the future, the fields will be completely removed.
+
+Kubernetes do not allow changing storage class for an existing persistent volumes.
```
PaulRMellor, line 63:
```suggestion
Kubernetes do not allow changing storage class for existing persistent volumes. 
```

079-deprecation-and-removal-of-storage-overrides.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Affected projects
+
+This proposal affects the Strimzi Cluster Operator only.
+
+## Backwards compatibility
+
+This proposal removes the support for storage overrides in Kafka clusters.
+It will impact all users using it and they will have to migrate to node pools with different storage classes.
+Kafka clusters not using the overrides or and any other operands supported by Strimzi will nto be impacted in any way.
```
PaulRMellor, line 82:
```suggestion
Kafka clusters not using the overrides and any other operands supported by Strimzi will not be impacted in any way.
```

079-deprecation-and-removal-of-storage-overrides.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Rejected alternatives
+
+### Different timeline
+
+The timeline between the deprecation of the overrides and the removal of their support is relatively short (likely 3-4 Strimzi versions).
+If we think this is a problem, we can choose a different timeline.
+For example, we can deprecated the overrides right now, but postpone when we drop the support for them:
```
PaulRMellor, line 90:
```suggestion
For example, we can deprecate the overrides right now, but postpone when we drop the support for them:
```

079-deprecation-and-removal-of-storage-overrides.md
```diff
@@ -0,0 +1,95 @@
...
+* Drop the support only at a later version (e.g. first Strimzi release after June 2025)
+* Drop the support only when we migrate to the `v1` CRD API
+
+This might give users more time to handle it.
+However, while the time window for the migration is relatively short, I believe it is sufficient as it for most existing users includes the migration to KRaft and to node pools and provides a good opportunity to get rid of this legacy feature.
```
PaulRMellor, line 95:
```suggestion
However, while the migration time window is relatively short, I believe it is sufficient for most existing users, as it includes the migration to KRaft and node pools, and provides a good opportunity to eliminate this legacy feature.
```

PaulRMellor (approved):
Proposal makes sense as this feature isn't needed now. I think the timeline works too, as it fits nicely with the node pools adoption.

fvaleri (commented):
Should we document the migration as we did for the listener configurations showing the old and the new equivalent configuration? I think it is implicit, but it's better to clarify in the proposal.

scholzj:
> Should we document the migration as we did for the listener configurations showing the old and the new equivalent configuration? I think it is implicit, but it's better to clarify in the proposal.

We have already similar guides for how to use the node pools. So I do not plan to do anything detailed. But I can add some bullet points to the release notes or something like that.

ppatierno (approved):
LGTM.

079-deprecation-and-removal-of-storage-overrides.md
```diff
@@ -0,0 +1,95 @@
...
+Kubernetes do not allow changing storage class for existing persistent volumes. 
+If any user doesn't migrate from the overrides in time, the existing Kafka nodes and persistent volumes will not be affected.
+Only when the PVC/PV is deleted or when new nodes will be added during a scale-up, the overrides will be ignored and the default storage class will be used.
+So the impact on existing users should be minimal.
+
+### Warnings
+
+With the deprecation of the storage overrides, the operator will be updated to issue warnings when the overrides are used.
+The warnings will be printed in regular logs from the Cluster Operator.
+And they will be also added to the conditions in the `Kafka` CR status.
```
tombentley, line 72: Will we document how the users can get rid of these warnings specifically? Or do we expect users to figure out the equivalent node pool configuration for themselves?

079-deprecation-and-removal-of-storage-overrides.md
```diff
@@ -0,0 +1,95 @@
...
+This proposal removes the support for storage overrides in Kafka clusters.
+It will impact all users using it and they will have to migrate to node pools with different storage classes.
+Kafka clusters not using the overrides and any other operands supported by Strimzi will not be impacted in any way.
+
+## Rejected alternatives
+
+### Different timeline
+
+The timeline between the deprecation of the overrides and the removal of their support is relatively short (likely 3-4 Strimzi versions).
+If we think this is a problem, we can choose a different timeline.
```
tombentley, line 89: Would the removal be covered by another Proposal, or is it part of this one? I guess I'm mostly interested in what the operator does when the override support is removed but the user has configured the overrides: Would that still just be "log a warning + add a condition", or would the behaviour then be different?

tombentley (commented):
Thanks @scholzj, mostly this looks fine, but I had a couple of questions.

079-deprecation-and-removal-of-storage-overrides.md
scholzj: They get rid of them by stopping using the overrides. We already have content covering how to migrate to node pools and use them for different configurations. What exactly is the new configuration depends on the use-case. I.e. why exactly you use these overrides. For example, in the situations where I saw this used/recommended it to be used, it was used as a _trick_ to control the scheduling of individual brokers. In such case, the proper replacement might be different node pools with affinity rules rather than a different storage class. But obviously there might be also situations where the user actually needed different storage class. All of these cases can be covered by NodePools. But there is not a single exact _conversion_ procedure as it depends what do you do. Also, from my experience, this is very niche thing not used by every other user.

So as replied to Federico's comment - we can definitely have some bullet points to guide the users. But I do not see any need to have some elaborate multipage guide.

scholzj: > Or do we expect users to figure out the equivalent node pool configuration for themselves?

To also answer the question directly ... yes, I would expect the users to figure the right configuration by themselves. Given the expected number of people using it, I also do not expect it to be an issue if they come up asking for help.

079-deprecation-and-removal-of-storage-overrides.md
scholzj: > Would the removal be covered by another Proposal, or is it part of this one?

This proposal includes the timeline -> on lines 59-61. The proposed timeline is to remove it together with ZooKeeper support (as in order to get to KRaft, everyone will need to use node pools and have a good chance to migrate away from the storage overrides as well). The place you are commenting on here is a rejected alternative. I'm basically trying to say here that I considered other timeline when it might be removed and rejected them. 

> I guess I'm mostly interested in what the operator does when the override support is removed but the user has configured the overrides: Would that still just be "log a warning + add a condition", or would the behaviour then be different?

As described on lines 63-66. When we start ignoring the overrides, nothing will happen right away with the existing volumes / brokers because the storage class is immutable. So the existing storage will keep exist as it was. What will change is that the storage override will be ignored when a new volume is created. That can happen for example when you scale-up the cluster with prepared override for the next node ID or when you delete the PV/PVC and let Strimzi create a new one.

Once we remove the support for the overrides, the API has to remain unchanged until v1 API is introduced. So users will still get the warnings based on the API field deprecations.

tombentley (approved):

fvaleri (approved):

scholzj:
This now has 4 binding and one non-binding +1 vote. If you plan to review this proposal, please do so by Monday EOB (CEST timezone). Otherwise, I will close the vote.

Frawless (approved):

scholzj:
Approved with 5 binding and 1 non-binding +1 votes.

PR #127, opened 2024-08-08
MirrorMaker Connector Offsets Support

katheris:


000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,390 @@
...
+
+Support for the "stop" endpoint was added via [PR 9095](https://github.com/strimzi/strimzi-kafka-operator/pull/9095).
+Currently users that are utilising the KafkaMirrorMaker2 resource cannot make use of the API endpoints for managing offsets.
+
+Proposal [#076][proposal-76] discusses how to add support for these API endpoints to the KafkaConnector resource.
+However it did not cover the KafkaMirrorMaker2 resource.
+
+## Motivation
+
+Previously if users wanted to list, update or delete offsets for a MirrorMaker connector they had to directly write to the Kafka Connect internal offsets topic.
```
scholzj, line 25:
```suggestion
Previously if users wanted to list, update or delete offsets for a MirrorMaker connector they had to directly read/write from/to the Kafka Connect internal offsets topic.
```

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,390 @@
...
+### New annotations
+
+All three actions will make use of a new annotation called `strimzi.io/connector-offsets`.
+The possible values will be `list`, `alter`, and `reset`.
+
+A second new annotation will be added called `strimzi.io/mirrormaker-connector`.
+It will be required when `strimzi.io/connector-offsets` is set.
+The value will be the name of the connector to apply the action to, for example `east-kafka->west-kafka.MirrorSourceConnector`.
+
+If the `strimzi.io/mirrormaker-connector` annotation is set, but the `strimzi.io/connector-offsets` annotation is missing the operator will take no action.
```
scholzj, line 131: Isn't it better to have a single annotation with some JSON structure? What if you would want to e.g. reset source and checkpoint connectors at the same time which seem slike something what might make sense?

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,390 @@
...
+To alter offsets we do not need the user to supply the entirety of these JSON objects.
+That is because some of the fields are already known to the operator.
+Specifically, we can omit:
+* For MirrorSourceConnector:
+  * `cluster`, this is equal to `spec.mirrors[].sourceCluster` in the KafkaMirrorMaker2 resource
+  * `topic`, this is the configuration topic for the underlying Kafka Connect cluster, by default named `mirrormaker2-cluster-configs`
+* For MirrorCheckpointConnector no fields can be omitted
+* For MirrorHeartbeatConnector:
+  * `partition.sourceClusterAlias`, this is equal to `spec.mirrors[].sourceCluster` in the KafkaMirrorMaker2 resource
+  * `partition.targetClusterAlias`, this is equal to `spec.mirrors[].targetCluster` in the KafkaMirrorMaker2 resource
```
scholzj, line 117: * I wonder if it is wise to omit the information
    * The MM2 cluster might have different source clusters. So event through you seem to referece a specific source cluster in the annotation, it might be confusing once you have the Config Map with the data and don't know anymore what cluster it is from. So I would consider keeping it and validating it when altering the offsets?
    * We rely on the format as defined by MM2. Once you start modifying it, it might cause problems when Kafka changes the format. E.g. if new fields with the names you want to omit are added or the original fields are renamed. If you stick with the Kafka format as is, we will be in a much better position.
* If we decide to keep it, it might be nice to include examples of the actual structures you will use.

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,390 @@
...
+If the user wants to see an updated set of offsets they will need to re-annotate the resource.
+
+The name of the ConfigMap the operator creates/updates will be set by the user in the KafkaMirrorMaker2 CR.
+The KafkaMirrorMaker2 CRD will be updated to contain a new top-level (i.e. directly under `spec`) property called `listOffsets` that must be set for the `strimzi.io/connector-offsets=list` annotation to take effect.
+The structure of the `listOffsets` property will be as below, to allow it to be extended in future if required:
+
+```yaml
+listOffsets:
+  toConfigMap:
+    name: my-connector-offsets
```
scholzj, line 151: I think you should place it with the connector configuration.
* That is where you stop the connector
* It will IMHO better match to how the other options work
* It will also fit the API model as you can inherit it with the Connect part of the implementation.

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,390 @@
...
+This means the connector name expected will have `->` replaced with `,`, similar to in the list offsets operation.
+
+The name of the ConfigMap the operator will use to construct the request body will be set by the user in the KafkaMirrorMaker2 CR.
+The KafkaMirrorMaker2 CRD will be updated to contain a new top-level (i.e. directly under `spec`) property called `alterOffsets` that must be set for the `strimzi.io/connector-offsets=alter` annotation to take effect.
+The structure of the `alterOffsets` property will be as below, to allow it to be extended in future if required:
+
+```yaml
+alterOffsets:
+  fromConfigMap:
+    name: my-connector-offsets
```
scholzj, line 273: Same as above about where in the structure should it be.

scholzj (commented):

000-mm2-connector-offsets-support.md
katheris: My understanding of Kubernetes annotations is they are intended to be a key/value pair where the key and value are both strings, are you suggesting the value string actually be JSON, e.g.
```
strimzi.io/mm2-connector-offsets: "{[{\"connector\": \"east-kafka->west-kafka.MirrorSourceConnector\", \"action\": \"list\"}, {\"connector\": \"east-kafka->west-kafka.MirrorCheckpointConnector\", \"action\": \"list\"}]}"
```

I worry that makes it quite a lot of work for the user to construct the JSON and a lot of room for error.

I did wonder about allowing the `strimzi.io/mirrormaker-connector` to be a comma separated list, but it brings the same difficulty that I listed in the rejected alternatives around making the `mirrormaker-connector` annotation optional, that the reconciliation is performed per connector and so it requires error handling if the action on one connector succeeds and one doesn't.

However, if we think it's a likely usecase to be able to apply actions to multiple connectors at once I can describe that error handling, I think it would likely just mean if something goes wrong the user has to reannotate the resource, so it's up to them to check the status for what succeeded and what didn't.

000-mm2-connector-offsets-support.md
katheris: Strimzi names the connectors as `<SOURCE_ALIAS>-><TARGET_ALIAS>.<CONNECTOR_NAME>` (e.g. `east-kafka->west-kafka.MirrorSourceConnector`) and since in the proposal the output is under an entry with that name I figured that would be sufficient to identify which specific connector the offsets are for, what do you think?

In terms of the format changing, we would still need a code change if we only validate it since the validation would start incorrectly failing, so is that a much better position?

I included examples where I explain how list and alter work for each connector type already, do you think I should also include it here in the proposal?

000-mm2-connector-offsets-support.md
katheris: Good point, that makes more sense, I've updated it.

000-mm2-connector-offsets-support.md
scholzj: I don't know, to me, this seems like something what make sense?
* When using multiple source clusters, I might want to reset all of their source connector offsets at once?
* When using source and offset mirroring, I might want to reset both the source and checkpoint connector offsets at the same point?

The point about JSON in the annotation is partially valid. Storing JSONs in annotation is common. But it does make it hard to annotate it from the command line (but is that how users will annotate it? Is it a bigger problem than having two annotations to be set, which if they are set one by one also causes problems and errors?).

000-mm2-connector-offsets-support.md
scholzj: > In terms of the format changing, we would still need a code change if we only validate it since the validation would start incorrectly failing, so is that a much better position?

My assumption is that the validation is only for a valid JSON as it is with the regular connectors. Although it is not completely clear from this proposal so it is probably worth clarifying it. That said, I think we should keep the validation behavior the same. That way we will not depend on Kafka so much for this.

The removal of the fields is tricky because if Kafka modifies it, how are you gonna decide what to do? Write another proposal to decide to remove some more fields? How will that affect the logic of adding it back to the JSON before passing it to Kafka if there are multiple versions on the Kafka side? That is what makes keeping it the same more future-proof for me.

> Strimzi names the connectors as <SOURCE_ALIAS>-><TARGET_ALIAS>.<CONNECTOR_NAME> (e.g. east-kafka->west-kafka.MirrorSourceConnector) and since in the proposal the output is under an entry with that name I figured that would be sufficient to identify which specific connector the offsets are for, what do you think?

TBH, I wondered if naming the fields in the Config Map based on connector name is the right choice. I did not comment about it because I decided that I'm not sure if it is a good or bad option. But relying on the field name only seems a risky business. That is also why I would keep it in the JSON.

000-mm2-connector-offsets-support.md
katheris: I'm happy to update the proposal so multiple connectors can be actioned at once, but I'd personally advocate for keeping with the two annotations, rather than having a more complex single annotation. I don't think there are big problems with if one is present and one not:

| `strimzi.io/connector-offsets` | `strimzi.io/mirrormaker-connectors` | outcome |
| -- | -- | -- |
| present | present | action applied |
| missing | missing | nothing happens |
| missing | present | nothing happens |
| present | missing | ? |

It's only the last state which is potentially a problem. For that I think either we could add a warning as I described, or if we are supporting multiple connectors, this could apply the action to all connectors in all mirrors.

You can use the `kubectl annotate` command to add multiple annotations at once

000-mm2-connector-offsets-support.md
katheris: Ah I had assumed we were validating the keys we expected, that is what I intended when I stated
```
The data supplied by the user will be validated and augmented with information from the KafkaMirrorMaker2 resource before it is submitted to the Connect API.
The expected format will match the output from the list offsets action.

For example for a MirrorSourceConnector the ConfigMap `<CONNECTOR_NAME>.json` data field must contain an entry that matches:
...
```

but I can see that wording isn't as explicit as it could be.

The reason I suggested more complex validation behaviour is I figured that the KafkaMirrorMaker2 resource gives a more hand-holding experience for people running MirrorMaker. So it made sense to me that it also gives more help with offset management. I guess it would be a tricky thing for Kafka to evolve, because any changes to the structure you would expect to be backwards compatible, or the MM connectors would also need to be updated to handle the missing field.

000-mm2-connector-offsets-support.md
scholzj: I don't think applying it to all connectors makes sense. That would be a big risk. Especially if you expect both annotation to be used. I would also expect at least a warning in the other case when only one annotation is present.

000-mm2-connector-offsets-support.md
scholzj: I personally would definitely aim for unification rather than for doing things differently.

katheris:
@scholzj I've made some updates addressing the two areas we were discussing

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,447 @@
...
+After the annotations are added, on the next reconciliation the operator will fetch the current offsets for the connector(s) and create Kubernetes ConfigMap(s) containing the JSON response(s).
+The operator will add an owner reference to the Kubernetes ConfigMap(s) pointing to the KafkaMirrorMaker2 resource that was annotated.
+If the ConfigMap(s) already exist the operator will patch the ConfigMap(s) with updated data and not add an owner reference.
+This means any existing keys in the ConfigMap(s) that the operator is not updating will remain as before.
+
+Since the operator reconciles connectors one-by-one, after each list operation the operator will update the `strimzi.io/mirrormaker-connector` annotation to remove the connector it has just listed offsets for.
+Once all list operations are complete the operator will then remove both the `strimzi.io/mirrormaker-connector` and `strimzi.io/connector-offsets` annotations from the KafkaMirrorMaker2 CR.
+This will make it easier to recover from errors.
```
scholzj, line 129: They should be all listed within the same reconciliation. So why do they need to be updated one by one? Why not update them just once at the end or keep it if something fails.

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,447 @@
...
+        listOffsets:
+          toConfigMap:
+            name: my-connector-offsets
+```
+
+The ConfigMap(s) will contain a single entry for each connector, containing the response received from the GET /connectors/{connector}/offsets endpoint.
+This will make it possible to use the same ConfigMap for different connectors.
+The name of the entry will be based on the connector name, however since 
+[`>` characters are not allowed in ConfigMap keys](https://kubernetes.io/docs/concepts/configuration/configmap/#configmap-object) the name will be altered to replace `->` with `,`.
+For example a connector called `east-kafka->west-kafka.MirrorSourceConnector` will have offsets placed under `east-kafka,west-kafka.MirrorSourceConnector.json`.
```
scholzj, line 162: Would it make more sense to just keep `-` for example? Using `,` as a separator seems strange in this place.

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,447 @@
...
+* If only one of the `strimzi.io/connector-offsets` and `strimzi.io/mirrormaker-connector` annotations is present on the KafkaMirrorMaker2 the operator will leave the annotation on the resource and update the KafkaMirrorMaker2 CR status to include a warning message similar to if `listOffsets` is missing.
+
+### Altering offsets
+
+To alter offsets the user will annotate the KafkaMirrorMaker2 resource with the new annotation `strimzi.io/connector-offsets` set to `alter`, and the `strimzi.io/mirrormaker-connector` annotation.
+After the annotations are added, on the next reconciliation the operator will call the PATCH /connectors/{connector}/offsets endpoint to alter the offsets for the connector(s).
+
+Since the operator reconciles connectors one-by-one, after each patch operation the operator will update the `strimzi.io/mirrormaker-connector` annotation to remove the connector it has just altered offsets for.
+Once all patch operations are complete the operator will then remove both the `strimzi.io/mirrormaker-connector` and `strimzi.io/connector-offsets` annotations from the KafkaMirrorMaker2 CR.
+This will make it easier to recover from errors.
```
scholzj, line 279: Same as above,

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,447 @@
...
+
+The operator will read in the entry called `<CONNECTOR_NAME>.json` from the ConfigMap.
+The entry name matches the entry name used in the list operation, so that a user can list offsets, then edit the ConfigMap the operator created, then request the offsets to be altered using that same ConfigMap.
+This means the connector name expected will have `->` replaced with `,`, similar to in the list offsets operation.
```
scholzj, line 283: Same as above.

scholzj (commented):
Same more nits. But LGTM otherwise.

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,447 @@
...
+* GET /connectors/{connector}/offsets
+* PATCH /connectors/{connector}/offsets
+* DELETE /connectors/{connector}/offsets
+* PUT /connectors/{connector}/stop
+
+Support for the "stop" endpoint was added via [PR 9095](https://github.com/strimzi/strimzi-kafka-operator/pull/9095).
+Currently users that are utilising the KafkaMirrorMaker2 resource cannot make use of the API endpoints for managing offsets.
+
+Proposal [#076][proposal-76] discusses how to add support for these API endpoints to the KafkaConnector resource.
+However it did not cover the KafkaMirrorMaker2 resource.
```
mimaison, line 21: Offsets are actually only used by `MirrorSourceConnector`, so I wonder if we should only allow these operations on that connector?

Resetting/altering offsets of `MirrorCheckpointConnector` and `MirrorHeartbeatConnector` has no effect. If you try changing offsets on these connectors they still validate the request has a valid format.

000-mm2-connector-offsets-support.md
katheris: The other option I thought of was changing the `->` to `--`, just because `-` often seems to be used in aliases, so I was trying to pick something that might make it easier to spot the separator

000-mm2-connector-offsets-support.md
katheris: @mimaison if they aren't used why do the `MirrorCheckpointConnector` and `MirrorHeartbeatConnector` store offsets in Kafka?

mimaison: The Connect runtime automatically does that for source connectors. But these source offsets are not used by these 2 connectors to function.

mimaison: See https://github.com/apache/kafka/pull/14005 for more details.

000-mm2-connector-offsets-support.md
scholzj: I guess `--` would work for me. I can imagine `.` as well I guess as it is already used in the name. But `,` looks strange to me.

000-mm2-connector-offsets-support.md
scholzj: Yeah, but ...
* You do also a lot of patching which can cause issues and rewrite other annotations
* Wouldn't the user want to reset them at the same time and if one fails reset them all again when working with multiple connectors?

I do not have completely strong feelings for this if you wanna keep it. But if I was writing this, I would do all at once at the end.

PS: How do we do it with task restarts? Isn't that similar?

000-mm2-connector-offsets-support.md
scholzj: > @scholzj from an API perspective presumably the other connectors would inherit the listOffsets and alterOffsets properties? So I'm thinking we allow that but just remove the annotations without taking any action if the user requests list/alter/reset on a connector that isn't MirrorSourceConnector. Then if the mirrormaker-connector annotation includes both e.g. a MirrorSourceConnector and MirrorCheckpointConnector, we perform the action for MirrorSourceConnector but not MirrorCheckpointConnector. What do you think?

I would not be worried about the API I guess. The question is what to do if the user tries to work with those offsets:
1) Just reset them even if it doesn't influence anything?
2) Raise a warning and keep the annotation? (in which case you can also add a warning condition probably)
3) Raise the warning and remove the annotation? (in which case it is basically a warning in log only)

I do not have strong opinions for any of these options to be honest. I would probably go with 2) if I was doing it myself. But it is tight between the options for me, so I'm fine to take your path 😉

katheris: After reading the comments on the PR that Mickael linked I'd be inclined for us to allow users to list/alter/reset offsets for all the MirrorMaker connectors. In the comments there is a discussion about whether users should be allowed to do these operations on MirrorCheckpoint and MirrorHeartbeat, or whether instead an UnsupportedOperationException should be thrown. The conclusion was that although the offsets aren't used by those connectors, there may still be reasons why a user would want to do those operations.

Given that Kafka allows it, I think we should also allow it. I'll update the proposal to capture this

000-mm2-connector-offsets-support.md
katheris: For restarting connectors/tasks, you can only restart a single connector/task at once, so the annotation is `"strimzi.io/restart-connector-task=<mirrormaker_connector_name>:<task_id>"`

scholzj (commented):
Thanks @katheris

scholzj (approved):

katheris:
@scholzj on reflection on your comments, and after reviewing the code and the behaviour of the restart annotation, I have changed the proposal back to require a single connector is specified. Are you happy with this change?

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,449 @@
...
+```
+
+### New annotations
+
+All three actions will make use of a new annotation called `strimzi.io/connector-offsets`.
+The possible values will be `list`, `alter`, and `reset`.
+
+A second new annotation will be added called `strimzi.io/mirrormaker-connector`.
+It will be required when `strimzi.io/connector-offsets` is set.
+The value will be the name of the connector to apply the action to, for example `east-kafka->west-kafka.MirrorSourceConnector,east-kafka->west-kafka.MirrorCheckpointConnector`.
```
scholzj, line 115: So I guess this should be just a single connector now?

```suggestion
The value will be the name of the connector to apply the action to, for example `east-kafka->west-kafka.MirrorSourceConnector`.
```

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,449 @@
...
+      heartbeatConnector:
+        #...
+        listOffsets:
+          toConfigMap:
+            name: my-connector-offsets
+```
+
+The ConfigMap will contain a single entry, containing the response received from the GET /connectors/{connector}/offsets endpoint.
+The name of the entry will be based on the connector name, however since 
+[`>` characters are not allowed in ConfigMap keys](https://kubernetes.io/docs/concepts/configuration/configmap/#configmap-object) the name will be altered to replace `->` with `--`.
```
mimaison, line 157: Connector names can include other special characters that are not allows in ConfigMap keys. Will they also get replaced? What happens if there's a collision?

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,449 @@
...
+  - lastTransitionTime: "2024-06-04T08:44:15.913138115Z"
+    message: Failed to reset the connector offsets for east-kafka->west-kafka.MirrorCheckpointConnector due to "message from endpoint".
+    reason: ResetOffsets
+    status: "True"
+    type: Warning
+  ```
+The operator will leave the annotations on the KafkaMirrorMaker2 resource until either the delete operation succeeds, or the user removes the annotations.
+This means the operator will retry the delete on every operation, allowing the condition to remain present for the user to see.
+
+Strimzi will shortcut and automatically fail to do the reset if the connector does not have it's `state` set as `stopped` in the KafkaMirrorMaker2 resource.
```
mimaison, line 400: nit: `it's` -> `its`

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,449 @@
...
+
+However, this would make the operator vulnerable to changes in the way the MirrorMaker connectors store their offsets so was rejected in favour of less strict validation.
+
+### Preventing the user from altering or reseting offsets for the MirrorCheckpointConnector and MirrorHeartbeatConnector
+
+The MirrorCheckpointConnector and MirrorHeartbeatConnector both store offsets in Kafka, however they do not actually use those offsets.
+MirrorMaker users may still track offsets to track the progress of those connectors, but since they are not used, it isn't immediately obvious why a user would want to reset or alter them.
+This [Apache Kafka PR](https://github.com/apache/kafka/pull/14005) talks about some of the reasons why these operations may still be desired by users.
+In addition, it is possible that in future these connectors will make use of these offsets.
+For these reasons this proposal explicitly allows users to perform list/alter/reset on all MirrorMaker connectors.
```
mimaison, line 446: I wonder if it's still worth documenting this behavior as I expect some users may try to use these operations.

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,449 @@
...
+If the `strimzi.io/mirrormaker-connector` annotation is set, but the `strimzi.io/connector-offsets` annotation is missing the operator will take no action but add a warning message.
+
+### Listing offsets
+
+To list the current offsets the user will annotate the KafkaMirrorMaker2 resource with the new annotation `strimzi.io/connector-offsets` set to `list`, and the `strimzi.io/mirrormaker-connector` annotation.
+After the annotations are added, on the next reconciliation the operator will fetch the current offsets for the connector and create a Kubernetes ConfigMap containing the JSON response.
+The operator will add an owner reference to the Kubernetes ConfigMap pointing to the KafkaMirrorMaker2 resource that was annotated.
+If the ConfigMap already exists the operator will patch the ConfigMap with updated data and not add an owner reference.
+This means any existing keys in the ConfigMap that the operator is not updating will remain as before.
+Once all list operation is complete the operator will then remove both the `strimzi.io/mirrormaker-connector` and `strimzi.io/connector-offsets` annotations from the KafkaMirrorMaker2 CR.
```
mimaison, line 126: nit: `all list operations are`?

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,449 @@
...
+        listOffsets:
+          toConfigMap:
+            name: my-connector-offsets
+```
+
+The ConfigMap will contain a single entry, containing the response received from the GET /connectors/{connector}/offsets endpoint.
+The name of the entry will be based on the connector name, however since 
+[`>` characters are not allowed in ConfigMap keys](https://kubernetes.io/docs/concepts/configuration/configmap/#configmap-object) the name will be altered to replace `->` with `--`.
+For example a connector called `east-kafka->west-kafka.MirrorSourceConnector` will have offsets placed under `east-kafka--west-kafka.MirrorSourceConnector.json`.
+This will make it possible to use the same ConfigMap for different connectors.
```
mimaison, line 159: Above it says the ConfigMap will contain a single entry. If I understand correctly, it will be one entry per connector?

000-mm2-connector-offsets-support.md
mimaison: This is really unfortunate because I believe Strimzi puts the `->` characters in the connector names.

000-mm2-connector-offsets-support.md
katheris: This should be `Once the list operation is`, since I changed the proposal to only accept one connector at a time.

000-mm2-connector-offsets-support.md
katheris: Yeah it is Strimzi that puts the `->` character in. As far as I understand it Strimzi doesn't allow users to specify the connector names for MirrorMaker. It puts the name as `SOURCE_ALIAS->TARGET_ALIAS.CONNECTOR` and the alias field has the regex `"^[a-zA-Z0-9\\._\\-]{1,100}$"` so I don't think we will have a problem. @scholzj do you agree?

000-mm2-connector-offsets-support.md
katheris: I updated the proposal to only accept a single connector at a time, so the ConfigMap will only contain one entry to begin with, but if you ran list multiple times with different connectors they would each get added to the ConfigMap as a separate entry

000-mm2-connector-offsets-support.md
katheris: Yeah we can certainly document that we would only recommend running these operations for MirrorSourceConnector

000-mm2-connector-offsets-support.md
mimaison: Ah ok, then the only problematic character should be `>`. I did not know Strimzi had restrictions on the cluster aliases. Thanks

mimaison (approved):
LGTM

see-quick (approved):
LGTM, Thanks

tombentley (approved):
LGTM, thank you.

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,449 @@
...
+```
+
+### New annotations
+
+All three actions will make use of a new annotation called `strimzi.io/connector-offsets`.
+The possible values will be `list`, `alter`, and `reset`.
+
+A second new annotation will be added called `strimzi.io/mirrormaker-connector`.
+It will be required when `strimzi.io/connector-offsets` is set.
+The value will be the name of the connector to apply the action to, for example `east-kafka->west-kafka.MirrorSourceConnector`.
```
ppatierno, line 115: Because the JSON field in the output won't have the -> anyway but it's replaced by --, does it really make sense to have the name this way? Maybe having the -- or just - since the beginning as the annotation value? I am just worried that for users and support we have to refer to this replacement and clarify it multiple times.

im-konge (approved):

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,449 @@
...
+
+If the `strimzi.io/mirrormaker-connector` annotation is set, but the `strimzi.io/connector-offsets` annotation is missing the operator will take no action but add a warning message.
```
PaulRMellor, line 117: what happens in the opposite scenario where the strimzi.io/connector-offsets annotation is set, but the strimzi.io/mirrormaker-connector annotation is missing? We say this later: 

_If only one of the `strimzi.io/connector-offsets` and `strimzi.io/mirrormaker-connector` annotations is present on the KafkaMirrorMaker2 the operator will leave the annotation on the resource and update the KafkaMirrorMaker2 CR status to include a warning message_

Maybe we don't need to mention anything here?

000-mm2-connector-offsets-support.md
PaulRMellor: I think it might be useful to note the naming convention you mentioned in a later comment: _Strimzi doesn't allow users to specify the connector names for MirrorMaker. It puts the name as SOURCE_ALIAS->TARGET_ALIAS.CONNECTOR_

PaulRMellor (approved):
Nice, clear proposal and a simple way to manage MM2 connector offsets 
Couple of minor suggestions

000-mm2-connector-offsets-support.md
katheris: @ppatierno so you're suggesting that the user provides the `--` version of the connector name for both the annotation and the entry in the ConfigMap? I think that does make sense, then they don't have to remember which is which. @scholzj does that change make sense to you?

scholzj: TBH, I don't that makes sense and I think that would be way more confusing. I would keep the correct name as MM2 uses it.

000-mm2-connector-offsets-support.md
```diff
@@ -0,0 +1,449 @@
...
+Strimzi should include mechanisms to let users manage offsets when utilising the KafkaMirrorMaker2 resources.
+Many users of MirrorMaker make use of this resource, rather than the KafkaConnector resource, so it's important to support it here as well.
+Similar, to [proposal #76][proposal-76], this proposal suggests introducing a new annotation on the KafkaMirrorMaker2 resource to trigger each operation.
+The operations are:
+* List offsets.
+* Alter offsets.
+* Reset offsets.
+
+The list offsets operation uses a ConfigMap for the output.
+The alter offsets operation uses a ConfigMap for the input.
```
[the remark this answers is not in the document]
katheris: @PaulRMellor I'm unsure about where you're suggesting adding the part about reset operation. Each operation applies to a specified connector, which is selected using an annotation as explained on line 46.

000-mm2-connector-offsets-support.md
scholzj: @ppatierno Any chance you could get back to this ^^^? You are the only one who seems to have some open comments. It would be great to have things resolved to be able to merge it.

ppatierno: I think what I missed here is that the format with the `->` is the "official" one used by KafkaConnect, right?
If now my understanding is correct, then I think the current proposal makes more sense.

ppatierno (approved):

scholzj:
This has not 6 binding and 1 non-binding +1 votes. If there are no new comments on Monday, I will merge it as approved.

fvaleri (approved):
Nice proposal,thanks.

scholzj:
Approved with 6 binding and 2 non-binding +1 votes.

PR #130, opened 2024-09-10
Add possibility to provide a template to generate `host` and `advertisedHost` fields

scholzj:
This proposal suggests a way how to (optionally) template the `host` and `advertisedHost` fields in the listener configurations in the Kafka CR. Thanks to that, Strimzi users should not need to specify these for every single broker if they are using some predictable pattern.

### DIFF
--- a/084-templating-host-and-advertisedHost-fields.md
+++ b/084-templating-host-and-advertisedHost-fields.md
@@ -93,9 +93,9 @@ These templates will support variables, allowing for flexible configurations:
 Strimzi will take these template fields and replace the variables with the corresponding value for each Kafka node.
 And the resulting value would be used for the (advertised) hostnames.
 
-The templates will be applied only to the per-broker values.
-The hostname for the bootstrap Ingress resource would still need to be specified directly and will never be generated from the template.
-(Advertised hostnames can be used only as per-broker configuration)
+The templates will only be applied to per-broker values. 
+The hostname for the bootstrap Ingress resource must be specified manually and will not be generated from the template. 
+The advertised hostnames are used only for per-broker configurations and cannot be configured for bootstrap.
 
 ### Handling conflicts
 
--- a/084-templating-host-and-advertisedHost-fields.md
+++ b/084-templating-host-and-advertisedHost-fields.md
@@ -68,7 +68,7 @@ Being required to specify these configurations for each broker has some negative
   The brokers might now have different numbers and might include gaps.
   The controller nodes share the same node ID space, but their hostnames do not need to be configured.
   All of these easily lead to confusion of how should this configuration look like.
-* When adding new brokers, user needs to make sure the `host` or `advertisedHost` fields will be added for the new brokers before scaling the cluster up.
+* When adding new brokers, users need to make sure the `host` or `advertisedHost` fields will be added for the new brokers before scaling the cluster up.
 
 But in most cases, the (advertised) hostnames often follow the same pattern.
 In the example above, you can see that the Ingress based listener uses the hostname pattern `my-cluster-kafka-<NODE-ID>.myingress.com`.
@@ -85,10 +85,10 @@ This proposal suggests adding two new fields to the `configuration` section of t
 * `advertisedHostTemplate`
 * `hostTemplate`
 
-These fields will be strings that will be used to configure templates that will be used to generate the hostnames.
-The templates will support several different _variables_:
-* `{nodeId}` will be replaced with the ID of the Kafka node to which the template is applied
-* `{nodePodName}` will be replaced with the name of the Kafka node (name of the Kubernetes Pod used for this node) to which the template is applied
+The fields will be strings used to define templates for generating hostnames.
+These templates will support variables, allowing for flexible configurations:
+* The `{nodeId}` variable will be replaced with the ID of the Kafka node to which the template is applied.
+* The `{nodePodName}` variable will be replaced with the Kubernetes pod name for the Kafka node where the template is applied.
 
 Strimzi will take these template fields and replace the variables with the corresponding value for each Kafka node.
 And the resulting value would be used for the (advertised) hostnames.
@@ -158,7 +158,7 @@ Originally, I considered supporting some additional variables in the template fi
 
 But at the end I decided to not include them in the proposal.
 These values do not differ between the different Kafka nodes and can be easily handled by the user.
-If the `Kafka` CR is generated from some kind template (e.g. Helm), the cluster name / namespace can be easily templated outside of Strimzi.
+If the `Kafka` CR is generated from some kind of template (e.g. Helm), the cluster name / namespace can be easily templated outside of Strimzi.
 If the `Kafka` CR is written manually, these can be easily specified by the user.
 
 If we see sufficient demand for these variables, we can easily add them later.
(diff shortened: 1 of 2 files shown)


084-templating-host-and-advertisedHost-fields.md
```diff
@@ -0,0 +1,165 @@
...
+## Proposal
+
+This proposal suggests adding two new fields to the `configuration` section of the listener:
+* `advertisedHostTemplate`
+* `hostTemplate`
+
+These fields will be strings that will be used to configure templates that will be used to generate the hostnames.
+The templates will support several different _variables_:
+* `{nodeId}` will be replaced with the ID of the Kafka node to which the template is applied
+* `{nodeName}` will be replaced with the name of the Kafka node (name of the Kubernetes Pod used for this node) to which the template is applied
```
tombentley, line 91: Generally a node can be thought of as having several 'names'. There's the pod name, but also host names (with and without qualification). It not really ambiguous here, because we're _configuring_ the host names. But if someone looked at the CR with the assumption that the `nodeName` was a host name then there's potential for confusion.

Another point to bear in mind, if the proposal is accepted then maybe we end up with other templates in CRs. Ideally we'd use a common set of variable names across all those places. In those other places `nodeName` might be more ambiguous.

For these reasons perhaps `nodePodName` would be better?

084-templating-host-and-advertisedHost-fields.md
```diff
@@ -0,0 +1,165 @@
...
+Strimzi will take these template fields and replace the variables with the corresponding value for each Kafka node.
+And the resulting value would be used for the (advertised) hostnames.
+
+The templates will be applied only to the per-broker values.
+The hostname for the bootstrap Ingress resource would still need to be specified directly and will never be generated from the template.
+(Advertised hostnames can be used only as per-broker configuration)
+
+### Handling conflicts
+
+It is possible that the user specifies both the `advertisedHostTemplate` / `hostTemplate` as well as the `advertisedHost` / `host` fields for one or more Kafka brokers.
```
tombentley, line 102: Can we not use `oneOf`?

scholzj: I'm not sure OneOf would work given the fields are in very different places (+ one of them is inside an array). But I do not think it is needed. The fields fit well together if you need to override the template for some reason. We just need to establish the priority.

084-templating-host-and-advertisedHost-fields.md
scholzj: This makes sense. I updated it to `nodePodName`.

084-templating-host-and-advertisedHost-fields.md
```diff
@@ -0,0 +1,165 @@
...
+## Proposal
+
+This proposal suggests adding two new fields to the `configuration` section of the listener:
+* `advertisedHostTemplate`
+* `hostTemplate`
+
+These fields will be strings that will be used to configure templates that will be used to generate the hostnames.
+The templates will support several different _variables_:
+* `{nodeId}` will be replaced with the ID of the Kafka node to which the template is applied
+* `{nodePodName}` will be replaced with the name of the Kafka node (name of the Kubernetes Pod used for this node) to which the template is applied
```
PaulRMellor, line 91:
```suggestion
* The `{nodeId}` variable will be replaced with the ID of the Kafka node to which the template is applied.
* The `{nodePodName}` variable will be replaced with the Kubernetes pod name for the Kafka node where the template is applied.
```

084-templating-host-and-advertisedHost-fields.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Supporting additional variables
+
+Originally, I considered supporting some additional variables in the template fields:
+* `{clusterName}` will be replaced with the name of the Kafka cluster (name of the `Kafka` CR)
+* `{namespace}` will be replaced with the namespace of the Kafka cluster (namespace where the `Kafka` CR exists)
+
+But at the end I decided to not include them in the proposal.
+These values do not differ between the different Kafka nodes and can be easily handled by the user.
+If the `Kafka` CR is generated from some kind template (e.g. Helm), the cluster name / namespace can be easily templated outside of Strimzi.
```
PaulRMellor, line 161:
```suggestion
If the `Kafka` CR is generated from some kind of template (e.g. Helm), the cluster name / namespace can be easily templated outside of Strimzi.
```

PaulRMellor (approved):
Looks good and simplifies the configuration.
I left a few minor suggestions and an observation on not including the bootstrap , but I think the rationale to add later if required is sound.

084-templating-host-and-advertisedHost-fields.md
```diff
@@ -0,0 +1,165 @@
...
+The templates will support several different _variables_:
+* `{nodeId}` will be replaced with the ID of the Kafka node to which the template is applied
+* `{nodePodName}` will be replaced with the name of the Kafka node (name of the Kubernetes Pod used for this node) to which the template is applied
+
+Strimzi will take these template fields and replace the variables with the corresponding value for each Kafka node.
+And the resulting value would be used for the (advertised) hostnames.
+
+The templates will be applied only to the per-broker values.
+The hostname for the bootstrap Ingress resource would still need to be specified directly and will never be generated from the template.
+(Advertised hostnames can be used only as per-broker configuration)
```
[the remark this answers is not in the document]
scholzj: I don't think the suggestion makes sense here ... in particular the `This is because` part.
* For the `host` field, the template is not used for the bootstrap `host` because the bootstrap `host` would use a different format/pattern. You cannot apply the template such as `broker-{nodeId}.myingress.com` to the bootstrap as there is no node ID. And you also have only one bootstrap `host`, so having a separate template makes no sense either.
* For the `advertisedHostname`, that is not supported at the bootstrap level because it is a configuration inside the Kafka configuration and can exist only per broker. (Whereas the `host` configuration is on the Kubernetes level)

Maybe we could use something like this?

```suggestion
The templates will only be applied to per-broker values. 
The hostname for the bootstrap Ingress resource must be specified manually and will not be generated from the template. 
The advertised hostnames are used only for per-broker configurations and cannot be configured for bootstrap.
```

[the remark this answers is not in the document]
PaulRMellor: works for me :+1:

084-templating-host-and-advertisedHost-fields.md
scholzj: @tombentley Is that ^^^ sufficient as an explanation here?

tombentley: I think it's ok for this proposal. 

It would be nice if we could one day leverage [CRD validation rules](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules) for this kind of thing.

fvaleri (approved):
LGTM. Thanks.

tombentley (approved):
Thanks @scholzj, LGTM

im-konge (approved):
LGTM, thanks

see-quick (approved):
LGTM 👍

Frawless (approved):

scholzj:
This now has 5 binding and 1 non-binding +1 votes. If there are no new comments tomorrow evening, I will close it as approved.

ppatierno (approved):
LGTM. This is a really big simplification. Thanks!

katheris (approved):

scholzj:
Approved with 6 binding and 2 non-binding votes.

PR #131, opened 2024-09-12
OpenFeature integration

see-quick:
This proposal introduces the potential integration of the Strimzi Kafka operator with the feature flagging system `OpenFeature` FlagD. Motivation comes from [1].

[1] - https://github.com/strimzi/strimzi-kafka-operator/issues/7520

### DIFF
--- a/084-open-feature-integration.md
+++ b/084-open-feature-integration.md
@@ -303,8 +303,7 @@ spec:
         defaultVariant: 'off'
         state: ENABLED
         targeting:
-          targeting:
-          if:
+          - if:
             - or:
                 - and:
                     - "==":
@@ -314,8 +313,9 @@ spec:
                     - "==":
                       - var: clusterName
                       - "kafka-cluster-b"
-            - "on"
-            - "off"
+                - "on"
+                - "off"
+          
 #  ...
 ```
 In this case, we have defined `feature-gate-x`, which is by default disabled (i.e., defaultVariant is set to `off`).
@@ -384,7 +384,7 @@ this.featureGates.fetchFeatureFlag("feature-gate-x", false, Boolean.class, evalu
 - **Faster Iterations/Testing:** Features can be tested and rolled out quickly, speeding up development cycles.
 - **Centralized Management:** feature flagging system integration allows centralized control of feature flags, simplifying management across multiple components.
 - **Scalability:** The approach scales efficiently for larger deployments without adding operational complexity.
-- **Backwards Compatibility:** The proposal maintains support for the existing `STRIMZI_FEATURE_GATES` method, ensuring a smooth transition.
+- **Backwards Compatibility:** The proposal maintains support for the existing `STRIMZI_FEATURE_GATES` method (i.e., env-var provider), ensuring a smooth transition.
 
 ### Potential Challenges
 
@@ -394,6 +394,7 @@ this.featureGates.fetchFeatureFlag("feature-gate-x", false, Boolean.class, evalu
 ## Affected/Not Affected Projects
 
 `Cluster Operator`, `Topic Operator` and `User Operator`; meaning the modification will be done in scope of `strimzi-cluster-operator` project.
+Moreover, `operator-common` is also affected, because we modify FeatureGates class.
 
 ## Questions
 
--- a/084-open-feature-integration.md
+++ b/084-open-feature-integration.md
@@ -103,19 +103,24 @@ public class FeatureGates {
     private static final String OPEN_FEATURE_PROVIDER_NAME_ENV = "OPEN_FEATURE_PROVIDER_NAME"; // Environment variable to choose OpenFeature provider
     private static final String STRIMZI_FEATURE_GATES_ENV = "STRIMZI_FEATURE_GATES";
     
+    // instance holder
+    private static FeatureGates instance;
+    
     private final Client featureClient;
     private final FeatureProvider provider;
 
     // When adding new feature gates, do not forget to add them to allFeatureGates(), toString(), equals(), and `hashCode() methods
     private FeatureGate continueOnManualRUFailure;
 
+    // getInstance() method impl.
+    
     /**
      * Constructs the feature gates configuration.
      *
      * @param featureGatesConfig String with a comma-separated list of enabled or disabled feature gates (mainly for testing purposes)
      * @param evaluationContext EvaluationContext
      */
-    public FeatureGates(String featureGatesConfig, final EvaluationContext evaluationContext) {
+    private FeatureGates(String featureGatesConfig, final EvaluationContext evaluationContext) {
         this.provider = this.getProviderFromEnv();
         OpenFeatureAPI.getInstance().setProvider(this.provider);
         this.featureClient = OpenFeatureAPI.getInstance().getClient();
--- a/084-open-feature-integration.md
+++ b/084-open-feature-integration.md
@@ -22,7 +22,12 @@ already implemented by OpenFeature folks. Some of the common characteristics of 
 3. Requires rolling updates for changes.
 
 The third one is the most important for us because if we remove it, we will benefit it from in testing and also reducing 
-overall time during changes features, which user might want to include in their infrastructure.
+overall time during changes features, which users might want to include in their infrastructure.
+The third point is particularly crucial because the requirement for rolling updates introduces delays in the deployment process. 
+Every time a feature gate is changed or updated, the entire system needs to be restarted, which can slow down testing and development cycles. 
+By removing this constraint, we could dramatically enhance testing efficiency, allowing for more rapid iterations and immediate validation of feature changes. 
+Additionally, it would reduce the overall deployment time for any new feature or change, providing users with a infrastructure that responds faster to updates and modifications. 
+This would lead to smoother operations, potentially fewer disruptions, improving both developer and user experiences.
 
 On the other side for the more complex users with need of flagging system, we would support of [FlagD](https://github.com/open-feature/flagd),
 which is pretty known because [OpenFeature is CNCF Incubating project](https://www.cncf.io/projects/openfeature/).
@@ -83,9 +88,8 @@ The overall setup would look like this:
  * Class for handling the configuration of feature gates
  */
 public class FeatureGates {
-
     private static final String CONTINUE_ON_MANUAL_RU_FAILURE = "ContinueReconciliationOnManualRollingUpdateFailure";
-    private static final String FLAGD_ENABLED_ENV_VAR = "FLAGD_ENABLED"; // Environment variable to toggle FlagD
+    private static final String OPEN_FEATURE_PROVIDER_NAME_ENV = "OPEN_FEATURE_PROVIDER_NAME"; // Environment variable to choose OpenFeature provider
 
     private final Client featureClient;
     private final FeatureProvider provider;
@@ -100,52 +104,40 @@ public class FeatureGates {
      */
     public FeatureGates(String featureGateConfig) {
         // Set the appropriate provider based on the environment variable
-        this.provider = isFlagDEnabled() ? new FlagdProvider() : new EnvVarProvider();
+        this.provider = getProviderFromEnv();
         OpenFeatureAPI.getInstance().setProvider(this.provider);
         this.featureClient = OpenFeatureAPI.getInstance().getClient();
-
-        // Validate and parse the featureGateConfig if it's provided
-        if (featureGateConfig != null && !featureGateConfig.trim().isEmpty()) {
-            List<String> featureGates;
-
-            // Validate the format of the feature gate configuration string
-            if (featureGateConfig.matches("(\\s*[+-][a-zA-Z0-9]+\\s*,)*\\s*[+-][a-zA-Z0-9]+\\s*")) {
-                featureGates = asList(featureGateConfig.trim().split("\\s*,+\\s*"));
-            } else {
-                throw new InvalidConfigurationException(featureGateConfig + " is not a valid feature gate configuration");
-            }
-
-            // Validate each feature gate in the config to ensure it is recognized
-            for (String featureGate : featureGates) {
-                featureGate = featureGate.substring(1); // Remove the + or - sign
-
-                // Only validate feature gates but do not apply them manually
-                switch (featureGate) {
-                    case CONTINUE_ON_MANUAL_RU_FAILURE:
-                        // This is a valid feature gate; continue with processing
-                        break;
-                    default:
-                        throw new InvalidConfigurationException("Unknown feature gate " + featureGate + " found in the configuration");
-                }
-            }
-        }
-
-        // Fetch feature gates using OpenFeature
-        boolean continueOnManualRUFailureValue = fetchFeatureFlag(CONTINUE_ON_MANUAL_RU_FAILURE, true, Boolean.class);
-        setValueOnlyOnce(continueOnManualRUFailure, continueOnManualRUFailureValue);
-
-        // Validate interdependencies (if any)
-        validateInterDependencies();
+        
+        // ...
     }
 
     /**
-     * Checks whether FlagD is enabled via environment variables.
+     * Maps the value of OPEN_FEATURE_PROVIDER_NAME to the corresponding provider.
      *
-     * @return True if FLAGD_ENABLED is set to "true", otherwise false.
+     * @return The corresponding FeatureProvider instance based on the environment variable.
      */
-    private boolean isFlagDEnabled() {
-        String flagDEnabled = System.getenv(FLAGD_ENABLED_ENV_VAR);
-        return flagDEnabled != null && flagDEnabled.equalsIgnoreCase("true");
+    private FeatureProvider getProviderFromEnv() throws InvalidOptions {
+        String providerName = System.getenv(OPEN_FEATURE_PROVIDER_NAME_ENV);
+
+        // Default to EnvVarProvider if the environment variable is not set
+        if (providerName == null || providerName.trim().isEmpty()) {
+            return new EnvVarProvider();
+        }
+
+        // Create a mapping between the environment variable and providers
+        Map<String, FeatureProvider> providerMap = new HashMap<>();
+        providerMap.put("flagd", new FlagdProvider());
+        providerMap.put("env-var", new EnvVarProvider());
+        providerMap.put("flagsmith", new FlagsmithProvider());
+        providerMap.put("configcat", new ConfigCatProvider());
+        providerMap.put("statsig", new StatsigProvider());
+        providerMap.put("unleash", new UnleashProvider());
+        providerMap.put("jsonlogic", new JsonlogicProvider());
+        providerMap.put("flipt", new FliptProvider());
+        providerMap.put("go-feature-flag", new GoFeatureFlagProvider());
+
+        // Return the corresponding provider or default to EnvVarProvider
+        return providerMap.getOrDefault(providerName.trim().toLowerCase(), new EnvVarProvider());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


084-open-feature-integration.md
```diff
@@ -0,0 +1,477 @@
...
+
+## Proposal
+
+This proposal suggests the integration of OpenFeature, providing two primary methods for feature management:
+
+For the current users (i.e., having classic ENV variable STRIMZI_FEATURE_GATES), we will use [Enviroment Variable Provider](https://github.com/open-feature/java-sdk-contrib/tree/main/providers/env-var)
+already implemented by OpenFeature folks. Some of the common characteristics of such provider are:
+1. Default method currently used. 
+2. Feature gates are set via environment variables.
+3. Requires rolling updates for changes.
```
scholzj, line 22: You link to an existing provider. But how does it provide the backwards compatibility for how we configure the feature gates?

084-open-feature-integration.md
```diff
@@ -0,0 +1,477 @@
...
+
+The third one is the most important for us because if we remove it, we will benefit it from in testing and also reducing 
+overall time during changes features, which user might want to include in their infrastructure.
```
scholzj, line 25: This probably deserves some elaboration?

084-open-feature-integration.md
```diff
@@ -0,0 +1,477 @@
...
+
+On the other side for the more complex users with need of flagging system, we would support of [FlagD](https://github.com/open-feature/flagd),
+which is pretty known because [OpenFeature is CNCF Incubating project](https://www.cncf.io/projects/openfeature/).
+Common characteristics of this provider are:
+1. New method proposed for more complex users, which needs better management over multiple features (of Strimzi)
+2. Integrates with FlagD for dynamic feature flagging
+3. Allows changes without the need for rolling updates.
```
scholzj, line 32: I thought the way it would work is that:
* We would have our own default provider for how we configure the feature gates
* Users can use any of the provider (and while we might bundle some, they might also bundle their own)

But you make it sound here like we need to explicitly pick that we are gonna support Flagd?

084-open-feature-integration.md
```diff
@@ -0,0 +1,477 @@
...
+```java
+// ...
+// imports omitted for brevity
+// ...
+
+/**
+ * Class for handling the configuration of feature gates
+ */
+public class FeatureGates {
+
```
scholzj, line 86: No need for empty line.

084-open-feature-integration.md
```diff
@@ -0,0 +1,477 @@
...
+    private static final String CONTINUE_ON_MANUAL_RU_FAILURE = "ContinueReconciliationOnManualRollingUpdateFailure";
+    private static final String FLAGD_ENABLED_ENV_VAR = "FLAGD_ENABLED"; // Environment variable to toggle FlagD
```
scholzj, line 88: Again, why does this not configure the _provider_ instead of enabling a particular provider?

084-open-feature-integration.md
```diff
@@ -0,0 +1,477 @@
...
+         | Feature Flags  |                      +---------------------------+
+         +----------------+
+
+Where in each component (i.e., ClusterOperator, UserOperator and TopicOperator), we will
+fetch feature flags dynamically from the OpenFeature API, which is managed by OpenFeature Operator.
+Each operator’s logic that is controlled and **centralized** by FeatureGates class (e.g., enabling new behaviors, managing rolling updates) 
+will dynamically receive flag updates from `FlagD` via the OpenFeature Operator.
+
+Example of the feature flags 
+This is configuration example of the current `STRIMZI_FEATURE_GATES`
```
scholzj, line 252: I don't understand what this means.

084-open-feature-integration.md
```diff
@@ -0,0 +1,477 @@
...
+    // ...
+}
+
+// ClusterOperator reconcile() loop
+reconcile() {
+    // ...
+    maybeUpdateFeatureGateD();
+    // ...
+}
+```
```
scholzj, line 395: This is not how the code is implemented in most real usecases. You wouldhave the changes nested deep in the code.

084-open-feature-integration.md
```diff
@@ -0,0 +1,477 @@
...
+    return get(FEATURE_GATES);
+}
+```
+These are fetched directly from `ConfigParameter`, which encapsulates a map implementation with additional utility methods (e.g., a parser to convert the string representation into a specified type).
+In this case, the `FeatureGates` type is defined as:
+```java
+ /**
+ * Configuration string with feature gates settings
+ */
+public static final ConfigParameter<FeatureGates> FEATURE_GATES = new ConfigParameter<>("STRIMZI_FEATURE_GATES", parseFeatureGates(), "", CONFIG_VALUES);
```
scholzj, line 67: You describe how the configuration is handled today. But the impact on the configuration classes is not really clear.

scholzj (commented):
I left some comments. But apart from them:
* I think the abiility to configure the feature gates per-cluster si much more important than the ability to configure the feautre gates dynamically.
* I don't understand the focus on Flagd. My assumption was that OpenFeature provides us a generic API for feature flagging that works with many existing systems and allows users to bring their own implementations as well. This PR reads as if it is not the case - should we in that case wait for some actual demand for Flagd first? I think it is fine to use Flagd as an example of how it might work with the other systems. But the proposal seems to hardcode it in.
* I would focus less on including long code segments and more on the actual impact show in shorter examples and/or explained with words.
* The backward compatibility is not clear to me.

see-quick:
> I think the abiility to configure the feature gates per-cluster si much more important than the ability to configure the feautre gates dynamically.

Okay, I will think about how we can achieve it and update the proposal. 

> I don't understand the focus on Flagd. My assumption was that OpenFeature provides us a generic API for feature flagging that works with many existing systems and allows users to bring their own implementations as well. This PR reads as if it is not the case - should we in that case wait for some actual demand for Flagd first? I think it is fine to use Flagd as an example of how it might work with the other systems. But the proposal seems to hardcode it in.

You are correct OpenFeature provides a generic API for feature flagging I was thinking that we would start with just FlagD support. But looking at this again it's better to let the user decide what provider he wants to use...So I will update the proposal accordingly to be more generic and not focused only on FlagD.

084-open-feature-integration.md
see-quick: Simply by using just `STRIMZI_FEATURE_GATES` env var in all operators as we do it currently.

084-open-feature-integration.md
see-quick: Updated :) thanks

084-open-feature-integration.md
see-quick: For me, it was the easiest option to do it at the start. But I agree it would be better to let for instance in 
```java
private static final String FLAGD_ENABLED_ENV_VAR = "FLAGD_ENABLED"; // Environment variable to toggle FlagD
```
instead of `FLAGD_ENABLED` here use `FlagProvider` name and in our code we could add multiple providers based on that ENV-VAR. Here is the list, which is supported by `OpenFeature` [1]. So there are plenty of them. 

[1] - https://mvnrepository.com/artifact/dev.openfeature.contrib.providers

084-open-feature-integration.md
see-quick: Sure, this should be the provider and not just focus on FlagD :). Thanks

084-open-feature-integration.md
see-quick: Eventually, we can have possibly:
```java
/**
     * Maps the value of OPEN_FEATURE_PROVIDER_NAME to the corresponding provider.
     *
     * @return The corresponding FeatureProvider instance based on the environment variable.
     */
    private FeatureProvider getProviderFromEnv() throws InvalidOptions {
        String providerName = System.getenv(OPEN_FEATURE_PROVIDER_NAME_ENV);

        // Default to EnvVarProvider if the environment variable is not set
        if (providerName == null || providerName.trim().isEmpty()) {
            return new EnvVarProvider();
        }

        // Create a mapping between the environment variable and providers
        Map<String, FeatureProvider> providerMap = new HashMap<>();
        providerMap.put("flagd", new FlagdProvider());
        providerMap.put("env-var", new EnvVarProvider());
        providerMap.put("flagsmith", new FlagsmithProvider());
        providerMap.put("configcat", new ConfigCatProvider());
        providerMap.put("statsig", new StatsigProvider());
        providerMap.put("unleash", new UnleashProvider());
        providerMap.put("jsonlogic", new JsonlogicProvider());
        providerMap.put("flipt", new FliptProvider());
        providerMap.put("go-feature-flag", new GoFeatureFlagProvider());

        // Return the corresponding provider or default to EnvVarProvider
        return providerMap.getOrDefault(providerName.trim().toLowerCase(), new EnvVarProvider());
    }
```

see-quick: and then everything would be handled by the OpenFeatureAPI Client for every provider.

Frawless: so adding new provider will require code change?

see-quick: Yes, but currently `OpenFeature` has only these providers, which I have listed. So I think it would be just a one-timer change. If in the future a new flagging system comes into the picture (of course then we would need to add it :))

084-open-feature-integration.md
see-quick: Updated :) Such attributes in config classes would be removed and logic will be moved to the `FeatureGates`

084-open-feature-integration.md
```diff
@@ -0,0 +1,417 @@
...
+
+    +------------------------------+      +------------------------------------+
+    | Centralized Feature Flagging |      |               Strimzi              |
+    |            Server            |      | (Cluster, User and Topic Operator) |
+    +------------------------------+      +------------------------------------+
+                   |                                        |
+                   |                                        |
+                   |   <----------- API Calls ----------->  |
+
+Where in each component (i.e., ClusterOperator, UserOperator and TopicOperator), we will fetch feature flags dynamically from the OpenFeature API, which is managed by feature flagging server.
```
tombentley, line 175: What happens in the communication to the server fails? I guess we're unable to proceed with the reconciliation. Thus the availability of the Strimzi operators becomes coupled to the availability of this feature flagging server. I don't think that's a problem for the proposal, but calling out explicitly for users.

084-open-feature-integration.md
```diff
@@ -0,0 +1,417 @@
...
+        variants:
+          'on': true
+          'off': false
+        defaultVariant: 'off'
+        state: ENABLED
+        targeting:
+          - if:
+            - or:
+                - and:
+                    - "==":
```
tombentley, line 309: It looks like this `and` operator only has a single operand, which is weird. Is this a mistake/bad example, or is it imposed by the CRD validation schema for the `FeatureFlag` kind?

084-open-feature-integration.md
```diff
@@ -0,0 +1,417 @@
...
+
+- **Flexibility:** Users can toggle features without redeploying or restarting services.
+- **Faster Iterations/Testing:** Features can be tested and rolled out quickly, speeding up development cycles.
+- **Centralized Management:** feature flagging system integration allows centralized control of feature flags, simplifying management across multiple components.
+- **Scalability:** The approach scales efficiently for larger deployments without adding operational complexity.
+- **Backwards Compatibility:** The proposal maintains support for the existing `STRIMZI_FEATURE_GATES` method (i.e., env-var provider), ensuring a smooth transition.
+
+### Potential Challenges
+
+- **Complexity:** Increased complexity in configuration management.
```
tombentley, line 391: This is the big one for me. If there were lots of users asking for the functionality that open feature provides then it's much easier to justify the complexity. But AFAIK no one is asking for this. That means the benefits are really internal at this point (the testing benefit).

084-open-feature-integration.md
```diff
@@ -0,0 +1,417 @@
...
+
+this.featureGates.fetchFeatureFlag("feature-gate-x", false, Boolean.class, evaluationContent)
+// this will return `true`, because our defined targeting above.
+```
+
+### Benefits
+
+- **Flexibility:** Users can toggle features without redeploying or restarting services.
+- **Faster Iterations/Testing:** Features can be tested and rolled out quickly, speeding up development cycles.
+- **Centralized Management:** feature flagging system integration allows centralized control of feature flags, simplifying management across multiple components.
```
tombentley, line 385: I've not looked closely at OpenFeature, but externalising the logic used for turning feature gates on and off could be a double edged sword. In particular, if I'm editing a `FeatureFlag` how do I know what will be affected? It could be 'nothing', or it could be 'everything' (it's not even limited at just being Strimzi things AFAICS). Does Open Feature help in figuring out the impact, or is it up to the user to piece it together?

I also worry a bit about discoverability: When feature flags are part of the operator deployment it's pretty obvious. When they're in some other resource then you need to know about the existence of  `FeatureFlag` (as a thing), as well as discovering which instances of it (could there be multiple ones?) the operator.

tombentley (commented):
Thanks @see-quick, I left a few comments. Right now although I can see _some_ value I'm not completely convinced the advantages outweigh the disadvantages.

084-open-feature-integration.md
see-quick: > What happens in the communication to the server fails? I guess we're unable to proceed with the reconciliation. Thus the availability of the Strimzi operators becomes coupled to the availability of this feature flagging server

Well this can be handled I think because we would end up in a scenario where:
```java
2024-09-27 09:10:18 ERROR GrpcConnector:149 - failed to connect to event stream, exhausted retries
```
client (strimzi) wants to reconnect because he is getting a `GrpcConnector` error. We can catch that and use just the default value of `FEATURE_GATE`; therefore, we no longer depend on feature flagging server availability. And if such error would disappear (feature flagging system is up) we continue with a normal state.

084-open-feature-integration.md
see-quick: Yeah, this is a mistake/bad example. The correct should be (locally I have fixed in my example files but didn't update proposal thanks):
```yaml
targeting:
          if:
            - or:
                - "==":
                   - var: clusterName
                   - "kafka-cluster-a"
                - "==":
                   - var: clusterName
                   - "kafka-cluster-b"
            - "on"
            - "off"
```

084-open-feature-integration.md
see-quick: > I've not looked closely at OpenFeature, but externalising the logic used for turning feature gates on and off could be a double edged sword. In particular, if I'm editing a FeatureFlag how do I know what will be affected? It could be 'nothing', or it could be 'everything' (it's not even limited at just being Strimzi things AFAICS). Does Open Feature help in figuring out the impact, or is it up to the user to piece it together?

 It up to the user to piece it together.  little offtopic: Maybe one option would have some examples in DOCs but I am not sure about their maintainability in terms that there are plenty of providers (flagd is just one of them). Another point is that OpenFeature Operator supports just `FlagD`. So other providers would need to run in some form of Deployment and they all have different configurations options. 
 
 > FeatureFlag (as a thing), as well as discovering which instances of it (could there be multiple ones?) the operator
 
 Yeah you can deploy multiple of them AFAIK. Firstly you have to define `FeatureFlagSource` e.g.,
 ```yaml
 apiVersion: core.openfeature.dev/v1beta1
kind: FeatureFlagSource
metadata:
  name: strimzi-feature-gates-source
  namespace: myproject
  labels:
    app: open-feature-demo
spec:
  sources:
    - source: strimzi-feature-gates-flags
      provider: kubernetes
 ```
 And in source you have to `reference` your created FeatureFlag CR (i.e., `strimzi-feature-gates-flags`). But again this is just for `FlagD` and in case of other provider you would need to configure it differently (and without use of Operator cause their operator for now supports just FlagD as feature flagging server).

084-open-feature-integration.md
```diff
@@ -0,0 +1,414 @@
...
+Conceptually, the communication between Strimzi and a feature flagging system can be illustrated as:
+
+
+    +------------------------------+      +------------------------------------+
+    | Centralized Feature Flagging |      |               Strimzi              |
+    |            Server            |      | (Cluster, User and Topic Operator) |
+    +------------------------------+      +------------------------------------+
+                   |                                        |
+                   |                                        |
+                   |   <----------- API Calls ----------->  |
```
ppatierno, line 173: Should the OpenFeatureAPI client being configured to connect to the server? Where does this configuration live? Something to be added to the `Kafka` CR?

ppatierno (commented):
I left just one comment even because I am really scared about the complexity that OpenFeature is bringing. Moving from just configuring an env var to deploy a feature flagging server + using a `FeatureFlag` custom resource (if using openfeature operator)  looks like to be a big step compared to the advantages. Even just using the env var provider so actually staying where we are seems to add complexity or anyway changing nothig. 
Are we really sure this is something compelling to the users? Are we really sure they are complaining about restarting stuff because of a feature flag enabled/disabled?
Behind this I see too much added complexity compared to the value, plus more code to maintain, documenting and testing.

see-quick:
> I left just one comment even because I am really scared about the complexity that OpenFeature is bringing. Moving from just configuring an env var to deploy a feature flagging server + using a `FeatureFlag` custom resource (if using openfeature operator) looks like to be a big step compared to the advantages. Even just using the env var provider so actually staying where we are seems to add complexity or anyway changing nothig. Are we really sure this is something compelling to the users? Are we really sure they are complaining about restarting stuff because of a feature flag enabled/disabled? Behind this I see too much added complexity compared to the value, plus more code to maintain, documenting and testing. Anyway I read about OpenFeature, but not is so much details. Maybe I am missing a lot from it.

First of all, if we just start with `EnvVarProvider` it does not include any complexity (if we do not count dependencies). We would simply remove dependency from Config classes and use just only `FeatureGates` class as source of truth. 

Secondly, that complexity is not so enormous compared to other features. 

> Are we really sure this is something compelling to the users? Are we really sure they are complaining about restarting stuff because of a feature flag enabled/disabled?

Well, it's not only about feature flag `enabled/disabled` but also controlling feature gates based on rules, which can user define (i.e., in namespace-0 with kafka-cluster-a -> feature-a-enabled) and in other namespace such feature would be disabled etc. 

> Behind this I see too much added complexity compared to the value, plus more code to maintain, documenting and testing.
Anyway I read about OpenFeature, but not is so much details. Maybe I am missing a lot from it.

The complexity here is just adding and modifying one class (i.e., `FeatureGates`), which would handle all about feature gates and will be treated as a singleton. All dependencies from config classes would be removed. And lastly, we would need to handle client configuration (for providers such as flagd, flagsmith etc.). But for start we can use default configuration (as I test it where client (`flagd`) connects to service name of feature flag IMO.

ppatierno:
> First of all, if we just start with EnvVarProvider it does not include any complexity (if we do not count dependencies). We would simply remove dependency from Config classes and use just only FeatureGates class as source of truth.

Well, adding dependencies on other projects, add complexity by default for me on different aspects. And also just using the EnvVarProvider would not bring any advantage compared to what we have today.

> Well, it's not only about feature flag enabled/disabled but also controlling feature gates based on rules, which can user define (i.e., in namespace-0 with kafka-cluster-a -> feature-a-enabled) and in other namespace such feature would be disabled etc.

Is there any demands for something like this?

see-quick:
> Well, adding dependencies on other projects, add complexity by default for me on different aspects. And also just using the EnvVarProvider would not bring any advantage compared to what we have today.

Sure. But that's not the goal of this proposal, `EnvVarProvider` is for using the current approach, which we are using. 

> Is there any demands for something like this?

Well [1]? 

[1] - https://github.com/strimzi/strimzi-kafka-operator/issues/7520

ppatierno:
[1] is an issue opened by Jakub not a demand coming from users. I was looking for tangible demand to understand if it's worth doing this.

im-konge:
Triaged on 31.10.2024: Because of the efforts in terms of maintainability, testability and complexity, we decided to close this now and maybe in the future re-open it, in case that it will bring more value.

PR #134, opened 2024-10-07
Add proposal for archiving Canary

im-konge:
This PR adds proposal for archiving Canary project as discussed on community call on 30th May 2024.

### DIFF
--- a/086-archive-canary.md
+++ b/086-archive-canary.md
@@ -1,0 +1,92 @@
+# Archive Canary project written in Go
+
+This proposal is about archiving the [Canary project written in Go](https://github.com/strimzi/strimzi-canary).
+
+## Current situation
+
+The Canary project, currently written in Go programming language, is component for monitoring the Kafka cluster.
+It provides functionality to periodically check the availability of the Kafka cluster during normal run, upgrades, downgrades, and rolling updates.
+That is done by connection check to the cluster, but also producer and consumer, that periodically do the message transmission between Kafka cluster and Canary.
+The Canary then provides Prometheus metrics and alerts for users or cluster admins to react on the Kafka cluster issues.
+
+## Motivation
+
+The Canary, as mentioned, is written in Go, however the Strimzi organization (and the engineers working on the projects inside this organization) is 
+focusing on Java programming language.
+That means the team is missing needed expertise in Go in order to solve various issues that Canary can have and has.
+Based on the previous proposal from PR [#58](https://github.com/strimzi/proposals/pull/58), the Sarama Kafka client library lacks of features that are available in official 
+Java Kafka client library - and the Sarama library is not the official Kafka library in comparison to the Java one.
+Additionally, there are multiple [issues created in the Canary project](https://github.com/strimzi/strimzi-canary/issues) without any comment or resolution for quite a long time.
+The project's dependencies weren't updated for two years now, meaning that there can be a lot of CVEs, unresolved issues with newer Kafka versions, and more.
+Finally, the inactivity on the project proves that there is not enough resources in order to maintain such project on level of fixing the issues and CVEs, neither to add more features to the tool.
+
+Because of the issues mentioned on the above lines, I'm proposing to archive the Canary project.
+
+## Canary in Java
+
+After few releases of Canary we realized that there is not enough expertise for the Go language and Sarama Kafka client doesn't provide functionality we need, 
+the [PR with proposal to move Canary to Java](https://github.com/strimzi/proposals/pull/58) was created.
+The proposal contained all issues with the current implementation together with proposed implementation and changes for the Canary in Java.
+In parallel with the proposal, the POC was written in Java and is available in the [im-konge/canary-java](https://github.com/im-konge/canary-java) repository.
+However, during the implementation process, we found out that few things are not possible using Java Kafka clients (for example the connection check that was one of the main features of Canary) and that
+metrics related to Sarama client are not relevant anymore.
+Changes like this would break the backwards compatibility, meaning that it would not be 1:1 copy of the Canary in Go.
+
+Other than that, we thought about changing the metrics provided by Canary to be more insightful, but we didn't agree on what the metrics should be.
+We had discussions with few community users about how the Canary can be more useful to them, but the users ended up writing their own Canary-like tool 
+with functionality useful to them.
+
+Because we didn't move forward with the proposal and agreement on the future of the Canary, after more than one year from the moment the proposal was created, we decided to close the 
+proposal, and we agreed to archive the project on [community meeting on May 30th 2024](https://youtu.be/UpStul__uCw?si=GTA5edXJEnGgxP1a).
+Additionally, we mentioned that in case that community finds Canary useful, provides information about what can be improved (in terms of metrics, functionality), and we will have enough
+capacity and resources to maintain it, we can open a new proposal for writing the Canary vol.2 in Java.
+
+## Proposal
+
+After this proposal is approved, we should:
+- archive the Canary project
+- remove the Canary install files from Strimzi operators repository - from both `/install` and `/packaging/install` folders
+- delete the mentions about the Canary project from the documentation and automation files (Makefiles)
+- inform users about archiving the project
+
+In case that community: 
+
+- finds Canary useful 
+- will be able to provide additional information about what metrics or functionality can be improved or used
+- we will be able to maintain the project
+
+we can write a new proposal for Canary vol.2 in Java.
+
+## Affected/not affected projects
+
+The only affected project is the [Canary](https://github.com/strimzi/strimzi-canary) that should be archived.
+In the Strimzi operators repository, the only affected parts are the [installation files](https://github.com/strimzi/strimzi-kafka-operator/tree/main/install/canary) and 
+[in development installation files](https://github.com/strimzi/strimzi-kafka-operator/tree/main/packaging/install/canary) that should be deleted by this proposal.
+
+## Compatibility
+
+The backwards compatibility is not relevant in this case, as the project will be archived and there is no other solution currently that would replace it.
+
+## Rejected alternatives
+
+### Maintaining Canary in Go and providing additional functionality
+
+One of the rejected alternative was to keep updating the Canary in Go and add more functionality to it.
+This was discussed and rejected because of:
+- lack of Go experts in the Strimzi organization
+- resources - it would take a lot of time to learn Go and properly testing every new change without knowing how it will work
+- missing functionality in the Sarama Kafka client
+
+### Maintaining Canary for dependency updates
+
+Another one alternative was to keep Canary project and updating the dependencies.
+Some of the dependency updates can be without breaking changes, but from time to time there are changes in the dependencies that requires additional 
+changes to the code, which brings us to the same situation as the previous alternative - someone would need to do the changes to the code, 
+test it properly and then release it.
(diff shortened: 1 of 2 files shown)


fvaleri (approved):
Makes sense. Thanks.

see-quick (approved):

086-archive-canary.md
```diff
@@ -0,0 +1,92 @@
...
+Because we didn't move forward with the proposal and agreement on the future of the Canary, after more than one year from the moment the proposal was created, we decided to close the 
+proposal, and we agreed to archive the project on [community meeting on May 30th 2024](https://youtu.be/UpStul__uCw?si=GTA5edXJEnGgxP1a).
+Additionally, we mentioned that in case that community finds Canary useful, provides information about what can be improved (in terms of metrics, functionality), and we will have enough
+capacity and resources to maintain it, we can open a new proposal for writing the Canary vol.2 in Java.
+
+## Proposal
+
+After this proposal is approved, we should:
+- archive the Canary project
+- remove the Canary install files from Strimzi operators repository - from both `/install` and `/packaging/install` folders
```
ppatierno, line 48: wondering if this should go at least in two steps, across two Strimzi releases. Deprecation (just in the CHANGELOG and/or documentation if any) and removal.

im-konge: I think we can, but how the install files will differ in comparison to other versions of Strimzi? Or do you think that many users are using the Canary from the operators' install files? I guess we don't even know if everything works as we didn't try the install files for a while.

scholzj: @ppatierno What do you expect to be the benefit? The last commit was 6 months ago, the last release was 2 years ago. So what would be the benefit of deprecating it first before removing it if it is unupdated for such a long time already? We are not going to delete any container images, so whatever works today would continue to work the same way regardless. Or do you expect it to gether new interest and new users/contributors?

scholzj, line 48: Strictly speaking, we should not remove them from `/install`. From there, they should be removed only as part of the next Strimzi release.

im-konge: Good point, I added it there. Thanks

ppatierno: I was just trying to follow the path we usually follow when removing something from the operator. It's not code or a feature but it's a component installed with some YAMLs. I can understand it has less impact but warning people that those installation files will be not available anymore in the future is good behaviour instead of removing them straightaway. If someone is using it they could make the decision to move away from it after our deprecation/removal warning.

Frawless (approved):

086-archive-canary.md
Frawless: The removal was discussed few times on community calls. We have the proposal now and as Jakub mentioned, everything will work just fine even after examples removal, only the install files will be gone. Released install files will be in-place until next release and because there are no changes for quite a long time, I don't see any problem in just removing it and in case anyone want to use it then just copy-out the examples and use it (that's what anyway every user should do). So I am up for deletion right after proposal approval.

086-archive-canary.md
```diff
@@ -0,0 +1,94 @@
...
+Additionally, there are multiple [issues created in the Canary project](https://github.com/strimzi/strimzi-canary/issues) without any comment or resolution for quite a long time.
+The project's dependencies weren't updated for two years now, meaning that there can be a lot of CVEs, unresolved issues with newer Kafka versions, and more.
+Finally, the inactivity on the project proves that there is not enough resources in order to maintain such project on level of fixing the issues and CVEs, neither to add more features to the tool.
+
+Because of the issues mentioned on the above lines, I'm proposing to archive the Canary project.
+
+## Canary in Java
+
+After few releases of Canary we realized that there is not enough expertise for the Go language and Sarama Kafka client doesn't provide functionality we need, 
+the [PR with proposal to move Canary to Java](https://github.com/strimzi/proposals/pull/58) was created.
```
PaulRMellor, line 28:
```suggestion
so the [PR proposing to move Canary to Java](https://github.com/strimzi/proposals/pull/58) was created.
```

086-archive-canary.md
```diff
@@ -0,0 +1,94 @@
...
+The proposal contained all issues with the current implementation together with proposed implementation and changes for the Canary in Java.
+In parallel with the proposal, the POC was written in Java and is available in the [im-konge/canary-java](https://github.com/im-konge/canary-java) repository.
+However, during the implementation process, we found out that few things are not possible using Java Kafka clients (for example the connection check that was one of the main features of Canary) and that
+metrics related to Sarama client are not relevant anymore.
+Changes like this would break the backwards compatibility, meaning that it would not be 1:1 copy of the Canary in Go.
+
+Other than that, we thought about changing the metrics provided by Canary to be more insightful, but we didn't agree on what the metrics should be.
```
PaulRMellor, line 35:
```suggestion
Other than that, we thought about changing the metrics provided by Canary to be more insightful, but we were unable to reach a consensus on the specific metrics to include.
```

086-archive-canary.md
```diff
@@ -0,0 +1,94 @@
...
+We had discussions with few community users about how the Canary can be more useful to them, but the users ended up writing their own Canary-like tool
```
PaulRMellor, line 36:
```suggestion
We had discussions with community users about how the Canary can be more useful to them, but the users ended up writing their own Canary-like tool 
```

086-archive-canary.md
```diff
@@ -0,0 +1,94 @@
...
+Changes like this would break the backwards compatibility, meaning that it would not be 1:1 copy of the Canary in Go.
+
+Other than that, we thought about changing the metrics provided by Canary to be more insightful, but we didn't agree on what the metrics should be.
+We had discussions with few community users about how the Canary can be more useful to them, but the users ended up writing their own Canary-like tool 
+with functionality useful to them.
+
+Because we didn't move forward with the proposal and agreement on the future of the Canary, after more than one year from the moment the proposal was created, we decided to close the 
+proposal, and we agreed to archive the project on [community meeting on May 30th 2024](https://youtu.be/UpStul__uCw?si=GTA5edXJEnGgxP1a).
+Additionally, we mentioned that in case that community finds Canary useful, provides information about what can be improved (in terms of metrics, functionality), and we will have enough
+capacity and resources to maintain it, we can open a new proposal for writing the Canary vol.2 in Java.
```
PaulRMellor, line 42:
```suggestion
If the community finds Canary valuable, provides feedback on how to improve its metrics and functionality, and we have sufficient capacity and resources, we will consider proposing a new project to rewrite Canary in Java.
```

086-archive-canary.md
```diff
@@ -0,0 +1,94 @@
...
+- delete the mentions about the Canary project from the documentation and automation files (Makefiles)
+- inform users about archiving the project
+
+In case that community: 
+
+- finds Canary useful 
+- will be able to provide additional information about what metrics or functionality can be improved or used
+- we will be able to maintain the project
+
+we can write a new proposal for Canary vol.2 in Java.
```
PaulRMellor, line 60:
```suggestion
If the community:

- finds Canary useful
- provides additional information about how to improve its metrics or functionality
- and we have the capacity to maintain the project

we can consider proposing a new project to rewrite Canary in Java.
```

086-archive-canary.md
```diff
@@ -0,0 +1,94 @@
...
+
+One of the rejected alternative was to keep updating the Canary in Go and add more functionality to it.
+This was discussed and rejected because of:
+- lack of Go experts in the Strimzi organization
+- resources - it would take a lot of time to learn Go and properly testing every new change without knowing how it will work
+- missing functionality in the Sarama Kafka client
+
+### Maintaining Canary for dependency updates
+
+Another one alternative was to keep Canary project and updating the dependencies.
```
PaulRMellor, line 84:
```suggestion
Another alternative was to keep Canary project and updating the dependencies.
```

086-archive-canary.md
```diff
@@ -0,0 +1,94 @@
...
+Some of the dependency updates can be without breaking changes, but from time to time there are changes in the dependencies that requires additional 
+changes to the code, which brings us to the same situation as the previous alternative - someone would need to do the changes to the code, 
+test it properly and then release it.
+Because of these issues, we decided to reject this alternative.
+
+### Rewrite Canary in Java
+
+Final alternative was to rewrite Canary in Java, but as was mentioned in the [Canary in Java](#canary-in-java), it would not be the same Canary as in Go,
```
PaulRMellor, line 92:
```suggestion
A final alternative was to rewrite Canary in Java, but as was mentioned in the [Canary in Java](#canary-in-java), it would not be the same Canary as in Go, 
```

PaulRMellor (approved):
Justification for archiving makes sense.
I left a few comments for readability as I reviewed.

im-konge:
Thanks a lot for the suggestions @PaulRMellor ! 🙂 All are now applied.

086-archive-canary.md
scholzj: I personally don't think the warning period is needed. But I think it is easy to say that the warning will be added in 0.44 and the files will be removed in 0.46 (with all the other major changes) if Paolo things it would be better. Even if it might be unnecessary it does not hurt anyone. 🤷

ppatierno: I thought more about this. Maybe I was overthinking on something which was not so requested. I don't want to block or make more complex the approach. Let's go with direct removal on next release.

ppatierno (approved):

086-archive-canary.md
```diff
@@ -0,0 +1,92 @@
+# Archive Canary project written in Go
```
scholzj, line 1: I find this quite confusing that you talk especially here in the title but also in the text about _Archive Canary project written in Go_. That suggests that there is some other Canary that is not going to be archived. But what the proposal really suggests is _Archive the Canary project_, period. This should be clear from the title as well as in the introduction paragraph (line 3).

086-archive-canary.md
```diff
@@ -0,0 +1,92 @@
...
+
+The Canary project, written in Go programming language, is a component for monitoring a Kafka cluster.
+It provides functionality to periodically check the availability of the Kafka cluster during normal run, upgrades, downgrades, and rolling updates.
+That is done by connection check to the cluster, but also producer and consumer, that periodically do the message transmission between Kafka cluster and Canary.
+The Canary then provides Prometheus metrics and alerts for users or cluster admins to react on the Kafka cluster issues.
+
+## Motivation
+
+The Canary, as mentioned, is written in Go, however the Strimzi organization (and the engineers working on the projects inside this organization) is focused on using the Java programming language.
+This means the team is missing the necessary expertise in Go in order to solve various issues in Canary.
```
scholzj, line 15: You do not need to change it. But I think this is not the real reason and it is just an excuse. The real reason is IMHO just not enough time and interest to keep maintaining it.

im-konge: Yeah I can add it as a main point. Thanks

086-archive-canary.md
im-konge: Good point.
I used the `Canary written in Go` often in the past, so that's why I wrote it here as well.
Anyway, I changed it to make it clear. Thanks

scholzj (approved):
Thanks. +1 (binding).

tombentley (approved):

im-konge:
> When it's approved, @im-konge are you going to take care of next steps to archive the project?

@ppatierno I will, thanks

katheris (approved):

PR #135, opened 2024-11-14
External Certificate Manager

katheris:
This proposal aims to allow Strimzi users to use an external certificate manager, specifically [cert-manager](https://cert-manager.io/), to manage certificates.

Related to https://github.com/strimzi/strimzi-kafka-operator/issues/929


087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+* Requiring that self-signed certificates cannot be used.
+
+## Proposal
+
+Strimzi will be updated to allow users to specify that certificates should be issued by an external certificate manager, rather than issued by the Cluster Operator.
+This proposal will specifically describe how this would work for cert-manager, however the user API for configuration will be written in a way that does not prevent other external certificate managers being added in the future.
+
+The proposal makes a few assumptions:
+* Strimzi will not be responsible for installing cert-manager, but we will document the versions of cert-manager that we have tested with.
+* Strimzi will not be responsible for creating `Issuer` or `ClusterIssuer` custom resources.
```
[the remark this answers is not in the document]
katheris: This is because there are lots of different issuers that work with cert-manager. So rather than Strimzi having to actively support all the different types, I've proposed that the user creates the `Issuer` or `ClusterIssuer` and handles supplying a Secret with the trusted certificates for the issuer they have chosen. That way Strimzi can work with any cert-manager issuer integrations.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+When the user creates the CA cert Secret they must add the `strimzi.io/ca-cert-generation` annotation to it.
+If the user updates the Secret to change the certificates included they must increment the annotation to inform Strimzi it has changed.
+Similar to today Strimzi will put the annotation on the pods (Kafka, ZooKeeper etc) to be able to spot when the generation has been changed.
+
+### Issuing certificates
+
+If the user has enabled this feature, when Strimzi needs to issue a certificate, instead of using the existing internal mechanism it will create a `Certificate` custom resource.
+Strimzi will wait during the reconciliation loop for the `Certificate` status to indicate that the certificate has been issued before continuing.
+When issuing cluster certificates (e.g for Kafka etc), once the certificate has been issued, Strimzi will annotate the cert-manager provided Secret with the `strimzi.io/server-cert-hash` annotation with the value being the hash of the certificate in the Secret.
```
tinaselenge, line 101: `issuing cluster certificates (e.g for Kafka etc)`  - I wonder if it would be useful to include the secret names for these certificates as an example.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+* The `Secret` that contains the CAs will be the same `Secret` currently used, so either `<CLUSTER_NAME>-cluster-ca-cert` or `<CLUSTER_NAME>-clients-ca-cert`.
+
+### Handling trust rollout
+
+Similar to today Strimzi will use the notion of a "generation" to determine whether to roll the cluster to pick up changes in either the `<CLUSTER_NAME>-cluster-ca-cert` or `<CLUSTER_NAME>-clients-ca-cert`.
+When the user creates the CA cert Secret they must add the `strimzi.io/ca-cert-generation` annotation to it.
+If the user updates the Secret to change the certificates included they must increment the annotation to inform Strimzi it has changed.
+Similar to today Strimzi will put the annotation on the pods (Kafka, ZooKeeper etc) to be able to spot when the generation has been changed.
+
+### Issuing certificates
```
tinaselenge, line 97:
```suggestion
### Issuing end-entity certificates
```

tinaselenge (approved):
Thanks for the proposal Kate, it looks good to me.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+
+None of the existing options allow the certificate management to be done completely separately from Strimzi.
+
+## Motivation
+
+Strimzi's primary purpose is to provide a way to run Apache Kafka clusters on Kubernetes.
+Although it is nice that it can manage certificates, it would be beneficial if the certificates could be managed by a dedicated certificate manager, such as [cert-manager](https://cert-manager.io/).
+This is a feature that is often requested, especially because many organizations have specific compliance requirements with regard to certificates, for example:
+* Requiring that CA private keys are not shared.
+* Requiring that self-signed certificates cannot be used.
```
scholzj, line 37: I don't think this really helps because the CA will be anyway bootstrapped as self-signed as it is today in most cases, and there is not much we can do about it.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+    generateCertificateAuthority: <boolean>
+    generateSecretOwnerReference: <boolean>
+    renewalDays: <integer> # days before notAfter when we should start renewal
+    certificateExpirationPolicy: <renew-certificate|replace-key>
+    certificateIssuer:
+      type: <internal|cert-manager.io> # (1)
+      issuerRef: # (2)
+        name: <string>
+        kind: <Issuer|ClusterIssuer>
+        group: <string> # cert-manager.io by default
```
scholzj, line 69: Should the type separation happen already at the `clusterCa` level? It would seem more logical to me.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+  The `name`, `kind`, and `group` properties will be copied over into the `Certificate` custom resource Strimzi creates.
+
+### User steps
+
+To make use of this new option the user will have to:
+
+1. Install cert-manager.
+2. Create an `Issuer` or `ClusterIssuer` custom resource.
+3. Create a `Secret` containing the CAs for Strimzi to trust.
+   Users can optionally use [trust-manager](https://cert-manager.io/docs/trust/trust-manager/) to create this Secret, but they are responsible for installing trust-manager, creating the `Bundle` CR and annotating the resulting Secret with the Strimzi cert annotation.
```
scholzj, line 84: Maybe we should instead (or next to this) describe the expectation of how the Secret should look like? IIRC, trust-manager creates a Secret will all CAs bundled into a single file? Is that supported / expected?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+4. Create a `Kafka` resource with `clusterCa.certificateIssuer` and/or `clientsCa.certificateIssuer` configured.
+
+Notes:
+* The `Secret` that contains the CAs will be the same `Secret` currently used, so either `<CLUSTER_NAME>-cluster-ca-cert` or `<CLUSTER_NAME>-clients-ca-cert`.
```
scholzj, line 88: I think we should again aim here for a separation of the Strimzi used Secrets and the user-provided Secrets. I.e. the user should provide it in _some_ custom secret and we should copy it ourself into `<CLUSTER_NAME>-cluster-ca-cert` or `<CLUSTER_NAME>-clients-ca-cert` if needed.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Handling trust rollout
+
+Similar to today Strimzi will use the notion of a "generation" to determine whether to roll the cluster to pick up changes in either the `<CLUSTER_NAME>-cluster-ca-cert` or `<CLUSTER_NAME>-clients-ca-cert`.
+When the user creates the CA cert Secret they must add the `strimzi.io/ca-cert-generation` annotation to it.
```
scholzj, line 93: * Does this work with the trust-manager suggested earlier?
* Can we work around it and manage the generation ourself? E.g. based on the hash of the user-provided certificate detect changes and bump the generation ourself?
* What is the impact on the `strimzi.io/ca-key-generation` given we now do not have the private key secret?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+If the user updates the Secret to change the certificates included they must increment the annotation to inform Strimzi it has changed.
+Similar to today Strimzi will put the annotation on the pods (Kafka, ZooKeeper etc) to be able to spot when the generation has been changed.
+
+### Issuing certificates
+
+If the user has enabled this feature, when Strimzi needs to issue a certificate, instead of using the existing internal mechanism it will create a `Certificate` custom resource.
+Strimzi will wait during the reconciliation loop for the `Certificate` status to indicate that the certificate has been issued before continuing.
```
scholzj, line 100: How long will we wait? The ussual operation timeout?

087-external-certificate-manager.md
scholzj, line 101: We do not control that Secret, so we should likely not annotate it / rely on its annotation. We need to cary it on the internal Secrets maybe?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+### Handling trust rollout
+
+Similar to today Strimzi will use the notion of a "generation" to determine whether to roll the cluster to pick up changes in either the `<CLUSTER_NAME>-cluster-ca-cert` or `<CLUSTER_NAME>-clients-ca-cert`.
+When the user creates the CA cert Secret they must add the `strimzi.io/ca-cert-generation` annotation to it.
+If the user updates the Secret to change the certificates included they must increment the annotation to inform Strimzi it has changed.
+Similar to today Strimzi will put the annotation on the pods (Kafka, ZooKeeper etc) to be able to spot when the generation has been changed.
+
+### Issuing certificates
+
+If the user has enabled this feature, when Strimzi needs to issue a certificate, instead of using the existing internal mechanism it will create a `Certificate` custom resource.
```
scholzj, line 99: Maybe we should make it clear that this is where we control the CN / SANs of the certificates?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+### Tracking changes to cluster end-entity certificates
+
+Cert-manager will be responsible for renewing all end-entity certificates.
+When a certificate is renewed cert-manager will update the related Secret.
+
+For cluster certificates (e.g. for Kafka etc), Strimzi will track and handle these changes using the `strimzi.io/server-cert-hash` annotation.
+During the reconciliation loop, even if all cluster end-entity certificates have been issued, Strimzi will patch the certificate Secrets with the correct `strimzi.io/server-cert-hash` annotation.
+The value of this annotation can then be compared with the value on the pods to determine whether the pods need to be restarted to pick up a new Secret.
+
+For user certificates (issued by the User Operator), the user will be responsible for making sure their applications notice cert-manager renewing the certificates and are updated to use the new certificate.
```
scholzj, line 113: I guess UO and the Clients CA deserves more attention here? Will the UO issue the `Certificate` resources? How will it keep the certificate Secrets? Or will `type: tls-external` be used here?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+Similar to today, Strimzi will also add this hash annotation to the pods to track whether they are mounting the latest version of the Secret.
+
+### Tracking changes to cluster end-entity certificates
+
+Cert-manager will be responsible for renewing all end-entity certificates.
+When a certificate is renewed cert-manager will update the related Secret.
+
+For cluster certificates (e.g. for Kafka etc), Strimzi will track and handle these changes using the `strimzi.io/server-cert-hash` annotation.
+During the reconciliation loop, even if all cluster end-entity certificates have been issued, Strimzi will patch the certificate Secrets with the correct `strimzi.io/server-cert-hash` annotation.
+The value of this annotation can then be compared with the value on the pods to determine whether the pods need to be restarted to pick up a new Secret.
```
scholzj, line 111: Not sure I follow the need to annotate the Secrets here. Normally, during the reconciliation:
* You take the hash of the certiicate
* Use the Hash to annotate the Pod in the Deployment / StrimziPodSet
* Either Kubernetes or Strimzi takes care of rolling the pod based on the Pod annotations being different

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+
+This feature will be optional and not disabled by default.
+
+### Migrating to this feature
+
+To start using this feature in an existing Kafka cluster the user must:
+1. Install cert-manager and create an `Issuer`.
+2. Pause reconciliation for their Kafka cluster.
+3. Update the `<CLUSTER_NAME>-cluster-ca-cert` and/or `<CLUSTER_NAME>-clients-ca-cert` Secrets to:
+   1. contain the CA(s) for the `Issuer` (keeping the old CA cert in the Secret still)
```
scholzj, line 129: I think this part is really important - `(keeping the old CA cert in the Secret still)` - and should be made more prominent? We should also cover the point when it should be removed.

scholzj (commented):
I left some nits. But the main thing missing seems to be how you plan to roll out the trust to the new CAs. There is no reference to `strimzi.io/ca-key-generation` or to rotation of private keys, so it is not clear how do you expect to handle it.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+spec:
+  clusterCa:
+    validityDays: <integer> # notBefore=now, notAfter=now + validityDays
+    generateCertificateAuthority: <boolean>
+    generateSecretOwnerReference: <boolean>
+    renewalDays: <integer> # days before notAfter when we should start renewal
+    certificateExpirationPolicy: <renew-certificate|replace-key>
+    certificateIssuer:
+      type: <internal|cert-manager.io> # (1)
+      issuerRef: # (2)
```
MichaelMorrisEst, line 66: The issuerRef element looks to be cert-manager specific rather than something relevant to any external cert manager which could become an issue when supporting other external managers (as mentioned above).
Perhaps the certificateIssuer should instead have a certManager specific sub-element, and add different similar elements in the future that are specific to other certificate managers, i.e. something like:
```yaml
clusterCa:
    certificateIssuer:
        certManager:
            issuerRef:
                name: <string>
                kind: <Issuer|ClusterIssuer>
                group: <string> # cert-manager.io by default
        someOtherManager: <-- future addition -->
            managerSpecificConfig:
                ...
        oneOf:
        - properties
            certManager{}
            someOtherManager{}
```
Or alternatively just allow a map of values to be specified, but that would be less user friendly

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+
+For user certificates (issued by the User Operator), the user will be responsible for making sure their applications notice cert-manager renewing the certificates and are updated to use the new certificate.
+
+## Affected/not affected projects
+
+This affects the Cluster Operator and User Operator.
+
+## Compatibility
+
+This feature will be optional and not disabled by default.
```
PaulRMellor, line 121:
```suggestion
This feature will be optional and enabled by default.
```

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Migrating to this feature
+
+To start using this feature in an existing Kafka cluster the user must:
```
PaulRMellor, line 125: What about a new cluster?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+When using this feature for the ClusterCa:
+* On the next reconciliation Strimzi will first roll the pods once to trust the new CA cert.
+* Strimzi will create `Certificate` resources for all the components and wait for the certificates to be issued.
+* Strimzi will roll the pods to use the new certificates.
+
+Once all the pods have been rolled the user can update the `<CLUSTER_NAME>-cluster-ca-cert` and/or `<CLUSTER_NAME>-clients-ca-cert` Secrets to remove the old CA cert.
+
+### Stopping using this feature
+
+To revert to user managed CAs the user will:
```
PaulRMellor, line 147:
```suggestion
To revert to user managed CAs the user must:
```

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+When using this feature for the ClusterCa:
+* On the next reconciliation Strimzi will first roll the pods once to trust the new CA cert.
+* Strimzi will issue new certificates for all the components.
+* Strimzi will roll the pods to use the new certificates.
+
+Once all the pods have been rolled the user can update the `<CLUSTER_NAME>-cluster-ca-cert` and/or `<CLUSTER_NAME>-clients-ca-cert` Secrets to remove the old CA cert.
+The user is responsible for removing the old `Certificate` resources and uninstalling cert-manager.
+
+Notes:
+* Today we do not document how to go from using user managed CAs to Strimzi managed CAs.
```
PaulRMellor, line 169: Maybe we should?

PaulRMellor (commented):
Looks good. I made a  few suggested changes to wording for clarity and readability. There are also a couple of questions for further clarification.

087-external-certificate-manager.md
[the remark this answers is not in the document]
Frawless: I think we just need to mention in in the docs properly as users could be confused from different projects where integration of CM works without creating any Issuer (afaiu operator creates self-sign Issuer when it is not created by users).

087-external-certificate-manager.md
katheris: @scholzj Do you mean like this?

```
spec:
  clusterCa:
    validityDays: <integer> # notBefore=now, notAfter=now + validityDays
    generateCertificateAuthority: <boolean>
    generateSecretOwnerReference: <boolean>
    renewalDays: <integer> # days before notAfter when we should start renewal
    certificateExpirationPolicy: <renew-certificate|replace-key>
    certificateIssuerType: <internal|cert-manager.io> # (1)
    certManagerIssuerRef: # (2)
      name: <string>
      kind: <Issuer|ClusterIssuer>
      group: <string> # cert-manager.io by default
```

087-external-certificate-manager.md
scholzj: I'm not sure it is even supported to go from custom CA to Strimzi CA. At least I do not remember anyone trying it / testing it. In the long term, I think the idea would be also to get rid of the internal CA management - so not sure this would be a issue from my point of view.

katheris:
@scholzj @tinaselenge @fvaleri @PaulRMellor @ppatierno @Frawless Thanks for all your comments. I've pushed an update to the proposal, the main changes apart from wording tweaks are:
* Update to copy certificates from cert-manager Secrets, not mount them.
* Handle trust rollout more directly in the Strimzi operator.
* Add diagrams.

I've left the User operator/clients CA part as TODO at the moment but I would appreciate any feedback on the cluster CA part of the proposal. I also do have a full diagram of a key replacement, but am still working on making it display in a way that's viewable. Let me know if that would be useful, or if the existing diagrams are clear enough.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+## Compatibility
+
+This feature will be optional and disabled by default.
+
+### Migrating to this feature
+
+To start using this feature in an existing Kafka cluster the user must:
+1. Install cert-manager and create an `Issuer`.
+2. Create Secrets to store the Cluster CA and/or Clients CA public certs
+3. Update the `Kafka` resource to have `clusterCa.type` and/or `clientsCa.type` set to `cert-manager.io`, and `certManagerIssuerRef` and `publicCert` configured.
```
fvaleri, line 173: They should also set `generateCertificateAuthority: false` as this is set to true by default, or we can simply ignore (see my previous comment).

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+4. Resume reconciliation.
```
fvaleri, line 174: Is pausing the reconciliation a missing step in this list?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+* On the next User Operator reconciliation Strimzi will create `Certificate` resources for all the existing `KafkaUser` custom resources.
+
+When using this feature for the ClusterCa:
+* On the next Cluster operator reconciliation Strimzi will copy the new certificate over to `<CLUSTER_NAME>-cluster-ca-cert` (renaming and keeping the old certificate), increment the cert generation annotation and add the key generation annotation.
+* Strimzi will roll the pods once to trust the new CA cert.
+* Strimzi will create `Certificate` resources for all the components and wait for the certificates to be issued.
+* Strimzi will copy over the new certificates into the pod Secrets.
+* Strimzi will roll the pods to use the new certificates.
+
+Once all the pods have the correct cert generation annotation Strimzi can update the `<CLUSTER_NAME>-cluster-ca-cert` and/or `<CLUSTER_NAME>-clients-ca-cert` Secrets to remove the old CA cert and delete the `<CLUSTER_NAME>-cluster-ca` and/or `<CLUSTER_NAME>-clients-ca` Secrets.
```
fvaleri, line 188: Shouldn't we roll the pods again here to "untrust" the old CA cert?

fvaleri (commented):
Thanks for the updates Kate.

The plus of this solution is to build up on the existing rollout logic, which is proven and well tested.

New diagrams are great. I see why you didn't put them inline, but maybe we can link them at the end of related sections.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+
+For both categories, to provide a secure, TLS-enabled setup by default when deploying Kafka clusters, Strimzi integrated its own CA operations into the Cluster Operator.
+The Cluster Operator accomplishes this by using openssl to generate self-signed root CA certificates and private keys which it then uses to directly sign end-entity (EE) certificates.
+This CA certificate has zero pathlen, which means it cannot sign any intermediate CA.
+A cluster CA and clients CA are generated.
+These CAs are only used for Kafka clusters and a unique instance of each CA is used for each Kafka cluster.
+
+In addition to Strimzi fully managing the certificates as described above, there are options for users to partially manage the certificates:
+* Users can [install and use their own CA certificate and private keys](https://strimzi.io/docs/operators/latest/deploying#installing-your-own-ca-certificates-str), instead of using the defaults generated by the Cluster Operator.
+  When using this option, both the CA certificate and private key must be provided, and Strimzi still issues the end-entity (EE) certificates that are presented by the components.
```
scholzj, line 24: Is it worth mentioning here that for the Clients CA, users can also:
* provide the CA public key
* provide a dummy value for the CA private key
* Issue their own user certificates out-of-band
* Use the `type: tls-external` users when they decide to use the KafkaUser resources and User Operator to manage ACLs and Qoutas

I think this pattern is used enough to make it worth mentioning here for clarity.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+* Users can [provide custom listener certificates](https://strimzi.io/docs/operators/latest/deploying#proc-installing-certs-per-listener-str) for TLS encryption.
+  This option only affects how user applications connect to Kafka.
+  It does not change how the Strimzi components connect to Kafka, or how the Kafka brokers connect to each other.
+
+None of the existing options allow all end-entity certificates to be issued by a tool that isn't Strimzi.
```
scholzj, line 29: As mentioned in the comment above, this is not entirely true for the Clients CA where there are plenty users issuing their own user certificates.

For Cluster CA, users can in theory pre-create the certificate Secrets. But it is not documented and getting the SANs of the EE certificates right is challenging. I know some users _tried this_, but I'm not aware of anyone actually doing this in real environments.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+      secretName: <string>
+      certificate: <string>
+```
+
+1. The`type` property will default to `internal` when not set and will use the existing behaviour, allowing backwards compatibility.
+   The option `cert-manager.io` will only be valid if `generateCertificateAuthority` is set to `false`.
+2. The property `certManagerIssuerRef` will only be used by Strimzi if `type` is set to `cert-manager.io`.
+  The `name`, `kind`, and `group` properties will be copied over into the `Certificate` custom resource Strimzi creates.
+3. The property `publicCert` will only be used by Strimzi if `type` is set to `cert-manager.io`.
+  The `secretName` and `certificate` properties will be used to locate the CA public certificate that must be trusted by Strimzi components in order to trust the end-entity certificates that cert-manager issues.
```
scholzj, line 80: I guess the way this will be implemented is that we will have different classes with different options (as we do for example for authorization, authentication ... and most of the other _typed_ APIs). So each type will have its own fields.

So maybe instead of a long list of fields that can be used one or the other type, you can simply provide two YAMLs for the two different types we will support today?

Alternative is to do what we have for listeners where we have a single class and the validation needs to be done in the operator code. But the main reason for that is that the listeners are defined in an array which is not the case for the CAs. So not sure there would be any benefit of doing that.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+    renewalDays: <integer> # days before notAfter when we should start renewal
+    certificateExpirationPolicy: <renew-certificate|replace-key>
+    type: <internal|cert-manager.io> # (1)
+    certManagerIssuerRef: # (2)
+      name: <string>
+      kind: <Issuer|ClusterIssuer>
+      group: <string> # cert-manager.io by default
+    publicCert: # (3)
+      secretName: <string>
+      certificate: <string>
```
scholzj, line 72: Should this support a list with multiple Secrets? Should it be ready to handle other forms of certificates such as ConfigMap? E.g.:

```yaml
    publicCert: # (3)
      - fromSecret:
          secretName: <string>
          certificate: <string>
```

This would:
* Allow to possibly load trusted certificates from more Secrets - would that be useful when migrating to a new CA/Issuer or something like that?
* Allow to extend the type in the future to something else than Secret (we can start with Secret only today, but we will have the option)

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+
+When a new Kafka cluster is created Strimzi will copy the CA public cert from the Secret identified in `clusterCa.publicCert` and `clientsCa.publicCert` to the `<CLUSTER_NAME>-cluster-ca-cert` and `<CLUSTER_NAME>-clients-ca-cert` Secrets respectively.
+Strimzi will add three annotations to the `<CLUSTER_NAME>-cluster-ca-cert` Secret:
+* `strimzi.io/ca-cert-hash` with the value being a hash of the certificate.
+* `strimzi.io/cluster-ca-key-generation` initially set to 0.
+* `strimzi.io/cluster-ca-cert-generation` initially set to 0.
+
+During a reconciliation Strimzi will check the hash of the certificate stored in the user's CA public cert Secret to see if an update is needed.
+If the certificate has changed Strimzi will rename the existing cert in the `<CLUSTER_NAME>-cluster-ca-cert` Secret to `ca-YYYY-MM-DDTHH-MM-SSZ.crt` and copy over the new certificate.
+It will also update the `strimzi.io/ca-cert-hash` and increment the `strimzi.io/cluster-ca-key-generation` and `strimzi.io/cluster-ca-cert-generation` annotations.
```
scholzj, line 113: Will the hash be based on the source Secret value or on the certificate itself? I.e. would for example some newline change trigger this? Or will it be safe against it?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### Issuing end-entity certificates
+
+If the user has enabled this feature, when Strimzi needs to issue a certificate, instead of using the existing internal mechanism it will create a `Certificate` custom resource.
```
scholzj, line 117: What does `If the user has enabled this feature` mean here? Isn't it always enabled when the `type: cert-manager.io` is used?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### Tracking changes to cluster end-entity certificates
+
+Cert-manager will be responsible for renewing all end-entity certificates.
+When a certificate is renewed cert-manager will update the related Secret.
+During a reconciliation Strimzi will check the hash of the certificate stored in cert-manager Secrets to see if an update is needed.
+
+When cert-manager renews an end-entity certificate Strimzi will copy over the new certificate into its own Secret.
+Strimzi will update the `strimzi.io/server-cert-hash` annotation to match the new certificate.
+Strimzi will update the `strimzi.io/cluster-ca-cert-generation` annotation to match the annotation on the `<CLUSTER_NAME>-cluster-ca-cert` Secret.
```
scholzj, line 136: This describes how do we detect that CM changed the certificate. Can you also describe how will Strimzi detect when it needs to change the certificate for some other reason? E.g. to add a new SAN to it?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### Handling CA key replacements
+
+During a CA key replacement the user must pause Strimzi reconciliation until the following are complete:
+* A new CA key has been generated
+* A new CA public certificate has been issued
+* The Secret provided to Strimzi containing the CA public certificate has been updated
+* Cert-manager has issued new end-entity certificates using the new CA public certificate
```
scholzj, line 144: Can you elaborate on why this is needed? And maybe also on what does happen here on the Cert Manager side? I assume you describe here a new CA used for existing issuer? 

A particularly interesting points:
* How will the user know that `Cert-manager has issued new end-entity certificates using the new CA public certificate`?
* Will the user include only the new CA in the trusted CAs? Or both the new and old and have the user delete the old CA later? I assume the second is correct, or?

----

Also, what about migration to a different Issuer? Should we have a chapter for that  as well?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+2. Roll the Kafka pods to trust the new CA cert, incrementing the key generation on the pods.
+3. Copy over the new Kafka certificates, updating the cert generation annotations on the Secrets.
+4. Roll the Kafka pods to use the new Kafka certificates, updating the cert generation annotations on the Pods.
+5. Then on the next reconciliation, since the Kafka pods now have correct cert and key generation, copy over the new operator certificate.
+6. Since the Kafka pods now have correct cert generation, remove the old CA public cert from the `<CLUSTER_NAME>-cluster-ca-cert` Secret
+
+I have created diagrams for the existing flow and the flow when using cert-manager for [certificate renewals](./images/087-cert-renewals.png) and the flow for when cert-manager issues [new end-entity certificates](./images/087-new-ee-certs.png).
+
+### User Operator/Clients CA
+TODO
```
scholzj, line 158: This would be an interesting part 🤔

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+To start using this feature in an existing Kafka cluster the user must:
+1. Install cert-manager and create an `Issuer`.
+2. Create Secrets to store the Cluster CA and/or Clients CA public certs
+3. Update the `Kafka` resource to have `clusterCa.type` and/or `clientsCa.type` set to `cert-manager.io`, and `certManagerIssuerRef` and `publicCert` configured.
+4. Resume reconciliation.
+
+When using this feature for the ClientsCa:
+* On the next Cluster operator reconciliation Strimzi will copy the new certificate over to `<CLUSTER_NAME>-clients-ca-cert` (renaming and keeping the old certificate), increment the cert generation annotation and add the key generation annotation.
+* Strimzi will roll the Kafka pods to trust the new CA cert.
+* On the next User Operator reconciliation Strimzi will create `Certificate` resources for all the existing `KafkaUser` custom resources.
```
scholzj, line 179: Will we expect the users to use the original CM secrets? Or will the UO copy the certs to its own secrets? (Probably more question for the TODO part about UO - but it might impact what happens to the old user secrets here)

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+* On the next reconciliation Strimzi will first roll the pods once to trust the new CA cert.
+* Strimzi will issue new certificates for all the components.
+* Strimzi will roll the pods to use the new certificates.
+
+Once all the pods have been rolled the user can update the `<CLUSTER_NAME>-cluster-ca-cert` and/or `<CLUSTER_NAME>-clients-ca-cert` Secrets to remove the old CA cert.
+The user is responsible for removing the old `Certificate` resources and uninstalling cert-manager.
+
+Notes:
+* Today we do not document how to go from using user managed CAs to Strimzi managed CAs.
+  For this reason I have not included how to go from cert-manager CAs to Strimzi managed CAs.
```
scholzj, line 215: Is the note irrelevant because you included it now?

scholzj (commented):
Thanks for the updates. I left some more comments, mainly about the formal side and some possibly missing scenarios being described.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+      name: <string>
+      kind: <Issuer|ClusterIssuer>
+      group: <string> # cert-manager.io by default
+    publicCert: # (3)
+      secretName: <string>
+      certificate: <string>
+```
+
+1. The`type` property will default to `internal` when not set and will use the existing behaviour, allowing backwards compatibility.
+   The option `cert-manager.io` will only be valid if `generateCertificateAuthority` is set to `false`.
```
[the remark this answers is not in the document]
katheris: Looking forwards, I think it's possible we could integrate with a certificate manager where Strimzi is the one that triggers the CA be generated so I think ignoring `.spec.clusterCa.generateCertificateAuthority` when `.spec.clusterCa.type` is not `internal` makes too big an assumption that any other `type` values added in future work the same as cert-manager. Also I think it would be confusing for the user of they can have `.spec.clusterCa.type` set to `cert-manager.io` and `.spec.clusterCa.generateCertificateAuthority` set to true. Wouldn't it be clearer to reject a configuration that doesn't make sense, rather than silently ignoring fields?

087-external-certificate-manager.md
katheris: @scholzj do you mean have two separate yamls in the proposal to make it clearer? Or restructuring the yaml to put the fields for the specific type under that type somehow?

087-external-certificate-manager.md
katheris: We would use the same mechanism as we have today, which is to extract the cert as an x509 certificate, then calculate the hash, so it will be safe against any extra whitespace in the source Secret value.

087-external-certificate-manager.md
katheris: Yeah I just meant when they have `type` set to `cert-manager.io`, I'll remove that part of the sentence as I don't think it adds anything

087-external-certificate-manager.md
katheris: The flow would actually work without the last bullet point so I'll remove that

I'll add some more details on the Cert Manager side and look at what happens when migrating to a different Issuer.

On this point:
> Will the user include only the new CA in the trusted CAs? Or both the new and old and have the user delete the old CA later? I assume the second is correct, or?

Since Strimzi is managing copying the CA cert over into the <CLUSTER_NAME>-cluster-ca-cert Secret the user just needs to provide the new one. Strimzi will hold onto the old one until it has seen new end-entity certificates have been issued and then remove it as described in the section that steps through what Strimzi does during the next reconciliation.

087-external-certificate-manager.md
katheris: I've included how to go from cert-manager CAs to user-managed CAs, but not how to go to Strimzi managed CAs.

087-external-certificate-manager.md
ppatierno: > For Cluster CA, users can in theory pre-create the certificate Secrets. But it is not documented and getting the SANs of the EE certificates right is challenging. I know some users tried this, but I'm not aware of anyone actually doing this in real environments.

For all these reasons I would not mentioned it here. I guess that referring to end-entity certificates, Kate is talking about the internal cluster components and not clients so the clients CA is out of the picture. Maybe we can clarify it here.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+
+The proposal makes a few assumptions:
+* Strimzi will not be responsible for installing cert-manager, but we will document the supported versions of cert-manager that we have tested with.
+* Strimzi will not be responsible for creating `Issuer` or `ClusterIssuer` custom resources.
+* Strimzi will not be responsible for requesting a CA certificate be issued, it will only interact with cert-manager to request end-entity certificates are issued.
+* Strimzi will create `Certificate` custom resources (to request a new end-entity certificate is issued) and will allow the user to influence the contents of these resources by exposing options in the `Kafka` custom resource.
+* Strimzi will not directly interact with the lower level `CertificateRequest` and `CertificateSigningRequests` custom resources.
+* When Strimzi creates a `Certificate` custom resource, cert-manager will issue the certificate within a reasonable amount of time such that Strimzi can wait during the reconciliation.
```
ppatierno, line 50: I was also wondering, what happens if the cert-manager is down or is not working properly. The Strimzi operator cannot wait forever so the reconciliation would fail, right? Then because the `Certificate` was already created, the operator will take care of it in the next reconciliation still waiting for cert-manager until a timeout again?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+### Internal handling of certificates
+
+Strimzi uses two generation annotations to track how Cluster CA certificates are trusted and used:
+* `strimzi.io/cluster-ca-key-generation` on Kafka pods to indicate the CA generation trusted by that pod
+* `strimzi.io/cluster-ca-cert-generation` on Secrets containing certs to indicate the CA generation that signed those certs
+
+The `strimzi.io/cluster-ca-cert-generation` annotation is also used on Kafka pods to indicate the CA generation that signed the certs it is currently presenting (i.e. the CA generation currently in use).
+
+When using cert-manager for issuing certificates Strimzi will continue to use these generations in the same way.
+Rather than directly mounting cert-manager Secrets and the user-provided CA cert Secret, Strimzi will copy the certificates into its own existing Secrets.
```
ppatierno, line 101: I would remove this sentence. Before reading the next paragraph where everything is well explained, I wrote the following questions: 

> "its own existing Secrets." ... are you referring to the xxxxx-cluster-ca-cert Secret?
Will the operator have a logic which at each reconciliation (periodically) watches the cert-manager Secret and update to its own? Or how is it going to work? Also, any reasons why not mounting the cert-manager Secret? For backward compatibility? (sorry too many questions).

Because the above sentence was not clear to me. To avoid confusion I would just remove it and leaving to the next paragraph the deep explanation of how things are going to work.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### Handling Cluster CA trust rollout
+
+When a new Kafka cluster is created Strimzi will copy the CA public cert from the Secret identified in `clusterCa.publicCert` and `clientsCa.publicCert` to the `<CLUSTER_NAME>-cluster-ca-cert` and `<CLUSTER_NAME>-clients-ca-cert` Secrets respectively.
+Strimzi will add three annotations to the `<CLUSTER_NAME>-cluster-ca-cert` Secret:
+* `strimzi.io/ca-cert-hash` with the value being a hash of the certificate.
+* `strimzi.io/cluster-ca-key-generation` initially set to 0.
+* `strimzi.io/cluster-ca-cert-generation` initially set to 0.
+
+During a reconciliation Strimzi will check the hash of the certificate stored in the user's CA public cert Secret to see if an update is needed.
```
ppatierno, line 111: is the user's CA public cert Secret what you called before the cert-manager Secret?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+If the certificate has changed Strimzi will rename the existing cert in the `<CLUSTER_NAME>-cluster-ca-cert` Secret to `ca-YYYY-MM-DDTHH-MM-SSZ.crt` and copy over the new certificate.
+It will also update the `strimzi.io/ca-cert-hash` and increment the `strimzi.io/cluster-ca-key-generation` and `strimzi.io/cluster-ca-cert-generation` annotations.
+
+#### Issuing end-entity certificates
+
+If the user has enabled this feature, when Strimzi needs to issue a certificate, instead of using the existing internal mechanism it will create a `Certificate` custom resource.
+Strimzi will specify the required CN/SANS in the `Certificate` resource for the end-entity certificate.
+Strimzi will specify the Secret for the certificate to be stored in as the existing name for the certificate Secret suffixed with `-cm`, for example `<CLUSTER_NAME>-cluster-operator-certs-cm`.
```
ppatierno, line 119: I had hard time to understand what this sentence means sorry. Maybe just a language barrier on my side :-(
Can you elaborate here or make it clearer to me please?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,241 @@
...
+* Strimzi will roll the pods to use the new certificates.
+
+Once all the pods have the correct cert generation annotation Strimzi can update the `<CLUSTER_NAME>-cluster-ca-cert` and/or `<CLUSTER_NAME>-clients-ca-cert` Secrets to remove the old CA cert and delete the `<CLUSTER_NAME>-cluster-ca` and/or `<CLUSTER_NAME>-clients-ca` Secrets.
+
+### Stopping using this feature
+
+To revert to user managed CAs the user must:
+1. Pause reconciliation for their Kafka cluster.
+2. Update the `<CLUSTER_NAME>-cluster-ca-cert` and/or `<CLUSTER_NAME>-clients-ca-cert` Secrets to:
+   1. contain their public CA cert (keeping the old cert-manager one)
```
ppatierno, line 195: what does it exactly mean? Are we moving to use Strimzi to generate the CAs right? What's "their public CA cert"?

087-external-certificate-manager.md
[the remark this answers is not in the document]
fvaleri: Agreed, rejecting and invalid configuration is better.

MichaelMorrisEst:
> @MichaelMorrisEst thanks for your comments. On the extensibility of the design, I've tried to write both the CRD and the way it's implemented such that we could add other certificate management options in future. However, I wasn't planning for it to be something a user can add at deploy time. My expectation is that we would add it directly to the codebase. The reason being that it is easier to reason about what is supported and make sure it's properly tested. Is that what you were expecting when you were asking about alternative implementations?

As an open source project, it will only ever be possible to add open source certificate management options directly in the Strimzi code base. I think it would be very useful to enable other options to be integrated without code changes. 
The Strimzi project would only be responsible for testing the mechanism works, it would be the responsibility of users integrating with other certificate manager tools to ensure their integration can and does fulfil the requirements of the Strimzi API

087-external-certificate-manager.md
scholzj: I do not think it is clear here it is only about the Cluster CA.

087-external-certificate-manager.md
scholzj: I guess if you wanna keep all the options as you suggest in https://github.com/strimzi/proposals/pull/135#discussion_r1926764306 the example is probably fine and you can ignore this.

scholzj:
> > @MichaelMorrisEst thanks for your comments. On the extensibility of the design, I've tried to write both the CRD and the way it's implemented such that we could add other certificate management options in future. However, I wasn't planning for it to be something a user can add at deploy time. My expectation is that we would add it directly to the codebase. The reason being that it is easier to reason about what is supported and make sure it's properly tested. Is that what you were expecting when you were asking about alternative implementations?
> 
> As an open source project, it will only ever be possible to add open source certificate management options directly in the Strimzi code base. I think it would be very useful to enable other options to be integrated without code changes. The Strimzi project would only be responsible for testing the mechanism works, it would be the responsibility of users integrating with other certificate manager tools to ensure their integration can and does fulfil the requirements of the Strimzi API

@MichaelMorrisEst I think open-source project is one aspect to consider. The other one is that it needs to have enough demand to justify the effort and maintenance. So, I think the extensibility makes sense.

The way that could be done would be to add some additional type `custom`  where you specify the class. I think the main challenge is to create a (sufficiently) stable interface that can be used to implement the plugins. I wonder if that is easier to design now or whether implementation of this proposal would help to define such interface as we will have two implementations to compare and base the pluggable interface on.

I guess there are some _dirty_ alternatives ...
* Integrate the other tools through Cert Manager itself which has its own pluggable interface as far as I understood. That might give a more versatile solution that is usable outside of Strimzi but obviously it would require to manage Cert Manager itself.
* Hook a custom operator on the Cert Manager custom resources and simply provide your certificates based on the Cert Manager `Certificate` resource through a custom tool.

These seem a bit more hacky. But I think they would provide a stable interface.

The whole problem moves to entirely different level when you would need to change the certificate handling completely including for example not storing them in Kubernetes Secrets (but loading them from something like Vault). As that would require the plugabiity not only in the operator code but also Kafka (Config Providers? Shell scripts?), Cruise Control (Shell scripts) or Kafka Exporter (Shell scripts) and so on. I'm not sure we would ever be able to get there - but who knows?

MichaelMorrisEst:
> > > @MichaelMorrisEst thanks for your comments. On the extensibility of the design, I've tried to write both the CRD and the way it's implemented such that we could add other certificate management options in future. However, I wasn't planning for it to be something a user can add at deploy time. My expectation is that we would add it directly to the codebase. The reason being that it is easier to reason about what is supported and make sure it's properly tested. Is that what you were expecting when you were asking about alternative implementations?
> > 
> > 
> > As an open source project, it will only ever be possible to add open source certificate management options directly in the Strimzi code base. I think it would be very useful to enable other options to be integrated without code changes. The Strimzi project would only be responsible for testing the mechanism works, it would be the responsibility of users integrating with other certificate manager tools to ensure their integration can and does fulfil the requirements of the Strimzi API
> 
> @MichaelMorrisEst I think open-source project is one aspect to consider. The other one is that it needs to have enough demand to justify the effort and maintenance. So, I think the extensibility makes sense.
> 
> The way that could be done would be to add some additional type `custom` where you specify the class. I think the main challenge is to create a (sufficiently) stable interface that can be used to implement the plugins. I wonder if that is easier to design now or whether implementation of this proposal would help to define such interface as we will have two implementations to compare and base the pluggable interface on.
> 
> I guess there are some _dirty_ alternatives ...
> 
> * Integrate the other tools through Cert Manager itself which has its own pluggable interface as far as I understood. That might give a more versatile solution that is usable outside of Strimzi but obviously it would require to manage Cert Manager itself.
> * Hook a custom operator on the Cert Manager custom resources and simply provide your certificates based on the Cert Manager `Certificate` resource through a custom tool.
> 
> These seem a bit more hacky. But I think they would provide a stable interface.
> 
> The whole problem moves to entirely different level when you would need to change the certificate handling completely including for example not storing them in Kubernetes Secrets (but loading them from something like Vault). As that would require the plugabiity not only in the operator code but also Kafka (Config Providers? Shell scripts?), Cruise Control (Shell scripts) or Kafka Exporter (Shell scripts) and so on. I'm not sure we would ever be able to get there - but who knows?

@scholzj 
I think implementing for the two implementations should give a good starting point for the interface and something that could be used as a basis to experiment with other implementations. Perhaps if the interface is created as part of this proposal it could be considered preliminary (behind a feature gate maybe?) and improved as necessary until we reach a point of stability.

Of the other suggestions the dependency on Cert Manager for the first suggestion limits its applicability, the second (custom operator for Cert Manager custom resources) looks more workable but it does seem a bit hacky

087-external-certificate-manager.md
katheris: Here I meant all certificates, so those associated with both the Cluster CA and Clients CA. I realise that theoretically it is possible by pre-creating the certificate Secrets, but as you say this is challenging. Is this sentence needed? Or is the motivation below sufficient so that I could remove the sentence here?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,196 @@
...
+
+Strimzi will be updated to allow users to specify that certificates should be issued by an external certificate manager, rather than issued by the Cluster Operator.
+This proposal will specifically describe how this would work for cert-manager, however the user API for configuration will be written in a way that does not prevent other external certificate managers being added in the future.
+
+The proposal makes a few assumptions:
+* Strimzi will not be responsible for installing cert-manager, but we will document the versions of cert-manager that we have tested with.
+* Strimzi will not be responsible for creating `Issuer` or `ClusterIssuer` custom resources.
+* Strimzi will create `Certificate` custom resources and will allow the user to influence the contents of these resources by exposing options in the `Kafka` custom resource.
+* Strimzi will not directly interact with the lower level `CertificateRequest` and `CertificateSigningRequests` custom resources.
+* When Strimzi creates a `Certificate` custom resource, cert-manager will issue the certificate within a reasonable amount of time such that Strimzi can wait during the reconciliation.
```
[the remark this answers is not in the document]
katheris: Here I'm talking about issuing end-entity certificates, as I described higher up.

I talk about time-outs and retries further down when going through the flow in more detail

087-external-certificate-manager.md
katheris: My current proposal doesn't allow for more than one certificate here. The proposed flow requires us to associate a generation id with one certificate. This field is for the latest trusted certificate, Strimzi will keep the previous certificate until it is sure everything trusts the new one. So I don't think we need to support more than one here? But could add it incase this changes in future, wdyt?

087-external-certificate-manager.md
katheris: This is covered in the section `User steps` earlier in the proposal

087-external-certificate-manager.md
katheris: This is moving to user managed CAs. So where the user provides the CA public and private keys

087-external-certificate-manager.md
katheris: Marking as resolved now I've added the UO section, feel free to reopen or add a new comment if there are still unanswered questions

katheris:
@ppatierno @scholzj I've pushed some updates and added the Clients CA/User Operator section

087-external-certificate-manager.md
katheris: I've rewritten this sentence, see if it makes more sense now

087-external-certificate-manager.md
katheris: No, they are two different things. I've rewritten so hopefully it makes more sense now

087-external-certificate-manager.md
ppatierno: Can we clarify it with just one additional sentence?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,294 @@
...
+1. Copy over the new CA public cert (keeping the old one as described previously) and update the hash and generation annotations.
+2. Roll the Kafka pods to trust the new CA cert, incrementing the `strimzi.io/cluster-ca-key-generation` annotation on the pods.
+3. Copy over the new Kafka certificates, updating the `strimzi.io/cluster-ca-cert-generation` annotations on the Secrets.
+4. Roll the Kafka pods to use the new Kafka certificates, updating the `strimzi.io/cluster-ca-cert-generation` annotations on the Pods.
+5. Then on the next reconciliation, since the Kafka pods now have correct cert and key generation, copy over the new operator certificate.
+6. Since the Kafka pods now have correct cert generation, remove the old CA public cert from the `<CLUSTER_NAME>-cluster-ca-cert` Secret
+
+> ![Renewing the cluster CA public cert](./images/087-cert-renewals.png)
+>
+> Fig 2: Existing and proposed workflow when the user provides a new cluster CA public cert
```
scholzj, line 162: Can you cover at which stage will this be done? Will that be done somewhere at the beginning of today's CaReconciler (and thus the first reocnciliation after unpausing will immediately start taking the steps to replace the CA)? Or soemwhere later in the reconciliation and essentially just _prepare things_ for the next reconciliation?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,294 @@
...
+Strimzi will update the `strimzi.io/server-cert-hash` annotation to match the new certificate.
+Strimzi will update the `strimzi.io/cluster-ca-cert-generation` annotation to match the `strimzi.io/ca-cert-generation` annotation on the `<CLUSTER_NAME>-cluster-ca-cert` Secret.
+
+Strimzi will review the `Certificate` resource every time it does a reconciliation to see if any changes to the requested certificate are needed, for example updating SANs.
+
+> ![Issuing new component end-entity certificates](./images/087-new-ee-certs.png)
+>
+> Fig 1: Proposed workflow when cert-manager issues new component end-entity certificates
+
+#### Handling CA key replacements
```
scholzj, line 144: Is there any difference between CA key replacement here and simple CA public key renewal? Given the later seems not to be covered in any other section, I assume all public key renewals are handled in the same way as key replacements (as we do with custom CAs today)? Assuming that is the case, it would be worth clarifying it here.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,294 @@
...
+
+#### Issuing KafkaUser certificates
+
+When a `KafkaUser` with `spec.authentication.type` set to `tls` is creates Strimzi  will create a `Certificate` custom resource.
+Strimzi will specify the required CN/SANS in the `Certificate` resource for the user certificate.
+Strimzi will specify the Secret for the certificate to be stored in as the existing name for the certificate Secret suffixed with `-cm`, for example `<USER_NAME>-cm`.
+Strimzi will request the certificate in both PEM and PKCS12 format.
+Strimzi will wait for the usual operation timeout during the reconciliation loop for the `Certificate` status to indicate that the certificate has been issued before continuing.
+Once the certificate has been issued, Strimzi will copy the certificate across from the cert-manager provided Secret into its own existing Secret.
+Strimzi will also copy across the ClientsCA public certificate.
```
scholzj, line 179: Do we still want to do the copying here? Or do we want the user to get it from the CM or (whoever actually produces it) directly? For the server certificates and CAs, we depend on them so we want to copy them out. The user certificate secrets actually exist only for the users. We do not care about them much otherwise. So I wonder if keeping them twice makes sense?

But maybe we still need it to track the generations etc.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,294 @@
...
+During a reconciliation Strimzi will check the hash of the certificate stored in cert-manager Secrets to see if an update is needed.
+Strimzi will copy over the new certificate into its own Secret and update the `strimzi.io/user-cert-hash` annotation to match the new certificate.
+
+> ![Issuing new user certificate](./images/087-new-renew-user-cert.png)
+>
+> Fig 3: Proposed workflow when cert-manager issues a new Kafka User certificate
+
+#### Handling CA cert renewals and key replacements
+
+During a reconciliation Strimzi will check the hash of the certificate stored in the user's CA public cert Secret to see if an update is needed.
```
scholzj, line 198: today we IMHO do not pass much information abotu the Clients CA to the UO and the UO instead uses Kube APi to fetch it from the Secret. But we might need to change it here as the UO will alos need to know what issuer to use etc.?

scholzj (commented):
Few more nits. mostly nits apart from some things around the User OPerator.

087-external-certificate-manager.md
katheris: @scholzj are you happy for me to mark this as resolved?

087-external-certificate-manager.md
katheris: Marking this as resolved since the proposal has changed since this comment was added

087-external-certificate-manager.md
```diff
@@ -0,0 +1,299 @@
...
+
+#### Issuing KafkaUser certificates
+
+The User operator will be updated to have four environment variables to configure the CA issuer:
+- `STRIMZI_CA_TYPE` with either the value `strimzi.io` or `cert-manager.io`
+- `STRIMZI_CM_ISSUER_NAME`
+- `STRIMZI_CM_ISSUER_KIND`
+- `STRIMZI_CM_ISSUER_GROUP`
+
+These will be set by the Cluster operator when the User operator is deployed as part of the Entity operator.
```
scholzj, line 181:
```suggestion
These will be set by the Cluster operator when the User operator is deployed as part of the Entity operator and when Clients CA is set to `type: cert-manager.io`.
```

087-external-certificate-manager.md
```diff
@@ -0,0 +1,299 @@
...
+      secretName: <string>
+      certificate: <string>
+```
+
+1. The`type` property will default to `strimzi.io` when not set and will use the existing behaviour, allowing backwards compatibility.
+   The option `cert-manager.io` will only be valid if `generateCertificateAuthority` is set to `false`.
+2. The property `certManagerIssuerRef` will only be used by Strimzi if `type` is set to `cert-manager.io`.
+  The `name`, `kind`, and `group` properties will be copied over into the `Certificate` custom resource Strimzi creates.
+3. The property `publicCert` will only be used by Strimzi if `type` is set to `cert-manager.io`.
+  The `secretName` and `certificate` properties will be used to locate the CA public certificate that must be trusted by Strimzi components in order to trust the end-entity certificates that cert-manager issues.
```
fvaleri, line 82: I think we can leverage the new CEL validation feature to do these checks at the API level. Not sure you want to mention this implementation detail, but I wanted to leave a note.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,299 @@
...
+* `strimzi.io/ca-cert-generation` initially set to 0.
+
+During a reconciliation Strimzi will check the hash of the certificate stored in `clusterCa.publicCert` Secret to see if an update is needed.
+If the certificate has changed Strimzi will rename the existing cert in the `<CLUSTER_NAME>-cluster-ca-cert` Secret to `ca-YYYY-MM-DDTHH-MM-SSZ.crt` and copy over the new certificate.
+It will also update the `strimzi.io/ca-cert-hash` and increment the `strimzi.io/ca-key-generation` and `strimzi.io/ca-cert-generation` annotations.
+
+#### Issuing end-entity certificates
+
+When Strimzi needs to issue a certificate, instead of using the existing internal mechanism it will create a `Certificate` custom resource.
+Strimzi will specify the required CN/SANS in the `Certificate` resource for the end-entity certificate.
```
fvaleri, line 120:
```suggestion
Strimzi will specify the required CN/SANs in the `Certificate` resource for the end-entity certificate.
```

087-external-certificate-manager.md
```diff
@@ -0,0 +1,299 @@
...
+
+The User operator will be updated to have four environment variables to configure the CA issuer:
+- `STRIMZI_CA_TYPE` with either the value `strimzi.io` or `cert-manager.io`
+- `STRIMZI_CM_ISSUER_NAME`
+- `STRIMZI_CM_ISSUER_KIND`
+- `STRIMZI_CM_ISSUER_GROUP`
+
+These will be set by the Cluster operator when the User operator is deployed as part of the Entity operator.
+
+When a `KafkaUser` with `spec.authentication.type` set to `tls` is creates Strimzi  will create a `Certificate` custom resource.
```
fvaleri, line 183:
```suggestion
When a `KafkaUser` with `spec.authentication.type` set to `tls` is created Strimzi will create a `Certificate` custom resource.
```

087-external-certificate-manager.md
```diff
@@ -0,0 +1,299 @@
...
+Strimzi will specify the required CN/SANS in the `Certificate` resource for the user certificate.
```
fvaleri, line 184:
```suggestion
Strimzi will specify the required CN/SANs in the `Certificate` resource for the user certificate.
```

fvaleri (approved):
LGTM. Thanks @katheris.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,299 @@
...
+Strimzi will request the certificate in both PEM and PKCS12 format.
+Strimzi will wait for the usual operation timeout during the reconciliation loop for the `Certificate` status to indicate that the certificate has been issued before continuing.
+
+#### Tracking changes to user certificates
+
+Cert-manager will be responsible for renewing all user certificates.
+When a certificate is renewed cert-manager will update the related Secret.
+Since the user's clients are directly using the cert-manager created Secret, Strimzi will take no action.
+
+#### Handling CA cert renewals and key replacements
```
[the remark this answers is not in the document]
ppatierno: I agree on this as well :-/

087-external-certificate-manager.md
```diff
@@ -0,0 +1,299 @@
...
+>
+> Fig 5: Proposed workflow when the user provides a new clients CA public cert
+
+## Affected/not affected projects
+
+This affects the Cluster Operator and User Operator.
+
+## Compatibility
+
+This feature will be optional and disabled by default.
```
ppatierno, line 217: Does it mean we are going to put this feature behind a feature gate? If yes, should we state it clearly and mentioning it?

087-external-certificate-manager.md
ppatierno: Ah right the section clearly is "To revert to user managed CAs the user must" :-D

087-external-certificate-manager.md
Frawless: From my understanding it won't be behind FG, but the default configuration will use Strimzi CA generation and users will need to explicitly configure cert-manager to use it.

ppatierno: Ok I mislead what "optional and disabled by default" means here, maybe we should highlight that if nothing is set, the "type" will be implicitly considered as the strimzi.io one?

Frawless (approved):
I am not security SME, but the feature proposal sounds good to me. +1

PaulRMellor (approved):
Thanks for addressing the comments, Kate.
I found the addition of the diagrams useful in showing the new flows.

087-external-certificate-manager.md
katheris: Noted, thanks :)

k-wall:
lgtm

087-external-certificate-manager.md
```diff
@@ -0,0 +1,324 @@
...
+
+Strimzi uses two generation annotations to track how Cluster CA certificates are trusted and Cluster CA issued certificates are used:
+* `strimzi.io/cluster-ca-key-generation` on Kafka pods to indicate the generation of the CA private key that signed the CA public cert trusted by that pod
+* `strimzi.io/cluster-ca-cert-generation` on Secrets containing certificates to indicate the generation of the CA public cert at the time the certificate was issued
+
+The `strimzi.io/cluster-ca-cert-generation` annotation is also used on Kafka pods to indicate the generation of the CA public cert associated with the certificate it is currently presenting.
+
+The following sections go into more detail about how we will use these annotations when cert-manager is issuing certificates.
+At a high level:
+* Strimzi will use cert path validation using the `java.security` libraries to determine whether a new Cluster CA public cert was signed with a new private key.
```
scholzj, line 105: Should we make it clear that this can be done from the public keys only? (I assume it can?)

087-external-certificate-manager.md
```diff
@@ -0,0 +1,324 @@
...
+    generateCertificateAuthority: <boolean>
+    generateSecretOwnerReference: <boolean>
+    renewalDays: <integer> # days before notAfter when we should start renewal
+    certificateExpirationPolicy: <renew-certificate|replace-key>
+    type: <strimzi.io|cert-manager.io> # (1)
+    certManagerIssuerRef: # (2)
+      name: <string>
+      kind: <Issuer|ClusterIssuer>
+      group: <string> # cert-manager.io by default
+    publicCert: # (3)
```
scholzj, line 72: I wonder if we can have some better name for this to make it more clea that this is the CA public certificate ...

087-external-certificate-manager.md
```diff
@@ -0,0 +1,324 @@
...
+      secretName: <string>
+      certificate: <string>
```
scholzj, line 74: A quick thought ... what about CAs based on multiple levels? (Root CA -> Intermediate CA -> Actual CA)? Will they be passed all in one file?

087-external-certificate-manager.md
Frawless: `publicCaCert` ?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,324 @@
...
+
+> ![Renewing the cluster CA public cert](./images/087-cert-renewals.png)
+>
+> Fig 1: Existing and proposed workflow when the user provides a new cluster CA public cert
+
+#### Issuing Cluster CA end-entity certificates
+
+When Strimzi needs to issue a certificate, instead of using the existing internal mechanism it will create a `Certificate` custom resource.
+Strimzi will specify the required CN/SANs in the `Certificate` resource for the end-entity certificate.
+Strimzi will set the `secretName` field in the `Certificate` resource as `<CERT_SECRET>-cm`, where `<CERT_SECRET>` is the name of the Secret Strimzi currently uses, for example `<CLUSTER_NAME>-cluster-operator-certs-cm`.
```
ppatierno, line 133: can we have an snippet example of such `Certificate` custom resource created by the operator?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,324 @@
...
+Strimzi will wait for the usual operation timeout during the reconciliation loop for the `Certificate` status to indicate that the certificate has been issued before continuing.
```
ppatierno, line 134: Just to make clear my understanding ... we are building a synchronous communication with cert-manager here, right? The operator creates the `Certificate` custom resource and then just wait (blocking the reconciliation) that the status is updated with the issued certificate or an error. We had such a similar kind of way with rebalancing at the beginning where, after the creation of a `KafkaRebalance`, the operator started an inner timer to poll the CC endpoint to check rebalancing status (within the same reconciliation). It didn't work well and we decided to leverage the periodic reconciliation. So `KafkaRebalance` is created and the reconciliation ends. On next reconciliation(s) the operator check the CC endpoint.
I was wondering if it should be the same here, not sure how long cert-manager can take to issue a certificate and if we want operator waiting in the reconciliation.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,324 @@
...
+
+> ![Issuing new component end-entity certificates](./images/087-new-ee-certs.png)
+>
+> Fig 2: Proposed workflow when cert-manager issues new component end-entity certificates
+
+#### Handling Cluster CA key replacements
+
+If during a reconciliation Strimzi determines (using cert path validation) that a new private key has been used, Strimzi will:
+1. Rename the existing cert in the `<CLUSTER_NAME>-cluster-ca-cert` Secret to `ca-YYYY-MM-DDTHH-MM-SSZ.crt` and copy over the new certificate.
+2. Update the `strimzi.io/ca-cert-hash` and increment the `strimzi.io/ca-cert-generation` and `strimzi.io/ca-key-generation` annotations.
```
ppatierno, line 168: it's happening on the Secret right? I would state it clearly.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,324 @@
...
+- `STRIMZI_CM_ISSUER_KIND`
+- `STRIMZI_CM_ISSUER_GROUP`
+
+These will be set by the Cluster operator when the User operator is deployed as part of the Entity operator and when `spec.clientsCa.type` is set to `cert-manager.io`.
+
+When a `KafkaUser` with `spec.authentication.type` set to `tls` is created Strimzi will create a `Certificate` custom resource.
+Strimzi will specify the required CN/SANs in the `Certificate` resource for the user certificate.
+Strimzi will specify the `Secret` for the certificate to be stored to match the name of the `KafkaUser` resource.
+Strimzi will request the certificate in both PEM and PKCS12 format.
+Strimzi will wait for the usual operation timeout during the reconciliation loop for the `Certificate` status to indicate that the certificate has been issued before continuing.
```
ppatierno, line 204: All the above "Strimzi" is the User Operator, can we mention it instead of Strimzi?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,324 @@
...
+
+#### Tracking changes to KafkaUser certificates
+
+Cert-manager will be responsible for renewing all user certificates.
+When a certificate is renewed cert-manager will update the related Secret.
+Since the user's clients are directly using the cert-manager created Secret, Strimzi will take no action.
+
+#### Handling Clients CA cert renewals and key replacements
+
+During a reconciliation Strimzi will check the hash of the certificate stored in the user's CA public cert Secret to see if an update is needed.
```
ppatierno, line 214: this is what we call "clients CA" right? Maybe using the same naming?

087-external-certificate-manager.md
```diff
@@ -0,0 +1,324 @@
...
+2. Create Secrets to store the Cluster CA and/or Clients CA public certs
+3. Update the `Kafka` resource to have `clusterCa.type` and/or `clientsCa.type` set to `cert-manager.io`, `clusterCa.generateCertificateAuthority` and/or `clientsCa.generateCertificateAuthority` set to `false`, and `certManagerIssuerRef` and `publicCert` configured.
+
+When using this feature for the ClientsCa:
+* On the next Cluster operator reconciliation Strimzi will copy the new certificate over to `<CLUSTER_NAME>-clients-ca-cert` replacing the old certificate and increment the cert generation annotation.
+* Strimzi will roll the Kafka pods to trust the new CA cert.
+* On the next User Operator reconciliation Strimzi will create `Certificate` resources for all the existing `KafkaUser` custom resources.
+
+When using this feature for the ClusterCa:
+* On the next Cluster operator reconciliation Strimzi will copy the new certificate over to `<CLUSTER_NAME>-cluster-ca-cert` (renaming and keeping the old certificate), increment the cert generation annotation and add the key generation annotation.
```
ppatierno, line 249: can we use the exact names of the annotations to be consistent across the rest of the proposal please? I think it's clearer.

087-external-certificate-manager.md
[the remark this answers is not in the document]
katheris: Marking as resolved as I updated the section titles so it should be clearer

087-external-certificate-manager.md
katheris: I think having the CA part is a good point, in future we might specify other certificates e.g. for calling a certificate manager API. I also get the point about a certificate is a public key, so using publicCert is not entirely accurate.

@scholzj @Frawless @ppatierno what about either `caCert` or `caPublicKey`? `caCert` aligns with the terminology we use in the docs today https://strimzi.io/docs/operators/latest/deploying#security-using-your-own-certificates-str but `caPublicKey` is more specific

ppatierno: I am for `caCert` at this point.

scholzj: `caCert` sounds good to me.

Frawless: +1 for `caCert`

087-external-certificate-manager.md
katheris: Yes, in the proposal I specified that the user should have them as a single file. That is also the behaviour when using the trust-manager from cert-manager.

087-external-certificate-manager.md
katheris: Yes that is correct. As you stated this does make the assumption that blocking is acceptable. So far from my testing the creation of the certificate is pretty quick but I guess it depends on the issuer. A rebalance can take a very long time, so I'm not sure how comparable the two actions are. The Kafka cluster cannot be created without the certificates so I think if possible it is better to wait, as otherwise it guarantees a delay in the deploying of the cluster.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,324 @@
...
+>
+> Fig 4: Existing workflow when the user provides a new clients CA public cert
+
+> ![Renewing/replacing the clients CA public cert/private key](./images/087-new-renew-replace-clientsca-cert.png)
+>
+> Fig 5: Proposed workflow when the user provides a new clients CA public cert
+
+## Affected/not affected projects
+
+This affects the Cluster Operator and User Operator.
```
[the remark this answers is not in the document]
katheris: My assumption was that we would use the fabric8 implementation, since it seems to be updated to the latest version as far as I could see. Where did you see it was not up to date?

087-external-certificate-manager.md
katheris: Ok, updated to be `caCert`

087-external-certificate-manager.md
```diff
@@ -0,0 +1,324 @@
...
+```yaml
+spec:
+  clusterCa:
+    validityDays: <integer> # notBefore=now, notAfter=now + validityDays
+    generateCertificateAuthority: <boolean>
+    generateSecretOwnerReference: <boolean>
+    renewalDays: <integer> # days before notAfter when we should start renewal
+    certificateExpirationPolicy: <renew-certificate|replace-key>
+    type: <strimzi.io|cert-manager.io> # (1)
+    certManagerIssuerRef: # (2)
```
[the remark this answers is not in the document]
katheris: That's a good point. I've updated it as you suggested, and put the `caCert` field under `certManager` as well.

087-external-certificate-manager.md
```diff
@@ -0,0 +1,324 @@
...
+   Users can optionally use [trust-manager](https://cert-manager.io/docs/trust/trust-manager/) to create this Secret, but they are responsible for installing trust-manager and creating the `Bundle` CR with a single `target` entry.
+4. Create a `Kafka` resource with `clusterCa.type` and/or `clientsCa.type` set to `cert-manager.io`, and `certManagerIssuerRef` and `publicCert` configured.
+
+### Internal handling of Cluster CA
+
+Strimzi uses two generation annotations to track how Cluster CA certificates are trusted and Cluster CA issued certificates are used:
+* `strimzi.io/cluster-ca-key-generation` on Kafka pods to indicate the generation of the CA private key that signed the CA public cert trusted by that pod
+* `strimzi.io/cluster-ca-cert-generation` on Secrets containing certificates to indicate the generation of the CA public cert at the time the certificate was issued
+
+The `strimzi.io/cluster-ca-cert-generation` annotation is also used on Kafka pods to indicate the generation of the CA public cert associated with the certificate it is currently presenting.
```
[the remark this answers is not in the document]
katheris: Updated

087-external-certificate-manager.md
ppatierno: > The Kafka cluster cannot be created without the certificates so I think if possible it is better to wait

Yeah good point. Let's start this way and see how it works.

087-external-certificate-manager.md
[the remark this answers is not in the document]
ppatierno: As far as I understood the model classes we have here https://github.com/kubernetes-client/java/tree/master/client-java-contrib/cert-manager/src/main/java/io/cert/manager/models are automatically generated by this script https://github.com/kubernetes-client/java/blob/master/client-java-contrib/cert-manager/update.sh#L34 which is still pointing to cert-manager v1.13.4 (one year ago) while the last release is v1.17.1. Even if it's also possible that CRDs didn't change that much? Anyway I would try to update the fabric8 client with latest cert-manager release.

087-external-certificate-manager.md
ppatierno: @katheris maybe did you miss this comment in the last batch of fixes? or do you have reasons for not putting a snippet example of `Certificate` resource?

katheris: @ppatierno yeah I did miss it, I've added the example now

ppatierno (approved):
LGTM. Thank you very much for the awesome work @katheris! And sorry to be so pedantic :-)

ppatierno:
This proposal has now 4 (binding) +1 and 2 (non-binding) +1. If there are no more comments by the end of tomorrow I will merge it.

@katheris can you changed the index on the file names, images and update the index file as well please? (from 87 it should be 99)
@MichaelMorrisEst do you have any additional comments?

katheris:
> @katheris can you changed the index on the file names, images and update the index file as well please? (from 87 it should be 99)

Done :)

ppatierno:
> Done :)

I got the wrong index ... congrats for making the 100th proposal! :-)

ppatierno:
This proposal now has 5 (binding) +1 and 2 (non-binding) +1. Approved!

PR #136, opened 2024-11-14
Add proposal for single step multi version downgrade

MichaelMorrisEst:
Related to [strimzi-kafka-operator/issues/10801](https://github.com/strimzi/strimzi-kafka-operator/issues/10801)


087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,45 @@
+# Support Single step multi version downgrade for Zookeeper based clusters 
+
+This proposal seeks to introduce support for multi-version downgrade of Strimzi in a single step where possible.
+The proposal only relates to Zookeeper based clusters.
```
scholzj, line 4: As I commented on the https://github.com/strimzi/strimzi-kafka-operator/issues/10801 issue. This does not seem to make any sense anymore:
* At best, this would land in Strimzi 0.45 with support for Kafka 3.8 and 3.9.
* These are the last versions with ZooKeeper support. ZooKeeper will be dropped in Strimzi 0.46 / Kafka 4.0
* So there will be never a downgrade of a ZooKeeper-based cluster that would utilize this feature

MichaelMorrisEst: I agree, doesn't make sense for Zookeeper at this stage. I've updated the proposal to address the same issue for KRaft based clusters instead

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,39 @@
...
+
+This proposal would make downgrade of Strimzi more user friendly, in specific, in the scenario where a single-step multi version upgrade has been performed and a need arises to rollback before the metadata version has been stepped. The proposal will allow the user perform the roll back in a single step rather than the multiple steps currently required.
+
+## Proposal
+
+During downgrade an attempt is made to read the Kafka information for the 'from' Kafka version by the 'to' version of the operator. When the 'from' Kafka version is not supported by the 'to' Strimzi version an error will be thrown because the version is unknown. Information such as the metadata version, interbroker protocol message format, log message format version are read from kafka-versions.yaml during the creation of a KafkaVersion object to represent the 'from' kafka version and the error message is generated as it cannot find the information for the unknown version.
+
+However, while this information is important to know for the 'to' version in upgrade and downgrade, and for the 'from' version in upgrade, it is not important for the 'from' version in the downgrade as it is not subsequently used.
```
scholzj, line 20: I don't agree it is not important. While currently, there are no explicit steps needed to do during downgrade, there might be in the future. So I think the statement that it is never needed is not true. (in the distant past, I think there were for example, some breaking ZooKeeper changes that might have included special steps at downgrade).

scholzj (commented):
Thanks for updating this proposal to KRaft. I think this would be a good feature to have and in fact I had it somewhere deep on my list. But I think the proposal should be extended to cover:
* More details of the implementation -> I assume you will need to somehow carry the information we are downgrading from unknown version in the code etc. This should be explained in more detail.
* System test strategy -> Describe how will it be tested etc. Saying that we do not expect any system tests is probably an option ... but even that should be covered.
* The risks tat are part of this proposal and how they will be documented
    * While most Kafka versions do not expect any special treatment when upgrading / downgrading, it is possible that there might be something needed for some versions. While for upgrade, this can be _simply implemented_ in the new operator version, for a downgrade like this it will not be supported by the old operator. As such, it is unclear how will such a downgrade be supported, which versions will support it, how deep will it be possible etc.
    * Similarly, there are many issues with the operator versions as well. For example, in one of the future Strimzi versions where the dynamic KRaft qourum configuration will be supported - the new operator will know how to swicth between the static qourum configuration and dynamic qourum configuration and possibly the other way around. But the old operator version will not have any knowledge of this and will not be able to do this and the downgrade would simply break.
    * I do not think ^^^ these are blockers. But these risks need to be covered by the proosal and it needs to describe how we will protect the users from it but also how will the expectations be set for the users. E.g. by explaining that this will be something to be used _at your own risk_ and that you are expected to test it for your combination of versions / configuration etc. This should likely also explain what exactly we test and what is just some code that does not block the downgrade, but offers not guarantees that it would work.

PS: If you write the proposal with a sentence-per-line, it will make it much easier to comment on it during the reviews than with paragraph-per-line. No need to change the existing proposal ... just something what might be helpful for the next one 😉.

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,39 @@
...
+- Downgrading of Kafka where the metadata version in use is greater than the maximum supported by the 'to' Kafka version is not supported
+
+## Motivation
+
+This proposal would make downgrade of Strimzi more user friendly, in specific, in the scenario where a single-step multi version upgrade has been performed and a need arises to rollback before the metadata version has been stepped. The proposal will allow the user perform the roll back in a single step rather than the multiple steps currently required.
```
PaulRMellor, line 14:
```suggestion
This proposal would make downgrade of Strimzi more user friendly, specifically, in the scenario where a single-step multi version upgrade has been performed and a need arises to rollback before the metadata version has been advanced. The proposal will allow the user perform the roll back in a single step rather than the multiple steps currently required.
```

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,39 @@
...
+However, while this information is important to know for the 'to' version in upgrade and downgrade, and for the 'from' version in upgrade, it is not important for the 'from' version in the downgrade as it is not subsequently used.
+
+Kafka does not support downgrade where the metadata version in use is higher than the highest metadata supported by the 'to' Kafka version or where there has been changes in the metadata between the 'from' and 'to' versions. There is already logic in place in Strimzi to prevent such a downgrade from happening. This proposal does not propose any changes to this functionality, it shall continue to be not supported to downgrade in such a scenario. 
+
+The proposed impact therefore is:
+- Add handling to not throw an exception when the 'from' version is not known in a downgrade.
+
+See https://github.com/strimzi/strimzi-kafka-operator/pull/10929
+
+## Affected/not affected projects
```
PaulRMellor, line 29:
```suggestion
## Affected projects
```

PaulRMellor (commented):
Thanks. I left a few comments on wording for clarity. I can return when the proposal covers the implementation detail and handling differences in functionality mentioned by Jakub

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,74 @@
...
+
+However, while this information is important to know for the 'to' version in upgrade and downgrade, and for the 'from' version in upgrade, it is not currently used for the 'from' version in the downgrade. It might be potentially useful in some some future downgrade scenarios, but none that are currently supported.
+
+Kafka does not support downgrade where the metadata version in use is higher than the highest metadata supported by the 'to' Kafka version or where there has been changes in the metadata between the 'from' and 'to' versions. There is already logic in place in Strimzi to prevent such a downgrade from happening. This proposal does not propose any changes to this functionality, it shall continue to be not supported to downgrade in such a scenario. 
+
+### Impacts:
```
scholzj, line 24: The proposal should not give options to choose from but it should describe the way how to do it. So from the A and B options, you should pick the ones that you think are the best and move the other options to the rejected alternetives (where others can see them during the review and say "Hey, I think this rejected one is actually beter").

In this case, I personally think we should stick with the A options for both cases.

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,74 @@
...
+An exception will be thrown if the requirement is not satisfied. 
+
+For example, if specific steps need to be implemented in the operator to upgrade/rollback from Kafka version 'x' to Kafka version 'y' and such steps are implemented in operator version 'n' then as part of the implementation in operator version 'n' to support the 'x' -> 'y' upgrade, the minOperatorVersionForRollback will bet set to 'n'.
+Any attempt to rollback a Kafka cluster from 'y' -> 'x' with operator version earlier than 'n' shall be rejected by the operator.
+
+### System Tests
+
+A new system test will be added with the following steps:
+- Deploy the latest operator
+- Deploy a kafka cluster with the highest supported kafka version but metadata version set to the appropriate version for the Kafka version the cluster will be downgrade to in the following steps
```
scholzj, line 60:
```suggestion
- Deploy a Kafka cluster with the highest supported Kafka version but metadata version set to the appropriate version for the Kafka version the cluster will be downgrade to in the following steps
```

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,74 @@
...
+- Downgrade the operator to the most recent version which does not support the kafka version used in the cluster.
```
scholzj, line 61:
```suggestion
- Downgrade the operator to the most recent version which does not support the Kafka version used in the cluster.
```

katheris (commented):
Overall this proposal looks good to me. I agree with Jakub that option A seems best in both scenarios, so you should update to move the rejected option to the `Rejected alternatives` section. Once that is done I would be happy to approve this proposal.

ppatierno:
A couple of comments ...
1. I agree with Kate and Jakub and the proposal should stick with one option and I am for option A as well.
2. I think you are using "rollback" and "downgrade" with the same meaning, right? Could we stick with just one term, maybe using "downgrade" consistently?

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Impacts:
+
+#### Handle unknown versions
+
+Proposal A
+KraftVerionChangeCreator.detectToAndFromVersions() invokes the version() method in KafkaVersion.Lookup which throws an exception if the requested version is not known.
+This exception can be caught in detectToAndFromVersions() in the rollback scenario and an instance of KafkaVersion created that has null values for the unknown data (such as metadata version).
+As this data is not currently read anywhere in the code the null values will not have any negative effect.
+However future code changes could result in NPE as it will not necessarily be obvious the values may be null.
```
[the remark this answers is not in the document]
MichaelMorrisEst: We do know the version number, just not anything else. The version number is useful (for example in KRaftVersionChnageCreator.prepareVersionChange(...) to determine if its an upgrade or downgrade), so it is useful to have the KafkaVersion object with the version number in it rather than having a null and losing the means to pass the 'from' version number

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,72 @@
...
+
+Kafka does not support downgrade where the metadata version in use is higher than the highest metadata supported by the 'to' Kafka version or where there has been changes in the metadata between the 'from' and 'to' versions. There is already logic in place in Strimzi to prevent such a downgrade from happening. This proposal does not propose any changes to this functionality, it shall continue to be not supported to downgrade in such a scenario. 
+
+### Impacts:
+
+#### Handle unknown versions
+
+KraftVerionChangeCreator.detectToAndFromVersions() invokes the version() method in KafkaVersion.Lookup which throws an exception if the requested version is not known.
+This exception can be caught in detectToAndFromVersions() in the downgrade scenario and an instance of KafkaVersion created that has null values for the unknown data (such as metadata version).
+As this data is not currently read anywhere in the code the null values will not have any negative effect.
```
scholzj, line 30: Maybe we should clarify here that we will compare the versions to confirm the assumption that it is a downgrade? (i.e. we will double-check that it is for example 4.5.0 to 4.1.0 and not 3.17.5 to 4.1.0)

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,72 @@
...
+It should also be documented that it is the responsibility of the user to ensure they validate the downgrade path they are taking to ensure no such issues exist. 
+
+
+### System Tests
+
+A new system test will be added with the following steps:
+- Deploy the latest operator
+- Deploy a Kafka cluster with the highest supported Kafka version but metadata version set to the appropriate version for the Kafka version the cluster will be downgrade to in the following steps
+- Downgrade the operator to the most recent version which does not support the Kafka version used in the cluster.
+- Downgrade the Kafka version used in the cluster to a version supported by the downgraded operator
```
scholzj, line 46: Thinking about this a bit more -> this ST makes not much 🤷‍♂️. The problem is that it would not really test the current version but the past / already released version. E.g., if we tested this today, we would start with Strimzi main branch / Kafka 3.9.0 and downgrade to Strimzi 0.44.0. If the downgrade would not work, it would be because of an issue with Strimzi 0.44. So it will not help us to discover any issues in the operator version we are actually testing.

I wonder if we should instead fake it? for example:
* Deploy Kafka 3.9.0 cluster with current main Strimzi branch
* Pause the reconciliation
* Change the annotations / Kafka CR status to some _unknown_ future version (e.g. 98.99.100)
* Unpause the reconciliation and have the operator handle the _downgrade_?

Or maybe we should give up on ST completely and have it covered by UT only? 🤷‍♂️

Any thoughts @im-konge @Frawless @see-quick?

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,72 @@
...
+## Compatibility
+
+N/A
+
+## Rejected alternatives
+
+Introducing similar functionality for Zookeeper based clusters rejected as the next version of Strimzi shall be the last to support Zookeeper and therefore there will never be a downgrade of a Zookeeper based cluster to that Strimzi version.
+
+#### Handle unknown versions
+Change the KafkaVersion class to an interface with two different implementations.
```
scholzj, line 61:
```suggestion
#### Handle unknown versions

Change the KafkaVersion class to an interface with two different implementations. 
```

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,72 @@
...
+
+#### Handle unknown versions
+Change the KafkaVersion class to an interface with two different implementations. 
+One implementation for the case where the Kafka version is known (which would be functionally equivalent to the current KafkaVersion class), and another implementation for unknown Kafka versions.
+The version() method in the Lookup class will return an instance of the appropriate class with tha calling method able to distinguish, when necessary, if the version is known or not based on the class of the returned instance.
+The methods in the implementation for the unknown versions will throw an exception for all methods for which the information is not available (i.e. all except the version method) to ensure an error is generated if any code mistakenly tries to use information that is not available for an unknown version. 
+
+#### Handling for specific versions
+Add a new field to KafkaStatus, named minOperatorVersionForRollback, which is set by the operator during Kafka upgrade when it is known that specific extra steps are needed for a downgrade to succeed.
```
scholzj, line 67:
```suggestion
#### Handling for specific versions

Add a new field to KafkaStatus, named minOperatorVersionForRollback, which is set by the operator during Kafka upgrade when it is known that specific extra steps are needed for a downgrade to succeed.
```

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,72 @@
...
+### Impacts:
+
+#### Handle unknown versions
+
+KraftVerionChangeCreator.detectToAndFromVersions() invokes the version() method in KafkaVersion.Lookup which throws an exception if the requested version is not known.
+This exception can be caught in detectToAndFromVersions() in the downgrade scenario and an instance of KafkaVersion created that has null values for the unknown data (such as metadata version).
+As this data is not currently read anywhere in the code the null values will not have any negative effect.
+
+#### Handling for specific versions
+It is possible that future Kafka versions might need extra steps to be executed for a particular upgrade/downgrade path.
```
scholzj, line 33:
```suggestion
#### Handling for specific versions

It is possible that future Kafka versions might need extra steps to be executed for a particular upgrade/downgrade path.
```

scholzj (commented):
@MichaelMorrisEst Thanks for the updates. I left some more formatting comments and two comments related to the content -> one is to confirm if we plan to do a check that it is actually a downgrade and other is about the ST where I realized it does not make much sense.

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,72 @@
...
+
+However, while this information is important to know for the 'to' version in upgrade and downgrade, and for the 'from' version in upgrade, it is not currently used for the 'from' version in the downgrade. It might be potentially useful in some some future downgrade scenarios, but none that are currently supported.
+
+Kafka does not support downgrade where the metadata version in use is higher than the highest metadata supported by the 'to' Kafka version or where there has been changes in the metadata between the 'from' and 'to' versions. There is already logic in place in Strimzi to prevent such a downgrade from happening. This proposal does not propose any changes to this functionality, it shall continue to be not supported to downgrade in such a scenario. 
+
+### Impacts:
```
katheris, line 24: There seems to be an empty heading here

087-single-step-multi-version-downgrade.md
katheris: +1

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,72 @@
...
+
+#### Handle unknown versions
+
+KraftVerionChangeCreator.detectToAndFromVersions() invokes the version() method in KafkaVersion.Lookup which throws an exception if the requested version is not known.
```
katheris, line 28:
```suggestion
KraftVersionChangeCreator.detectToAndFromVersions() invokes the version() method in KafkaVersion.Lookup which throws an exception if the requested version is not known.
```

katheris (approved):
Couple of nits but otherwise looks good to me

087-single-step-multi-version-downgrade.md
MichaelMorrisEst: Yes it groups the level 4 headings below. I can add a sentence just to make it clearer

087-single-step-multi-version-downgrade.md
MichaelMorrisEst: Yes thats a problem alright. I will update to state UT only

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,81 @@
...
+
+#### Handling for specific versions
+
+It is possible that future Kafka versions might need extra steps to be executed for a particular upgrade/downgrade path.
+Similarly future versions of the operator might be capable of making configuration changes that previous operator versions would not be capable of undoing. 
+
+Documentation of the downgrade procedure for any upgrade/downgrade paths where such issues arise should document the specific risks for those paths and how or if it is possible to downgrade back in those scenarios.
+It should also be documented that it is the responsibility of the user to ensure they validate the downgrade path they are taking to ensure no such issues exist. 
+
+
```
scholzj, line 42:
```suggestion

```

scholzj (approved):
LGTM. Thanks for the proposal.

087-single-step-multi-version-downgrade.md
```diff
@@ -0,0 +1,81 @@
...
+- Downgrading of Kafka where the metadata version in use is greater than the maximum supported by the 'to' Kafka version is not supported
+
+## Motivation
+
+This proposal would make downgrade of Strimzi more user friendly, specifically, in the scenario where a single-step multi version upgrade has been performed and a need arises to downgrade before the metadata version has been advanced. The proposal will allow the user perform the roll back in a single step rather than the multiple steps currently required.
```
ppatierno, line 14:
```suggestion
This proposal would make downgrade of Strimzi more user friendly, specifically, in the scenario where a single-step multi version upgrade has been performed and a need arises to downgrade before the metadata version has been advanced. The proposal will allow the user perform the downgrade in a single step rather than the multiple steps currently required.
```

ppatierno (approved):
LGTM. Just a nit.

Frawless (approved):

scholzj:
This proposal has now 4 binding +1 votes. Given it has been open for some time, if there are no new comments, we should try to get it closed this week. Si if you haven't looked at it and are interested in it, please do so soon.

see-quick (approved):

im-konge (approved):

PaulRMellor (approved):

scholzj:
@MichaelMorrisEst I wanted to update the proposal number and add it to the index in README.md. But I do not seem to be able to push into the PR for some reason. Can you please do it so that I can close the vote and merge it? The next proposal number is  `093` right now. But it might change depending on when you get to it. Thanks.

MichaelMorrisEst:
> @MichaelMorrisEst I wanted to update the proposal number and add it to the index in README.md. But I do not seem to be able to push into the PR for some reason. Can you please do it so that I can close the vote and merge it? The next proposal number is `093` right now. But it might change depending on when you get to it. Thanks.

@scholzj Done, thanks

ppatierno:
This proposal has now 7 binding +1 votes, so it's approved. 
@MichaelMorrisEst thanks! looking forward to the implementation!

MichaelMorrisEst:
Thanks everyone for taking the time to review and for providing feedback

PR #139, opened 2024-11-25
Add proposal for adopting Kafka Connect health check endpoint

fvaleri:
This proposal describes an approach for adopting the health check endpoint for Kafka Connect and Mirror Maker 2 components.


087-adopt-connect-health-endpoint.md
```diff
@@ -0,0 +1,84 @@
+# Adopt the Kafka Connect health check endpoint
+
+In a Kubernetes node, the kubelet component uses the configured liveness probe to know when to restart a container (i.e. deadlock).
+Instead, it uses the configured readiness probe to know when a container is ready to start accepting traffic (i.e. startup phase completed).
```
scholzj, line 4: The `Instead` does not seem to fit into how the text flows? Also, the test seems a but confusing with regards to startup probes. Maybe you could just focus that it is used to verify if the pod is ready to accept traffic?

087-adopt-connect-health-endpoint.md
```diff
@@ -0,0 +1,84 @@
...
+A common pattern for liveness probes is to use the same HTTP endpoint as for readiness probes.
```
scholzj, line 5: I don't think that is a common pattern ... but if you have nothing better, it is an option.

087-adopt-connect-health-endpoint.md
```diff
@@ -0,0 +1,84 @@
...
+  "status": "healthy",
+  "message": "Worker has completed startup and is ready to handle requests."
+}
+```
+
+## Proposal
+
+Strimzi will support the Kafka 3.8 branch for an extended period of time ([SIP-77](https://github.com/strimzi/proposals/blob/main/077-support-for-kafka-4.0.md#extended-support-for-kafka-39)) due to the upcoming major Kafka release.
+
+The `/health` endpoint will be used for Kafka Connect and Mirror Maker 2 components only when the following conditions will be true at the same time:
```
scholzj, line 63: So, will it be used only for the readiness probe? Or for both readiness and liveness? ny changes tot he default timing of the probes?

scholzj (commented):
Thanks for the proposal. I left some comments. Could you please also clarify what kind of user-impact we expect? What is the chance that users will actuall see that the probe changed? That defines whether we for example need to introduce this through a feature gate etc.

087-adopt-connect-health-endpoint.md
fvaleri: Getting rid of this as not important in this context.

087-adopt-connect-health-endpoint.md
```diff
@@ -0,0 +1,84 @@
...
+$ kubectl exec my-cluster-kafka-0 -- curl -s http://my-connect-cluster-connect-api:8083/health | jq
+{
+  "status": "healthy",
+  "message": "Worker has completed startup and is ready to handle requests."
+}
+```
+
+## Proposal
+
+Strimzi will support the Kafka 3.8 branch for an extended period of time ([SIP-77](https://github.com/strimzi/proposals/blob/main/077-support-for-kafka-4.0.md#extended-support-for-kafka-39)) due to the upcoming major Kafka release.
```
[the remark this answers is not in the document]
fvaleri: Removing the extended support part (my bad), I still think that it would be good to provide `/health` endpoint support as early as possible. The change to get rid of the 3.8 endpoint fallback logic is very simple, and there will be unit tests failing as a reminder. Wdyt?

087-adopt-connect-health-endpoint.md
```diff
@@ -0,0 +1,84 @@
...
+
+The `/health` endpoint will be used for Kafka Connect and Mirror Maker 2 components only when the following conditions will be true at the same time:
+
+1. `.spec.image` is null
```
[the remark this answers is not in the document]
fvaleri: Ok, sounds good.

087-adopt-connect-health-endpoint.md
```diff
@@ -0,0 +1,84 @@
...
+The only affected project is the Cluster Operator, in particular Kafka Connect and Mirror Maker 2 components.
+
+## Compatibility
+
+This change is fully backwards compatible as no further restrictions are created by the `/health` endpoint.
+
+## Rejected alternatives
+
+Wait until Kafka 3.8 is out of support in Strimzi and then switch from `/` to `/health`.
+It is important to adopt new Kafka features as early as possible, and this approach would add a big delay considering Kafka 3.8 extended support.
```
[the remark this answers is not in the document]
fvaleri: Ok. See my other comment on this. Thanks.

087-adopt-connect-health-endpoint.md
fvaleri: For both, added. Instead, the following line was already present:

> Default liveness and readiness probes configuration won't change.

scholzj:
> > Could you please also clarify what kind of user-impact we expect? What is the chance that users will actuall see that the probe changed?
> 
> I don't see any impact on users, also considering that the `/` endpoint will still be available, but not used anymore when Kafka 3.9+ is used. Added a note.

The impact is not about the `/` endpoint not being available. But the proposal argues that the new endpoint behaves better/differently. In that case, one can expect other behavior from it - e.g. the perceived startup taking longer - and that might mean that users will need to change their probe settings compared to what they have today or face the issue that their Connect / MM2 deployments might start failing.

fvaleri:
@scholzj I added a commit with my user impact tests, but I see GH is taking a lot of time to process this commit (see "Processing updates" under PR title). I guess we just have to wait.

087-adopt-connect-health-endpoint.md
```diff
@@ -0,0 +1,88 @@
...
+When Kafka 3.8 branch will be out of support, we will be able to get rid of the 3.8 fallback logic, and always use the `/health` endpoint.
+Unit tests related to this feature will fail when Kafka 3.8 support will be removed from the code base.
+
+Default liveness and readiness probes configuration won't change.
+
+## Affected/not affected projects
+
+The only affected project is the Cluster Operator, in particular Kafka Connect and Mirror Maker 2 components.
+
+## Compatibility
```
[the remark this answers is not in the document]
fvaleri: > You explain above how the new health check is better ... but here, you say it does not matter. 

I'm not saying that. I'm just saying that performance is not a concern.

> The test you did below is nice, but I assume that it is a very simple and small cluster with the default configuration.

We can repeat that with a more realistic setup if that helps to gain confidence, but a feature gate seems to be an overkill to me.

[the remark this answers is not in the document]
scholzj: > We can repeat that with a more realistic setup if that helps to gain confidence

It is not about _realistic_ setup. It is about the extreme setups where it is most likely to break someone's setup.

PaulRMellor (approved):
The proposal is very clear and sounds like a useful update.
I guess the question is over when to introduce.
Maybe it's better to introduce it when we drop support for 3.8 to reduce any potential risks or testing overheads and tie the change to that

fvaleri:
@scholzj @PaulRMellor your feedback should be addressed.

087-adopt-connect-health-endpoint.md
```diff
@@ -0,0 +1,86 @@
...
+
+| Endpoint | Pod ready time in seconds | Response time in ms |
+|----------|---------------------------|---------------------|
+| /        | 65                        | 3.0286              |
+| /health  | 62                        | 3.7525              |
+
+## Rejected alternatives
+
+Switch to the `/health` endpoint while still supporting Kafka 3.8.
+This would allow to get the new feature early, but at the cost of additional complexity.
```
PaulRMellor, line 86:
```suggestion
This would allow the new feature to be adopted earlier, but with the cost of additional complexity.
```

PaulRMellor (approved):
Great. Thanks @fvaleri 
I just left a minor suggestion on wording

087-adopt-connect-health-endpoint.md
```diff
@@ -0,0 +1,86 @@
...
+  "status": "healthy",
+  "message": "Worker has completed startup and is ready to handle requests."
+}
+```
+
+## Proposal
+
+Strimzi currently supports both Kafka 3.8 and 3.9 releases.
+
+As the `/health` endpoint was introduced in Kafka 3.9, we will wait for Kafka 3.8 to go out of support in Strimzi before doing the switch.
```
scholzj, line 62: Maybe you can specifically mention Strimzi 0.46 as that is where the bulk of the Kafka 4.0 and ZooKeeper removal changes will land.

087-adopt-connect-health-endpoint.md
```diff
@@ -0,0 +1,87 @@
+# Adopt the Kafka Connect health check endpoint
+
+In a Kubernetes node, the kubelet component uses the configured liveness probe to know when to restart a container (i.e. deadlock).
```
tombentley, line 3:
```suggestion
In a Kubernetes node, the kubelet component uses the configured liveness probe to know when to restart a container. 
With a suitably written application this can be used to restart a container if the applications process enters an illegal state (e.g. deadlock).
```

tombentley (approved):
Thanks @fvaleri

see-quick (approved):
LGTM. Thanks

ppatierno (approved):
LGTM. +1 from me.

087-adopt-connect-health-endpoint.md
fvaleri: Thanks Tom.

scholzj:
This proposal has no 5 binding +1 votes. If there are no new comments on Monday EOB, we should close it as approved.

scholzj:

PR #141, opened 2024-12-11
Add Connect support to test-container

mimaison:


090-add-connect-to-test-container.md
```diff
@@ -0,0 +1,153 @@
...
+
+With [test-container](https://github.com/strimzi/test-container) and [test-clients](https://github.com/strimzi/test-clients), it's possible to build test environments for all Apache Kafka components but Kafka Connect. 
+
+Being able to easily start Kafka Connect clusters would be useful for testing client-side components, such as [metrics-reporter](https://github.com/strimzi/metrics-reporter) and [strimzi-kafka-oauth](https://github.com/strimzi/strimzi-kafka-oauth), with Kafka Connect.
+
+## Proposal
+
+Create 2 new classes in the `io.strimzi.test.container` package of `test-container`:
+- `StrimziConnectCluster`: This represents a Kafka Connect cluster. This is the main API users will call to create and interact with Kafka Connect clusters.
+- `StrimziConnectContainer`: This represents a Kafka Connect worker.
```
scholzj, line 17: Do we really need two different public classes? Wouldn't `StrimziConnectCluster` be sufficient? For the Kafka clusters, this duality seems weird and tends to each class supporting different features etc.

090-add-connect-to-test-container.md
```diff
@@ -0,0 +1,153 @@
...
+        public StrimziConnectClusterBuilder withKafkaVersion(String kafkaVersion) { }
+
+        /**
+         * Whether to include the FileStream connectors.
+         * If not called, the FileStream connectors are enabled.
+         *
+         * @param includeFileConnectors Use false to not include the FileStream connectors
+         * @return the current instance of {@code StrimziConnectClusterBuilder} for method chaining
+         */
+        public StrimziConnectClusterBuilder withIncludeFileConnectors(boolean includeFileConnectors) { }
```
see-quick, line 104:
```suggestion
        public StrimziConnectClusterBuilder withIncludeFileConnectors() { }
```
No need to have a boolean here I think :). By default `false` and if you call such method you simply assign true to a variable.

090-add-connect-to-test-container.md
see-quick: Yeah, I think maybe if we just expose `StrimziConnectCluster` and having core functionality in the `StrimziConnectContainer` class as we have it for `StrimziKafkaContainer` could be beneficial ... in terms that you can easily create StrimziConnectCluster with 1 replica (i.e., `StrimziConnectContainer`) or multiple replicas...

scholzj: What is hard on using the `StrimziConnectCluster` cluster with single replica? I think we should not have two public APIs to do the same as it proved very disfunctional for the Kafka brokers.

see-quick: > What is hard on using the StrimziConnectCluster cluster with single replica? I think we should not have two public APIs to do the same as it proved very disfunctional for the Kafka brokers.

By the previous comment, I meant that we would have `StrimziConnectContainer` (which would not be a part of the public API). It would be used to simplify the `StrimziConnectCluster` class, where you would do just the same as we do for StrimziKafkaCluster (i.e., start multiple StrimziKafkaContainer).

mimaison:
Thanks for the reviews. I pushed an update that:
- removes `StrimziConnectContainer`. I kept `getWorkers()` but it now returns a collection of `GenericContainer`
- adds a new method to `StrimziKafkaCluster` to get bootstrap servers that can be used from other containers

scholzj (approved):
LGTM, +1. Thanks for the proposal @mimaison.

see-quick (approved):

tombentley (approved):
LGTM, thanks @mimaison

090-add-connect-to-test-container.md
Frawless: So what will be output of this? Are we keep the proposed behavior, change it in favor of what Maros propose, or mix it? Guess any other connection will be mounted via volumes?

mimaison: I'm happy with the current proposal. For other connectors, users will need to mount the necessary JARs and use `withAdditionalConnectConfiguration()` to properly set the `plugin.path` configuration accordingly.

Frawless: I think better way is to remove boolean and just name it properly. So `withoutFileConnectors()` for example as default value will be true. That way user can turn it off and load their own File connectors. WDYT?

090-add-connect-to-test-container.md
```diff
@@ -0,0 +1,144 @@
...
+        /**
+         * Sets the Kafka cluster the Kafka Connect cluster will use to.
+         * @param kafkaCluster the {@link StrimziKafkaCluster} instance
+         * @return the current instance of {@code StrimziConnectClusterBuilder} for method chaining
+         */
+        public StrimziConnectClusterBuilder withKafkaCluster(StrimziKafkaCluster kafkaCluster) { }
+
+        /**
+         * Sets the number of Kafka Connect workers in the cluster.
+         * If not called, the cluster has a single broker
```
ppatierno, line 68: "single worker"?

ppatierno (approved):
Just one nit but LGTM. Thanks!

090-add-connect-to-test-container.md
mimaison: Fixed

090-add-connect-to-test-container.md
katheris: What's the reason for having the file connectors opt out rather than opt in? Given that today you have to actively add the file connectors when running Kafka yourself? Are we expecting most tests to need the file connectors?

mimaison: Out of the built-in connectors these are the easiest to run, as the MirrorMaker connector require 2 Kafka clusters (technically HeartbeatConnector only requires 1 cluster but still it's confusing to use it this way).
I expect the FileStreams connectors to satisfy most of the requirements (I think it should be enough to do the tests I want in metrics-reporter). Then users with more advanced requirements can inject and use their own connectors if needed.

scholzj:
@katheris Do you have something more on this? Or can we start slowly closing the vote on this proposal? Thanks.

Frawless (approved):

katheris (approved):
LGTM

scholzj:
Approved with 7 binding +1 votes. Thanks for the proposal.

PR #142, opened 2024-12-12
Exposing name resolution configurations for pods

gertschouten:


090-support-dns-config.md
```diff
@@ -0,0 +1,41 @@
+<!-- This template is provided as an example with sections you may wish to comment on with respect to your proposal. Add or remove sections as required to best articulate the proposal. -->
+
```
scholzj, line 2: I guess you can remove this.

090-support-dns-config.md
```diff
@@ -0,0 +1,41 @@
...
+## Proposal
+
+By appending Strimzi's common PodTemplate in the Strimzi api with additional properties dnsPolicy (to convey ClusterFirst, ClusterfirstWithHostNet, Default and None as name resolution policies) and dnsConfig (to specify nameservers, conditional subdomains to search for and options to decide resolution behaviors), the system configuration file /etc/recolv.conf for Strimzi resources can be controlled, thereby deciding how domain name resolution should be performed on a resource level. 
+Strimzi's CRDs generated from from PodTemplate will thereby allow for user input for name resolution customization.
+By appending these dnsConfig and dnsPolicy properties to WorkloadUtils in the Strimzi cluster operator for PodBuilder and PodTemplateSpecBuilder so that these will propagate to Pods controlled by the operator, the propagation from Strimzi CRD to the Pod resource will be accounted for.
+Provisioned Pods will thereby utilize name resolution configuration based on the user input if specified and fall back to defaults, similar to the current situation, in case no dnsConfig or dnsPolicy is specified by the user. 
+
```
scholzj, line 28: Can you provide a YAML example of how it will be used?

scholzj (commented):
Thanks for the proposal. I think it looks mostly good. I just left two comments.

scholzj:
You will also need to fix the DCO signoff -> the instructions should be under the _Details_ link next to the  DCO status in the PR status section.

PaulRMellor (commented):
This proposal seems useful and addresses certain use cases, though I’m not familiar with the technical aspects of the DNS resolution process. I’ve left a few comments regarding wording for clarity. YAML example would be a great addition.

090-support-dns-config.md
gertschouten: Addressed in [10ba3f8](https://github.com/strimzi/proposals/pull/142/commits/10ba3f8f70da291a4836d9a7acf0da80c5b30d51)

090-support-dns-config.md
gertschouten: Addressed in [10ba3f8](https://github.com/strimzi/proposals/pull/142/commits/10ba3f8f70da291a4836d9a7acf0da80c5b30d51)

090-support-dns-config.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Affected/not affected projects
```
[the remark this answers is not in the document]
gertschouten: Explicitly listed affected projects and removed MirrorMaker from that list in [10ba3f8](https://github.com/strimzi/proposals/pull/142/commits/10ba3f8f70da291a4836d9a7acf0da80c5b30d51)

090-support-dns-config.md
```diff
@@ -0,0 +1,72 @@
...
+nameserver 192.0.2.1
+search ns1.svc.cluster-domain.example my.dns.search.suffix
+options ndots:2 edns0
+```
+
+## Affected/not affected projects
+
+Affected:
+- io.strimzi.api for the PodTemplate.
+- io.strimzi.cluster-operator for resources that utilize PodTemplate: kafka, kafkabridge, kafkaconnect, kafkamirrormaker2, kafkanodepool.
```
scholzj, line 64: Maybe you can just list the affected modules?

```suggestion
- `api` module in the operator project
- `cluster-operator` module in the operator project
```

090-support-dns-config.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Compatibility
+
+Non-specified dnsConfig and dnsPolicy will not propagate these properties to the Pod resources and thereby comply to the current situation. 
+
+## Rejected alternatives
+
+-
```
scholzj, line 72: You can simply remove this if there are no rejected alternatives.

090-support-dns-config.md
```diff
@@ -0,0 +1,72 @@
...
+      dnsConfig:
+        nameservers:
+          - 192.0.2.1 
+        searches:
+          - ns1.svc.cluster-domain.example
+          - my.dns.search.suffix
+        options:
+          - name: ndots
+            value: "2"
+          - name: edns0y:
```
scholzj, line 50: Is the `y:` at the end expected? You do not have it below in the resulting configuration.

090-support-dns-config.md
```diff
@@ -0,0 +1,72 @@
...
+```
+When the Pod above is created, the container gets the following contents in its /etc/resolv.conf file:
+
+```
+nameserver 192.0.2.1
+search ns1.svc.cluster-domain.example my.dns.search.suffix
+options ndots:2 edns0
```
scholzj, line 57: Does this example actually work? Doesn't it break a MM2 deployment with multiple replicas which would communicate over Kubernetes hostnames and need to be hooked into the Kubernetes DNS?

scholzj (commented):
Some more questions / comments. But LGTM otherwise.

tombentley (approved):
LGTM, thanks @gertschouten!

090-support-dns-config.md
gertschouten: Addressed in [ca41f9f](https://github.com/strimzi/proposals/pull/142/commits/ca41f9f49af1f60f883c6c5880ea477a3de1c3e3)

090-support-dns-config.md
gertschouten: Addressed in [c478556](https://github.com/strimzi/proposals/pull/142/commits/c478556265a8dfb09b008544cf91e9ed07523df0)

090-support-dns-config.md
gertschouten: This example should work for MM2 instances that form a cluster (MM2 replicas in a deployment) as these are client to their configured source and target Kafka clusters but are not depending on direct inter-communication between each other.
This should be the case since MM2 does not share state or replicate metadata directly between instances. State for progress and coordination is stored in the source and target Kafka clusters. Coordination itself should also be handled in the communication between the MM2 instances as a clients and the brokers as its servers, rather than in-between the instances directly.

The dnsConfig and dnsPolicy settings control how the processes in the MM2 pods as clients resolve DNS names for their servers: Kafka brokers in the source cluster and Kafka brokers in the target cluster. As long as in this example, the nameserver 192.0.2.1 is able to resolve DNS queries for both the source and target cluster brokers, which in the case of the example would be external server names, resolution should be successful for these clients. The search and options configuration highlight optional DNS configuration that allows f.e. search domain expansion if needed.

MM2 instances that expose servers, e.g. REST API endpoints which need to be accessed by clients inside the same cluster (other pods or services within the Kubernetes cluster) will have DNS name resolution through Kubernetes' internal DNS system, using the service name. These are still hooked into the Kubernetes DNS and is independent of the dnsConfig and dnsPolicy set on the MM2 pods.
The resolution for those clients will depend on the specific DNS configuration set for them, which, by default, is ClusterFirst, with resolution handled by Kubernetes DNS.

ppatierno (approved):
LGTM. Thanks!

090-support-dns-config.md
scholzj: The communication between Connect / Mirror Maker 2 nodes is happening also on the REST API as there are some requests being forwarded between the nodes. I do not have it deployed anywhere right now, but IIRC it is using the Pod DNS names assigned by Kubernetes through the headless service.

scholzj: I guess there are two concerns about this ...
* We want to make sure it really solves your problem
* The example YAML is important to show the structure of the API - but if it is not easily applicable for other users, we should make there some note for exampel that _special DNS settings_ are required for this to actually keep the MM2 working or something like that to not confuse other users reading it.

gertschouten: Ah right. This part?
The REST API is used not only by users to monitor / manage Kafka Connect. In distributed mode, it is also used for the Kafka Connect cross-cluster communication. Some requests received on the follower nodes REST API will be forwarded to the leader node REST API. In case the URI under which is given host reachable is different from the URI which it listens on, the configuration options rest.advertised.host.name, rest.advertised.port and rest.advertised.listener can be used to change the URI which will be used by the follower nodes to connect with the leader. When using both HTTP and HTTPS listeners, the rest.advertised.listener option can be also used to define which listener will be used for the cross-cluster communication. When using HTTPS for communication between nodes, the same ssl.* or listeners.https options will be used to configure the HTTPS client.

scholzj: Yeah, exactly.

gertschouten: I see.
For the broker this configuration is exposed on the Kafka CRD by advertisedHostTemplate. But for KafkaConnect and MirrorMaker2 resources it gets set to the cluster-internal DNS name here, but is currently not configurable through the CRD:
[ADVERTISED_HOSTNAME](https://github.com/strimzi/strimzi-kafka-operator/blob/main/docker-images/kafka-based/kafka/scripts/kafka_connect_run.sh#L17 )
Is that correct?

gertschouten: What are your thoughts on this one Jakub:
a. Per namespace a deployment of coredns with a ClusterIP service with a chosen IP address.
b. A. gets configured to forward svc.cluster.local, cluster.local adresses to the cluster nameserver (.10) and to forward all other to the external nameserver that can resolve the brokers external to the cluster but for the respective environment.
c. The Pod template configured to use the IP assigned to the service in a. as its nameserver. I.e. utilize the scope of this proposal.

No sidecars. 
No need for routing intra-cluster traffic with help of Routes etc. 
Strimzi operator's scope limited to "just" allow for c., where the actual nameserver and its forwarding still is able to handle both resolution for cluster-internal names as well as forwarding those that are external to the external nameserver relevant for given namespace but all handled/provisioned seperately.

scholzj: I'm honestly not really a DNS expert. The DNS name we use in the Connect configuration is in a format like this `my-connect-connect-0.my-connect-connect.myproject.svc`. So I guess it should work, but not 100% sure.

As for the proposal, maybe you can just add some note saying something like that the exact configuration might depend on user's environment or something like that. So it is more important to understand if this really helps you at the end or not.

gertschouten: I successfully ran a test today.

A CoreDns deployment specific to the namespace configured to forward DNS queries for the private.confluent.cloud domain to the external nameserver (in our case a Route53 resolver on AWS) and to forward any other queries to the nameservers configured in /etc/resolv.conf (to fall back on defaults).
The deployment runs the CoreDns on a tcp and udp listener on port 5353 and a service of type ClusterIP in front of it listening on port 53.
Then a client Pod, configured like the example, pointing to the ClusterIP of the service as its nameserver.

Both the Confluent brokers external to the cluster and the cluster-internal Strimzi brokers and other services internal to the cluster got resolved successfully.

So the proposal is viable and will help us in the end: In the shared cluster a per-namespace CoreDNS deployment that resolves external brokers relevant to the specific namespace which forwards any other queries as per default configuration in its /etc/resolv.conf. Then on the Strimzi end, the ability to set such custom nameserver instead of using the default cluster-wide resolver.

I updated the example and noted that the example will depend on the user's specific environment.

PaulRMellor (approved):
Thanks for the addition of the example.

scholzj (approved):
LGTM. Thanks.

scholzj:
This has now 5 binding +1 voted. If there are no new comments, I will close the vote on Friday.

see-quick (approved):
Thanks 👍

Frawless (approved):

scholzj:
Approved with 7 binding +1 votes. Thanks for the proposal.

PR #143, opened 2024-12-19
Add proposal for integrating Bridge with Metrics Reporter

fvaleri:
Add proposal for integrating Bridge with Metrics Reporter.


090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Motivation
+
+We want to support the Strimzi Metrics Reporter as an alternative way of configuring metrics across all components.
+In order to provide a consistent user experience, the HTTP Bridge should align with all the other components in terms of metrics configuration. 
+
+## Proposal
+
+As specified in SIP-064, a new metrics config type called `strimziMetricsReporter` will be added to `metricsConfig` schema along with its configuration.
+Only the Strimzi Metrics Reporter's `prometheus.metrics.reporter.allowlist` property will be configurable by users when using the Cluster Operator.
```
ppatierno, line 39: I think it's still confusing referring to the Cluster Operator here. Can we just explain the schema referring to the bridge (maybe mentioning that the same is going to be used for the Cluster Operator)?

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,113 @@
...
+      values:
+        allowList:
+          - "kafka_log.*"
+          - "kafka_network.*"
+```
+
+### HTTP Bridge
+
+The HTTP Bridge project will be updated to support multiple types of metrics configurations.
+A new `bridge.metrics` property will be available, and will only accept `jmxPrometheusExporter` and `strimziMetricsReporter`.
```
ppatierno, line 57:
```suggestion
A new `bridge.metrics` property will be available within the application.properties configuration file and will only accept `jmxPrometheusExporter` and `strimziMetricsReporter`.
```

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,113 @@
...
+When set, the user will get a warning suggesting to use the new `bridge.metrics` configuration.
+In case they are both set, the `bridge.metrics` configuration will take precedence over `KAFKA_BRIDGE_METRICS_ENABLED`.
+
+The `MetricsReporter` class will be updated to also include a new `StrimziCollectorRegistry` that will work similarly to the `JmxCollectorRegistry`.
+The `StrimziCollectorRegistry` will include a reference to `PrometheusRegistry.defaultRegistry`, which is the same instance used by the Strimzi Metrics Reporter to collect metrics.
+
+The `kafka_bridge_config_generator.sh` script is used to generate the image configuration based on environment variables.
+This script will be updated to also include `bridge.metrics` and related configurations (see the following section).
+
+The Bridge will also try to load the JMX Exporter configuration file from the path specified by the `bridge.metrics.jmx.exporter.config.path` property.
```
ppatierno, line 81: This sounds to be an addition to the enablement of the Strimzi metrics reporter.
I mean, we could add the Strimzi metrics reporter and leaving the current JMX reporter still working with the hard-coded configuration. I am not against it but if we want to have both things we should make it clearer that this is a new addition and not needed for the Strimzi metrics reporter which is the main goal of the proposal.

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,113 @@
...
+If not specified or not found, it will fall back to the hard coded configuration.
+
+### Cluster Operator
```
ppatierno, line 84: I guess this title should reflect what the Strimzi cluster operator will do to configure the bridge. Can we use a better title?

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,113 @@
...
+
+1. `STRIMZI_METRICS`: This will contain the `metricsConfig` types to enable (one of `jmxPrometheusExporter` and `strimziMetricsReporter`).
+2. `KAFKA_BRIDGE_METRICS_JMX_CONFIG`: Used with `STRIMZI_METRICS=jmxPrometheusExporter` to pass the configuration file path.
+3. `KAFKA_BRIDGE_METRICS_SMR_CONFIG`: Used with `STRIMZI_METRICS=strimziMetricsReporter` to pass the plugin configuration.
+
+The `enableMetrics` property will be deprecated and removed in a future release.
+In case they are both set, `metricsConfig` will take precedence over `enableMetrics`.
+
+All Strimzi Metrics Reporter configurations will be locked down with the exception of `allowList`.
+Unlike the other major components, the metrics endpoint port will be set to 8080, which is the one already used to expose the REST API.
```
ppatierno, line 98: the metrics endpoint should be disabled not set to 8080 otherwise it will conflict with the bridge one, or?

090-integrate-bridge-with-metrics-reporter.md
ppatierno, line 57: I would also expect that not defining it just defaults to the current behaviour by using JMX right? Should we state it here clearly?

090-integrate-bridge-with-metrics-reporter.md
fvaleri: Initially, this wasn't part of the proposal, but then I realized that the JMX Exporter configuration is required when using the shared `metricsConfig` schema, and we have it in all the metrics examples. Additionally, I think it adds a useful feature for the standalone Bridge users. Wdyt?

ppatierno: I am fine with it, but let's make it clear it's an addition not something needed for the Strimzi metrics reporter support.

ppatierno: Anyway when you said "it's required" ... don't we have any way to fall back using the default "hard coded" configuration? I think it's mentioned somewhere else in the proposal, or?

090-integrate-bridge-with-metrics-reporter.md
fvaleri: Current default is `KAFKA_BRIDGE_METRICS_ENABLED=false`, so when `bridge.metrics` will not be set, then the metrics endpoint will be disabled. This behavior is not changing.

090-integrate-bridge-with-metrics-reporter.md
fvaleri: Yes, I just expressed this poorly. Fixing.

090-integrate-bridge-with-metrics-reporter.md
ppatierno: but we are deprecating the env var so it means that if you don't change the logic to default right now we have to do that in the future when the env var is not there anymore?

fvaleri:
@ppatierno thanks for the review. I reworked the proposal to always present Bridge changes first, and addressed the other comments.

090-integrate-bridge-with-metrics-reporter.md
fvaleri: > let's make it clear it's an addition not something needed for the Strimzi metrics reporter support

Ok, will do.

> don't we have any way to fall back using the default "hard coded" configuration? 

Correct. Let me fix.

090-integrate-bridge-with-metrics-reporter.md
ppatierno: Just to clarify what I mean ...
The implementation code will take care of KAFKA_BRIDGE_METRICS_ENABLED=false to understand that no metrics have to be exposed. But at some point the code should take care of "the bridge.metrics field is NOT set" to have the same (no metrics). So we should plan to switch to the second check when the env var is removed.

fvaleri: This is already covered by the draft implementation. Let's discuss that when we get to it.

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,117 @@
...
+This is different from how the metrics endpoint is enabled in all the other major components deployed by the Cluster Operator.
+The Cluster Operator exposes metrics through the [Prometheus JMX Exporter](https://github.com/prometheus/jmx_exporter), which can be configured using the shared `metricsConfig` schema.
+This schema has a type property that only allows the `jmxPrometheusExporter` value, and a reference to a ConfigMap containing its configuration:
+
+```sh
+metricsConfig:
+  type: jmxPrometheusExporter
+  valueFrom:
+    configMapKeyRef:
+      name: $COMPONENT_NAME-metrics
```
ppatierno, line 28: maybe a note to highlight what `$COMPONENT_NAME` means here and possible values?

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,117 @@
...
+The following example will be provided as a comment in the default application.properties file:
+
+```sh
+bridge.metrics=strimziMetricsReporter
+kafka.metric.reporters=io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter
+kafka.prometheus.metrics.reporter.listener.enable=false
+kafka.prometheus.metrics.reporter.allowlist=.*
+```
+
+The `KAFKA_BRIDGE_METRICS_ENABLED` environment variable will be deprecated and removed in a future release.
```
ppatierno, line 61: should we make a clearer plan. I.e. deprecated in 0.32.0 (next release when we have the Strimzi metrics reporter) and removed in 0.33.0 or 0.34.0? Of course the releases can be shifted if we don't get this proposal implemented on time for the 0.32.0 release.

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,117 @@
+# Integrate Bridge with Metrics Reporter
+
+In [SIP-064](https://github.com/strimzi/proposals/blob/main/064-prometheus-metrics-reporter.md) we introduced the [Strimzi Metrics Reporter](https://github.com/strimzi/metrics-reporter).
+This is a Kafka MetricsReporter plugin that directly exposes metrics in Prometheus format via an HTTP endpoint.
+
+## Current situation
+
+The HTTP Bridge allows users to enable or disable the metrics endpoint using the `KAFKA_BRIDGE_METRICS_ENABLED` environment variable.
+When this variable is set to true, the Bridge creates the JMX Exporter's JmxCollector class with hard coded configuration.
```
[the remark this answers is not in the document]
fvaleri: Sure.

090-integrate-bridge-with-metrics-reporter.md
fvaleri: Yes, this is confusing. Let me fix it.

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,117 @@
...
+metricsConfig:
+  type: jmxPrometheusExporter
+  valueFrom:
+    configMapKeyRef:
+      name: $COMPONENT_NAME-metrics
+      key: metrics-config.yml
+```
+
+At runtime, the above configuration enables a Java agent which exposes Kafka JMX metrics in Prometheus format through an HTTP endpoint on port 9404.
+Note that this agent depends on the [Kafka JMX Reporter](https://github.com/apache/kafka/blob/3.9.0/clients/src/main/java/org/apache/kafka/common/metrics/JmxReporter.java) plugin, which is enabled by default.
```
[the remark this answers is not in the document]
fvaleri: "Requires" should be better.

090-integrate-bridge-with-metrics-reporter.md
fvaleri: Ok, I will propose a schedule so that we can discuss.

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,117 @@
...
+The StrimziCollectorRegistry will include a reference to PrometheusRegistry.defaultRegistry, which is the same instance used by the Strimzi Metrics Reporter to collect metrics.
+
+The kafka_bridge_config_generator.sh script is used to generate the image configuration based on environment variables.
+This script will be also updated to include `bridge.metrics` and related configurations (see the following section).
+
+The Bridge will try to load the JMX Prometheus Exporter configuration file from the path specified by the `bridge.metrics.jmx.exporter.config.path` property.
+If the property is not specified or the file is not found, the Bridge will fall back to the hard coded configuration.
+This feature is not strictly required to support the Strimzi Metrics Reporter, but will be used by the Cluster Operator.
+
+### Cluster Operator integration
```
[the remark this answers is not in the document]
fvaleri: I will add a dedicated section. For `auto.include.jmx.reporter` there is no mention in the Strimzi Metrics Reporter proposal, so we will leave that to the user. Additionally, that property will be removed in Kafka 4.0, and the user would need to be explicitly add the JMX metrics plugin when setting any custom plugin.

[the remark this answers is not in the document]
ppatierno: But we could end in a Strimzi version with Kafka 3.9 support, so the `auto.include.jmx.reporter` would be still needed, or?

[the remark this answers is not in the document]
fvaleri: Given it's deprecated and not required to enable the Strimzi Metrics Reporter, I would leave this configuration to the user through `.spec.kafka.config`. Wdyt?

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,121 @@
...
+kafka.prometheus.metrics.reporter.allowlist=.*
+```
+
+The MetricsReporter class will be updated to also include a new StrimziCollectorRegistry that will work similarly to the JmxCollectorRegistry.
+The StrimziCollectorRegistry will include a reference to PrometheusRegistry.defaultRegistry, which is the same instance used by the *Reporter* to collect metrics.
+
+The kafka_bridge_config_generator.sh script is used to generate the image configuration based on environment variables.
+This script will be also updated to include `bridge.metrics` and related configurations (see the following section).
+
+The Bridge will try to load the Exporter's configuration file from the path specified by the `bridge.metrics.jmx.exporter.config.path` property.
```
PaulRMellor, line 69:
```suggestion
The Bridge will try to load the configuration file of the *Exporter* from the path specified by the `bridge.metrics.jmx.exporter.config.path` property.
```

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,121 @@
...
+If the property is not specified or the file is not found, the Bridge will fall back to the hard coded configuration.
```
PaulRMellor, line 70:
```suggestion
If the property is not specified or the file is not found, the Bridge will fall back to the hard-coded configuration.
```

090-integrate-bridge-with-metrics-reporter.md
PaulRMellor: I'm not clear on this property and the context here? How is it specified and under what circumstances? Is it specified in application.properties if we choose jmxPrometheusExporter?

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,121 @@
...
+This feature is not strictly required to support the *Reporter*, but will be used by the Cluster Operator.
+
+### Cluster Operator integration
+
+The KafkaBridge's CRD will also support `metricsConfig` with the addition of `strimziMetricsReporter` type.
```
PaulRMellor, line 75:
```suggestion
The `KafkaBridge` CRD will also support `metricsConfig` with the addition of `strimziMetricsReporter` type.
```

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,121 @@
...
+This is how the Strimzi Reporter configuration will look like:
```
PaulRMellor, line 76:
```suggestion
This is how the Strimzi *Reporter* configuration will look like:
```

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,121 @@
...
+Three new environment variables will be introduced to pass the metrics configuration to the Bridge's container:
+
+- `STRIMZI_METRICS`: This will contain the `metricsConfig` types to enable (one of `jmxPrometheusExporter` and `strimziMetricsReporter`).
+- `KAFKA_BRIDGE_METRICS_JMX_CONFIG`: Used with *Exporter* to pass the configuration file path.
+- `KAFKA_BRIDGE_METRICS_SMR_CONFIG`: Used with *Reporter* to pass the plugin configuration.
+
+The *Exporter* configuration file will be stored in a ConfigMap and mounted in the Bridge's container.
+The full configuration file passed to the Bridge's container will be `/opt/strimzi/custom-config/metrics-config.yml`.
+
+Metrics will be exposed through the Bridge's HTTP server, so the *Reporter*'s listener will be disabled.
```
PaulRMellor, line 97:
```suggestion
Metrics will be exposed through the Bridge's HTTP server, so the listener of the *Reporter* will be disabled.
```

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Bridge
+
+The `KAFKA_BRIDGE_METRICS_ENABLED` environment variable will be deprecated in 0.32.0 and removed in 0.34.0.
+When set, the user will get a warning suggesting to use the `bridge.metrics` property.
+In case they are both set, `bridge.metrics` will take precedence over `KAFKA_BRIDGE_METRICS_ENABLED`.
+
+### Cluster Operator
+
+The `enableMetrics` property in KafkaBridge CRD will be deprecated in 0.46.0 and removed in 0.48.0.
```
PaulRMellor, line 114:
```suggestion
The `enableMetrics` property in `KafkaBridge` CRD will be deprecated in 0.46.0 and removed in 0.48.0.
```

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,121 @@
...
+When set, the user will get a warning suggesting to use the `metricsConfig` configuration.
+In case they are both set, `metricsConfig` will take precedence over `enableMetrics`.
+
+## Rejected alternatives
+
+Lock down and automate the *Reporter* configurations for the standalone Bridge.
+This was rejected because the Bridge allows user to customize Kafka clients and plugins using the application.properties file, which includes commented examples.
```
PaulRMellor, line 121:
```suggestion
This was rejected because the Bridge allows users to customize Kafka clients and plugins using the application.properties file, which includes commented examples.
```

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,121 @@
...
+The following example will be documented and also provided as a comment in the default application.properties file:
+
+```sh
+bridge.metrics=strimziMetricsReporter
+kafka.metric.reporters=io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter
+kafka.prometheus.metrics.reporter.listener.enable=false
+kafka.prometheus.metrics.reporter.allowlist=.*
+```
+
+The MetricsReporter class will be updated to also include a new StrimziCollectorRegistry that will work similarly to the JmxCollectorRegistry.
```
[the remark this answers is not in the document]
fvaleri: Sure. Creating a collector registry for each metrics type ensure better isolation, and makes it easier to add or remove metrics types.

090-integrate-bridge-with-metrics-reporter.md
fvaleri: This configuration property will be used by the Cluster Operator to pass the user-provided Exporter's configuration file to the Bridge (see `metricsConfig` schema). The same property can also be leveraged by standalone Bridge users.

fvaleri:
@ppatierno and @PaulRMellor your comments should be addressed, but let me know if you have any other concern. It would also be good to get some more feedback from other maintainers. Thanks.

PaulRMellor (approved):
Updates look good :+1:

ppatierno (approved):
LGTM.

im-konge (approved):

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,133 @@
...
+The following example will be documented and included, with comments, in the default `application.properties` file.
+The `kafka.` prefix is used by the Bridge to pass the following configuration to the internal Kafka clients.
+
+```sh
+# uncomment the following lines to enable Strimzi Reporter metrics, check the documentation for more details
+bridge.metrics=strimziMetricsReporter #1
+kafka.metric.reporters=io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter #2
+kafka.prometheus.metrics.reporter.listener.enable=false #3
+kafka.prometheus.metrics.reporter.allowlist=.* #4
+```
```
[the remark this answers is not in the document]
ppatierno: I guess this part of the properties file is an example when the bridge is used on bare metal/vm. In such a case, it's the user to provide the entire configuration (their own application.properties file). The bridge itself doesn't never change the config file on startup to add "something" to it.
One thing we could imagine is that the bridge doesn't update the file but add these value at runtime (in memory) when configuring the internal Kafka clients. My only doubt is that it won't be reflected in what the user configured so the user doesn't know about that.

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,133 @@
...
+2. Kafka configuration used to set the Kafka metrics reporter implementation.
+3. *Reporter* configuration used to disable the HTTP listener (metrics are exposed through the Kafka Bridge listener).
+4. *Reporter* configuration used to configure a comma separated list of regex patterns to filter metrics to collect.
+
+The MetricsReporter class will be updated to also include a new `StrimziCollectorRegistry` that will work similarly to the `JmxCollectorRegistry`.
+Creating a collector registry for each metrics type ensure better isolation, and makes it easier to add or remove metrics types.
+The `StrimziCollectorRegistry` will include a reference to `PrometheusRegistry.defaultRegistry`, which is the same instance used by the *Reporter* to collect metrics.
+
+The `kafka_bridge_config_generator.sh` script is used to generate the Kafka Bridge image configuration based on environment variables.
+This script will be also updated to include `bridge.metrics` and related configurations, as described in the following integration section.
```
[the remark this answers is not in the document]
ppatierno: Yeah. This proposal was written down before I started the work on moving bridge configuration. I think there is some changes needed here @fvaleri based on that PR.

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,133 @@
+# Integrate Kafka Bridge with Metrics Reporter
+
+In [SIP-064](https://github.com/strimzi/proposals/blob/main/064-prometheus-metrics-reporter.md), we introduced the [Strimzi Metrics Reporter](https://github.com/strimzi/metrics-reporter).
```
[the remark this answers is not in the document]
tinaselenge: can we use just use `Strimzi Proposal 064`?

090-integrate-bridge-with-metrics-reporter.md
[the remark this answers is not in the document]
fvaleri: We can use the same logic that we have in the Operator, with locked down Reporter configuration and a new `bridge.metrics.allowList` property.

090-integrate-bridge-with-metrics-reporter.md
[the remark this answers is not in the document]
fvaleri: Ok, but I will wait for that PR to settle down.

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,133 @@
...
+This is how the *Reporter* configuration will look like:
+
+```sh
+spec:
+  metricsConfig:
+    type: strimziMetricsReporter
+    values:
+      allowList:
+        - "kafka_log.*"
+        - "kafka_network.*"
```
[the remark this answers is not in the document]
fvaleri: Yes, the default is `.*`. Example updated.

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Affected/not affected projects
+
+The affected projects are Cluster Operator and Kafka Bridge.
+
+## Compatibility
+
+### Bridge
+
+The `KAFKA_BRIDGE_METRICS_ENABLED` environment variable will be deprecated in 0.32.0 and removed in 0.34.0.
```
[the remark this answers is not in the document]
fvaleri: We usually use versions, but I'm fine with dates if you think it's better. Do you mean to deprecate next release and remove in January 2026?

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,133 @@
...
+When set, the user will get a warning suggesting to use the `bridge.metrics` property.
+In case they are both set, `bridge.metrics` will take precedence over `KAFKA_BRIDGE_METRICS_ENABLED`.
+
+### Cluster Operator
+
+The `enableMetrics` property in `KafkaBridge` CRD will be deprecated in 0.46.0 and removed in 0.48.0.
```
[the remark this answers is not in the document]
fvaleri: > I guess we can also keep it working until removed in v1 API easily to simplify it for the users.

Ok, there is no reason to remove it before v1. Updated.

> explained how would the migration work as well

What do you mean exactly with migration?

090-integrate-bridge-with-metrics-reporter.md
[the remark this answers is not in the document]
scholzj: Maybe I missed some of the arguments you are trying to make and maybe my comment was not clear, sorry. It of course does not matter when the operator configures it. But when I configure it myself, why do I need to set both
```
bridge.metrics=strimziMetricsReporter #1
kafka.metric.reporters=io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter #2
```

They seem to do the same - enable the Strimzi metrics reporter, or? So what is the difference between them?

090-integrate-bridge-with-metrics-reporter.md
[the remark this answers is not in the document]
scholzj: Yeah, I think that makes sense.

090-integrate-bridge-with-metrics-reporter.md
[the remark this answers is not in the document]
scholzj: That is not really what I meant. And it relates to the PR from @OwenCorrigan76 as well. I think we should choose one of the ways here:
1. Make the allowList required
2. Have a good default for them for each component. (And I guess `.*` is not the right default) This would be hardcoded in the code base and change as we for example update our dashabords.

Having a flat default `.*` everywhere is IMHO a bad option, because historically our experience is that users want to limit the metrics exporter:
* For performance reasons in Kafka (this might improve with our own reporter, but I expect it would still be an issue at some amount of metrics)
* For performance reasons in Prometheus (the metrics can be dropped by Prometheus after scraping, but that is inefficient and harder to configure)

090-integrate-bridge-with-metrics-reporter.md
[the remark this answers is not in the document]
scholzj: We do - but the Bridge minor releases do not have the same set cadence as the operator. The BRidge already used time based definition for example in https://github.com/strimzi/proposals/blob/main/071-deprecate-bridge-openapi-2.md

090-integrate-bridge-with-metrics-reporter.md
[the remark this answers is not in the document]
scholzj: My expectation is that `enableMetrics` will be replaced by:
```yaml
  metricsConfig:
    type: jmxPrometheusExporter
```

But that is missing from this proposal and it is not clear if that is intentional or unintentional.

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,125 @@
...
+The Kafka Bridge will try to load a custom Exporter's configuration file from the path specified by the `bridge.metrics.exporter.config.path` property.
+If the property is not specified or the file is not found, the Kafka Bridge will fall back to the hard-coded configuration file.
+This feature is not strictly required to support the *Reporter*, but will be used by the Cluster Operator.
+
+### Cluster Operator integration
+
+The `KafkaBridge` CRD will also support `metricsConfig` with the addition of `strimziMetricsReporter` type.
+This is how the *Reporter* configuration will look like:
+
+```sh
```
scholzj, line 87:
```suggestion
```yaml
```

090-integrate-bridge-with-metrics-reporter.md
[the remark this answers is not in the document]
fvaleri: The `bridge.metrics` property is the analogous of the `metricsConfig.type` field we have in the CO schemas. Maybe the property name is confusing, but it was inspired by the already existing `bridge.tracing`. It will be used to initialize the Bridge's `MetricsRporter` with the right `CollectorRegistry`.

The `kafka.metric.reporters` property is basically the Kafka `metric.reporters` with the `kafka.` prefix used by the Bridge to group broker configurations.

090-integrate-bridge-with-metrics-reporter.md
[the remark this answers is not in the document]
fvaleri: Ok, makes sense. Will update the proposal.

090-integrate-bridge-with-metrics-reporter.md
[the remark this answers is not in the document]
fvaleri: Yes, I will make it explicit.

090-integrate-bridge-with-metrics-reporter.md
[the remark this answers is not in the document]
scholzj: But why do we need the user to configure both? You say it is `analogous of the metricsConfig.type field` -> but there you configure only the `metricsConfig.type` field.

090-integrate-bridge-with-metrics-reporter.md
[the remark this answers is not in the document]
fvaleri: Bridge defaults added. We will address the other components in the related PRs.

fvaleri:
@scholzj I think I addressed all your comments, let me know what you think. Thanks.

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Cluster Operator
+
+The `enableMetrics` property in `KafkaBridge` CRD will be deprecated in the next release and removed with the release of Strimzi v1 API.
+When set, the user will get a warning suggesting to use the `metricsConfig` configuration.
+In case they are both set, `metricsConfig` will take precedence over `enableMetrics`.
+
+#### Migration
+
+When `enableMetrics` will be removed, *Exporter* users will have to create a secret with its configuration, and reference the secret as shown in the following example:
```
scholzj, line 132:
```suggestion
When `enableMetrics` will be removed, *Exporter* users will have to create a config map with its configuration, and reference it as shown in the following example:
```

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,149 @@
...
+spec:
+  metricsConfig:
+    type: jmxPrometheusExporter
+    valueFrom:
+      configMapKeyRef:
+        name: my-metrics
+        key: my-metrics-config.yml
+```
+
+A full example that include the *Exporter* configuration secret will be provided.
```
scholzj, line 144:
```suggestion
A full example that include the *Exporter* configuration config map will be provided.
```

scholzj (approved):
Two nits. But LGTM otherwise. Thanks.

090-integrate-bridge-with-metrics-reporter.md
```diff
@@ -0,0 +1,149 @@
...
+When set, the user will get a warning suggesting to use the `bridge.metrics` property.
+In case they are both set, `bridge.metrics` will take precedence over `KAFKA_BRIDGE_METRICS_ENABLED`.
+
+#### Migration
+
+When `KAFKA_BRIDGE_METRICS_ENABLED` will be removed, *Exporter* users will have to set `bridge.metrics=jmxPrometheusExporter` in their `application.properties`.
+
+### Cluster Operator
+
+The `enableMetrics` property in `KafkaBridge` CRD will be deprecated in the next release and removed with the release of Strimzi v1 API.
```
katheris, line 126: If this proposal is accepted we should make sure to raise an issue to remove the `enableMetrics` property when we release the Strimzi v1 API so this isn't forgotten

fvaleri: The `enableMetrics` property will be marked as deprecated, so I guess it will be automatically eligible for removal in v1.

scholzj: Yeah - whoever works on the `v1` API would be expected to remove all the deprecated APIs.

see-quick (approved):
LGTM, Thanks

ppatierno:
This proposal has now 6 +1 binding votes. I will leave this open until the end of today for further comments and then merge.

Frawless (approved):

tinaselenge (approved):

PR #147, opened 2025-01-27
Proposal for splitting metrics-reporter into modules

mimaison:
This proposes splitting the metrics-reporter project into client and server modules so it's easier to evolve in the future.


093-split-metrics-reporter-into-modules.md
```diff
@@ -0,0 +1,51 @@
...
+
+I'm also considering adding support for [KIP-714](https://cwiki.apache.org/confluence/display/KAFKA/KIP-714%3A+Client+metrics+and+observability) to metrics-reporter. Again this feature is only implemented by server-side metrics reporters. Adding support for this feature will be addressed in a separate proposal.
+
+## Proposal
+
+I propose splitting the project into 2 Java modules:
+
+1. `client-metrics-reporter`: This module will only depend on kafka-clients and the Prometheus libraries. It will be used by Apache Kafka clients (Producer, Admin, Consumer, Connect and Streams) by setting the `metric.reporters` configuration to `io.strimzi.kafka.metrics.ClientMetricsReporter`. All the existing metrics-reporter configurations will stay the same. The differences are:
```
fvaleri, line 21: Why not `KafkaClientMetricsReporter`, `KafkaServerMetricsReporter` and `YammerServerMetricsReporter`? It just reads better.

fvaleri (commented):
Thanks @mimaison.  I don't see a huge impact on the ongoing efforts, assuming this proposal is approved and we will have a release soon after. Can you clarify this in the proposal?

093-split-metrics-reporter-into-modules.md
```diff
@@ -0,0 +1,51 @@
...
+
+Another benefit of having separate modules is the ability to enforce dependencies. Depending on where the reporter runs, a Kafka client or a Kafka server, different dependencies are available at runtime. We need to make sure that logic used by clients does not load any classes only available on servers. We hit this issue in the past ([issue #48](https://github.com/strimzi/metrics-reporter/issues/48)). This is not something that can easily be tested so currently it relies on maintainers being diligent when making code changes to ensure we don't hit this issue again in the future.
+
+I'm also considering adding support for [KIP-714](https://cwiki.apache.org/confluence/display/KAFKA/KIP-714%3A+Client+metrics+and+observability) to metrics-reporter. Again this feature is only implemented by server-side metrics reporters. Adding support for this feature will be addressed in a separate proposal.
```
scholzj, line 15: This is a bit confusing for me. Shouldn't this be client-side only as the clients send the metrics through this KIP? Or maybe I misunderstood things? This might anyway deserve clarification (or removal as I think you have enough justification even without it 😉).

093-split-metrics-reporter-into-modules.md
```diff
@@ -0,0 +1,51 @@
...
+   - the reporter class name (it used to be `io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter`)
+   - the dependency to add to the classpath
+   
+   This reporter will use its metric context to validate it runs in a client and will fail at startup if it's not the case.
+
+2. `server-metrics-reporter`: This module will depend on the client-metrics-reporter module and also on Apache Kafka server JARs required to capture Yammer metrics (as described in [Proposal #64](https://github.com/strimzi/proposals/blob/main/064-prometheus-metrics-reporter.md)). It can be used by Apache Kafka servers (brokers and controllers) by setting the `metric.reporters` configuration to `io.strimzi.kafka.metrics.ServerKafkaMetricsReporter` and the `kafka.metrics.reporters` configuration to `io.strimzi.kafka.metrics.ServerYammerMetricsReporter`. All the existing metrics-reporter configurations will stay the same. The differences are:
```
scholzj, line 27: So, this would essentially provide the client metrics (AFAIK they matter for the replication clients etc.) + additional server metrics. did I got it right?

093-split-metrics-reporter-into-modules.md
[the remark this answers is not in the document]
mimaison: The Connect runtime and Streams are considered clients. To collect their metrics you would still follow the instructions from https://github.com/strimzi/metrics-reporter?tab=readme-ov-file#kafka-connect-and-kafka-streams but use `io.strimzi.kafka.metrics.ClientMetricsReporter` as the class name.

093-split-metrics-reporter-into-modules.md
mimaison: I'm not sure I understand the question. Can you clarify what you mean?

093-split-metrics-reporter-into-modules.md
scholzj: So, in order to support the KIP-714, the server module would in theory read the metrics from the clients and expose them as Prometheus metrics in the Kafka broker?

mimaison: Yes exactly.

scholzj: Ahh, ok. That makes sense (or it actually does not in the bigger picture, but I get the context for the module split now and understand it :-D).

093-split-metrics-reporter-into-modules.md
scholzj: What I'm wondering about - and maybe my understanding is wrong - the metrics exposed from the broker include some client metrics as well as the replication is done internally by a Kafka client?

093-split-metrics-reporter-into-modules.md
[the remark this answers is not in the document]
scholzj: So essentially, when used in Connect, the client reporter will automatically handle the additional Connect metrics as well?

scholzj: Maybe we can add `prometheus` to the package name? That would keep  E.g. `io.strimzi.kafka.metrics.prometheus.ClientMetricsReporter`. (I suspect we might need other technologies as well one (maybe not so distant) day :-/)

093-split-metrics-reporter-into-modules.md
mimaison: > the replication is done internally by a Kafka client

No, that's not quite right, replication is not really done by internal Kafka clients. Brokers share some metrics with clients but I consider all broker metrics to be that _broker metrics_, not a set of client + broker metrics.

The new `io.strimzi.kafka.metrics.ServerKafkaMetricsReporter` would basically collect all the metrics that at the moment are collected by `io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter` when it runs in a broker.

Right now, `io.strimzi.kafka.metrics.ServerKafkaMetricsReporter` would not do anything more than `io.strimzi.kafka.metrics.ClientMetricsReporter`. It would just extend it and not override any method (as today `io.strimzi.kafka.metrics.ServerKafkaMetricsReporter` can run on both clients and servers). Then when we start adding new features, for example [support for dynamic configurations](https://github.com/strimzi/metrics-reporter/issues/55), then `io.strimzi.kafka.metrics.ServerKafkaMetricsReporter` would have extra logic unique to the server side, but still handle all the broker metrics.

093-split-metrics-reporter-into-modules.md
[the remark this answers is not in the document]
mimaison: Yes, like `io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter` does at the moment.

fvaleri: > we might need other technologies as well

In that case, I would expect a config parameter to configure the output format with a default value set to Prometheus for backwards compatibility.

mimaison: If we are to add extra reporters in the future, I expect users would pick the format by using different classes. For example if you use `io.strimzi.kafka.metrics.prometheus.ClientMetricsReporter` the export format is Prometheus. So I wouldn't add a new configuration to the reporter itself.

093-split-metrics-reporter-into-modules.md
scholzj: Ok, thanks for the explanation.

093-split-metrics-reporter-into-modules.md
fvaleri: Ok, makes sense. Didn't know Prom had push-based model.

scholzj: I don't think it is Prometheus itself - it would be more in the OpenTelemetry realm. But I think the push-based metrics are becoming more common.

ppatierno: AFAIK, the is Prometheus Agent which is able to use the push-based model so writing metrics remotely as well.
Regarding the naming I am fine with `ClientMetricsReporter` but maybe within a `prometheus` package as suggested. It would be clearly Kafka even because the FQDM has `kafka` in the package as well.

093-split-metrics-reporter-into-modules.md
```diff
@@ -0,0 +1,49 @@
...
+
+The project will publish 2 artifacts, one per module. The build will also produce archives including all the dependencies:
+- The client-metrics-reporter archive will contain the client-metrics-reporter JAR and all its dependencies
+- The server-metrics-reporter archive will contain the server-metrics-reporter JAR and all its dependencies (including the client-metrics-reporter JAR)
+## Affected/not affected projects
+
+In addition of `metrics-reporter`, this will also impact:
+
+- `strimzi-kafka-operator`: There is a PR ongoing to add support for the metrics-reporter. This will need to be updated to use the new class names and dependency.
+- `strimzi-kafka-bridge`: The proposal to add support for the metrics-reporter is not impacted, but the implementation will need to be updated to use the new class names and dependency.
```
ppatierno, line 39: @fvaleri what should we do with https://github.com/strimzi/strimzi-kafka-bridge/pull/954 then? Should we wait for this proposal being accepted, implemented and metrics reporter released? Or @mimaison this is going to take more time so it's bettere to proceed with a first metrics reporter integration within the bridge?

mimaison: If you are close to merge the PR, then you go ahead.
This proposal is not accepted yet, then we need to merge the code, and do a release.

093-split-metrics-reporter-into-modules.md
mimaison: I like the suggestion of having a `prometheus` package. If we have consensus on that point, I'll make the change.

tombentley (approved):
LGTM, thanks @mimaison

093-split-metrics-reporter-into-modules.md
mimaison: I've pushed an update changing the packages to `io.strimzi.kafka.metrics.prometheus`.

scholzj (approved):
Thanks @mimaison. +1 binding.

ppatierno (approved):
LGTM. Thanks Mickael!
When this proposal is approved and merged we should try to have the implementation and the release done before a new operator and bridge release so that we don't need to make breaking changes as described in your compatibility section imho.

mimaison:
I have most of the code ready, so I should be able to open a PR next week.

im-konge (approved):

fvaleri (approved):
Thanks.

Frawless (approved):

093-split-metrics-reporter-into-modules.md
ppatierno: I know this could be a breaking change compared to the 0.1.0 but I think it's better doing it now.
+1 from me to change the modules and artifacts names to include prometheues naming as well. @tombentley @scholzj @Frawless @fvaleri @im-konge are you ok with that? (tagging you because already approved the proposal).

scholzj: You should probably clarify what exactly are you talking about.

ppatierno: I assume that @mimaison metns that instead of having `io.strimzi.metrics-reporter-0.1.0.jar` as today, we would have `io.strimzi.prometheus.metrics-reporter-0.2.0.jar` instead. In the future, a different implementation would have a separate artifact with a different name (replacing "prometheus" with something else). This way if you want to use just the prometheus one you can get only the specific dependency and not bringing with you stuff which is unrelated but belongs to a different implementation.

scholzj: In any case if you decide to include Prometheus in the module name, it should absolutely not be in the group ID but only in the artifact ID.

im-konge: I guess that nobody knows how many dependencies would be useful or useless in different implementations. Also in what point in the future should we add another implementation? Having the name in the artifacts and modules now would look like there is another JAR containing different implementation (even if there is just Prometheus one).

I'm fine with adding `prometheus` package. For changing the names of the module and artifact - it could be changed in the future when another implementations will be added and the split will be needed, right?

PaulRMellor (approved):
Motivation and proposal are clear and well-justified

scholzj:
This has now 6 binding and one nob-binding +1 votes. If there are no new comments by tomorrow, I will proceed and merge this as approved.

scholzj:

PR #149, opened 2025-02-14
Add support for VolumeAttributeClass

venkatesh2090:
Proposal to add support for `VolumeAttributeClasses`.


095-add-support-volumeattributesclassname.md
```diff
@@ -0,0 +1,105 @@
+<!-- This template is provided as an example with sections you may wish to comment on with respect to your proposal. Add or remove sections as required to best articulate the proposal. -->
+
```
scholzj, line 2: I guess you can delete the comment and the empty line.

095-add-support-volumeattributesclassname.md
```diff
@@ -0,0 +1,105 @@
...
+# Add `volumeAttributesClassName` to the storage configuration
+
+Add the ability to set `volumeAttributesClassName` in the configuration for `PersistentClaimStorage`.
+`VolumeAttributesClasses` provide the ability to decouple storage parameters like
+IOPS, throughput, fstype or any other cloud specific ones from the `StorageClass`.
+
+
```
scholzj, line 9:
```suggestion

```

095-add-support-volumeattributesclassname.md
```diff
@@ -0,0 +1,105 @@
...
+      cpu: 1
+      memory: 512Gi
+    limits:
+      cpu: 500m
+      memory: 256Gi
+  storage:
+    type: persistent-claim
+    size: 16Gi
+    class: ebs
+    volumeAttributesClassName: ebs-fast
```
scholzj, line 62: Givin Storage Class name is just `class`, I wonder if we should name this field `volumeAttributesClass` only (without the `Name` at the end).

095-add-support-volumeattributesclassname.md
```diff
@@ -0,0 +1,105 @@
...
+
+
+## Current situation
+
+It is  not possible to set the `volumeAttributesClassName` using [`PersistentClaimStorage`](https://github.com/strimzi/strimzi-kafka-operator/blob/c1b20f726dddbcd2a070c2eeb14fd30902027aec/api/src/main/java/io/strimzi/api/kafka/model/kafka/PersistentClaimStorage.java).
+This is currently being managed using various annotations
+
+## Motivation
+
+Kubernetes v1.31 added a new method of configuring storage parameters for `PersistentVolumes` (PV)
```
scholzj, line 17: What happens on Kube 1.31 while this feature is disabled or on older Kubernetes versions:
* When the field is not set -> I expect nothing will happen and everything will work as before?
* When the field is set -> Does it throw an error? Is it just silently ignored?

scholzj (commented):
Thanks for the proposal. I think it looks mostly good. I just left few comments.

095-add-support-volumeattributesclassname.md
venkatesh2090: > When the field is not set -> I expect nothing will happen and everything will work as before?

Yes, it just functions like before.

> When the field is set -> Does it throw an error? Is it just silently ignored?

It ignores the field. Looks like when I dump the yaml, the PVC doesn't even have that field.

095-add-support-volumeattributesclassname.md
venkatesh2090: Sure will change it :+1: 

I was thinking might be easier to correlate with a PVC if the fields had the same name. But since the `class` field is already not 1-1, I can agree as well.

095-add-support-volumeattributesclassname.md
```diff
@@ -0,0 +1,102 @@
+# Add `volumeAttributesClassName` to the storage configuration
+
+Add the ability to set `volumeAttributesClassName` in the configuration for `PersistentClaimStorage`.
```
ppatierno, line 3: Someone reading could not understand what `PersistentVolumeClaim` is here. Of course, it's part of the Strimzi api, it's one of the model classes but need to be clarified.

095-add-support-volumeattributesclassname.md
```diff
@@ -0,0 +1,102 @@
...
+`VolumeAttributesClasses` provide the ability to decouple storage parameters like
+IOPS, throughput, fstype or any other cloud specific ones from the `StorageClass`.
+
+## Current situation
+
+It is  not possible to set the `volumeAttributesClassName` using [`PersistentClaimStorage`](https://github.com/strimzi/strimzi-kafka-operator/blob/c1b20f726dddbcd2a070c2eeb14fd30902027aec/api/src/main/java/io/strimzi/api/kafka/model/kafka/PersistentClaimStorage.java).
+This is currently being managed using various annotations
```
ppatierno, line 10: What does this sentence mean exactly?
"This" is referring to what?
"various annotations" is ambiguos.

095-add-support-volumeattributesclassname.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Motivation
+
+Kubernetes v1.31 added a new method of configuring storage parameters for `PersistentVolumes` (PV)
+using [`VolumeAttributesClass`](https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/) (VAC).
+`PersistentVolumeClaims` (PVC) can then refer to them using `volumeAttribtesClassName` along with their
```
ppatierno, line 16:
```suggestion
`PersistentVolumeClaims` (PVC) can then refer to them using `volumeAttributesClassName` along with their
```

095-add-support-volumeattributesclassname.md
```diff
@@ -0,0 +1,102 @@
...
+To accommodate this change, the `PersistentClaimStorage` (PCS) API needs an additional field `volumeAttributesClass`.
+When this field changes in the PCS, the Cluster Operator (CO) can map it to the generated PVC's
+`volumeAttributesClassName` and let the CSI Driver take care of provisioning. There shouldn't be
+a need to validate the VAC since the parameter names depend on the cloud provider.
+
+The [`external-provisioner`](https://github.com/kubernetes-csi/external-provisioner) `csi-provisioner`
+usually checks whether the `driverName` in the VAC is the same as the `provisioner` in the SC.
+The operator could check that pre-emptively, but cannot depend on the CSI driver to use that implementation
+of a csi-provisioner. The user would be responsible for making sure the VACs and the SCs are configured
+correctly before configuring the PCS.
```
ppatierno, line 33: What exactly happens if the user doesn't make sure for a proper configuration?

095-add-support-volumeattributesclassname.md
```diff
@@ -0,0 +1,102 @@
...
+  storageClassName: ebs
+  volumeAttributesClassName: ebs-fast
+```
+
+Here the user is expected to have a VAC named ebs-fast which has the same `driverName` as the
+`provisioner` of the SC named ebs. They are also expected to have the right parameters for
+the corresponding cloud provider's CSI driver.
+
+When a user decides to update the `volumeAttributesClassName` of a PVC, the CSI driver will
+apply the changes. This might result in a cloud provider error being thrown as well. In that
```
ppatierno, line 86: so does the operator report the error within the `Kafka` status? how does the user notice the mistake in order to fix it?

095-add-support-volumeattributesclassname.md
fvaleri: I guess the error will surface to the Kafka status. An example would help.

095-add-support-volumeattributesclassname.md
fvaleri: Can you provide an example?

fvaleri (commented):
Hi @venkatesh2090. The overall approach LGTM and I think this is a useful feature.

095-add-support-volumeattributesclassname.md
venkatesh2090: Looks like I was incorrect here. I was just referring to the description in strimzi/strimzi-kafka-operator#10462 that requested this proposal. I'll update the document.

095-add-support-volumeattributesclassname.md
venkatesh2090: The error is posted on the PVC's status. The user is expected to check the PVC if the pod isn't being scheduled properly. Making it appear in Kafka would mean we are making redundant checks, making sure the PVC is in the ready status (which I'm not sure is currently being done or not). Will update the doc with this info

I have added an example at the end for an AWS EBS backed volume.

095-add-support-volumeattributesclassname.md
venkatesh2090: The example is at the end. I have copied it from the PVC's status. Will update the doc with this info

ppatierno:
@venkatesh2090 in order to make the review easier, can you split one sentence per line please? also avoiding to split the sentences in the middle and to a new line, thanks.

095-add-support-volumeattributesclassname.md
ppatierno: You are actually putting a WARN example which is not clear to me where it is visible, some logs? somewhere? On the PVC describe? We were wondering if the user should have a feedback on the Kafka custom resource status that something wrong happened.

095-add-support-volumeattributesclassname.md
```diff
@@ -0,0 +1,107 @@
+# Add `volumeAttributesClassName` to the storage configuration
+
+Volumes in Strimzi are configured using the `PersistentClaimStorage` model. This is part
+of the Strimzi API's `Kafka` and `KafkaNodePool` among others. This proposal goes over the
+ability to set `volumeAttributesClassName` in the configuration for `PersistentClaimStorage`.
+`VolumeAttributesClasses` provide the ability to decouple storage parameters like
```
ppatierno, line 6:
```suggestion
`VolumeAttributesClasse` provides the ability to decouple storage parameters like
```

095-add-support-volumeattributesclassname.md
ppatierno: Usually the user doesn't need to take a look at the operands (like PVCs in this case) or at least they can get details from there if they want but having a central point for warnings/errors which is the `Kafka` custom resource is much better.

scholzj: I would avoid any further processing as the exact errors might differ depending on the driver etc. I think this is pretty advanced stuff and the users using this should be able to monitor their PVCs etc.

scholzj (approved):
LGTM. Thanks for the proposal. +1 (binding).

095-add-support-volumeattributesclassname.md
ppatierno: but the error should be in the PVC status not depending on the driver, that's my understanding. So why we can't reflect that in the Kafka CR? (like we reflect errors from other "external systems" in the proper custom resource, i.e. Cruise Control errors into the KafkaRebalance).

scholzj: I do not think it is always the case that it will be there. You also don't know how it will be listed there. I'm not sure what other external systems are you talking about. I think if PVC resizing fails for example, it would also not be necessarily shown in the Kafka CR status as it is hard to detect in all cases.

PaulRMellor (commented):
Thanks. The proposal is clear and makes sense.
I'm not too familiar with the storage config details, but I left a couple of questions for my understanding as a reviewer.

im-konge (approved):
I'm not an expert, but the proposal LGTM. 
+1 (binding)

095-add-support-volumeattributesclassname.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Motivation
+
+Kubernetes v1.31 added a new method of configuring storage parameters for `PersistentVolumes` (PV) using [`VolumeAttributesClass`](https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/) (VAC). `PersistentVolumeClaims` (PVC) can then refer to them using `volumeAttributesClassName` along with their corresponding `storageClassName`. This decouples storage parameters specification from the `StorageClass` (SC) into the VAC. In a PVC, the `storageClassName` field is immutable, whereas the `volumeAttributesClassName` isn't. This makes it possible to dynamically reconfigure the PV without losing data.
+
+## Proposal
+
+To accommodate this change, the `PersistentClaimStorage` (PCS) API needs an additional field `volumeAttributesClass`. When this field changes in the PCS, the Cluster Operator (CO) can map it to the generated PVC's `volumeAttributesClassName` and let the CSI Driver take care of provisioning. There shouldn't be a need to validate the VAC since the parameter names depend on the cloud provider.
+
```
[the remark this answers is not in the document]
venkatesh2090: If the error is apparent, e.g. the VAC is incorrect, the pods won't schedule and I think it is safe to expect the users will have alerting for deployment / replicaset mismatch. There are metrics for those. As for ones that don't immediately break the backing storage, there should be kubelet metrics for the volumes as well that say what the state of a PVC is. imo a user should be responsible for any misconfiguration with the CSI side of configuration.

There could be a generic check for if the volume status != failed or something. But that should happen across the controller, regardless of if there is or isn't a VAC. It is out of scope for this proposal.

095-add-support-volumeattributesclassname.md
```diff
@@ -0,0 +1,80 @@
...
+## Current situation
+
+It is  not possible to set the `volumeAttributesClassName` using [`PersistentClaimStorage`](https://github.com/strimzi/strimzi-kafka-operator/blob/c1b20f726dddbcd2a070c2eeb14fd30902027aec/api/src/main/java/io/strimzi/api/kafka/model/kafka/PersistentClaimStorage.java). To change any configuration parameters of the physical volume backing a `PersistentVolumeClaim`, users have to rely on the `StorageClass`'s `spec.parameters` field. This couples the parameters and the class.
+
+## Motivation
+
+Kubernetes v1.31 added a new method of configuring storage parameters for `PersistentVolumes` (PV) using [`VolumeAttributesClass`](https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/) (VAC). `PersistentVolumeClaims` (PVC) can then refer to them using `volumeAttributesClassName` along with their corresponding `storageClassName`. This decouples storage parameters specification from the `StorageClass` (SC) into the VAC. In a PVC, the `storageClassName` field is immutable, whereas the `volumeAttributesClassName` isn't. This makes it possible to dynamically reconfigure the PV without losing data.
```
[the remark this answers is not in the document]
venkatesh2090: The feature itself is fairly breakage free. If the feature gate is enabled and the `volumeAttributeClassName` field is present on the PVC, it is completely ignored when applying.

Otherwise, it is just applied and the CSI driver takes over for any validations.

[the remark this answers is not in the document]
venkatesh2090: > Should we note that VolumeAttributesClass is cluster-wide resource like StorageClass?

I can add that info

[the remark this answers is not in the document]
venkatesh2090: Before I push anything, does pushing dismiss any approvals?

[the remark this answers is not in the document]
scholzj: That depends on the amount of changes you do. Explaining tht `VolumeAttributesClass` is cluster wide resource would not cause any issues I guess.

095-add-support-volumeattributesclassname.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Motivation
+
+Kubernetes v1.31 added a new method of configuring storage parameters for `PersistentVolumes` (PV) using [`VolumeAttributesClass`](https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/) (VAC). These VACs are a cluster-wide resource. `PersistentVolumeClaims` (PVC) can then refer to them using `volumeAttributesClassName` along with their corresponding `storageClassName`. This decouples storage parameters specification from the `StorageClass` (SC) into the VAC. In a PVC, the `storageClassName` field is immutable, whereas the `volumeAttributesClassName` isn't. This makes it possible to dynamically reconfigure the PV without losing data.
+
+## Proposal
+
+To accommodate this change, the `PersistentClaimStorage` (PCS) API needs an additional field `volumeAttributesClass`. When this field changes in the PCS, the Cluster Operator (CO) can map it to the generated PVC's `volumeAttributesClassName` and let the CSI Driver take care of provisioning. There shouldn't be a need to validate the VAC since the parameter names depend on the cloud provider.
```
ppatierno, line 15: I would highlight it's a String field.

ppatierno (approved):
I left one nit. No need to review again. +1 (binding) from me.

see-quick (approved):
LGTM, Thanks.

PaulRMellor (approved):
Thanks for the responses to my questions.

scholzj:
This proposal now has 5 binding and one non-binding +1 votes. If there are no more comments, I will close it as approved tomorrow.

Frawless (approved):

scholzj:

PR #150, opened 2025-02-25
Deprecate and remove `type: opa` authorization in Kafka CR

scholzj:
This proposal suggests deprecation and removal of the Open Policy Agent (OPA) authorization (`type: opa`) as it can be easily replaced by `type: custom` authorization.


095-deprecate-OPA-authorization.md
```diff
@@ -0,0 +1,87 @@
...
+## Backwards compatibility
+
+Users using the `type: opa` authorization will be impacted by this changes as they will need to start using custom container images and update the Kafka CR resources.
+Other users will not be impacted.
+
+## Rejected alternatives
+
+As an alternative path, we can consider dropping the OPA support completely already before the `v1` CRD API.
+For example drop the binaries and stop using the `type: opa` authorization already in an earlier Strimzi version such as Strimzi 0.48.
+However, I decide to start the proposal with the removal with the removal of the `v1beta2` API version.
```
im-konge, line 87:
```suggestion
However, I decide to start the proposal with the removal of the `v1beta2` API version.
```

im-konge (approved):
The deprecation and removal makes sense to me. Also, I think that removing it as part of removal of the `v1beta2` API  support looks good to me. Thanks!

095-deprecate-OPA-authorization.md
```diff
@@ -0,0 +1,87 @@
...
+* Deprecate the `type: opa` authorization
+* Update the documentation to not use this deprecated field and use the `type: custom` authorization instead
+* Update the `CHANGELOG.md` file and documentation to inform users about this deprecation
+* Have warnings raised by the Cluster Operator when the `type: opa` authorizer is used
+
+While deprecated, we will still continue bundling the OPA authorizer plugin as part of Strimzi.
+
+As part of Strimzi `v1` CRD API, we will remove `type: opa` authorization from the `v1` API.
+
+Finally, in the first Strimzi version that drops the support for the `v1beta2` API, we will:
```
ppatierno, line 66: The previous sentence mentions to remove `opa` type as part of `v1` CRD.
The above one mentions a Strimzi version which drops `v1beta2` API.
How are they related? I mean, there will be a Strimzi version supporting both `v1` and `v1beta2` where `opa` type is removed (because of `v1`) but the following things will be still in place like bundling plugin, because still `v1beta2`? Can you elaborate and maybe clarify in the proposal please?

095-deprecate-OPA-authorization.md
```diff
@@ -0,0 +1,87 @@
+# Deprecate and remove `type: opa` authorization in Kafka CR
+
+This proposal suggests to deprecate and later remove the Open Policy Agent (OPA) authorization (`type: opa`).
+
+## Current situation
+
+Strimzi currently support [Open Policy Agent authorization plugin](https://github.com/StyraInc/opa-kafka-plugin) for Kafka brokers.
```
see-quick, line 7:
```suggestion
Strimzi supports [Open Policy Agent authorization plugin](https://github.com/StyraInc/opa-kafka-plugin) for Kafka brokers.
```

095-deprecate-OPA-authorization.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+## Motivation
+
+Supporting the `type: opa` authorization and bundling the plugin in our images is not for free:
+* We need to maintain the code in the Cluster Operator
+* With every new Kafka release, we need to make sure the dependencies are aligned between Kafka, other plugins and the OPA Authorizer plugin
+* We need to maintain the system tests and make sure we use an reasonably up-to-date OPA version
+
+While there seem to be some users using the `type: opa` authorization, it does not seem to be widely used.
```
see-quick, line 28:
```suggestion
While some users appear to be using type: opa authorization, it does not seem to be widely adopted.
```

095-deprecate-OPA-authorization.md
```diff
@@ -0,0 +1,87 @@
...
+This proposal affects the Strimzi Cluster Operator, System Tests, and the documentation.
+
+## Backwards compatibility
+
+Users using the `type: opa` authorization will be impacted by this changes as they will need to start using custom container images and update the Kafka CR resources.
+Other users will not be impacted.
+
+## Rejected alternatives
+
+As an alternative path, we can consider dropping the OPA support completely already before the `v1` CRD API.
```
see-quick, line 85:
```suggestion
As an alternative, we could consider dropping OPA support entirely before the v1 CRD API release
```

fvaleri (approved):
LGTM. I fully agree with the motivation and plugin strategy.

095-deprecate-OPA-authorization.md
scholzj: The `v1` will not support `type: opa` API anymore. But while `v1beta2` is still supported, the API will be there. So the support needs to remain in the code and in the container images.

scholzj: I tried to make it more clear in the text.

ppatierno: Ignore my last comment, I was adding it but while you were updating the proposal. Your new text is fine with me ;-)

ppatierno (approved):
LGTM.

ShubhamRwt (approved):

095-deprecate-OPA-authorization.md
```diff
@@ -0,0 +1,88 @@
...
+While deprecated, we will still continue bundling the OPA authorizer plugin as part of Strimzi.
+
+When the Strimzi `v1` CRD API is added, it will not support the `type: opa` anymore.
+But as the `type: opa` authorization will be still part of the `v1beta2` API, the support in Cluster Operator and in container images has to remain.
+
+Only in the first Strimzi version that drops the support for the `v1beta2` API and supports the `v1` API only, we will:
+* Stop bundling the OPA authorizer plugin in the Strimzi container images
+* Remove the production code for configuring the OPA authorization
+* Remove the OPA system test
+* Update the documentation
```
PaulRMellor, line 71:
```suggestion
* Update the documentation to remove the `type: opa` authorization content.
```

PaulRMellor (approved):
Considering the alternative approach and current adoption of the feature, the change makes sense

Frawless (approved):

scholzj:
This has now 4 binding and 2 non-binding +1 votes. If there are no more comments I will merge it as approved tomorrow EOB.

scholzj:
Approved with 4 binding and 2 non-binding votes.

PR #151, opened 2025-02-26
Remove Travis-CI from Strimzi Kafka Bridge and Strimzi Kafka OAuth repos

Frawless:
This proposal suggests to drop current usage of [Travis CI](https://www.travis-ci.com/) in [strimzi-kafka-bridge](https://github.com/strimzi/strimzi-kafka-bridge) and [strimzi-kafka-oauth](https://github.com/strimzi/strimzi-kafka-oauth) repositories.


096-drop-travis-ci-and-testing-for-ppc-and-s390x.md
```diff
@@ -0,0 +1,51 @@
...
+_Note that Strimzi's Travis CI quota is sponsored by IBM._
+
+## Proposal
+
+This proposal suggests removing the existing Travis CI configurations from the Strimzi Kafka Bridge and Strimzi Kafka OAuth repositories, effectively eliminating Travis CI usage from the Strimzi organization.
+Regarding the current support matrix, we will not make any changes, meaning that all currently supported architectures — `amd64`, `arm64`, `ppc64le`, and `s390x` — will continue to be supported by the Strimzi project.
+
+We will newly use the following terminology:
+- the `amd64` architecture will have the status `supported`, which means that all available tests will be run on this architecture on regular basis as we do now
+- the `ppc64le`, `s390x`, and `arm64` architectures will have the status `supported, not tested`, meaning that tests will not be regularly executed on these architectures
```
im-konge, line 33: `arm64` can be tested using TF in the operators repo. Or you meant it will be `supported, not tested` because we are not using TF in the operators repo regularly + there is no TF in these two repositories?

im-konge (approved):
LGTM, just one question.

096-drop-travis-ci-and-testing-for-ppc-and-s390x.md
Frawless: Yes, we do not run the tests regularly and we do not use it for running unit and IT tests. I decided to make it more restrictive in the proposal than it is in reality, but it is still true - builds and the main part of testing is done only for `amd64`

> there is no TF in these two repositories
I think the app is installed, but yes, there is no config right now. I mentioned also som reasons in `Rejected Alternatives`

im-konge: Okay, thanks.

scholzj, line 33: I think this proposal should clarify where will this be listed / documented for the users.

scholzj (commented):
I left one nit. I think it would be also good to clarify what the builds you talk about do - they do only the Java builds for Bridge and OAuth which is the least risky part of the whole process. The part where the testing on the other platforms would really matter most are the STs in the operators repo and we never had that. So it does not change in any way.

I think having that in the proposal would be useful for the context of what you are removing. As you are essentially removing a problematic CI and clarifying our testing scope in a way it should have been done already in the past.

As a general note, arm64 is pretty popular because of its cost. So it would be great to bering that into the fully supported platforms. But that is not a topic for this proposal.

096-drop-travis-ci-and-testing-for-ppc-and-s390x.md
Frawless: Do we have somewhere list of supported archs? I am not able to find it. My initial idea was to specify it within the list, but if it doesn't exist, should we mentioned in somewhere in the docs or maybe jus ton website? README came to my mind as well, so adding the list to README of the projects might be good place as well?

scholzj (approved):

see-quick (approved):
Make sense. LGTM.

ppatierno (approved):
LGTM. Thanks.

096-drop-travis-ci-and-testing-for-ppc-and-s390x.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Motivation
+
+In the past, we adopted Azure Pipelines as the primary CI system for the Operators repository, using it to run tests, build documentation, and create artifacts and images.
+The main reason for this transition was the improved resource quotas available for CNCF projects, which allowed us to run pipelines with far fewer restrictions than we had in Travis CI. 
+However, Azure Pipelines does not provide agents for the `ppc64le` and `s390x` architectures.
+This was not an issue for Operators, as we build multi-architecture images on `amd64` using _docker buildx_. 
+However, for Strimzi Kafka Bridge and Strimzi Kafka OAuth, we decided to retain the testing pipelines for `ppc64le` and `s390x`.
+
+In recent months, we have encountered several unexpected issues with Travis CI that required us to submit support tickets to restore our quota.
```
PaulRMellor, line 13: does this mean the quotas ran out or were applied incorrectly?

096-drop-travis-ci-and-testing-for-ppc-and-s390x.md
```diff
@@ -0,0 +1,63 @@
...
+These issues are often not immediately noticed when they occur but are only discovered weeks later. 
+It seems that no one even realizes that some pipelines fail to execute from time to time.
+Another scenario is that jobs fail due to various Travis CI issues, yet we simply ignore the results and merge the PRs anyway.
```
PaulRMellor, line 16: The CI issues were spurious or could not be determined?  or not worth the time to raise?

PaulRMellor (approved):
The motivation makes sense.
I left a couple of questions to clear up some ambiguities

096-drop-travis-ci-and-testing-for-ppc-and-s390x.md
Frawless: That's a good question. I think the quota on Travis CI has some kind of validity period in form of credits and it has to be manually renewed by Travis team from time to time, but I didn't find any order in that so I don't have proper answer.

096-drop-travis-ci-and-testing-for-ppc-and-s390x.md
Frawless: I think we just do not realize we have a problems for some time as Travis is not our main CI tool we use and it would be much better for us to use just one CI tool if possible. Travis kinda adds additional work that we would like to avoid.

Frawless:
This proposal is now approved with 6 binding and 1 non-binding votes. I will keep it open until EOD and then merge.

PR #152, opened 2025-03-06
Redesign restart events

katheris:


101-redesign-restart-events.md
```diff
@@ -0,0 +1,81 @@
...
+  reportingInstance: strimzi-cluster-operator-55d66bf7bd-htjtd
+  # ...
+```
+
+1. The `regarding` field in the [Event API](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/event-v1/#Event) is changed to the Kafka (or Connect or MM2) resource.
+  The `regarding` field maps to `involvedObject` in the output.
+2. The restart message will be updated to include the Pod name to make it easier to identify the affected Pod when listing events.
+3. The `related` field is added pointing to the Pod that is being rolled.
+
+
```
scholzj, line 68:
```suggestion

```

101-redesign-restart-events.md
```diff
@@ -0,0 +1,81 @@
...
+## Affected/not affected projects
+
+This only affects the Strimzi cluster operator.
+
+## Compatibility
+
+This will be a change for users, however since the events being emitted aren't versioned there is no clear way to indicate this change.
+However, we can introduce this change at the same time as the v1 API.
+When the v1 API lands users are likely to be reviewing the changelog more thoroughly than for other releases, increasing the likelihood that they will not be caught out by the change.
```
scholzj, line 77: I'm fine with this. But you can also consider doing it in 0.46 which has the CHANGELOG full of similar changes.

scholzj (approved):
Thanks for the proposal. I left some nits, but I'm +1 overall.

ppatierno (approved):
LGTM. But should we consider having such events for other pods like bridge, cruise control and entity operator?

fvaleri (commented):
Are we also updating or adding log messages?

In [#10958 ](https://github.com/strimzi/strimzi-kafka-operator/issues/10958) we mention the possibility to remove events with the benefit of simplified maintenance. I'm ok to keep them, but maybe we should add the removal as rejected alternative with some motivation.

katheris:
> But should we consider having such events for other pods like bridge, cruise control and entity operator?

@ppatierno I think we could, however I think that is outside the scope of this proposal and related issue.

> Are we also updating or adding log messages?

@fvaleri the proposal includes updating the restart reason to include the pod name, however there are no other log message changes included. The issue was around the events being hard to find, not the operator logs being unclear.

fvaleri (approved):
LGTM. Thanks for addressing my comments.

101-redesign-restart-events.md
katheris: I've updated it to say we should aim for 0.46 release, but if we miss that use the release where the v1 API lands

tinaselenge (approved):
LGTM :)

Frawless (approved):

im-konge (approved):

ppatierno:
This proposal has now 4 +1 (binding) votes and 2 +1 (non binding). Waiting one more day for last minute additional feedback, then @katheris I think you can merge it.

PR #153, opened 2025-03-10
Respect maintenance time windows in cert-manager integration

katheris:
Update to proposal [100](https://github.com/strimzi/proposals/blob/main/100-external-certificate-manager.md#tracking-changes-to-cluster-ca-end-entity-certificates) to account for maintenance time windows.

### DIFF
--- a/101-cert-manager-maintenance-windows.md
+++ b/101-cert-manager-maintenance-windows.md
@@ -1,0 +1,67 @@
+# cert-manager Maintenance Windows
+
+Proposal [100](https://github.com/strimzi/proposals/blob/main/100-external-certificate-manager.md) proposed integrating Strimzi with [cert-manager](https://cert-manager.io/) as an option for managing certificates.
+Proposal 100 did not mention anywhere the option for users to specify maintenance time windows.
+This proposal describes how maintenance time windows should be incorporated into the cert-manager integration implementation.
+
+## Current situation
+
+Strimzi allows users to specify maintenace time windows to manage when the Cluster Operator will renew certificates.
+This setting only applies when Strimzi is managing the Cluster/Clients CA, not when it is provided by the user.
+
+## Motivation
+
+When using cert-manager for certificate management, Strimzi will not have any control over when certificates are renewed.
+The Kafka Pods must be restarted to present new certificates, and users should be able to control when that is done.
+
+## Proposal
+
+When Strimzi is configured to use cert-manager for managing certificates (as described in proposal 100), users can specify a maintenance time window to influence when Strimzi will start using new Cluster CA end-entity certificates.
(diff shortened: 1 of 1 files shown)


scholzj (commented):
I guess I will not -1 it if everyone else approves it. But I think this is wrong. The maintenance time window is used only for situations when Strimzi decides to do something on its own. It is not used for inputs from outside. 

Cert-manager doing something is an input from outside and while I understand that someone might want to avoid that causing a rolling update in a wrong time, it should be addressed at Cert Manager side and not on our side.

Doing this in Strimzi is a slippery slope:
* It will not be clear anymore which actions fall under the maintenance window
* It will be hard to argue why other things should not fall under the maintenance window
* It does not differ from many other situations when something from outside (or someone as that is never clear which change is done by a human and which by some automated system) changes the configuration and we react to it.

(Also, it does not relate to the linked issue which is IMHO about using custom listener certificate and not custom CA)

katheris:
@scholzj thanks for sharing your thoughts, it's definitely a good point about whether this can be addressed on the cert-manager side. I have started investigating whether it is possible or not, and if not whether they would be open to adding such a feature.

Specifically on your point about determining what falls under the maintenance window, as far as I know today maintenance time windows are only used for determining when Strimzi renews certificates (triggering the Pods to roll). For me it feels like Strimzi determining when to start using new certificates from cert-manager is a similar type of decision so I don't agree that it will make it harder to defend not having maintenance windows apply elsewhere.

> Also, it does not relate to the linked issue which is IMHO about using custom listener certificate and not custom CA

Although the linked issued was about custom listener certificates, my understanding from talking to the user was that if there was a cert-manager integration then they would switch to using it and then would run into this issue. So I agree it does not directly resolve the issue they raised, but I think it is related.

scholzj:
> Specifically on your point about determining what falls under the maintenance window, as far as I know today maintenance time windows are only used for determining when Strimzi renews certificates (triggering the Pods to roll). For me it feels like Strimzi determining when to start using new certificates from cert-manager is a similar type of decision so I don't agree that it will make it harder to defend not having maintenance windows apply elsewhere.

I do not think you can see it as the same thing when in one case we renew the certificate and in the other case someone else (Cert-Manager) does it. That makes it similar to a change caused by the infrastructure for example rather than to our current CA renewal.

> Although the linked issued was about custom listener certificates, my understanding from talking to the user was that if there was a cert-manager integration then they would switch to using it and then would run into this issue. So I agree it does not directly resolve the issue they raised, but I think it is related.

It is a bit confusing, but the way the user describes it they use a server certificate signed by their _Corporate CA_. They will not give Strimzi access to such CA to use it directly and they will not let Strimzi issue certificates with our CNs. That much seems pretty clear from the comments.

scholzj:
@katheris Any update on this? Should it be closed?

scholzj:
Triaged on 4.9.2025: We decided to close it for the time being. But we migth get back to it and revisit it later if there is more demand from users. (longer discussion in the YouTube recording.)

PR #154, opened 2025-03-19
Proposal for configuring username as secret in KafkaClientAuthenticationPlain

joystern13:
…

### DIFF
--- a/101-username-as-secret.md
+++ b/101-username-as-secret.md
@@ -1,0 +1,61 @@
+# Username as secret in `KafkaClientAuthenticationPlain`
+
+Add the ability to populate username from a Kubernetes secret in the `KafkaClientAuthenticationPlain` configuration.
+
+## Current situation
+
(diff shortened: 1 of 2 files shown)


scholzj (commented):
Thanks for the proposal.

I do not think the username is secret and I do not think Kafka is treating it as such. It can show up in many different configs and logs. So I do not think this would solve all your problems.

But this idea comes up from time to time. If we decide that this is something we should support, we should definitely have consistency accross our APIs. So should we accept this proposal and its implementation, it should propose and implement this for all places where the username is specified. E.g. also for SCRAM-SHA authentications and so on.

scholzj (commented):
You should also provide more details on the planned implementation. You cover only the API, which is important, but there is obviously also the actual implementation behind it that should be covered.

ppatierno:
Thanks for this proposal which makes sense but as Jakub stated we would need more details about the implementation and not only a higher view of the API.

PR #163, opened 2025-07-23
doc: adding proposal for support Kafka KIP-745, new restart parameters for kafka connectors

rlanhellas:
This proposal aims to add support to [Kafka KIP-745](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=181308623)


rlanhellas:
@katheris , I updated my proposal in order to try to answer your questions. Just one thing I didn't add is about MirrorMaker. IMO MirrorMaker should be a different proposal

105-add-restart-parameters-to-kafka-connectors.md
```diff
@@ -0,0 +1,84 @@
...
+
+The final method should look like this:
+
+```java
+@SuppressWarnings({ "rawtypes" })
+private Future<List<Condition>> maybeRestartConnector(Reconciliation reconciliation, String host, KafkaConnectApi apiClient, String connectorName, CustomResource resource, List<Condition> conditions) {
+    if (hasRestartAnnotation(resource, connectorName)) {
+        boolean restartIncludeTasks = hasRestartInlcudeTastksAnnotation(resource, connectorName);
+        boolean restartOnlyFailedTasks = hasRestartOnlyFailedTastksAnnotation(resource, connectorName);
+        LOGGER.debugCr(reconciliation, "Restarting connector {}", connectorName);
```
scholzj, line 53: Later in the code, it might be worth incuding the parameters in this log message

105-add-restart-parameters-to-kafka-connectors.md
```diff
@@ -0,0 +1,84 @@
...
+        return VertxUtil.completableFutureToVertxFuture(apiClient.restart(host, port, connectorName, restartIncludeTasks, restartOnlyFailedTasks))
+                .compose(ignored -> removeRestartAnnotation(reconciliation, resource)
```
scholzj, line 55: Will we remove the new annotations as well? Or will we keep them?

105-add-restart-parameters-to-kafka-connectors.md
```diff
@@ -0,0 +1,84 @@
...
+                        conditions.add(StatusUtils.buildWarningCondition("RestartConnector", message));
+                        return Future.succeededFuture(conditions);
+                    });
+    } else {
+        return Future.succeededFuture(conditions);
+    }
+}
+```
+
+As you can see on code above, `strimzi.io/restart` is required in order to process the restart. We accept any sort of combinations between these two new annotations,
```
scholzj, line 70: I wondered a bit if this makes sense. If it would be better to treat all these annotations separately -> i.e. to have them trigger the restart on their own.

The advatage of that might be that it would be easier to use for the users. The way the proposal stands right now, it will be critical for the users to set the annotations at once. E.g.:
```
kubectl annotate kctr my-connector strimzi.io/restart=true strimzi.io/restart-include-tasks
```

Because if they set them separately such as:
```
kubectl annotate kctr my-connector strimzi.io/restart=true
kubectl annotate kctr my-connector strimzi.io/restart-include-tasks
```

Then the `restart-include-tasks` will likely be ignroed as the operator will react on the first annotation already.

----

However, while this might be tricky for users, at the end the current proposal might be more future proof as it might in the future allow to add more parameters and combine them easily. So I think it is fine as it is. But:
* I wanted to raise this to the other reviewers
* It might be worth adding the alternative to rejected alternative to make it properly documented that it ewas considered.

scholzj (commented):
Thanks for the proposal. I left some nits, but it looks mostly good.

105-add-restart-parameters-to-kafka-connectors.md
scholzj: One reason why I would remove them is the consistency. But there is also the concern about what other options might be added in the future and what impact they might have if they are not remove and re-used by mistake.

But I think it also creates pretty complicated user experience. Imagine the following scenario ...
1. `Ops person 1` restarted all tasks last week and the `strimzi.io/restart-include-tasks=true` annotation was left there
2. This week, other `ops person 2` wants to restart the failed tasks.
3. So they go to the docs and see that to restart failed tasks, they should run the following command:
   ```
   kubectl annotate kctr my-connector strimzi.io/restart=true strimzi.io/restart-only-failed=true
   ```
4. But suprise - there is the `strimzi.io/restart-include-tasks=true` annotation left

So it seems to me like removing the annotations would create more predictable outcome and easier UX and documentation.

scholzj: Well, the separate use depends on the ordering in the first place, so not great. But you can remove them only once the restart was done, so not sure that is a problem and it changes how that can be done.

rlanhellas: I see your point . Do you think we should have a annotation to keep annotations ? something like restart-keep-query-parameters. So by default we remove all but if explicitly user set this to true we keep the annotations referring the query parameters.

scholzj: I would not add annotation to keep annotations 😉.

But in general, as I said - we can keep it as you suggested and we can see what others think. But it should be documented to make it clear what the plan is. (And possibly you can include the alternative in the _Rejected alternatives_ section.)

scholzj: But you can absolutely add things like that to the rejected alternatives. That is the way to show the reviewers what the other options are and why do you think they are not optimal. And it might be that at the end the concensus will be around one of the rejected alternatives. But that is the way how to make sure the reviewers are aware of the other way that were considered if nothing else and can better decide.

rlanhellas: Right, my other idea would be let user add arguments as Kafka KIP suggest, example:

strimzi.io/restart="includeTasks=true"

105-add-restart-parameters-to-kafka-connectors.md
```diff
@@ -0,0 +1,84 @@
...
+```java
+return VertxUtil.completableFutureToVertxFuture(apiClient.restart(host, port, connectorName, false, false))
+```
+
+The same way we check if there is a restart annotation, we will check if there is a `strimzi.io/restart-include-tasks` and `strimzi.io/restart-only-failed` annotations, and set the values accordingly, 
+something like this:
+
+```java
+boolean restartIncludeTasks = hasRestartInlcudeTastksAnnotation(resource, connectorName);
+boolean restartOnlyFailedTasks = hasRestartOnlyFailedTastksAnnotation(resource, connectorName);
```
katheris, line 42:
```suggestion
boolean restartIncludeTasks = hasRestartIncludeTasksAnnotation(resource, connectorName);
boolean restartOnlyFailedTasks = hasRestartOnlyFailedTasksAnnotation(resource, connectorName);
```

105-add-restart-parameters-to-kafka-connectors.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+The final method should look like this:
+
+```java
+@SuppressWarnings({ "rawtypes" })
+private Future<List<Condition>> maybeRestartConnector(Reconciliation reconciliation, String host, KafkaConnectApi apiClient, String connectorName, CustomResource resource, List<Condition> conditions) {
+    if (hasRestartAnnotation(resource, connectorName)) {
+        boolean restartIncludeTasks = hasRestartInlcudeTastksAnnotation(resource, connectorName);
+        boolean restartOnlyFailedTasks = hasRestartOnlyFailedTastksAnnotation(resource, connectorName);
```
katheris, line 52:
```suggestion
        boolean restartIncludeTasks = hasRestartIncludeTasksAnnotation(resource, connectorName);
        boolean restartOnlyFailedTasks = hasRestartOnlyFailedTasksAnnotation(resource, connectorName);
```

105-add-restart-parameters-to-kafka-connectors.md
```diff
@@ -0,0 +1,84 @@
...
+
+We should be able to support customizing this parameters then user can choose behavior according with your requirements.
+
+## Proposal
+
+If you take a look on other very famous products like nginx, they allow most of configs using [annotations](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/). Then my proposal is to do the same here, and we should have the following:
+
+strimzi.io/restart # already exist
+strimzi.io/restart-include-tasks # new, with boolean value
+strimzi.io/restart-only-failed # new, with boolean value
```
katheris, line 19: To make it clear that these are options of the first annotation would the names `strimzi.io/restart.includeTasks` and `strimzi.io/restart.onlyFailed` be worth considering?

105-add-restart-parameters-to-kafka-connectors.md
scholzj: I think JSON forat would be more flexible toward the future if we decide to change the existing annotation's value.

katheris: @scholzj you mean something like:
```
strimzi.io/restart="{'includeTasks':'true','onlyFailed':'true}"

Comment (rlanhellas) [reply 105-add-restart-parameters-to-kafka-connectors.md:55]: I like the idea of using value inside `strimzi.io/restart` keeping everything in a single annotation, but tbh I think using json inside turn things harder, because of single and double quotes mixing. I would prefer what @katheris  suggested:

`strimzi.io/restart="includeTasks=true&onlyFailed=true"`

File: 105-add-restart-parameters-to-kafka-connectors.md (hunk 1 above)
Comment (rlanhellas) [reply 105-add-restart-parameters-to-kafka-connectors.md]: Added on proposal

File: 105-add-restart-parameters-to-kafka-connectors.md (hunk 2 above)
Comment (scholzj) [reply 105-add-restart-parameters-to-kafka-connectors.md:55]: > Also, I would assume it will always be an HTTP REST API call so I see the JSON usage overkill instead of just having a usual HTTP query parameters like strimzi.io/restart-params="includeTasks=true&onlyFailed=true".

What if you need to encode somehting more complicated such as JSON itself or some more complicated structure that needs escaping etc.?

Comment (katheris) [reply 105-add-restart-parameters-to-kafka-connectors.md:55]: So far the only things added to the API are query parameters, so unless/until they add something where a body is required then we won't need a more complex structure. I think this does add risk for the user. If they create an annotation with JSON which is invalid then we would assume they weren't passing JSON and still do the restart, but without their required query parameters.

Since the KafkaConnector CR is supposed to provide a more Kubernetes native experience so I do feel a bit hesitant about having an actual HTTP query parameter string. Also having `strimzi.io/restart-params="includeTasks=true&onlyFailed=true"` feels like overkill, since I would only provide the `includeTasks` part if I wanted it to be true.

So what if we have one annotation where you can use:
`strimzi.io/restart=true` for the current behaviour
`strimzi.io/restart=includeTasks` to restart connector and tasks
`strimzi.io/restart=onlyFailed` to restart connector but only if failed
`strimzi.io/restart=includeTasks&onlyFailed` to restart connector and tasks, but only if failed

We could even have the code just look for the presence of `includeTasks` and `onlyFailed` in the annotation value, then users could actually use any separator they want.

It's not as future proof as the JSON option, but it gives a set of annotation options that should be easy for users to apply if they are doing it manually, we avoid having to do JSON parsing and it hopefully has a smaller chance of the user missing a quote or a colon and then getting unexpected behaviour. @rlanhellas @scholzj @ppatierno what do you think?

Comment (rlanhellas) [reply 105-add-restart-parameters-to-kafka-connectors.md:55]: Just adding my 5 cents here. I'm more inclined to this proposal:

```
strimzi.io/restart=true for the current behaviour
strimzi.io/restart=includeTasks to restart connector and tasks
strimzi.io/restart=onlyFailed to restart connector but only if failed
strimzi.io/restart=includeTasks&onlyFailed to restart connector and tasks, but only if failed
```

So we keep backward compatibility, and give the user the ability to add parameters if needed.

Comment (ppatierno) [reply 105-add-restart-parameters-to-kafka-connectors.md:55]: For me the params annotations is a kind of configuration for the restart. Of course, if you want to restart, it happens when you applies the `strimzi.io/restart` annotation so if you have params already, they will be used, otherwise not. I see it logical imho.

Comment (katheris) [reply 105-add-restart-parameters-to-kafka-connectors.md:55]: >wdym? is it that much different from strimzi.io/restart=includeTasks&onlyFailed. I don't get why it's overkill.

The part I feel is overkill is having `=true`, we never expect the user to set `strimzi.io/restart-params=includeTasks=false&onlyFailed=false`, so why require them to include the `=true` part.

@ppatierno I see your point about the current annotation being a "boolean", but semantically the annotations aren't strictly boolean. Today I can set `strimzi.io/restart=foobar` and it wouldn't do the restart and wouldn't throw any sort of error (based at least on my understanding of the [code](https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/operator/assembly/AbstractConnectOperator.java#L771)). Given that Kubernetes annotations are designed to be String key/value pairs I think augmenting the value feels somewhat natural.

I'm not against having a separate annotation for the parameters, it just seems unnecessary when we could incorporate it into a single annotation. I also wonder if it would be confusing for users why `restart-params` can be used with the `restart` annotation, but not the `restart-task` annotation.

Comment (ppatierno) [reply 105-add-restart-parameters-to-kafka-connectors.md:55]: > The part I feel is overkill is having =true, we never expect the user to set strimzi.io/restart-params="includeTasks=false&onlyFailed=false", so why require them to include the =true part.

I was just proposing to use the same query string approach, where you need key-value pairs.

> Today I can set strimzi.io/restart=foobar and it wouldn't do the restart and wouldn't throw any sort of error (based at least on my understanding of the code). Given that Kubernetes annotations are designed to be String key/value pairs I think augmenting the value feels somewhat natural.

And this makes me more confused than before then. I see that Kube annotations are k/v pairs where values are just String but I guess it's just to be flexible but I would need an API design where configurations are kind of "strongly typed". I think in Strimzi we already have several annotation acting like booleans (i.e. `strimzi.io/force-rebuild`, `strimzi.io/rebalance-template`, ...) I am not sure we would really change them to host other values. We are going to break users operations if they are using/parting these values as booleans.

> I also wonder if it would be confusing for users why restart-params can be used with the restart annotation, but not the restart-task annotation.

I didn't remember about the restart-task annotation, but in that I case it would be adding a `restart-task-params` if needed imho.

Anyway, I am not going to block the proposal with a -1, it's just that using the same annotation I see concerns in terms of backward compatibility. But if, as KafkaConnect expert, you don't see them, I am fine.

Comment (rlanhellas) [reply 105-add-restart-parameters-to-kafka-connectors.md:55]: @ppatierno , @katheris , @scholzj , do you guys have any decision here ? I can update the proposal accordingly

Comment (mimaison) [reply 105-add-restart-parameters-to-kafka-connectors.md:55]: I would also prefer a single annotation.

Regarding the format, I find the `VALUE1&VALUE2` syntax a bit unusual. I guess it comes from the REST syntax.

Looking at the [Well-Known Labels, Annotations and Taints](https://kubernetes.io/docs/reference/labels-annotations-taints/) list, it seems JSON is often used when the value is complex. There are also a few examples using comma separated values `VALUE1,VALUE2`.

Comment (rlanhellas) [reply 105-add-restart-parameters-to-kafka-connectors.md:55]: Perfect. Everyone here agree with this proposal? @katheris  @scholzj  @ppatierno 

`strimzi.io/restart=includeTasks,onlyFailed`

Comment (katheris) [reply 105-add-restart-parameters-to-kafka-connectors.md:55]: Yep works for me.
Given the number of +1s on this thread already I think you can feel free to update the proposal, rather than waiting for people to respond to this comment

Comment (rlanhellas) [reply 105-add-restart-parameters-to-kafka-connectors.md:59]: proposal was updated @katheris , @scholzj , @ppatierno , @im-konge  , if you can take a look.

File: 105-add-restart-parameters-to-kafka-connectors.md [hunk 7]
```diff
@@ -0,0 +1,89 @@
...
+
+Currently when adding annotation restart these two new parameters are always false when calling the Kafka Connect API.
+
+## Motivation
+
+We should be able to support customizing this parameters then user can choose behavior according with your requirements.
+
+## Proposal
+
+The ideia is to allow users to use combination of arguments, like these:
```
scholzj, line 15: Typo ...

```suggestion
The idea is to allow users to use combination of arguments, like these: 
```

105-add-restart-parameters-to-kafka-connectors.md
```diff
@@ -0,0 +1,89 @@
...
+
+The ideia is to allow users to use combination of arguments, like these: 
+
+```yaml
+strimzi.io/restart=includeTasks,onlyFailed          # restart with args: includeTasks=true and onlyFailed=true
+strimzi.io/restart=includeTasks                     # restart with args: includeTasks=true and onlyFailed=false
+strimzi.io/restart=onlyFailed                       # restart with args: includeTasks=false and onlyFailed=true
+strimzi.io/restart=true                             # restart with args: includeTasks=false and onlyFailed=false
+strimzi.io/restart=false,includeTasks,onlyFailed    # do not restart
+strimzi.io/restart=true,includeTasks,onlyFailed     # restart with args: includeTasks=true and onlyFailed=true
```
scholzj, line 23: Do we really want/need these two? I would say that you should just not set the annotatoin or set it to false to not restart. This just complicates the parsing, but it also makes questions about the order ... will you parse `includeTasks,false,onlyFailed` in the same way? etc.

105-add-restart-parameters-to-kafka-connectors.md
```diff
@@ -0,0 +1,89 @@
...
+}
+```
+
+So, we won't change this behavior, instead we will set the values of `includeTasks` and `onlyFailed` based on the user expectations, since today we always set them to false:
+
+```java
+return VertxUtil.completableFutureToVertxFuture(apiClient.restart(host, port, connectorName, false, false))
+```
+
+The first change will be on the method `hasRestartAnnotation`, so it can return true or false according with mixed cases showed previously. Next step will be create methods to check if `includeTasks` and `failedTasks` are set, suggestion:
```
scholzj, line 42: You should specify what will happen when it includes some other values. E.g. `strimzi.io/restart=includeTasks,onlyRunning`

105-add-restart-parameters-to-kafka-connectors.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Compatibility
+
+We will keep default value as false for both variables, keeping backward compatibility for users not using the new annotations.
+
+## Rejected alternatives
+There are other alternatives considered and the reason why not chosen is as follows:
+
+1. Remove new annotations after connector restarted.
+    - Removing all annotations after restart the connector would require to use always pass new arguments in one single command, it means: 3 annotations in one-shot. What could lead to confusion and mistakes.
```
scholzj, line 86: This should be probably removed / updated -> as there is no new annotation now, removing it is not rejected. But using a different annotation for the flags is rejected alternative.

105-add-restart-parameters-to-kafka-connectors.md
rlanhellas: IMO , if user set a not existent argument we should fail the parse and log and error, not restarting the connector. So `strimzi.io/restart=includeTasks,onlyRunning` will fail to parse and log error to use fix it.

scholzj: That sounds reasonable -> please include it int he proposal.

rlanhellas: just did it, please check again

105-add-restart-parameters-to-kafka-connectors.md
rlanhellas: done

105-add-restart-parameters-to-kafka-connectors.md
rlanhellas: updated the proposal

scholzj:
> > > I think this looks mostly good for Connect. Did we dropped Mirror Maker 2 somewhere along the way? Or was it never part of this? I wonder how easy it would be to implement this for Mirror Maker 2 in some compatible fashion.
> > 
> > 
> > It was never part of this, IMO we should have a different proposal for MM2 , maybe doing reference for this one
> 
> I think separate proposal is fine. But these are not unrelated things. So they need to be considered in context to make sure the _interface_ gives the same _look and feel_.

@katheris Would something like `strimzi.io/restart-connector=<mirrormaker_connector_name>:includeTasks,onlyFailed` or `strimzi.io/restart-connector=<mirrormaker_connector_name>=includeTasks,onlyFailed` work in MM2 context?

rlanhellas:
> > > I think this looks mostly good for Connect. Did we dropped Mirror Maker 2 somewhere along the way? Or was it never part of this? I wonder how easy it would be to implement this for Mirror Maker 2 in some compatible fashion.
> > 
> > 
> > It was never part of this, IMO we should have a different proposal for MM2 , maybe doing reference for this one
> 
> I think separate proposal is fine. But these are not unrelated things. So they need to be considered in context to make sure the _interface_ gives the same _look and feel_.

Agree, since MM2 use the same method `maybeRestartConnector` I need to care about not breaking the MM2 restart connector.

Frawless (approved):

see-quick (approved):
Looks good! 👍 Thanks for the proposal @rlanhellas ! 💯

scholzj (approved):

PR #165, opened 2025-07-30
Added proposal to use Kafka Admin API to list registered brokers and manage their unregistration

ppatierno:
This proposal aims to replace the usage of the `registeredNodeIds` field within a `Kafka` status with listing registered nodes via the Kafka Admin API as per [KIP-1073](https://cwiki.apache.org/confluence/display/KAFKA/KIP-1073:+Return+fenced+brokers+in+DescribeCluster+response) implemantion available starting from Apache Kafka 4.0.0.
It aims to fix https://github.com/strimzi/strimzi-kafka-operator/issues/11667 (and also https://github.com/strimzi/strimzi-kafka-operator/issues/11477 as side effect).


106-list-unregister-nodes-admin-api.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Proposal
+
+This proposal suggests removing the custom registration tracking and instead relying on Apache Kafka's built-in registration tracking available via the Admin API.
+
+### Key changes
+
+The key changes are:
+
+* remove the `status.registeredNodeIds` field from the `Kafka` custom resource. The operator will no longer maintain its own list of registered nodes.
```
scholzj, line 48: You probably should not remove it for backwards compatibility reasons. Just stop using it and deprecate it. Maybe you did not meant it that way, but it probably deserves more clarity.

ppatierno: To be honest I meant that but yes I am wrong. We need to deprecate it first. We also stated it in the proposal 081. Thanks for pointing this out.

106-list-unregister-nodes-admin-api.md
```diff
@@ -83,9 +85,8 @@ The `KafkaReconciler` will stop using `status.registeredNodeIds` and instead rel
 ## Compatibility
 
 This feature will come in the next Strimzi release with the support for Apache Kafka 4.x versions only (with KIP-1073 implementation available).
-When the operator is upgraded, it removes the `status.registeredNodeIds` field from the custom resource for existing clusters.
-This may impact users relying on that field for GitOps or monitoring.
-However, [Strimzi proposal 081](https://github.com/strimzi/proposals/blob/main/081-unregistration-of-KRaft-nodes.md) clearly stated that the field was temporary and subject to removal once Kafka provided native support.
+When the operator is upgraded, it won't use the `status.registeredNodeIds` field anymore but still filling it with the registered nodes.
```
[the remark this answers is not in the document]
ppatierno: We are deprecating it but I would assume it will still list the registered nodes as today for backward compatibility if users are using it for any reasons from the `Kafka` CR status. Why providing a field which is just empty?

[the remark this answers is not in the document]
scholzj: Ahh, ok. Sounds good then 👍.

TBH, I do not know if this has any use for anyone else. I expect that nobody else uses it. So not having the field in the actual custom resources is fine.

But removing it from the CRD might cause issues at Kube level. It would also cause infinite reocnciliation when the new CRD is used with older operator. (e.g. in situations like this: https://strimzi.io/blog/2025/04/10/phased-strimzi-upgrade-example/). So that cannot be done until `v1`.

katheris (approved):

106-list-unregister-nodes-admin-api.md
```diff
@@ -0,0 +1,93 @@
+# Use Kafka Admin API to list registered nodes and manage their unregistration
+
+This proposal suggests leveraging [KIP-1073](https://cwiki.apache.org/confluence/display/KAFKA/KIP-1073:+Return+fenced+brokers+in+DescribeCluster+response) to use the Kafka Admin API for listing all registered nodes in an Apache Kafka cluster (including the fenced ones) and and using that information for managing nodes unregistration.
```
Frawless, line 3:
```suggestion
This proposal suggests leveraging [KIP-1073](https://cwiki.apache.org/confluence/display/KAFKA/KIP-1073:+Return+fenced+brokers+in+DescribeCluster+response) to use the Kafka Admin API for listing all registered nodes in an Apache Kafka cluster (including the fenced ones) and using that information for managing nodes unregistration.
```

106-list-unregister-nodes-admin-api.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Proposal
+
+This proposal suggests removing the custom registration tracking and instead relying on Apache Kafka's built-in registration tracking available via the Admin API.
+
+### Key changes
+
+The key changes are:
+
+* deprecate the `status.registeredNodeIds` field from the `Kafka` custom resource. The field will remain the CRD but the operator won't set it in the `Kafka` status.
```
Frawless, line 48:
```suggestion
* deprecate the `status.registeredNodeIds` field from the `Kafka` custom resource. The field will remain in the CRD but the operator won't set it in the `Kafka` status.
```

Frawless (approved):

106-list-unregister-nodes-admin-api.md
```diff
@@ -0,0 +1,93 @@
...
+During a reconciliation, the operator goes through the following steps:
+
+* get current nodes IDs (both brokers and controllers) from the `KafkaCluster.nodes()` as is done today. It also extracts which ones are controller only nodes (not in mixed mode).
+* call `KafkaNodeUnregistration.listRegisteredNodes` method using the Admin client connected to the brokers' bootstrap service to retrieve all registered brokers.
+* build a `nodeIdsToUnregister` list with the nodes IDs to be unregistered.
+* unregister the nodes via the `KafkaNodeUnregistration.unregisterNodes` method as is done today.
+
+The `nodeIdsToUnregister` list is created in the following way: 
+
+* define `previousNodeIds` as the set of all registered brokers returned by Kafka.
```
tinaselenge, line 64: should this be `previousBrokerNodeIds`?

106-list-unregister-nodes-admin-api.md
```diff
@@ -0,0 +1,93 @@
...
+* define `fencedControllerOnlyNodeIds` as the subset of fenced nodes that are now controller-only (i.e., no longer brokers from being mixed-node).
```
tinaselenge, line 65: this sounds a bit confusing to me because fence state only applies to brokers, but I can't think of better name. Maybe it doesn't matter that much.

106-list-unregister-nodes-admin-api.md
```diff
@@ -0,0 +1,93 @@
...
+## Proposal
+
+This proposal suggests removing the custom registration tracking and instead relying on Apache Kafka's built-in registration tracking available via the Admin API.
+
+### Key changes
+
+The key changes are:
+
+* deprecate the `status.registeredNodeIds` field from the `Kafka` custom resource. The field will remain in the CRD but the operator won't set it in the `Kafka` status.
+* use the Kafka Admin client's `describeCluster()` method (with the option of including the fenced nodes) to retrieve the list of currently registered nodes. This call is done within a newly added `KafkaNodeUnregistration.listRegisteredNodes` method.
```
tinaselenge, line 49:
```suggestion
* use the Kafka Admin client's `describeCluster()` method (with the option of including the fenced nodes) to retrieve the list of currently registered broker nodes. This call is done within a newly added `KafkaNodeUnregistration.listRegisteredNodes` method.
```

tinaselenge: Should the method be also listRegisteredBrokerNodes since this method would specifically calling broker endpoint only?

106-list-unregister-nodes-admin-api.md
tinaselenge: Yeah I understand. I guess we can explain that in a comment/Java doc when implementing it. I will resolve this comment since I couldn't come up with a suggestion.

tinaselenge (approved):
Thanks for the proposal. I have left some minor comments but it looks good to me in general.

106-list-unregister-nodes-admin-api.md
ppatierno: Agree.

106-list-unregister-nodes-admin-api.md
ppatierno: Agree.

ppatierno:
@tinaselenge thanks for the good points. I refactored a little bit the proposal to make it clearer it's only about brokers unregistration as it's the current situation (there is no such controllers unregistration yet, even because no scale down supported).
@scholzj @im-konge @katheris @Frawless because you already approved, can you take a look at the last commit and let me know if you are ok with the changes?

scholzj:
@ppatierno I'm ok with these changes.

ppatierno:
Cool thanks folks! With still 4 +1 (bindings) and 1 +1 (non binding) I am going to wait until the EOD for additional feedback before updating the list and merging.

PR #167, opened 2025-08-04
Added Proposal Kafka Preview Feature Support in Strimzi

Harithasridhar1306:
This proposal introduces native and optional support in Strimzi for managing Kafka preview features in a declarative and persistent way.

### DIFF
--- a/108-kafka-preview-features.md
+++ b/108-kafka-preview-features.md
@@ -45,7 +45,7 @@ As a result:
 
 By supporting preview features through the Kafka CR, Strimzi can offer a seamless, GitOps-compatible, and maintainable experience for all cluster types.
 
-Proposal:
+## Proposal
 I propose to add native declarative support for preview features in the Kafka CR.
 
 CRD Changes


108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+Kafka preview features (such as kafka.shared.consumer) need to be manually enabled using the kafka-features.sh or the Kafka Admin API.
+This works in static clusters but presents issues in dynamic, testing or GitOps-managed environments.
+There is currently no way to express these features declaratively within the Kafka custom resource.
+Without declarative support, the following issues arise:
+
+- Feature drift from manual changes.
+
+- Loss of settings in ephemeral environments.
+
+- Inability to track changes through Git.
```
scholzj, line 14:
```suggestion
* Feature drift from manual changes.
* Loss of settings in ephemeral environments.
* Inability to track changes through Git.
```

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+These features are disabled by default and must be explicitly enabled on each broker to become usable.
+
+Kafka stores the state of these features in the metadata quorum (ZooKeeper or KRaft, depending on the deployment).
+The feature state includes:
+
+- Feature name (e.g., kafka.shared.consumer)
+
+- Version 
+
+- Enabled/disabled flag
```
scholzj, line 27:
```suggestion
* Feature name (e.g., kafka.shared.consumer)
* Version 
*  Enabled/disabled flag
```

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+
+Currently, these features are enabled via:
+
+The kafka-features.sh  tool
+
+The Kafka Admin API (Feature API)
```
scholzj, line 33:
```suggestion
* The `kafka-features.sh` tool
* The Kafka Admin API (Feature API)
```

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
+## Declarative Management of Kafka Preview Features in Strimzi
+
+## Current Situation:
+
+Kafka preview features (such as kafka.shared.consumer) need to be manually enabled using the kafka-features.sh or the Kafka Admin API.
```
scholzj, line 5:
```suggestion
Kafka preview features (such as `kafka.shared.consumer`) need to be manually enabled using the ¨kafka-features.sh` or the Kafka Admin API.
```

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+The kafka-features.sh  tool
+
+The Kafka Admin API (Feature API)
+
+However, these tools require manual execution after cluster creation, which is incompatible with automated GitOps workflows or testing environments.
+
+This proposal introduces a way to declare desired preview features via the Kafka CR, allowing the Strimzi operator to apply and reconcile the correct state automatically.
+
+
+## Motivation:
```
scholzj, line 40:
```suggestion

## Motivation:

```

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+CRD Changes
+Extend the Kafka CRD with a new previewFeatures section under .spec.kafka:
+
+spec:
+  kafka:
+    version: 3.7.0
+    previewFeatures:
+      - name: kafka.shared.consumer
+        enabled: true
+        version: 1
```
scholzj, line 69: Format this as code with ```yaml

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+
+- Feature drift from manual changes.
+
+- Loss of settings in ephemeral environments.
+
+- Inability to track changes through Git.
+
+Kafka Preview Features Explained
+Kafka preview features are opt-in capabilities introduced in newer Kafka versions to allow early access to experimental or upcoming functionality.
+These features are disabled by default and must be explicitly enabled on each broker to become usable.
```
scholzj, line 18: So, what happens when the features GA? How will we validate them? How will the error handling look like?

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+- Declarative and GitOps-friendly.
+
+- Drift protection through automatic reconciliation.
+
+- Consistent environments across deployments.
+
+Considerations
+- Requires changes to the Kafka CRD schema.
+
+- Operator logic must be extended to handle feature state reconciliation.
```
scholzj, line 88: You should provide details of how exactly will the operator do it -> which phase, which classes, etc. Also, do we need some trakcing of the current state of these features?

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+Proposal:
+I propose to add native declarative support for preview features in the Kafka CR.
+
+CRD Changes
+Extend the Kafka CRD with a new previewFeatures section under .spec.kafka:
+
+spec:
+  kafka:
+    version: 3.7.0
+    previewFeatures:
```
scholzj, line 66: Are these 100% previews of new features only?

scholzj (commented):
I left some comments and questions. In general, fixing the formatting as I siggested in the comments would improve the readability so much ...

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+- Teams cannot easily maintain consistent environments across dev/staging/prod.
+
+- Manual intervention is needed after each deployment or cluster recreation.
+
+By supporting preview features through the Kafka CR, Strimzi can offer a seamless, GitOps-compatible, and maintainable experience for all cluster types.
+
+Proposal:
+I propose to add native declarative support for preview features in the Kafka CR.
+
+CRD Changes
```
tinaselenge, line 60:
```suggestion
### CRD Changes
```

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+Extend the Kafka CRD with a new previewFeatures section under .spec.kafka:
```
tinaselenge, line 61:
```suggestion
Extend the Kafka CRD with a new preview features section under `.spec.kafka`:
```

tinaselenge (commented):
Thank you for the proposal. I left some comments.

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+      - name: kafka.shared.consumer
+        enabled: true
+        version: 1
+
+Operator Behaviour
+1) On reconciliation, Strimzi queries the current preview feature state using the Kafka Admin API.
+
+2) If a mismatch is detected, the Operator applies the desired state using the Admin API or kafka-features.sh.
+
+3) Changes are reconciled continuously to prevent drift.
```
im-konge, line 76: Also, how it will work during upgrades of the Kafka version? Is there a way how to check what are the available features that can be enabled and in case that in the list of the `previewFeatures` inside the `.spec.kafka` will be something wrong (not available anymore or graduated to GA), should it be ignored and mentioned in the status of the Kafka CR?

im-konge: The same thing should be with the downgrade, some of the features will not be available if you will do downgrade of the Kafka cluster from one version to the other, so that should be handled as well.

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+
+Benefits
+- Declarative and GitOps-friendly.
+
+- Drift protection through automatic reconciliation.
+
+- Consistent environments across deployments.
```
im-konge, line 83:
```suggestion
- Declarative and GitOps-friendly.
- Drift protection through automatic reconciliation.
- Consistent environments across deployments.
```

But I guess you already mentioned the benefits before.

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+
+Considerations
+- Requires changes to the Kafka CRD schema.
+
+- Operator logic must be extended to handle feature state reconciliation.
+
+- Validation rules needed to ensure correct feature name, version, and enabled state.
```
im-konge, line 90: How the validation will look like? How the user will be informed that there is something wrong? Also, should we keep some information about enabled features inside the `.status` section of the Kafka CR?

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+- Not declarative through the Kafka CR.
+
+- No automatic reconciliation or drift protection.
+
+- More error-prone due to lack of CR validation.
+
+## Affected Projects:
+- strimzi-kafka-operator — Kafka CRD schema and reconciliation logic.
+
+- strimzi-images — Potential updates to include Admin API handling for preview features.
```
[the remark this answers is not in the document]
im-konge: I guess the "potential" is wrong word TBH.
With every change to the code (and every build of the image) you will have some updates in comparison with the current images.
Also, it's not a project, so I would remove this.

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Compatibility:
+
+The previewFeatures field will be optional, and clusters without it will behave as they do today.
```
im-konge, line 107:
```suggestion
The `previewFeatures` field will be optional, and clusters without it will behave as they do today.
```

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Rejected alternatives:
+External Script or Init Container
+- A user-managed init container or script could run kafka-features.sh at broker startup.
+- This would be configured via a ConfigMap, environment variables, or a mounted script.
```
im-konge, line 112: This is good, but you should maybe mention why it is rejected alternative.

108-kafka-preview-features.md
Harithasridhar1306: > Also, how it will work during upgrades of the Kafka version? Is there a way how to check what are the available features that can be enabled and in case that in the list of the `previewFeatures` inside the `.spec.kafka` will be something wrong (not available anymore or graduated to GA), should it be ignored and mentioned in the status of the Kafka CR?

 Upgrade Scenario:

Queries the broker’s available features via Admin API.

Compares them with .spec.kafka.previewFeatures.

For each entry in .spec:

If still preview and supported → apply as requested.

If graduated to GA → skip enabling (Kafka enables it by default); record in .status with reason=FeatureGraduatedToGA.

If removed → skip enabling; record in .status with reason=FeatureUnavailable.

The cluster stays running without error, but users get visible feedback in status.conditions.

is what i was thinking -open to suggestions

Harithasridhar1306: > The same thing should be with the downgrade, some of the features will not be available if you will do downgrade of the Kafka cluster from one version to the other, so that should be handled as well.

Downgrade Scenario:

If a feature is no longer supported in the downgraded version, the operator disables it via Admin API and records in .status with reason=FeatureUnsupportedInVersion.

is what i was thinking -open to suggestions

108-kafka-preview-features.md
Harithasridhar1306: The operator will validate requested features before applying them. If there's an invalid feature name, unsupported version, or other mismatch, the operator will:

Skip the problematic entry to avoid cluster failure, and

Record an entry under .status.previewFeatures including: feature name, requested versus applied version, state (Enabled, Skipped, etc.), and a clear reason (InvalidFeatureName, UnsupportedVersion).
This lets users immediately see what was applied and why something failed.

 - name: kafka.old.preview
      requestedVersion: 1
      appliedVersion: null
      state: Skipped
      reason: "Feature not available in Kafka 3.5."
    - name: kafka.new.featureX
      requestedVersion: 2
      appliedVersion: null
      state: Skipped
      reason: "Invalid feature name."
      
      
    hope this answers your question

108-kafka-preview-features.md
```diff
@@ -0,0 +1,112 @@
...
+
+- Validation rules needed to ensure correct feature name, version, and enabled state.
+
+Limitations
+
+- Not declarative through the Kafka CR.
+
+- No automatic reconciliation or drift protection.
+
+- More error-prone due to lack of CR validation.
```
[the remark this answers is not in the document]
Harithasridhar1306: sorry, my bad ---this limitation was for the rejected version

108-kafka-preview-features.md
Harithasridhar1306: right
i shall add that too

108-kafka-preview-features.md
Harithasridhar1306: Validation:

1. During reconciliation, Strimzi will query the broker’s supported features via the Kafka Admin API (DescribeFeatures).
2. If the feature is GA and still listed in .spec.kafka.previewFeatures, the operator will check whether:
3. Kafka still allows managing it through the UpdateFeatures API.
4. The requested version is valid for the current Kafka version.


Error Handling:
- If the feature is GA, **enabled by default**, and cannot be disabled, the operator will skip applying the change.Record a CannotBeDisabled condition in .status with a reason explaining that the feature is GA and mandatory.

- If the feature is **GA but supports version changes or disabling**, the operator will reconcile as per the CR.
- If the CR requests an unsupported version (either new or not available in this version), the operator will--Skips applying the change.
- Record a VersionNotSupported condition in .status listing the available versions from Kafka’s feature metadata.

108-kafka-preview-features.md
Harithasridhar1306: The .spec.kafka.previewFeatures field will only control features that Kafka exposes via the Feature API — whether they are in preview or already GA but still configurable.

1. If the feature is GA but still configurable:
The operator will allow the user to set it explicitly in previewFeatures even though the name might be misleading, we can rename it to kafkaFeatures if needed.

2. If the feature is GA and not configurable anymore:
The operator will ignore the CR entry for that feature, set a Warning condition in .status explaining that it’s GA and mandatory, and proceed with the rest.

3. If the feature is GA with multiple versions:
Users can still request a specific version in .spec.kafka.previewFeatures and the operator will attempt to set it if Kafka permits.

108-kafka-preview-features.md
im-konge: That's good, could you please add it to the proposal? I guess it's important to know how it will be handled. Thanks :)

108-kafka-preview-features.md
im-konge: Yeah you should mention it in the proposal, so it is covered.

108-kafka-preview-features.md
scholzj: TBH, I'm not sure it is that simple.

I can see how when the feature moves to GA we just issue some warning condition. But if it enabled some feature and that feature version is not supported because it was replaced with a new version, it sounds like ignoring it by throwing a warning would essentially just disable the feature? That seems wrong if there is newer version? What does Kafka itself do in such situation?

Also, I think this is where you need to introduce proper context from how the Kafka community expects to use these _features_. Will each Kafka version support multiple versions of some feature (it sounds like it from the API you propose and that Kafka uses?)? Is the _GA_ just a special version of the feature?

I think understanding it properly is important. And if it is not clear how Kafka will use it, we might not be able to decide on this without screwing up our API. So we would be blocked. To be honest, I think I didi not properly understood how this will evolve (maybe I missed somehting, maybe the problem is that Kafka community itself did not understood how this will evolve) when I wrote https://github.com/strimzi/proposals/blob/main/061-kraft-upgrades-and-downgrades.md as the metadata version seems to be related to this as well.

_As a sidenote: You seem to talk about the status conditions in response to multiple comments. But there is nothing about it in the proposal. If that is how you plan to deal with these, it is important to cover it there as that is the only way how to have the discussion centered in a single place. And not have the discussion independently in multiple places and possibly lead to conflicting outcomes._

108-kafka-preview-features.md
```diff
@@ -0,0 +1,99 @@
+## Declarative Management of Kafka Preview Features in Strimzi
+
+## Current Situation:
+
+Kafka preview features (such as `kafka.shared.consumer`) need to be manually enabled using the ¨kafka-features.sh` or the Kafka Admin API.
+This works in static clusters but presents issues in dynamic, testing or GitOps-managed environments.
```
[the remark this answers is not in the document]
Harithasridhar1306: By static clusters, I mean long-lived, relatively stable Kafka environments where the configuration is applied once and rarely changes—for example, production clusters that are provisioned manually or with minimal automation. In these cases, enabling a preview feature manually through kafka-features.sh or the Admin API is manageable, since the cluster is not frequently recreated.

By dynamic clusters, I mean short-lived or continuously reconciled environments, such as clusters for testing, CI/CD pipelines, or GitOps-managed deployments where the operator ensures the declared state is always applied. In these scenarios, manually running kafka-features.sh isn’t practical because the feature enablement would be lost on cluster re-creation or drifted from the GitOps source of truth.

108-kafka-preview-features.md
```diff
@@ -0,0 +1,99 @@
...
+  kafka:
+    version: 3.7.0
+    previewFeatures:
+      - name: share.version
+        enabled: true
+        version: 1
+
+### Operator Behaviour
+1) On reconciliation, Strimzi queries the current preview feature state using the Kafka Admin API.
+2) If a mismatch is detected, the Operator applies the desired state using the Admin API or `kafka-features.sh`.
```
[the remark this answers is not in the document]
Harithasridhar1306: You’re absolutely right — the Operator should rely on the Kafka Admin API rather than invoking kafka-features.sh. The reference to kafka-features.sh in the proposal was meant only as an example of how users interact with preview features, not as an implementation detail for Strimzi. For the reconciliation logic itself, the plan is to query and update feature states through the Admin API.

108-kafka-preview-features.md
scholzj: > If Kafka allows multiple versions of the same feature to coexist (as the API suggests), the operator will treat GA as the “stable” version and attempt to reconcile to the closest supported version available.

That is not really how declarative management works. You cannot just decide to use something what the operator things is _similar_ and ignore the difference. There is also probably no way how to determine what is the _closest_ version.

> Strimzi will not silently disable the feature — instead, it will record a VersionNotSupported condition in .status and list which versions Kafka still supports, so the user is guided to update the CR. If Kafka makes the GA version mandatory and non-configurable, the operator will skip reconciling that field and surface a CannotBeDisabled condition in .status to make the user aware.

So, will this fail the reocnciliation and for example block an upgrade?

108-kafka-preview-features.md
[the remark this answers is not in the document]
scholzj: I do not think this is a terminology we use anywhere, so you should explain it in the proposal.

However, I think you got one thing wrong:
* The _static_ production clusters often use GitOps and often cannot use the `kafka-features.sh` command because the only proper and approved way to change things is through GitOps processes (i.e. they would for example need to manage it through a Kubernetes Job calling `kafka-features.sh` which is submitted through GitOps).
* The _dynamic_ shortlived clusters would nto want to use `kafka-features.sh` every five minutes to set it. But then - you IMHO do not use that often GitOps to create a cluster for a very short time and then delete it again through GitOps. So the GitOps part of tje argument here - I do not get it.

108-kafka-preview-features.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Kafka Preview Features Explained
+Kafka preview features are opt-in capabilities introduced in newer Kafka versions to allow early access to experimental or upcoming functionality.
+Early access or Preview features that are not production-ready yet, are disabled by default and must be explicitly enabled on each broker to become usable.
+
+Kafka stores the state of these features in the metadata quorum (ZooKeeper or KRaft, depending on the deployment).
+The feature state includes:
+* Feature name (e.g., kafka.shared.consumer)
+* Version 
+*  Enabled/disabled flag
```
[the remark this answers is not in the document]
Harithasridhar1306: I completely agree that we cannot and should not change the Kafka API itself. My intention with the “enabled/disabled” simplification is not to introduce a new abstraction that overrides Kafka’s semantics, but rather to describe how Strimzi might present the existing finalized version level in a more user-friendly way. Internally, it would still rely on the Kafka Admin API fields (FinalizedVersionLevel, SupportedMinVersion, etc.) without altering their meaning. I’ll make sure to clarify this better in the proposal so it’s clear that we’re only mapping what Kafka already exposes, not creating a divergent API surface.

108-kafka-preview-features.md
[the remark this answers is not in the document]
Harithasridhar1306: I see your point about the terminology not being commonly used, and I’ll adjust the proposal to avoid introducing “static/dynamic” and instead describe clusters in terms of lifecycle and management style.

You’re also right that production clusters often rely on GitOps as the only approved mechanism for changes, so running kafka-features.sh directly wouldn’t fit those processes. I’ll reframe the explanation to emphasize that:
Prod clusters: usually GitOps-managed, requiring any feature enablement to be expressed declaratively, rather than direct CLI commands.

Non-prod clusters : recreated frequently, so manual commands don’t persist across rebuilds. While GitOps is less common here, automation via CI pipelines or scripts is needed instead of manual steps.

scholzj:
Discussed on the community call on 8.1.2026: This should be closed as there does not seem to be any progress.

PR #148, opened 2025-02-09
Proposal: Support for Custom Authentication Type in Kafka Clients

YuanShisong:

PR #168, opened 2025-08-12
Support for `type: custom` client authentication

scholzj:
This proposal adds support for new `type: custom` client authentication. It will allow users to use ustom authentication mechanisms such as the Amazon MSK IAM based authentication in the strimzi operands based on the Kafka clients (Connect, Bridge, MM2). It replaces the previous dead / unsuccessful proposals such as https://github.com/strimzi/proposals/pull/148 or https://github.com/strimzi/proposals/pull/42.


109-custom-client-authentication.md
```diff
@@ -0,0 +1,156 @@
...
+    config:
+      ssl.keystore.location: /mnt/my-custom-tls-cert/keystore.p12
+      ssl.keystore.password: changeme
+      ssl.keystore.type: PKCS12
+  ```
+
+### Custom authentication plugins
+
+Strimzi will not include any custom authentication plugins in its container images.
+Users will be responsible for adding them on their own.
```
PaulRMellor, line 109: we don't need to say anything on where they'll reside or how they're found?

109-custom-client-authentication.md
```diff
@@ -0,0 +1,156 @@
...
+* `ssl.keystore.`
+* `sasl.`
+
+These options are otherwise forbidden and cannot be set in the regular _configuration_ sections (i.e. in `.spec.config` or `.spec.clusters.config`).
+Other options which might be needed by the custom authentication mechanisms can be configured in the regular _configuration_ sections.
+That way, we make sure that users are able to configure the custom mechanisms, but cannot use it to override the other configuration options blocked by Strimzi (e.g. the REST API configuration for Connect) by mistake or intention.
+
+Use of TLS and the configuration of trusted certificates will continue to be done through the corresponding `tls` sections.
+It will not be configurable through the custom authentication mechanism.
+
```
PaulRMellor, line 70: shall we say how invalid config is handled?

PaulRMellor (approved):
The proposal justifies the case for introducing this and the proposed solution is clearly described

109-custom-client-authentication.md
scholzj: I added a link to the docs which cover this: https://strimzi.io/docs/operators/latest/full/configuring.html#con-common-configuration-volumes-reference

109-custom-client-authentication.md
scholzj: I extended the proposal.

109-custom-client-authentication.md
```diff
@@ -0,0 +1,156 @@
...
+    config:
+      sasl.mechanism: OAUTHBEARER
+      sasl.jaas.config: org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required;
+      sasl.login.callback.handler.class: CustomAuthenticateCallbackHandler
+  ```
+* Custom mTLS authentication (when the key is injected by an external mechanism rather than loaded from a Kubernetes `Secret`):
+  ```yaml
+  authentication:
+    type: custom
+    sasl: false
```
ppatierno, line 99: Is the sasl field going to be mandatory or defaulting to false if omitted?

ppatierno (approved):
LGTM. I left just one question/comment.

109-custom-client-authentication.md
scholzj: Good point. It will be optional and default to false. I clarified it in the proposal text. Thanks.

scholzj:

PR #169, opened 2025-08-19
Updated the unregister nodes API proposal due to implementation changes

ppatierno:
After some feedback on the corresponding implementation PR https://github.com/strimzi/strimzi-kafka-operator/pull/11729 to simplify the implementation, this PR updates the proposal to reflect such changes.
It's not something we usually do but I would like to avoid having a so stale proposal compared to the implementation.


im-konge (approved):

Frawless (approved):

ppatierno:
This is not a real proposal but a change on an already approved one. Anyway we have 4 +1 bindings already. I will wait EOD before merging this PR.

PR #170, opened 2025-08-20
Add proposal for using Buildah instead of Kaniko after its archivation

im-konge:
This proposal aims to resolve the issue with archivation of Kaniko - mentioned in issue https://github.com/strimzi/strimzi-kafka-operator/issues/11579 .
I already implemented PoC (without tests and some more checks) here - https://github.com/im-konge/strimzi-kafka-operator/tree/connect-build-buildah - if someone is interested.


111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,138 @@
...
+
+The main motivation behind this proposal is to resolve the issue that Kaniko is archived, and it will not receive any updates (no bug or CVE fixes).
+Currently, it seems to not be a problem, but when some critical CVE or bug arises, we have to fix it.
+Also, we need a support in case that there will be something blocking us on newer versions of Kubernetes, which we would not get in this case anymore.
+
+## Proposal
+
+For the Connect Build feature on Kubernetes, we will use [Buildah](https://buildah.io/) as a replacement of Kaniko.
+Buildah is well-supported, widely used tool for building container images.
+Also, it seems that OpenShift uses Buildah in Build API and its process - which makes Buildah a good alternative to Kaniko, with guarantees that it will be supported and maintained further.
```
see-quick, line 20:
```suggestion
It also appears that OpenShift’s Build API relies on Buildah, which makes Buildah a strong alternative to Kaniko, with the added assurance of long-term support and maintenance.
```

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,138 @@
...
+Other than that, Buildah doesn't need any workaround in order to run it on Kubernetes rootlessly - the only thing that is needed during the build and push stages is to specify `--storage-driver=vfs` option.
+VFS (Virtual File System) driver ensures user-space implementation, it stores each layer as a full copy of the files, but without need of root permissions.
+The only trade-off is the build time (it's slower) and consumption of disk space.
+But it seems that it is similar to what Kaniko does as well.
+
+We will use the official Buildah images on Quay.io - currently `quay.io/buildah/stable:v1.40.1` - and similarly to Kaniko executor image - we will pull the official image, tag it with the version and push it to the Strimzi repository on Quay.
+That's useful because of our versioning and we will have the image ready even if someone on Buildah side decides to remove it from their Quay repository.
+Users then can specify their own Buildah image using `STRIMZI_DEFAULT_BUILDAH_IMAGE` (similarly to what is possible today with Kaniko).
+
+The usage of Buildah will be gated behind feature gate called `UseConnectBuildWithBuildah` that will have following schedule:
```
see-quick, line 30: Do we need a feature gate for this, or is it too disruptive for users currently relying on Kaniko?

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,138 @@
...
+|----------------|-------------------|--------|
+| 0.48.0         | 0.51.0            | 0.54.0 |
+
+The build of the Connect image will be done the same way as today, the main difference will be in the commands used in the build and push process.
+Kaniko executor did everything at once, which can be done by Buildah as well, but I rejected to have it done in one command because:
+
+- We will not get the SHA of image after the push.
+- Once pushed, the image will not be stored locally, so we cannot check the SHA using some different command.
+
+The SHA of the image is needed to keep the compatability with previous implementation using Kaniko.
```
see-quick, line 42:
```suggestion
The SHA of the image is needed to keep compatibility with the previous implementation using Kaniko.
```

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,138 @@
...
+buildah build --file=/dockerfile/Dockerfile --tag=__IMAGE__ --storage-driver=vfs __ADDITIONAL_BUILD_OPTS__
+buildah push --storage-driver=vfs --digestfile=/tmp/digest __ADDITIONAL_PUSH_OPTS__ __IMAGE__
+buildah images --digests --filter=digest=sha256:$(cat /tmp/digest) --format='{{.Name}}@{{.Digest}}' > /dev/termination-log
+```
+- `__IMAGE__` is placeholder for user desired name of the image (with registry, repository, and possibly tag)
+- `__ADDITIONAL_BUILD_OPTS__` is placeholder for user desired additional build options
+- `__ADDITIONAL_PUSH_OPTS__` is placeholder for user desired additional push options
+
+We need to take these three steps in order to get the SHA of the image together with correct repository.
+That will work in all cases - even considering cases when user will not specify the tag and will keep the default (`latest`).
```
see-quick, line 61:
```suggestion
This will work in all scenarios, including when the user does not specify a tag and the default (latest) is used.
```

111-use-buildah-instead-of-kaniko.md
see-quick: After a bit of thinking and reading compatibility section, I think this is a good decision (i.e., use feature gate).

see-quick (approved):
Thanks @im-konge, +1 from me. I just have a few minor stylistic comments worth considering 👍

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,138 @@
...
+Buildah has option to store the SHA to specify file on path using `--digestfile` option - but only when `push` command is used.
+It will just output the SHA, not the full image, so we need to build it from the SHA and built image and output it to `/dev/termination-log` in order to have it inside the message of the completed Pod.
+
+We will use following commands:
+
+```shell
+buildah build --file=/dockerfile/Dockerfile --tag=__IMAGE__ --storage-driver=vfs __ADDITIONAL_BUILD_OPTS__
+buildah push --storage-driver=vfs --digestfile=/tmp/digest __ADDITIONAL_PUSH_OPTS__ __IMAGE__
+buildah images --digests --filter=digest=sha256:$(cat /tmp/digest) --format='{{.Name}}@{{.Digest}}' > /dev/termination-log
+```
```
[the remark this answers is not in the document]
im-konge: They are executed as the block of commands inside the `args` field of the Pod.
Something like this:
```
spec:
  containers:
  - args:
    - /bin/bash
    - -c
    - |
      buildah build --file=/dockerfile/Dockerfile --tag=__IMAGE__ --storage-driver=vfs
      buildah push --storage-driver=vfs --digestfile=/tmp/digest __ADDITIONAL_PUSH_OPTS__ __IMAGE__
      buildah images --digests --filter=digest=sha256:$(cat /tmp/digest) --format='{{.Name}}@{{.Digest}}' > /dev/termination-log
```
Or did you mean it differently?

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,138 @@
...
+
+We need to add two new fields to `DockerOutput` model in order to support additional options for both build and push options.
+Those new fields will be:
+- `additionalBuildahBuildOptions` - for additional options to the `build` command. Allowed options will be: `--annotation, --authfile, --cert-dir, --creds, --decryption-key, --env, --label, --logfile, --manifest, --retry-delay, --secret, --security-opt, --timestamp, --tls-verify`
+- `additionalBuildahPushOptions` - for additional options to the `push` command. Allowed options will be: `--authfile, --cert-dir, --creds, --format, --quiet, --remove-signatures, --retry, --retry-delay, --sign-by, --tls-verify`
+
+As for the Kaniko additional options, if the user-specified Buildah options will contain forbidden options (or not known), user will be notified by message inside the `.status` section of `KafkaConnect` resource.
+
+In case that Buildah will be used (enabled by the feature gate) and user specifies some Kaniko additional options, they will be ignored and user will be notified with warning inside the `.status` section of `KafkaConnect` CR.
```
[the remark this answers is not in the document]
im-konge: Yes, I will mention it there. Thanks

111-use-buildah-instead-of-kaniko.md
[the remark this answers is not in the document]
scholzj: That is what I meant, yes. Although it makes me wonder if having some script would make things cleaner?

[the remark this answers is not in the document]
im-konge: I thought about the same thing, also having our custom image based on the Buildah image. However, I think it would be better to have it in the code rather than some custom image, so if users would like to use their own buildah image, they will not depend on our custom image with script. But yes, it's another solution which would be maybe better.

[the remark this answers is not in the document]
Frawless: how it will behave if the first command fails? I mean is `set -e` set by default?

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,138 @@
...
+
+Even though it's change in terms of the tool that builds the images, it will be gated behind the feature gate, so users can adapt to this change through few releases.
+Until the feature gate is GA, we will keep the Kaniko implementation in place, ensuring the backwards compatability.
+If the Buildah feature gate will be used, the `.spec.build.additionalKanikoOptions` field will be ignored and user will be notified in `.status` section of the `KafkaConnect` CR.
+Once the Buildah feature gate will be promoted to Beta, we will deprecate the `.spec.build.additionalKanikoOptions` field.
+Finally, after Buildah feature gate will be promoted to GA, we will remove the implementation of Kaniko, together with the Makefile for tagging the Kaniko executor image.
+
+## Rejected alternatives
+
+### Only one implementation of Connect Build - Buildah on both Kubernetes and OpenShift
```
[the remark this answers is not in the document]
Frawless: Wouldn't also some restrictive Kubernetes envs will cause same troubles to buildah as OpenShift?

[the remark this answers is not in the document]
scholzj: SWhy does it need to run as anyuid if it does not need root?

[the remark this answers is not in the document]
im-konge: TBH I'm not sure about the internals, previously when I tried with runAsUser 1000 there were things like `Error during unshare(CLONE_NEWUSER): Function not implemented` when it tries to create some user namespaces.  I found various blogs about creating own SCC, but it didn't actually work.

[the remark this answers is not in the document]
im-konge: Yeah you are right.
Anyway, I tried it without the anyuid config and it seems to work fine as well as it is.
So maybe we can continue without that RoleBinding.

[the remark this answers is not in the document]
im-konge: So I guess I should add it to PoC code and run few tests on OCP with Buildah in order to see if it really works and then I can add it to proposal (removing it from the `Rejected alternatives`).
WDYT? :)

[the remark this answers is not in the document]
im-konge: After few days of struggle with the OCP implementation to the PoC, it finally works.
However, I had to do some changes (in comparison to the YAMLs with files) in the operator and add some ClusterRoles/RoleBindings for the operator and the builder Pod itself. And that's because I removed the SecurityContext part from the Pod's YAML - which causes the most issues :).

First thing is, that we have to change the security context, because otherwise OCP will add the "most restricted" context to the Pod - and based on my tests (and like 3 days struggling with workarounds how to enable it on restricted profile) it doesn't work at all. The default is `restricted-v2` which doesn't let you do anything (I mean anything in terms of the build, which needs more than OCP allows to do). The `non-root-v2` is not fitting as well. The only profile with which worked is the `anyuid` one. That's because the mounts are for `root`, Buildah needs to create user namespace during build of the image, and also some of the things are pre-created for user `builder` or `build` (I don't remember the exact one), with which you will end up with "unknown user id XXXXXXXX" during the build of an image.

Now to add this security profile to the builder Pod, Strimzi CO needs ClusterRole for allowing to "bind" this profile to the ServiceAccount. All the new ClusterRoles and ClusterRoleBindings will be attached at the end of this comment [1]. Other than this `anyuid` profile, Strimzi CO needs permissions to add `image-builder` role to the connect-build SA. That's mainly for OpenShift, so the SA can push to the internal OCP registries (and CO can mount the SA secret (containing `dockercfg`) to the builder Pod and used it).

Now, once these ClusterRoles and ClusterRoleBindings are applied for CO, Strimzi can - in OpenShift - create RoleBindings for the connect-builder SA and it will work fine. I tried few tests and it worked well.

Thanks to this I was able to keep the `Dockerfile` for Buildah as it was, without need of more hacking and things like entrypoints etc. - because TBH, I did a lot of hacking of the `Dockerfile`, just to see if with those hacks the build will probably work on the `restricted-v2` profile - but surprise, it will not.

Unfortunately, without custom SecurityContext (and the `anyuid` role) it will not work - and I tried a lot of things in order to skip need of adding `anyuid` to the builder's Pod, but without luck.

IDK how huge blocker or security risk (I guess it will be a security risk) it would be to have this ClusterRoles for our operator and have the `anyuid` for the builder Pod. But at least I tried 🤷 
Let me know what you think about it.

Also, I think that having this on OCP will in _some way_ work as with the Build API, I'm not sure how much restrictive it actually is (I think that it needs the more or less same things if it uses Buildah) and IMHO the Connect Build feature is useful for development etc. In case that users will like to have it restricted and will not like to use this, they can use the Image Volume feature or build the image outside of the cluster. But that's just my opinion.

NOTE: The `image-builder` role will be added just for the Namespace where the build is happening. So it will not be "cluster wide access" or something like that. Same with the `anyuid`, it's done using RoleBindings in the particular Namespace.

[1] - https://gist.github.com/im-konge/2a684e2f8a4a22cb6a2889f7ee8d86d6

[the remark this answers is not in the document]
scholzj: Ok, so ... I take that it does not work under the restricted profile on Kubernetes either, or? Because that is what I understood as possible with Buildah. But if it does not work under restricted v2 on OCP, it should not on Kube either.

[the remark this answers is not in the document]
im-konge: If the Kubernetes cluster will have the same restricted env as OCP by default does, then it will not work as well, yes. (TBH I'm not sure how much Kaniko worked on restricted env before as well..)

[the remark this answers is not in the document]
scholzj: Well, you can try it: https://kubernetes.io/docs/concepts/security/pod-security-standards/ ... its part of Kubernetes itself.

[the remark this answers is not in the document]
scholzj: But it would be worth explaning it in the proposals, as this is actually one of the blockers with Kaniko and fully supporting the restricted security policy and why I wanted originally to investigate Buildah even before the Kaniko archival.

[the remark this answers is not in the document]
im-konge: Maybe I'm doing something wrong. But based on the blog about running Buildah on OCP - the official Buildah one - talks about `anyuid`.
I can try to contact someone from Buildah in order to see if it's possible to run on restricted-v2 or not.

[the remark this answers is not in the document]
im-konge: FYI I opened discussion on Buildah repo https://github.com/containers/buildah/discussions/6359
Let's see. I will check also what SCC Build API uses.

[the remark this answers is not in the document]
im-konge: I checked what Build API uses for SCC and it's `openshift.io/scc: privileged`.
The `securityContext` looks like this:
```
    securityContext:
      capabilities:
        add:
        - CHOWN
        - DAC_OVERRIDE
        drop:
        - ALL
      privileged: false
      runAsGroup: 0
      runAsNonRoot: false
      runAsUser: 0
      seccompProfile:
        type: RuntimeDefault
```
so it's not `restricted-v2` or `non-root-v2` as I tried.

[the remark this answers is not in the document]
im-konge: I can try to run it in the same way as Build API runs it. In that case it wouldn't be _that_ different. But if you think that we shouldn't invest more time to it, I can just add it to rejected alternatives.

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,138 @@
...
+- `ConnectBuildOperator`
+- `KafkaConnectAssemblyOperator`
+
+These classes contain the logic of the Connect Build feature, thus they are affected the most.
+We will need also changes to `DockerOutput` model and to the `KafkaConnect` CRD in order to support two new fields for specifying the additional Buildah options.
+
+## Compatibility
+
+Even though it's change in terms of the tool that builds the images, it will be gated behind the feature gate, so users can adapt to this change through few releases.
+Until the feature gate is GA, we will keep the Kaniko implementation in place, ensuring the backwards compatability.
```
ppatierno, line 88:
```suggestion
Until the feature gate is GA, we will keep the Kaniko implementation in place, ensuring the backwards compatibility.
```

111-use-buildah-instead-of-kaniko.md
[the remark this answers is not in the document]
scholzj: > I didn't have time to check it with Kaniko, I will do that I guess tomorrow or something.

Check what with Kaniko? Kaniko does not work with the restricted profile, I know that. So no need to check if that is what you meant.

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,179 @@
...
+Other than that, Buildah doesn't need any workaround in order to run it on Kubernetes rootlessly - the only thing that is needed during the build and push stages is to specify `--storage-driver=vfs` option.
+VFS (Virtual File System) driver ensures user-space implementation, it stores each layer as a full copy of the files, but without need of root permissions.
+The only trade-off is the build time (it's slower) and consumption of disk space.
+But it seems that it is similar to what Kaniko does as well.
+
+We will use the official Buildah images on Quay.io - currently `quay.io/buildah/stable:v1.40.1` - and similarly to Kaniko executor image - we will pull the official image, tag it with the version and push it to the Strimzi repository on Quay.
+That's useful because of our versioning and we will have the image ready even if someone on Buildah side decides to remove it from their Quay repository.
+Users then can specify their own Buildah image using `STRIMZI_DEFAULT_BUILDAH_IMAGE` (similarly to what is possible today with Kaniko).
```
ppatierno, line 28: Should we mention that the `STRIMZI_DEFAULT_KANIKO_EXECUTOR_IMAGE` env var will be removed from the deployment YAML of a new operator release and that and if the user has their own deployment YAML, it will be ignored and should be replaced by `STRIMZI_DEFAULT_BUILDAH_IMAGE`?

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,179 @@
...
+    - `--quiet` - when writing the output image, suppress progress output.
+    - `--remove-signatures` - don't copy signatures when pushing images.
+    - `--retry` - number of times to retry in case of failure during image pull (the base one).
+    - `--retry-delay` - duration between retry attempts.
+    - `--sign-by` - sign the pushed image using the GPG key that matches the specified fingerprint.
+    - `--tls-verify` - skip TLS verification for insecure registries.
+
+As for the Kaniko additional options, if the user-specified Buildah options will contain forbidden options (or not known), user will be notified by message inside the `.status` section of `KafkaConnect` resource, the `InvalidResourceException` will be thrown (logged inside the operator log) and the build will fail.
+
+If Buildah is enabled through the feature gate and the user provides Kaniko-specific options, those options will be ignored, and a warning will be reported in the .status section of the KafkaConnect CR.
```
ppatierno, line 95:
```suggestion
If Buildah is enabled through the feature gate and the user provides Kaniko-specific options, those options will be ignored, and a warning will be reported in the `.status` section of the `KafkaConnect` CR.
```

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,179 @@
...
+Once the feature gate will be moved to beta, the additional Kaniko options field will be deprecated.
+
+Finally, this feature will be available on Kubernetes only - it will not be available on OpenShift, which is mentioned as one of the rejected alternative.
+So the implementation and usage of OpenShift Build API will be kept.
+
```
ppatierno, line 100: Regarding the above paragraph could we have a YAML example?

im-konge: YAML example of what exactly? Sorry I have some YAMLs that I used for the PoC on OCP - ClusterRoles, Rolebindings... so I'm not sure what exactly do you mean :D

111-use-buildah-instead-of-kaniko.md
ppatierno: I am fine with your update in the proposal.

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,229 @@
...
+          - type: maven
+            group: org.apache.kafka
+            artifact: connect-file
+            version: 4.1.0
+```
+
+### Running Buildah in restricted environment
+
+The Buildah, similarly to Kaniko, has issues running in restricted environment.
+To run the build and push commands, it needs some capabilities like - `cap_kill`, `cap_setgid`, and `cap_setuid`.
```
scholzj, line 154: It is suprising push needs something special as that is just transfering some data to some remote registry.

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,229 @@
...
+        artifacts:
+          - type: maven
+            group: org.apache.kafka
+            artifact: connect-file
+            version: 4.1.0
+```
+
+### Running Buildah in restricted environment
+
+The Buildah, similarly to Kaniko, has issues running in restricted environment.
```
scholzj, line 153: You should likely explain what restricted environment is?

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,229 @@
...
+In case that users will use Kaniko, the Buildah environment variable will be ignored, and vice versa.
+The users will be responsible for changing these environment variables based on the mode in which they run the Connect build - in case that they have customized Deployment files.
+
+The usage of Buildah will be gated behind feature gate called `UseConnectBuildWithBuildah` that will have following schedule:
+
+| Alpha (opt-in) | Beta (default-on) | GA     |
+|----------------|-------------------|--------|
+| 0.49.0         | 0.52.0            | 0.55.0 |
+
+The build of the Connect image will be done the same way as today, the main difference will be in the commands used in the build and push process.
```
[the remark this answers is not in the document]
im-konge: Will try that :)

111-use-buildah-instead-of-kaniko.md
scholzj: Well, i assumed you tried that it is the push which requires these and not the build?

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,229 @@
...
+- `<IMAGE>` is placeholder for user desired name of the image (with registry, repository, and possibly tag)
+- `<ADDITIONAL_BUILD_OPTS>` is placeholder for user desired additional build options
+- `<ADDITIONAL_PUSH_OPTS>` is placeholder for user desired additional push options
+
+We need to take these three steps in order to get the SHA of the image together with correct repository.
+This will work in all scenarios, including when the user does not specify a tag and the default (`latest`) is used.
+
+We need to add two new fields to `DockerOutput` model in order to support additional options for both build and push options.
+Those new fields will be:
+- `additionalBuildahBuildOptions` - for additional options to the `build` command.
```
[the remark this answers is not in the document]
ppatierno: I got what you mean and yes I agree.

111-use-buildah-instead-of-kaniko.md
scholzj: Well if it faield in the build then you should probably say that it is the build that needs it then.

111-use-buildah-instead-of-kaniko.md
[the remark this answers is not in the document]
scholzj: Well, the APi change should be part of a proposal. So it should be likely covered in this proposal.

im-konge:
@scholzj hopefully I resolved your comments, I will need to just check the multi-stage builds with Maven artifacts, but in the meantime - can you please check if the text I added makes more sense to you now? Thanks

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,234 @@
...
+    config.storage.topic: connect-cluster-configs
+    status.storage.topic: connect-cluster-status
+    config.storage.replication.factor: -1
+    offset.storage.replication.factor: -1
+    status.storage.replication.factor: -1
+  build:
+    output:
+      type: docker
+      image: my-internal-registry:5000/repository/image:latest
+      additionalBuildahBuildOptions:
```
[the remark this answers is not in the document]
im-konge: Yeah I just forgot to change it there :) good catch. Thanks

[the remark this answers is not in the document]
im-konge: I pushed the change + changed also the Compatability section to reflect the changes I've made to the proposal. Thanks Kate :)

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,232 @@
...
+This will work in all scenarios, including when the user does not specify a tag and the default (`latest`) is used.
+
+Together with this feature, and upcoming v1 API, we will do following changes inside the `DockerOutput` model:
+
+1. we will deprecate `additionalKanikoOptions` field as the name of it is connected directly to Kaniko and more generic name can be used (this field will be removed as part of the v1 API).
+2. we will add `additionalOptions` field which will be used from now on and will cover additional options for both Kaniko and Buildah.
+
+The `additionalKanikoOptions` will still be used until the Buildah feature gate is moved to Beta stage, but warning about deprecation will be added inside the `.status` section of the `KafkaConnect` CR.
+Additionally, if both of the fields will be filled with values, `additionalOptions` takes precedence.
+Once the `UseConnectBuildWithBuildah` feature gate is moved to Beta, the `additionalKanikoOptions` will be completely ignored, the warning about deprecation will slightly change to "is deprecated and ignored" in case that it will be used.
```
scholzj, line 76: I do not think this makes sense. Although I guess the `v1` will actually drop it around the same time if my proposal is approved.

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,232 @@
...
+
+We will use following commands:
+
+```shell
+buildah build --file=/dockerfile/Dockerfile --tag=<IMAGE> --storage-driver=vfs
+buildah push --storage-driver=vfs --digestfile=/tmp/digest <ADDITIONAL_OPTS> <IMAGE>
+buildah images --digests --filter=digest=sha256:$(cat /tmp/digest) --format='{{.Name}}@{{.Digest}}' > /dev/termination-log
+```
+- `<IMAGE>` is placeholder for user desired name of the image (with registry, repository, and possibly tag)
+- `<ADDITIONAL_OPTS>` is placeholder for user desired additional options for the `buildah push` command
```
scholzj, line 64: So, the build does not need the additional options?

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,232 @@
...
+
+As in case of Kaniko, Buildah will have also some allowed options that can be specified during the `push` phase:
+- `--authfile` - path to the authentication file (usually created using the `buildah login`).
+- `--cert-dir` - path to directory with certificates that should be used for connecting to registries.
+- `--creds` - the `[username[:password]]` to use to authenticate with the registry if required.
+- `--retry` - number of times to retry in case of failure during image pull (the base one).
+- `--retry-delay` - duration between retry attempts.
+- `--tls-verify` - skip TLS verification for insecure registries.
+
+If the user-specified Buildah options will contain forbidden options (or not known), user will be notified by message inside the `.status` section of `KafkaConnect` resource, the `InvalidResourceException` will be thrown (logged inside the operator log) and the build will fail.
```
scholzj, line 99: So, you should probably just say that reconciliation will fail with `InvalidResourceException`. Or do you plan to add some special messages tot he status?

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,239 @@
...
+
+1. we will deprecate `additionalKanikoOptions` field as the name of it is connected directly to Kaniko and more generic name can be used (this field will be removed as part of the v1 API).
+2. we will add two new fields - `additionalBuildOptions` and `additionalPushOptions` - to cover additional options for both phases
+
+The `additionalKanikoOptions` will still be used with Kaniko until the `UseConnectBuildWithBuildah` feature gate moves to GA or until it is removed in the CRD v1 API, whatever comes first.
+An automated warning about deprecation will be added inside the `.status` section of the `KafkaConnect` CR in case it is used as for all other deprecated fields.
+Instead of `additionalKanikoOptions`, users should use `additionalBuildOptions`, but not `additionalPushOptions`.
+We will not split the Kaniko's additional options to two groups, and we will keep everything configured in one.
+That's because Kaniko doesn't have the build and push phases split, but everything is done in one command.
+The difference between the build and push options in Kaniko is that the `-pull` suffix is applied for options used only for the build phase.
```
scholzj, line 80: I don't really get the meaning here. Seems like it needs more detials or should be deleted?

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,239 @@
...
+
+Buildah, in the other hand, has these phases split and has its own set of options for both.
+In order to allow specifying different options and their values for both of these phases, we will add two fields.
+Additionally, if user will use Kaniko and specify both `additionalKanikoOptions` and `additionalBuildOptions`, `additionalBuildOptions` takes precedence.
+
+For Buildah, we will check only the `additionalBuildOptions` and `additionalPushOptions` fields.
+In case that user will specify `additionalKanikoOptions` with Buildah enabled feature gate, two conditions with warnings will be added into `.status` section of `KafkaConnect` CR - one about deprecation and one about `additionalKanikoOptions` being ignored, as Buildah is used.
```
scholzj, line 87: I don't think this needs any additional status warning. They will find out pretty quickly.

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,239 @@
...
+
+- `additionalPushOptions`
+  - `--authfile` - path to the authentication file (usually created using the `buildah login`).
+  - `--cert-dir` - path to directory with certificates that should be used for connecting to registries.
+  - `--creds` - the `[username[:password]]` to use to authenticate with the registry if required.
+  - `--retry` - number of times to retry in case of failure during image pull (the base one).
+  - `--retry-delay` - duration between retry attempts.
+  - `--tls-verify` - skip TLS verification for insecure registries.
+
+If the user-specified Buildah options will contain forbidden options (or not known), condition with warning will be added inside the `.status` section of `KafkaConnect` resource, the `InvalidResourceException` will be thrown (logged inside the operator log) and the build will fail.
```
scholzj, line 108: This makes no sense. You either ignore them with warning. Or you fail the reconciliation. Not both things. Failing the reconciliation will take care of the visibility without any additional warnings in logs or status.

What do we do for Kaniko? Fail everything or ignore with a warning?

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,239 @@
...
+
+Once the feature gate will be moved to beta, we change the environment variable from `STRIMZI_DEFAULT_KANIKO_EXECUTOR_IMAGE` to `STRIMZI_DEFAULT_BUILDAH_IMAGE` in our deployment files.
```
scholzj, line 110: This makes no sense. When you introduce this, you add the Buildah image variable. A feature gate moving to beta in general only changes the default state of the feature gate. The image variables stay there till the end. And when it moves to GA, you remove the Kaniko one.

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,239 @@
...
+    config.storage.replication.factor: -1
+    offset.storage.replication.factor: -1
+    status.storage.replication.factor: -1
+  build:
+    output:
+      type: docker
+      image: my-internal-registry:5000/repository/image:latest
+      additionalOptions:
+        - "--tls-verify=false"
+        - "--quiet"
```
scholzj, line 149: I guess this should be updated?

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,239 @@
...
+- `KafkaConnectAssemblyOperator`
+
+These classes contain the logic of the Connect Build feature, thus they are affected the most.
+We will need also changes to `DockerOutput` model and to the `KafkaConnect` CRD in order to support two new fields for specifying the additional Buildah options.
+
+## Compatibility
+
+Even though it's change in terms of the tool that builds the images, it will be gated behind the feature gate, so users can adapt to this change through few releases.
+Until the feature gate is GA, we will keep the Kaniko implementation in place, ensuring the backwards compatability.
+The `.spec.build.additionalKanikoOptions` field will be deprecated from the start, however it will be usable until the Buildah feature gate is moved to Beta phase.
```
scholzj, line 181:
```suggestion
The `.spec.build.additionalKanikoOptions` field will be deprecated from the start, however it will be usable until the Buildah feature gate is moved to Beta phase or removed when moving to the `v1` API.
```

scholzj (commented):
Looks good in general, but some more comments.

111-use-buildah-instead-of-kaniko.md
im-konge: I just wanted to mention that the options for Kaniko's `build` phase are suffixed with `-pull` for example. But I guess it doesn't make sense to have this sentence there. Let me delete it.

111-use-buildah-instead-of-kaniko.md
im-konge: I checked the validation method for Kaniko and there is just the exception:

```
        if (!forbiddenOptions.isEmpty())    {
            throw new InvalidResourceException(".spec.build.additionalKanikoOptions contains forbidden options: " + forbiddenOptions);
        }
```
so it fails the reconciliation. I will change the sentence accordingly.

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+This is more described in the [Running Buildah in restricted environment](#running-buildah-in-restricted-environment) section of this proposal.
+
+We will use the official Buildah images on Quay.io - currently `quay.io/buildah/stable:v1.40.1` - and similarly to Kaniko executor image - we will pull the official image, tag it with the version and push it to the Strimzi repository on Quay.
+That's useful because of our versioning and we will have the image ready even if someone on Buildah side decides to remove it from their Quay repository.
+The full name of our Buildah image will be `quay.io/strimzi/buildah:OPERATOR_VERSION` - for `main` branch it will use the `latest` tag, otherwise it will be tagged by the operator version for the particular release.
+
+Users then can specify their own Buildah image using `STRIMZI_DEFAULT_BUILDAH_IMAGE` (similarly to what is possible today with Kaniko).
+In case that users will use Kaniko, the Buildah environment variable will be ignored, and vice versa.
+The users will be responsible for changing these environment variables based on the mode in which they run the Connect build - in case that they have customized Deployment files.
```
scholzj, line 35: Is this the cause of the misunderstanding in https://github.com/strimzi/proposals/pull/170#discussion_r2347247825? I'm not sure how exactly you meant it. But both `STRIMZI_DEFAULT_BUILDAH_IMAGE` and `STRIMZI_DEFAULT_KANIKO_EXECUTOR_IMAGE` need to be set from day one.

111-use-buildah-instead-of-kaniko.md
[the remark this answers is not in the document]
scholzj: > > Have you tried multi-stage builds we use for Maven artifacts? Does that work?
>
> Will try that :)

Just to double check ... does this work? I assume you tried it?

scholzj (approved):

111-use-buildah-instead-of-kaniko.md
im-konge: Ah okay, so both should be in the Deployment file from day one, right? I understood it before that there should be just the Kaniko one and once the Buildah feature gate is moved to Beta, the Kaniko env variable will be replaced with the Buildah one. So yeah in case that there will be both, it doesn't make any sense to change it between the phases.

111-use-buildah-instead-of-kaniko.md
im-konge: This was previously applied comment that in case users will have their own Deployment files, they are reponsible for changing the env variables (for Buildah etc.). Plus I replied on the previous comment you had https://github.com/strimzi/proposals/pull/170#discussion_r2347454160

111-use-buildah-instead-of-kaniko.md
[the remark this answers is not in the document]
im-konge: I tried that before and it worked, I wanted to update the PoC and run all of our STs in order to see that everything will work as expected.

111-use-buildah-instead-of-kaniko.md
scholzj: Yes, both should be there from day one.

see-quick (approved):
LGTM 👍

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+These classes contain the logic of the Connect Build feature, thus they are affected the most.
+We will need also changes to `DockerOutput` model and to the `KafkaConnect` CRD in order to support two new fields for specifying the additional Buildah options.
+
+## Compatibility
+
+Even though it's change in terms of the tool that builds the images, it will be gated behind the feature gate, so users can adapt to this change through few releases.
+Until the feature gate is GA, we will keep the Kaniko implementation in place, ensuring the backwards compatability.
+The `.spec.build.additionalKanikoOptions` field will be deprecated from the start, however it will be usable until the Buildah feature gate is moved to Beta phase or removed when moving to the `v1` API.
+After that, the `additionalKaninkoOptions` field will be completely ignored and later (with v1 API) it will be removed.
+But we will give users enough time to change between `additionalKanikoOptions` and `additionalOptions` before we reach the Beta phase, so it shouldn't be a problem.
```
ppatierno, line 180: I guess it has to be replaced by `additionalBuildOptions` and `additionalPushOptions`?

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+Currently, it seems to not be a problem, but when some critical CVE or bug arises, we have to fix it.
+Also, we need a support in case that there will be something blocking us on newer versions of Kubernetes, which we would not get in this case anymore.
+
+## Proposal
+
+For the Connect Build feature on Kubernetes, we will use [Buildah](https://buildah.io/) as a replacement of Kaniko.
+Buildah is well-supported, widely used tool for building container images.
+It also appears that OpenShift’s Build API relies on Buildah, which makes Buildah a strong alternative to Kaniko, with the added assurance of long-term support and maintenance.
+Other than that, Buildah doesn't need any workaround in order to run it on Kubernetes rootlessly - the only thing that is needed during the build and push stages is to specify `--storage-driver=vfs` option.
+VFS (Virtual File System) driver ensures user-space implementation, it stores each layer as a full copy of the files, but without need of root permissions.
```
PaulRMellor, line 22:
```suggestion
The VFS (Virtual File System) driver provides a user-space implementation. It stores each layer as a full copy of the files, but does not require root permissions.
```

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+The only trade-off is the build time (it's slower) and consumption of disk space.
+
+Even though we will run Buildah with these options, which should make it rootless, it still needs some capabilities to run without issues during the build phase.
+That means we cannot run the build Pod with restricted Pod Security profile, as these capabilities are dropped in these profiles.
+This is more described in the [Running Buildah in restricted environment](#running-buildah-in-restricted-environment) section of this proposal.
+
+We will use the official Buildah images on Quay.io - currently `quay.io/buildah/stable:v1.40.1` - and similarly to Kaniko executor image - we will pull the official image, tag it with the version and push it to the Strimzi repository on Quay.
```
PaulRMellor, line 29:
```suggestion
We will use the official Buildah images on Quay.io (currently `quay.io/buildah/stable:v1.40.1`). As with the Kaniko executor image, we will pull the official image, tag it with the version, and push it to the Strimzi repository on Quay.
```

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+That's useful because of our versioning and we will have the image ready even if someone on Buildah side decides to remove it from their Quay repository.
```
PaulRMellor, line 30:
```suggestion
That's useful because of our versioning. We will have the image ready even if someone on the Buildah side decides to remove it from their Quay repository.
```

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+The full name of our Buildah image will be `quay.io/strimzi/buildah:OPERATOR_VERSION` - for `main` branch it will use the `latest` tag, otherwise it will be tagged by the operator version for the particular release.
+
+Users then can specify their own Buildah image using `STRIMZI_DEFAULT_BUILDAH_IMAGE` (similarly to what is possible today with Kaniko).
+In case that users will use Kaniko, the Buildah environment variable will be ignored, and vice versa.
+The users will be responsible for changing these environment variables based on the mode in which they run the Connect build - in case that they have customized Deployment files.
+
+The usage of Buildah will be gated behind feature gate called `UseConnectBuildWithBuildah` that will have following schedule:
```
PaulRMellor, line 37:
```suggestion
The usage of Buildah will be gated behind a feature gate called `UseConnectBuildWithBuildah` that will have following schedule:
```

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+
+| Alpha (opt-in) | Beta (default-on) | GA     |
+|----------------|-------------------|--------|
+| 0.49.0         | 0.52.0            | 0.55.0 |
+
+The build of the Connect image will be done the same way as today, the main difference will be in the commands used in the build and push process.
+Kaniko executor did everything at once, which can be done by Buildah as well, but I rejected to have it done in one command because:
```
PaulRMellor, line 44:
```suggestion
Kaniko executor did everything at once, which can be done by Buildah as well, but I decided not to implement it as a single command because:
```

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+
+- We will not get the SHA of image after the push.
+- Once pushed, the image will not be stored locally, so we cannot check the SHA using some different command.
+
+The SHA of the image is needed to keep the compatibility with the previous implementation using Kaniko.
+Kaniko, once the image is build and pushed, is able to return the full name of the image together with the SHA to particular path.
```
PaulRMellor, line 50:
```suggestion
Kaniko, once the image is built and pushed, is able to return the full name of the image together with the SHA to a specified path.
```

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+That is done using the `--image-name-with-digest-file` option of the Kaniko executor.
+Unfortunately, Buildah doesn't have such option in case that you want to build and push the image in one command (that is done using the `buildah build` command, in order to push directly, you need to prefix the image with `docker://`).
+Buildah has option to store the SHA to specify file on path using `--digestfile` option - but only when `push` command is used.
```
PaulRMellor, line 53:
```suggestion
Buildah has an option to store the SHA in a specified file using the `--digestfile` option, but only when the `push` command is used.
```

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+1. we will deprecate `additionalKanikoOptions` field as the name of it is connected directly to Kaniko and more generic name can be used (this field will be removed as part of the v1 API).
+2. we will add two new fields - `additionalBuildOptions` and `additionalPushOptions` - to cover additional options for both phases
+
+The `additionalKanikoOptions` will still be used with Kaniko until the `UseConnectBuildWithBuildah` feature gate moves to GA or until it is removed in the CRD v1 API, whatever comes first.
+An automated warning about deprecation will be added inside the `.status` section of the `KafkaConnect` CR in case it is used as for all other deprecated fields.
+Instead of `additionalKanikoOptions`, users should use `additionalBuildOptions`, but not `additionalPushOptions`.
+We will not split the Kaniko's additional options to two groups, and we will keep everything configured in one.
+That's because Kaniko doesn't have the build and push phases split, but everything is done in one command.
+
+Buildah, in the other hand, has these phases split and has its own set of options for both.
```
PaulRMellor, line 81:
```suggestion
Buildah, on the other hand, has these phases split and has its own set of options for both.
```

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+    offset.storage.replication.factor: -1
+    status.storage.replication.factor: -1
+  build:
+    output:
+      type: docker
+      image: my-internal-registry:5000/repository/image:latest
+      additionalBuildOptions:
+        - "--tls-verify=false"
+      additionalPushOptions:
+        - "--quiet"
```
PaulRMellor, line 146: I don't see option this in the list above?

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+          - type: maven
+            group: org.apache.kafka
+            artifact: connect-file
+            version: 4.1.0
+```
+
+### Running Buildah in restricted environment
+
+To successfully run Buildah in container on Kubernetes, the Pod needs to have some capabilities.
+The capabilities are `cap_kill`, `cap_setgid`, and `cap_setuid`, which are used in the `build` phase.
```
PaulRMellor, line 159: Are these the only options? Or only required options?
Should we describe why and what these options are for?

PaulRMellor, line 159:
```suggestion
The required capabilities are `CAP_KILL`, `CAP_SETGID`, and `CAP_SETUID,` which are used during the build phase.”
```

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+In case that user will use restricted Pod Security profile, Buildah will not work - as these capabilities are dropped in this profile.
+That is similar to Kaniko - which is not able to run in this restricted profile as well.
+Additionally, that is a reason why we will not implement Buildah for OpenShift - and it's described as [one of the rejected alternatives](#only-one-implementation-of-connect-build---buildah-on-both-kubernetes-and-openshift).
+
+## Affected/not affected projects
+
+The one and only affected project is `strimzi-kafka-operator` repository, especially following classes:
```
PaulRMellor, line 166:
```suggestion
The only affected project is the `strimzi-kafka-operator` repository, particularly the following classes:
```

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+- `KafkaConnectBuild`
+- `ConnectBuildOperator`
+- `KafkaConnectAssemblyOperator`
+
+These classes contain the logic of the Connect Build feature, thus they are affected the most.
+We will need also changes to `DockerOutput` model and to the `KafkaConnect` CRD in order to support two new fields for specifying the additional Buildah options.
+
+## Compatibility
+
+Even though it's change in terms of the tool that builds the images, it will be gated behind the feature gate, so users can adapt to this change through few releases.
```
PaulRMellor, line 176:
```suggestion
Even though it's a change in terms of the tool that builds the images, it will be gated behind the feature gate, so users can adapt to this change through few releases.
```

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+However, when I tried it on OpenShift, I found out that we would need to manage multiple other resources and grant more roles to the Cluster Operator than we have today.
+For OpenShift, the Cluster Operator would have to create two `RoleBinding`s for the Connect build's `ServiceAccount` - one for granting the `image-builder` role (for getting credentials to push to the internal OpenShift registries), and one for granting the `anyuid` security context.
+That would mean that Cluster Operator would need another `ClusterRole` that would grant it permissions to `bind` these roles to the `ServiceAccount`.
+
+The `anyuid` SCC is needed for Buildah during the image build.
+I don't know the exact internals, but based on my findings, Buildah needs to create an user namespace and also manipulate with UIDs.
+The `anyuid` SCC usage and why it's needed is briefly described in the [Buildah's tutorial about running Buildah on OCP](https://github.com/containers/buildah/blob/main/docs/tutorials/05-openshift-rootless-build.md#create-service-account-for-building-images).
```
PaulRMellor, line 194:
```suggestion
The use of the `anyuid` SCC and why it's needed is briefly described in [Buildah's tutorial about running Buildah on OCP](https://github.com/containers/buildah/blob/main/docs/tutorials/05-openshift-rootless-build.md#create-service-account-for-building-images).
```

111-use-buildah-instead-of-kaniko.md
```diff
@@ -0,0 +1,236 @@
...
+
+The alternative of using Image Volumes feature is connected to the previous section.
+As I already mentioned, the Image Volumes feature is not widespread, and it's not supported by all Kubernetes versions Strimzi supports.
+Even if it's possible solution for the future, we need to wait until it's used and supported widely before moving to it.
+
+### Using other tools like BuildKit
+
+During implementing and testing the PoC with Buildah, I tried also different tools like BuildKit.
+BuildKit is mainly rootful, but it can be used rootless.
+However, in order to build an image inside the container, it needs some hacking and workarounds, which are not that straighforward as in case of Buildah.
```
PaulRMellor, line 235:
```suggestion
However, in order to build an image inside the container, it needs some hacking and workarounds, which are not that straightforward as in case of Buildah.
```

111-use-buildah-instead-of-kaniko.md
PaulRMellor: What's the upgrade policy?

111-use-buildah-instead-of-kaniko.md
im-konge: But I will have a look if there is some information for what are those being used. Good point Paul, thanks

111-use-buildah-instead-of-kaniko.md
im-konge: Good point, it seems that I accidentally removed it 🤦 . Thanks for pointing it out.

111-use-buildah-instead-of-kaniko.md
scholzj: I'm not sure what is meant with upgrade policy in this case.

* Given the comment is at the line talking about the Buildah version we will use, does it mean `when we will upgrade Buildah to use a newer version?` @PaulRMellor? In which case the answer probably should be that we should aim to use the latest stable version (if nothing else for the CVE fixes etc.). This is also what we used for Kaniko. It would be worth adding to the proposal.

* But @im-konge seems to be talking about Strimzi upgrade and FG changes. So I'm not sure if that is based on some more insight about how the comment was meant. That is a bit harder to answer, as we do not know for sure how users are using it. My assumption would be that most users are using it without any additional options. In such case, when the feature gate moves, it should hopefully be completely transparent for them, and they might not even know that we now use Buildah instead of Kaniko. If they already had some special requirements with Kaniko and needed to use the additional options (e.g. because of using unsigned certificates etc.), they would likely need to figure out what options will they need for Buildah and switch. This would need to be done manually by them. So we should definitely promote this change in the release notes (and CHANGELOG), in the _What's new_ videos or blog posts, and encourage them to test it early.

----

An alternative to the feature gate might be to add a new type `type: buildah` and deprecate/remove `type: docker` (representing Kaniko). But I think this would not bring any advantages. It would force everyone to change their custom resources. While with the feature gate, we hope many users might not need to do anything and not even notice it. So I think the FG is a better choice.

Frawless (approved):

111-use-buildah-instead-of-kaniko.md
PaulRMellor: Yes, I meant for Buildah in this instance

im-konge:
This proposal has now 5 binding +1 votes. I will wait until tomorrow (Friday) EOD before merging this (and changing the file name + README).

im-konge:

PR #171, opened 2025-08-21
Add proposal for supporting SSL for KafkaBridge

tinaselenge:
POC:
https://github.com/strimzi/strimzi-kafka-bridge/compare/main...tinaselenge:strimzi-kafka-bridge:ssl-support?expand=1
https://github.com/strimzi/strimzi-kafka-operator/compare/main...tinaselenge:bridge-ssl?expand=1 

Related to https://github.com/strimzi/strimzi-kafka-bridge/issues/939


1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,166 @@
...
+## Proposal
+
+This proposal adds new configurations to enable SSL server and specify locations of certificate and key files in PEM format so that client connectionts to HTTP Bridge is encryped. The new configurations will be added to the current `http` prefixed configurations:
+- http.ssl.enable
+- http.ssl.keystore.location
+- http.ssl.keystore.key.location
+- http.ssl.protocol
+- http.ssl.enabled.protocols
+
+When `http.ssl.enable` is set to true, Kafka Bridge server would be started with ssl enabled and the key and certificate loaded from the locations defined by `http.ssl.keystore.location` and `http.ssl.keystore.key.location` configurations. If `http.port` configuration is not set, the server port will be set to `8443` by default. This will allow users to make HTTPS connections from their clients, and the server would not accept unsecured connections.
```
scholzj, line 18: I think this requires a bit more consideration ...
* Can we really afford to have only the TLS listener? What about healthchecks, metrics, etc.
* If we have both PLAIN and TLS listener, what about the security (not jsut TLS, but in the future also Auth and so on)?
* So maybe we need to split-of metrics and healthchecks into a separate listener in the first place?

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,166 @@
...
+
+Users can enable SSL by setting the new field `spec.http.sslConfigurations`. Under this new field, they have to set `generateCertficiate` field that accepts boolean value and is set to false by default. If it is not set in the CR, that means the user chooses to provide their own certificate via Kubernetes secret, in which case, they have to set `certificateAndKey` to refer to the secret that contains the certificate and key. 
+
+If `generateCertificate` is set true, the operator will issue the certificate using the same CA for their Kafka cluster. When operator is reconciling KafkaBridge resource, it is not aware of which Kafka cluster the HTTP Bridge is connected to. So if the Kafka cluster is running in a different namespace, therefore the CA secrets are not in the same namespace as the KafkaBridge resource, the operator needs to know where to fetch them. In this case users have to also set `caNamespace`. However, if their Kafka cluster and KafkaBridge resources are running in the same namespace, they don't need to set this field, as the operator would look for them in the namespace of the KafkaBridge resource by default.
```
scholzj, line 26: I personally would definitely avoid this.
* The resources were designed as independent and IMHO they should stay that way
* How will you handle situations when the cluster is deleted etc.?
* How will you know which cluster to pick from given namespace?
* How can you guarantee that this is _future proof_ with regards to the future Cert Manager support in the Kafka CR? Also, some of the thoughts in the past were that after/if we manage to get proper support for Cert Manager, we would be able to deprecate and remove our own CA management anyway.

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,166 @@
...
+This proposal adds new configurations to enable SSL server and specify locations of certificate and key files in PEM format so that client connectionts to HTTP Bridge is encryped. The new configurations will be added to the current `http` prefixed configurations:
+- http.ssl.enable
+- http.ssl.keystore.location
+- http.ssl.keystore.key.location
+- http.ssl.protocol
+- http.ssl.enabled.protocols
+
+When `http.ssl.enable` is set to true, Kafka Bridge server would be started with ssl enabled and the key and certificate loaded from the locations defined by `http.ssl.keystore.location` and `http.ssl.keystore.key.location` configurations. If `http.port` configuration is not set, the server port will be set to `8443` by default. This will allow users to make HTTPS connections from their clients, and the server would not accept unsecured connections.
+
+If `http.ssl.protocol` is configured, the protocol will be added to the current or default list of enabled protocols. If `http.ssl.enabled.protocols` is configured, the list will replace the default list of enabled secure transport protocols that Vertx sets.
```
scholzj, line 20: Aren't there some existing Vert.x options to configure this? If not, do we need the same options also for cipher suites? As that is another common thing users configure.

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,166 @@
...
+spec:
+  replicas: 3
+  bootstrapServers: <cluster_name>-cluster-kafka-bootstrap:9092
+  # HTTP configuration (required)
+  http:
+    port: 443
+    sslConfigurations:
+      # This will override the default list of enabled protocols.
+      # As of Vertx version 5.0.3, the default list is  "TLSv1, TLSv1.1, TLSv1.2, TLSv1.3".
+      sslEnabledProtocols: "TLSv1.2,TLSv1.3"
```
scholzj, line 127: I understand we probably want dedicated API for seleting the keystore. But do we want it also for all the other options? Why not keep things such as protocol configuration for a regular `config` section?

scholzj (commented):
Just some quick thoughts.

scholzj:
The way we setup the certificates seems also important to whether we ever expect to have some better scalability or not I guess.

1xx-enable-ssl-for-kafka-bridge.md
tinaselenge: Thanks for the feedback Jakub. 
> Can we really afford to have only the TLS listener? What about healthchecks, metrics, etc.
>So maybe we need to split-of metrics and healthchecks into a separate listener in the first place?

You right, we have to consider the metrics and health endpoints for monitoring and readiness/liveness. As you suggested, we can have both PLAIN and TLS listener and then when ssl enabled, the PLAIN listener is configured for only `/healthy`, `/ready` and `/metrics` endpoints and the TLS listener is for the other endpoints. 

> If we have both PLAIN and TLS listener, what about the security (not jsut TLS, but in the future also Auth and so on)?
I was going to scope the auth part out of this proposal but perhaps should mention different auth mechanism we could support such as HTTP Basic Authentication and JWT in the future improvement. I will make an update on this.

1xx-enable-ssl-for-kafka-bridge.md
tinaselenge: Good points! Tbh, I have gone back and forth on how we would or should do certificate management for another CR, since this is not something we do today. We only manage certificates for our internal components such as CC, kafka-exporter etc. So what you are suggesting is that, the users would have to provide the certAndKey and manage their own certificate, at least for now? When Cert Manager support is completed, perhaps we could see how we can use it to issue a certificate for Bridge resource in the future? This would certainly simplify the proposal more.

1xx-enable-ssl-for-kafka-bridge.md
tinaselenge: So that it can be used for Kafka clients as well?

1xx-enable-ssl-for-kafka-bridge.md
scholzj: > I was going to scope the auth part out of this proposal but perhaps should mention different auth mechanism we could support such as HTTP Basic Authentication and JWT in the future improvement. I will make an update on this.

Right. But even with auth being out of scope for this proposal, we need to think about it a bit. What my concern is is the following ... if we now create two interfaces -> plain and TLS with the idea that the plain one will be used for healthchecks and metrics, it does not seem to work long time. Because you will face the same issue once again:
* If you enable authentication on the plaintext interface, healcthchecks and metrics will stop working
* Having a plaintext interface always enabled for metrics and healthchecks, means you cannot enforce TLS either.

So, maybe, we need in the first lace to separate healthchecks and metrics from the regular listeners and move them to completely separate endpoint.

1xx-enable-ssl-for-kafka-bridge.md
scholzj: No, I'm more concerned that overtime you would have lot of different fields here in the API with different options just to cover everyones requirement. Having it in ther `config` makes this easier as that is just a `Map`.

1xx-enable-ssl-for-kafka-bridge.md
ppatierno: I agree on moving healthchecks and metrics endpoints to a different HTTP plain interface. It should be the same for the `/openapi` endpoint as well. Said that if we are allowing two different interfaces, PLAIN and TLS, we would also need an additional `http.ssl.port` (other than the `http.port`). Also ... should we make the port related to the healthchecks, metrics and openapi endpoint being configurable as well? If we hard code such port, it could be a problem for users who run bridge on bare metal/VMs and that port is already owned by something else. Maybe an additional `http.management.port` (or something similar better than `management`? But it seems to be used in other frameworks, i.e. SpringBoot AFAIU). Of course it's going to have an impact on the operator as well to configure the additional port and the Kube probes.

1xx-enable-ssl-for-kafka-bridge.md
ppatierno: Also taking into account that the Kafka cluster could be even running on a totally different Kube instance, cloud, bare metal/VMs and so on (not sure about the point of doing that but it's possible).
At the same time I could expect users asking for some automation, at least when the Kafka cluster is within the same namespace of the bridge (which is most of the times?). Should maybe the bridge have an additional annotation like `strimzi.io/cluster: my-cluster` as we have on KafkaTopic and KafkaUser to make clear the cluster it's referring to. Maybe making the `bootstrapServers` field optional in such a case. This way the operator would know which CA to use to generate a cert for the bridge. But in general it depends on when ceret-manager will land, because this part of the code could become obsolete soon.

tinaselenge:
> The way we setup the certificates seems also important to whether we ever expect to have some better scalability or not I guess.

With this proposal, the same certificate and key (via referenced secret in the CR) will be volume mounted into each KafkaBridge pod if it is scaled to multiple replicas, which is not different than how TLS authentication is configured for KafkaBridge talking to Kafka. My understanding is that KafkaBridge is not truly scalable in terms of consumers, but I'm not sure what is the expectation at the moment, in terms of improving that and what to consider for the certificate setup.

1xx-enable-ssl-for-kafka-bridge.md
tinaselenge: I have added `http.ssl.port` but left the management endpoints to carry on using the http.port so that it doesn't have an impact on kube probes and compatibility issue.

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,112 @@
...
+- http.ssl.remove.cipher.suite
+
+When `http.ssl.enable` is set to true, HTTP Bridge server will be started with SSL enabled and the key and certificate loaded from the locations defined by `http.ssl.keystore.location` and `http.ssl.keystore.key.location` configurations. If `http.ssl.port` configuration is not set, the server port will be set to `8443` by default. This will allow users to make TLS encrypted connections from their clients for all the HTTP Bridge endpoints for client operations. However, connections to the following endpoints for monitoring will stay unencrypted:
+- `/healthy`
+- `/ready`
+- `/metrics`
+- `/openapi`
+- `/openapiv2`
+- `/openapiv3`
+- `/info`
```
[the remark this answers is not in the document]
ppatierno: I think with option 2 we have an opportunity to separate administrative endpoints from operational ones. Even scraping metrics on the same HTTP interface where you are sending and receiving messages could have an impact on performance. Separating the HTTP listeners could be better (and this is even not related to the TLS addition). Anyway let's see what others think.

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,112 @@
...
+
+When `http.ssl.enable` is set, these endpoints will have a separate listener, that is always PLAINTEXT. It will use the port defined by the existing `http.port` that is set to `8080` by default. The purpose of these endpoints is to be exposed only internally for the health checks, getting OpenAPI specification and metrics collection. In the future, we are likely to implement authentication and authorization for HTTP Bridge. By having separate listeners, these monitoring endpoints will not be affected by such changes.
+
+If `http.ssl.protocol` is configured, the protocol will be added to the current or default list of enabled protocols. If `http.ssl.enabled.protocols` is configured with a comma separated list, it will replace the default list of enabled secure transport protocols that Vertx sets.
+
+Vertx uses the default cipher suites provided by the underlying JDK SSL/TLS engine and allows users to append a cipher suite to the ordered suites or remove one from it. HTTP Bridge users can use the new configurations `http.ssl.add.cipher.suite` and `http.ssl.remove.cipher.suite` to add or remove a cipher suite from the default list of enabled cipher suites.
```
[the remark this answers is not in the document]
ppatierno: Like for the SSL protocols, we can try to have something similar to Kafka with just `ssl.cipher.suites` which is a comma separated list and if not set it defaults to the available ones (which has to be seen what it exactly means for Vert.x).?

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,112 @@
...
+- `/ready`
+- `/metrics`
+- `/openapi`
+- `/openapiv2`
+- `/openapiv3`
+- `/info`
+
+When `http.ssl.enable` is set, these endpoints will have a separate listener, that is always PLAINTEXT. It will use the port defined by the existing `http.port` that is set to `8080` by default. The purpose of these endpoints is to be exposed only internally for the health checks, getting OpenAPI specification and metrics collection. In the future, we are likely to implement authentication and authorization for HTTP Bridge. By having separate listeners, these monitoring endpoints will not be affected by such changes.
+
+If `http.ssl.protocol` is configured, the protocol will be added to the current or default list of enabled protocols. If `http.ssl.enabled.protocols` is configured with a comma separated list, it will replace the default list of enabled secure transport protocols that Vertx sets.
```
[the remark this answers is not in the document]
tinaselenge: From reading the [Vertx Java doc](https://vertx.io/docs/apidocs/io/vertx/core/net/SSLOptions.html), I don't think these 2 methods are used differently. It is either used to overwrite the default list of protocols or append a protocol to the ordered list of protocols Vertx server would accept. This means we cannot mimic `ssl.protocol` of Kafka, so I think your suggestion of having just `http.ssl.enabled.protocols` makes sense. 

[`"TLSv1", "TLSv1.1", "TLSv1.2", "TLSv1.3"`](https://vertx.io/docs/apidocs/io/vertx/core/net/SSLOptions.html#DEFAULT_ENABLED_SECURE_TRANSPORT_PROTOCOLS) is the default list of protocols Vertx uses. Do we need to set this to the default of Kafka by default?

1xx-enable-ssl-for-kafka-bridge.md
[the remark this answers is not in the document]
tinaselenge: I think that would make sense. There is no bulk setter for ciphers suites, but we can take a list of cipher suites, loop through them with addEnabledCipherSuite() method. Now thinking about it, `removeEnabledCipherSuite` is probably not that useful, as it's something we set up before starting the server. 

>  (which has to be seen what it exactly means for Vert.x)?
It seems like Vertx does not set any defaults so probably defaults provided by the JDK is used.

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,110 @@
...
+When `http.ssl.enable` is set to true, HTTP Bridge server will be started with SSL enabled and the key and certificate loaded from the locations defined by `http.ssl.keystore.location` and `http.ssl.keystore.key.location` configurations. If `http.ssl.port` configuration is not set, the server port will be set to `8443` by default. This will allow users to make TLS encrypted connections from their clients for all the HTTP Bridge endpoints for client operations. However, connections to the following endpoints for monitoring will stay unencrypted:
+- `/healthy`
+- `/ready`
+- `/metrics`
+- `/openapi`
+- `/openapiv2`
+- `/openapiv3`
+- `/info`
+
+When `http.ssl.enable` is set, these endpoints will have a separate listener, that is always PLAINTEXT. It will use the port defined by the existing `http.port` that is set to `8080` by default. The purpose of these endpoints is to be exposed only internally for the health checks, getting OpenAPI specification and metrics collection. In the future, we are likely to implement authentication and authorization for HTTP Bridge. By having separate listeners, these monitoring endpoints will not be affected by such changes.
```
scholzj, line 32: As I said before, this just postpones the issue for when authentication is introduced. I think the proposal should be more forward-thinking and use a completely separate port for the metrics and healthchecks.

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,110 @@
...
+- http.ssl.enabled.cipher.suites
+
+When `http.ssl.enable` is set to true, HTTP Bridge server will be started with SSL enabled and the key and certificate loaded from the locations defined by `http.ssl.keystore.location` and `http.ssl.keystore.key.location` configurations. If `http.ssl.port` configuration is not set, the server port will be set to `8443` by default. This will allow users to make TLS encrypted connections from their clients for all the HTTP Bridge endpoints for client operations. However, connections to the following endpoints for monitoring will stay unencrypted:
+- `/healthy`
+- `/ready`
+- `/metrics`
+- `/openapi`
+- `/openapiv2`
+- `/openapiv3`
+- `/info`
```
scholzj, line 30: Maybe it is worth explaining why should these not be on the SSL listener as well. I guess I can understand that for metrics and possibly the ready / healthy endpoints. Not so sure about the OpenAPI endpoints and the /info endpoint. Wouldn't they be useful for clients as well?

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,110 @@
...
+Let's take a look at an example of KafkaBridge CR with the new fields:
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaBridge
+metadata:
+  name: my-bridge
+spec:
+  replicas: 3
+  bootstrapServers: <cluster_name>-cluster-kafka-bootstrap:9092
+  http:
```
scholzj, line 51: Why not have a separate section `https`?

1xx-enable-ssl-for-kafka-bridge.md
ppatierno: Currently `/info` provides only bridge version, while the OpenAPI endpoints provide the full OpenAPI YAML definition which is useful in case of integration with API management systems like 3scale (a use case I worked in the past) for dynamic discovery of the operations. Having them behind the SSL could make such integration harder (or break it) if the other side doesn't allow SSL connection first.

scholzj: I don't know TBH. That was how it started before I started digging into why are we removing these from 8080/8443. As far as I'm concerned, we can do just healthcecks and metrics.

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,137 @@
...
+
+The connection between HTTP Bridge and Kafka cluster can be secured however, currently there is no support for securing the connections to HTTP Bridge from clients. The current recommendation is to use firewalls or API gateways to secure the client connections. This requires users to make additional configurations and does not provide the best user experience. 
+
+Moreover, some of the endpoints are for internal use only, such as `/healthy`, `/ready` and `/metrics` but they are exposed on the same listener and the same port as the other endpoints that are used by external clients. This means the endpoints for internal use only are also exposed to external clients. And if SSL is enabled, it will introduce more operational complexity as these internal endpoints would need to be configured with TLS and certificates.
+
+## Proposal
+This proposal also adds new configurations to enable SSL server and specify locations of certificate and key files in PEM format so that client connections to HTTP Bridge can be encrypted. The new configurations will be added to the current `http` prefixed configurations:
+- http.ssl.enable
+- http.ssl.keystore.location
+- http.ssl.keystore.key.location
```
scholzj, line 19: What is the difference between `http.ssl.keystore.location` and `http.ssl.keystore.key.location`? Also, should we:
* Have some type field? What formats will we support? You seem to mention PEM only. But given how common JKS/PKCS12 is in Java land, we should be at least ready to support them. They would also need a password field I guess.
* Have some field to pass the keystore public and private key directly? E.g. `http.ssl.keystore.key` and `http.ssl.keystore.certificate.chain`?
* Would it make more sense to have the options named as `https`? E.g. `https.enable`?

1xx-enable-ssl-for-kafka-bridge.md
scholzj: Why will we use `http.port`? Why not allow using both HTTP and HTTPs in parallel?

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,137 @@
...
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaBridge
+metadata:
+  name: my-bridge
+spec:
+  replicas: 3
+  bootstrapServers: <cluster_name>-cluster-kafka-bootstrap:9092
+  http:
+    port: 8443
+    adminPort: 8081 #NEW FIELD
```
scholzj, line 55: Should this be really configurable? I would probably argue against it. I don't think similar ports are configurable in other operands.

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,137 @@
...
+    sslEnable: true #NEW FIELD
+    certificateAndKey: #NEW FIELD
+      secretName: my-secret
+      certificate: public.crt
+      key: private.key
+...
+```
+The liveness and readiness probes will connect to the port defined by `adminPort` (or 8080 if not set).
```
scholzj, line 63: Should this be 8081 instead as that is the default for the admin port?

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,137 @@
...
+Moreover, some of the endpoints are for internal use only, such as `/healthy`, `/ready` and `/metrics` but they are exposed on the same listener and the same port as the other endpoints that are used by external clients. This means the endpoints for internal use only are also exposed to external clients. And if SSL is enabled, it will introduce more operational complexity as these internal endpoints would need to be configured with TLS and certificates.
+
+## Proposal
+This proposal also adds new configurations to enable SSL server and specify locations of certificate and key files in PEM format so that client connections to HTTP Bridge can be encrypted. The new configurations will be added to the current `http` prefixed configurations:
+- http.ssl.enable
+- http.ssl.keystore.location
+- http.ssl.keystore.key.location
+- http.ssl.enabled.protocols
+- http.ssl.enabled.cipher.suites
+- http.admin.port
```
scholzj, line 22: It does not really have any admin options. Should we try to find some better name for it?

1xx-enable-ssl-for-kafka-bridge.md
tinaselenge: sorry my mistake, it is 8081.

1xx-enable-ssl-for-kafka-bridge.md
scholzj: Probably `http.internal.port` or `http.healthmetrics.port` makes most sense. @ppatierno WDYT?

ppatierno: I see `http.healthmetrics.port` too much specific. I liked the `management` one but I guess it's going to be refused because too similar to `admin`. I would go with `internal` at this point ... even if it's not so clear what actually it provides (it just says, it's internal).

scholzj: I get the ugliness of all the alternatives we have here.  But do we have really anything better? I mean it does not do any management, or? Would it do some management one day?

1xx-enable-ssl-for-kafka-bridge.md
ppatierno: > But they are odd ones to be configured under http prefix when running a standalone bridge. I think they should have been under kafka prefix because will we need to enable consumer or producer specifically only for HTTP or HTTPS?

The `kafka.` prefix is used for all configuration related to the internal Kafka consumer, producer and admin. Any configuration from there, is just used to configure the corresponding Kafka clients.
The `http.consumer.enabled` and `http.producer.enabled` represents the HTTP part of them. I remember it was a community ask to be able to disable consumer or producer HTTP endpoint (in order to get `503` `Service Unavailable`) if not enabled.

> If we allow running both, will there users who would want to run only HTTPS ? If so we would need to have an option to disable HTTP.

I would agree with Jakub. I think for security concerns we'll have for sure people enabling HTTPS and ... disabling HTTP (which is actually not supported). What I am not sure is if people we'll enable them both at the same time.
I think that compared with the Kafka listeners we had more problems because we have much more integration within Kubernetes, so you can have Ingress, Routes, etc etc ...
For the bridge, it can run as standalone as well, so I don't think we can have more than HTTP and HTTPS. Everything you want to expose it outside of Kubernetes is up to you, even the operator doesn't provide anything (just an internal Service).

1xx-enable-ssl-for-kafka-bridge.md
ppatierno: I was taking a look around for other project and it looks like Spring and Quarkus are using the `management` term for this kind of thing. AFAIU Quarkus has a `management` port exposing health, metrics, info and openapi (very similar to the ones we have) https://quarkus.io/guides/management-interface-reference and https://www.baeldung.com/quarkus-management-interface
Spring seems to expose much more still through a `management` port.
At this point, could this name be considered quite common if used by this very well know development framework? Also should we revisit the presence of /info and /openapi on it?

scholzj: Ok, I guess I can live with it if others use the same bad name as well :-). Maybe one day we will add some management to it, some dynamic reconfiguriation or something 🤷

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,148 @@
...
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaBridge
+metadata:
+  name: my-bridge
+spec:
+  replicas: 3
+  bootstrapServers: <cluster_name>-cluster-kafka-bootstrap:9092
+  http:
+    port: 8443
+    sslEnable: true
```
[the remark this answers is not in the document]
im-konge: Yeah that's something what I was thinking about as well.

scholzj (commented):
Few more nits. Mostly formatting. The only real point for discussion is about the SSL protocols. Looks good otherwise.

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,148 @@
...
+      config: #NEW FIELD
+        ssl.enabled.protocols: "TLSv1.1, "TLSv1.2", "TLSv1.3"
+        ssl.enabled.cipher.suites: "TLS_DHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384"
+...
+```
+
+## Future Improvements
+
+### Support more formats for the keystore
+We can support formats such as `PKSC12` and `JKS` as they are common for Java applications. If we support these formats, then we would need to add the following configurations:
```
PaulRMellor, line 96:
```suggestion
We can support formats such as `PKCS12` and `JKS` as they are common for Java applications. If we support these formats, then we would need to add the following configurations:
```

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,148 @@
...
+- http.ssl.keystore.type
+- http.ssl.keystore.password
+
+### Support mTLS
+In the future, we can support mutual TLS for the client connections to HTTP Bridge as well. This could be implemented by adding a field `trustedCertificates`, for example:
```
PaulRMellor, line 101:
```suggestion
In the future, we can support mutual TLS for the client connections to HTTP Bridge as well. This could be implemented by adding a new `trustedCertificates` property in the `tls` section, for example:
```

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,148 @@
...
+        certificate: public.crt
+        key: private.key
+      trustedCertificates: #NEW FIELD
+        - secretName: my-cluster-cluster-cert
+          pattern: "*.crt"
+        - secretName: my-cluster-cluster-cert
+          pattern: "*.crt"    
+...
+```
+This is not in the scope of this proposal, as we will likely to support multiple authentication mechanisms in the future.
```
PaulRMellor, line 125:
```suggestion
This is not in the scope of this proposal, but we are likely to support multiple authentication mechanisms in the future.
```

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,148 @@
...
+
+In the future, we are also likely to implement authentication and authorization for HTTP Bridge. By having separate listeners, the internal endpoints will not be affected by such changes and will make it easier to implement them.
```
PaulRMellor, line 127:
```suggestion
We also plan to implement authentication and authorization for the HTTP Bridge. By using separate listeners, internal endpoints are isolated from these changes, making implementation simpler and less disruptive.
```

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Automating certificate management
+
+Users can choose to allow the operator to manage the certificate instead of providing their own certificate and key. This can be implemented by adding another boolean field on the KafkaBridge CR such as `generateCertificate`. The operator then could issue the certificate using the same CA for their Kafka cluster as we do today for internal components such as Cruise Control and Kafka Exporter, however with the upcoming support for cert-manager, the way the operator manages certificates will change. This is the reason we are leaving this feature for future improvement.
```
PaulRMellor, line 131:
```suggestion
Users can choose to allow the operator to manage the certificate instead of providing their own certificate and key. This can be implemented by adding another boolean field on the `KafkaBridge` CR such as `generateCertificate`. The operator could issue certificates using the same CA for their Kafka cluster, as is currently done for internal components like Cruise Control and Kafka Exporter. However, with upcoming support for cert-manager, certificate management is expected to change. For this reason, we are deferring this feature for implementation in the future.
```

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Affected/not affected projects
+
+- strimzi-kafka-bridge
+- strimzi-cluster-operator
+
+## Compatibility
+
+This will introduce a breaking change for users who collect metrics on port 8080 or the port defined by `http.port` configuration. Also users who run Bridge on baremetals and do their own manual healtchecks using `/healthy` and `/ready` endpoints will be affected as well. These users have to reconfigure their healthcheck and monitoring system to use port 8081 or reconfigure their Bridge server with `http.management.port`.
```
PaulRMellor, line 140:
```suggestion
This will introduce a breaking change for users who collect metrics on port 8080 or the port defined by `http.port` configuration. Also users who run the HTTP Bridge on bare metal and do their own manual health checks using `/healthy` and `/ready` endpoints will be affected as well. These users have to reconfigure their health check and monitoring system to use port 8081 or reconfigure their Bridge server with `http.management.port`.
```

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Rejected alternatives
+
+### Automating certificate management as part of this proposal
+
+- Currently the operator does not manage certificates for other custom resources, as they are designed to be independent, it only manages certificates for its internal components such as Cruise Control and Kafka Exporter, that are deployed as part of the Kafka custom resource. This will introduce a new behaviour to the operator and would require taking other custom resources such as KafkaConnect into consideration as well, because we would potentially need to manage certificates for it too. We are currently in the process of refactoring how the operator manages certificates to support external certificate manager as proposed [here](https://github.com/strimzi/proposals/pull/135), therefore it is not the right time to introduce such new behaviour, as it would not be future proof. Instead, it was decided to leave it for future improvement.
```
PaulRMellor, line 146:
```suggestion
- Currently, the Cluster Operator does not manage certificates for other custom resources, as they are designed to operate independently. It only handles certificates for internal components, such as Cruise Control and Kafka Exporter, which are deployed as part of the `Kafka` custom resource. Extending certificate management to other resources, like `KafkaConnect`, would introduce new behavior and require broader changes. Since we’re looking at refactoring certificate handling to support external managers (see the proposal for the [External Certificate Manager](https://github.com/strimzi/proposals/pull/135)), now is not the right time to introduce this change. To ensure future compatibility, we’ve deferred it as a potential improvement.
```

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,148 @@
...
+
+- Continue using a single interface and only use separate listeners when SSL is enabled so that we don't introduce a breaking change to existing users. Given that we will be introducing the first major version of Bridge, introducing a breaking change is acceptable and it makes much more sense to have separate listeners for better separation of concerns regardless of TLS. In the future, we are likely to implement authentication and authorization for HTTP Bridge. By having separate listeners, th admin endpoints will not be affected and implementation will be simpler.
```
PaulRMellor, line 148:
```suggestion
- Continue using a single interface and only use separate listeners when SSL is enabled so that we don't introduce a breaking change to existing users. Given that we will be introducing the first major version of Bridge, introducing a breaking change is acceptable and it makes much more sense to have separate listeners for better separation of concerns, , whether or not TLS is enabled. In the future, we are likely to implement authentication and authorization for HTTP Bridge. By having separate listeners, the admin endpoints will not be affected and implementation will be simpler. 
```

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,155 @@
...
+- http.ssl.enable
+- http.ssl.keystore.certificate.location
+- http.ssl.keystore.key.location
+- http.ssl.keystore.certificate.chain
+- http.ssl.keystore.key
+- http.ssl.enabled.protocols
+- http.ssl.enabled.cipher.suites
+- http.management.port
+
+When `http.ssl.enable` is set to `true`, the HTTP Bridge server starts with SSL enabled and the new configurations will allow users to define locations of the keystore files or certificate chain and key in `PEM` format. Additional formats such as `PKCS12` and `JKS` might be supported in the future if users request them.
```
ppatierno: So AFAIU the user should set `http.ssl.keystore.certificate.location` together with `http.ssl.keystore.key.location` OR `http.ssl.keystore.certificate.chain` together with `http.ssl.keystore.key`. Is my understanding right? Should we make it more explicit?

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,155 @@
...
+
+`http.ssl.enabled.protocols` can be configured with a comma separated list of enabled secure transport protocols that the server will accept from connecting clients. If not set by the user, the server will use `TLSv1.2,TLSv1.3` as the default. 
+
+`http.ssl.enabled.cipher.suites` can be configured with a comma separated list of cipher suites that the server will support. If not set, the default list of cipher suites provided by the underlying JDK SSL/TLS engine will be used.
```
ppatierno: Same questions as for the protocols.

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,155 @@
...
+
+If the existing `http.port` is not set and `http.ssl.enable` is set to `true`, the server port will be set to `443` by default. This will allow users to make TLS encrypted connections from their clients for all the HTTP Bridge endpoints for client operations.
+
+This proposal also adds support for a separate listener for internal endpoints such as:
+- `/healthy`
+- `/ready`
+- `/metrics`
+
+The internal endpoints will use the port defined by the new configuration property `http.management.port`, defaulting to 8081 if unspecified. These endpoints will no longer be accessible via the default listener port defined by `http.port` or port 8080. All other endpoints will continue to use the existing listener on port 8080, or 443 if SSL is enabled.
```
ppatierno: Did you try what happens if ... the user DOESN'T enable TLS so they get default HTTP port 8080 but then they also set `http.management.port` to 8080? If you start two servers in the bridge, it's not going to work because the first one will keep the port and the other one will find the port already used. At the same time, as user, I would expect this to work. I just want operations and admin on the same 8080 port as today. Have you taken into account this scenario in terms of implementation?

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,155 @@
...
+  http:
+    port: 8443
+    tls: #NEW FIELD
+      serverCertChainAndKey: #NEW FIELD
+        secretName: my-secret
+        certificate: public.crt
+        key: private.key
+...
+```
+The liveness and readiness probes will connect to port 8081 (instead of `http.port` or 8080). If users have metrics set up, they would need to use port 8081 to connect to the `metrics` endpoint and cannot reconfigure this port number as it's only for internal use.
```
ppatierno: So we don't allow to change the management port within the `KafkaBridge` custom resource? Why?

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Motivation
+
+The connection between HTTP Bridge and Kafka cluster can be secured. However, there is currently no support for securing client connections to the HTTP Bridge. The current recommendation is to use firewalls or API gateways to secure the client connections. This requires users to make additional configurations and does not provide the best user experience. 
+
+Moreover, some of the endpoints are for internal use only, such as `/healthy`, `/ready` and `/metrics` but they are exposed on the same listener and the same port as the other endpoints that are used by external clients. This means the endpoints for internal use only are also exposed to external clients. And if SSL is enabled, it will introduce more operational complexity as these internal endpoints would need to be configured with TLS and certificates.
+
+## Proposal
+
+This proposal introduces adds new configurations to enable an SSL server and load a keystore certificate and key in PEM format so that client connections to HTTP Bridge can be encrypted. The new configurations will be added to the existing `http` prefixed configurations:
```
see-quick:
```suggestion
This proposal introduces new configurations to enable an SSL server and load a keystore certificate and key in PEM format so that client connections to HTTP Bridge can be encrypted. The new configurations will be added to the existing `http` prefixed configurations:
```

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,155 @@
...
+- http.ssl.enable
```
see-quick: Just out of curiosity, adding the `https` instead of `http.ssl` prefix wouldn't be better?

1xx-enable-ssl-for-kafka-bridge.md
```diff
@@ -0,0 +1,155 @@
...
+- http.ssl.enabled.cipher.suites
+- http.management.port
+
+When `http.ssl.enable` is set to `true`, the HTTP Bridge server starts with SSL enabled and the new configurations will allow users to define locations of the keystore files or certificate chain and key in `PEM` format. Additional formats such as `PKCS12` and `JKS` might be supported in the future if users request them.
+
+`http.ssl.enabled.protocols` can be configured with a comma separated list of enabled secure transport protocols that the server will accept from connecting clients. If not set by the user, the server will use `TLSv1.2,TLSv1.3` as the default. 
+
+`http.ssl.enabled.cipher.suites` can be configured with a comma separated list of cipher suites that the server will support. If not set, the default list of cipher suites provided by the underlying JDK SSL/TLS engine will be used.
+
+If the existing `http.port` is not set and `http.ssl.enable` is set to `true`, the server port will be set to `443` by default. This will allow users to make TLS encrypted connections from their clients for all the HTTP Bridge endpoints for client operations.
```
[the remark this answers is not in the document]
tinaselenge: This was also discussed early on, I will find and shared the link here. When running bridge with the operator, we actually enforce `8443` so maybe it is better to make it `8443` for standalone as well?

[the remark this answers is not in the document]
tinaselenge: Initially the proposal had 8443, but I remember changing it based on a discussion with @ppatierno but can't remember or find it. I think we had decided that 443 was more standard HTTPS port?

PR #172, opened 2025-08-31
Add the `v1` version to `KafkaTopic` and `KafkaUser` CRD APIs

scholzj:
This PR opens a proposal about adding the new `v1` API version for Kafka Topic and Users and explains why I think it is important to start early with these based on our previous experience. However, it is also important that we know for sure how should they look like. So if anyone thinks there are some additional things to deprecate or change, they should use this proposal (comment on it) to make sure it is covered.

### DIFF
--- a/112-add-v1-api-for-topics-and-users.md
+++ b/112-add-v1-api-for-topics-and-users.md
@@ -1,0 +1,112 @@
+# Add the `v1` version to `KafkaTopic` and `KafkaUser` CRD APIs
+
+As part of move to the Strimzi `v1` CRD API and to Strimzi 1.0.0, this proposal introduces the `v1` versions for the `KafkaTopic` and `KafkaUser` resources.
+
+## Motivation
+
+One of the lessons we learned from the introduction of the Strimzi `v1beta2` CRD API and from the migration from the `v1alpha1` / `v1beta1` APIs, is that the `KafkaTopic` and `KafkaUser` resources require special care.
+The reason for it is that they are operated by detached User and Topic Operators.
+When the Strimzi upgrade happens, in the first phase:
+* The CRDs are updated
+* The Cluster Operator Pod is rolled
+
+While these two events are not synchronized and are _eventually consistent_, they happen at roughly the same time.
+The User and Topic Operators on the other hand, they are rolled only much later as part of the Kafka cluster reconciliation.
+This can be many minutes after the CRDs were updated.
+
+Additionally, the Topic and User Operator are deleting their Users / Topics from the Kafka cluster when the resources are deleted.
+This is different from the Cluster Operator, which in general does not delete the operands as they are deleted by Kubernetes Garbage Collection.
+(While this does not apply to the `KafkaConnector` resources which are deleted by the operator through the Connect REST API, the `KafkaConnector` resources are reconciled only when the corresponding `KafkaConnect` resource exists.)
+As a result, if the Cluster Operator does not see the custom resources for some time, it will not delete anything.
+In contrast, the User and Topic Operators would delete any topic or user configuration when they don't see the corresponding CRs (in case of the User Operator) or when the perceive that they have been deleted (in case of the Topic Operator).
+
+This had significant impact on the introduction of `v1beta2` and migration from `v1beta1` to `v1beta2` APIs.
+We used the following schedule:
+* Strimzi 0.22 introduced the `v1beta2` API, but the operators kept using the old `v1beta1` API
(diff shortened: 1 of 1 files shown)


112-add-v1-api-for-topics-and-users.md
```diff
@@ -0,0 +1,112 @@
...
+
+_Note:_
+_The schedule for the migration to `v1beta2` API was heavily influenced by Kubernetes dropping support for the `v1beta1` CRD API._
+_That forced us to proceed very quickly._
+
+This schedule worked for the Cluster Operator.
+But it did not worked for the Topic and User Operators.
+When the old API versions were removed from the Kubernetes cluster, we had still the Topic and User Operators from the previous Strimzi version running.
+And they were still using the old APIs.
+As a result, they perceived all their `KafkaUser` and `KafkaTopic` resources to be deleted and deleted them from the Kafka cluster.
```
fvaleri, line 36: Is this referring to a scenario in which we store `v1` and completely drop `v1beta2` as opposed to mark as not served? If `v1beta2` is present but not served, the operator should get 404 for these CRs and skip reconciliation (no resource deletion in Kafka).

Instead, the old TO was deleting topics in Kafka when the corresponding CR version was not served anymore because it was doing 2way reconciliation.

fvaleri (commented):
Thanks. The proposal looks good. As the schema changes are limited, I think it is fine to skip migration. I left a couple of comments for you to consider.

112-add-v1-api-for-topics-and-users.md
scholzj: > This happened when the `v1beta1` / `v1`alpha1` were completely removed. To be honest, I'm not 100% sure how it would work with the unidirectional TO today. But definitely nothing changed in UO. However, I think the operators got a DELETE event so it might not change even for UTO.

fvaleri: Ok. This is what I thought. A delete event would cause topic deletions in Kafka, unless we mark all v1beta2 KTs as unmanaged. In that case topics are not deleted in Kafka because delete events for unmanaged KTs are just ignored. This is something you may add to the proposal as a note.

scholzj: Yeah - the thing is that the operator (and pretty much Kubernetes itself) does not really distinguish whether the topic is `v1beta2` or `v1`. While the different versions might have a different schema, the version is more related to the API calls than the resources. (Honestly, I think the API versioning is really screwed up 🙄) 

So when you use the `v1` API, you will see all the topic resources whether they are `v1` or `v1beta2`. And the same when you use the `v1beta2` API. So when you remove the old API version, the resources would "disappear" regardless what version they were created as.

So it is not about marking the `v1beta2` resources as unmanaged as there are the same resources as the `v1` resources. It is mainly about using the `v1` API calls before we remove the old APIs.

112-add-v1-api-for-topics-and-users.md
```diff
@@ -0,0 +1,112 @@
...
+We should also ideally try to leave more time between the versions and not do all there steps in three subsequent Strimzi releases.
+That should give users more freedom and avoid forcing them to upgrade through 3 subsequent Strimzi versions.
+
+However, given we are now aware of the increased sensitivity of the User and Topic Operators, we should try to introduce the `v1` version for them as early as possible.
+That should make it easier for our users to migrate to the `v1` API while keeping more relaxed upgrade schedules (i.e. skipping some Strimzi versions).
+This should be also reasonably straight forward for the `KafkaUser` and `KafkaTopic` resources as there are minimal changes expected to them.
+
+## Proposal
+
+We will introduce the `v1` versions of the `KafkaUser` and `KafkaTopic` CRDs in Strimzi 0.49.0.
```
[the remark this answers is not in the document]
scholzj: Hmm, I did not know about that. **Thanks for raising it and pointing it out.** I will have to investigate what it really does and whether we already support this (not as far as I know) an how can we add support for it. I will incorporate this into the proposal.

I still think it makes sense to deprecate the old versions all together in one release, once all of them are ready. Especially given in case of topics, there is nothing to do really. And in case of users there is not much to do for most users either.

scholzj:
FYI: I'm going to close this for the time being and open a new version later. While the proposal part for the `v1` version stands, I need to think a bit more about the timing of the operator switch to the `v1` API for the topics and users, as I don't think some of the reasoning stands and do not want to mislead anyone. Credit to @fvaleri, whose deprecation comment actually triggered some considerations.

PR #173, opened 2025-09-04
Proposal for supporting `connections.max.reauth.ms` configuration for SCRAM listeners

tinaselenge:
Related to: https://github.com/strimzi/strimzi-kafka-operator/issues/5639

### DIFF
--- a/1xx-reauth-config-for-scram-listener.md
+++ b/1xx-reauth-config-for-scram-listener.md
@@ -1,0 +1,46 @@
+# Support `connections.max.reauth.ms` configuration for SCRAM listeners
+
+## Current situation
+
+Apache Kafka provides the [`connections.max.reauth.ms`](https://kafka.apache.org/documentation/#brokerconfigs_connections.max.reauth.ms) configuration parameter which can be set at broker level for all listeners or a specific listener.
+
+Currently, Strimzi doesn't have support for specifying such a parameter at both broker or listener level because the `listener.` prefix is part of the list of the forbidden configurations for the `spec.kafka.config` section.
+
+However, this configuration is already supported for [OAuth](https://strimzi.io/docs/operators/latest/full/configuring.html#type-KafkaListenerAuthenticationOAuth-reference) at listener level under `authentication` field and can be set for [`custom` type authentication](https://strimzi.io/docs/operators/latest/full/configuring.html#type-KafkaListenerAuthenticationCustom-schema-reference) under `listenerConfig` field. 
+
+## Motivation
+
+The `connections.max.reauth.ms` is a useful configuration for SCRAM authentications enabled on a specific listener to define the maximum lifetime of a client connection before re-authentication is required. This ensures that clients with long lived connections periodically re-authenticate in case passwords already changed and reduces the risk of having connections with expired or compromised credentials. Since this is already supported/can be set for both OAuth and `custom` type authentications, it makes sense to support it for SCRAM authentications too.
+
+## Proposal
+
+`maxSecondsWithoutReauthentication` field already exists for `KafkaListenerAuthenticationOAuth` class, so for consistency, the same field will be added `KafkaListenerAuthenticationScramSha512` class. User can set the configuration shown in the following example:
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+  namespace: myproject
+spec:
+  kafka:
+    # ...
+    listeners:
+      - name: plain
+        port: 9092
+        type: internal
+        tls: true
+        authentication:
+          type: scram-sha-512
+          maxSecondsWithoutReauthentication: 3600
+```
+
+## Affected/not affected projects
+
+- strimzi-cluster-operator
+
+## Compatibility
+
+As this is a new configuration, there is no backwards compatibility issues relating to this proposal.
+
+## Rejected alternatives


scholzj (commented):
Thanks for the proposal ... I think that in general this would work fine.

But I wonder if we should consider some alternatives before we decide to add some new API field for this ...
* Should we let users to configure this cluster-wide in `.spec.kafka.config`? Are there usecases where it should be configured on a per-listener basis?
* Should we find a way how to allow users configure per-listener options in `.spec.kafka.config`? (Right now, we can make it configurable at the cluster level, but not with the per-listener prefixes)
* Users can probably use the custom authentication for this. Isn't that sufficient? Given Kafka has this disabled and does not seem to see this as an important option, maybe it is niche thing among Strimzi users as well?

Might be worth adding these at least to the alternatives for everyone's consideration.

It is also worth noting that https://github.com/strimzi/strimzi-kafka-operator/issues/11764 suggests considering deprecating the Oauth authentication exactly because it became just a huge pile of various options. While this proposal does not do this for SCRAM-SHA, it is a reason to consider if we want to handle these options through new API fields or not.

1xx-reauth-config-for-scram-listener.md
```diff
@@ -0,0 +1,46 @@
+# Support `connections.max.reauth.ms` configuration for SCRAM listeners
+
+## Current situation
+
+Apache Kafka provides the [`connections.max.reauth.ms`](https://kafka.apache.org/documentation/#brokerconfigs_connections.max.reauth.ms) configuration parameter which can be set at broker level for all listeners or a specific listener.
+
+Currently, Strimzi doesn't have support for specifying such a parameter at both broker or listener level because the `listener.` prefix is part of the list of the forbidden configurations for the `spec.kafka.config` section.
```
PaulRMellor, line 7:
```suggestion
Currently, Strimzi supports setting `connections.max.reauth.ms` broker-wide through `spec.kafka.config`. Per-listener settings cannot be configured this way, because `listener.` properties are forbidden.
```

PaulRMellor: Assuming we're actually okay at the broker level

1xx-reauth-config-for-scram-listener.md
```diff
@@ -0,0 +1,46 @@
...
+
+However, this configuration is already supported for [OAuth](https://strimzi.io/docs/operators/latest/full/configuring.html#type-KafkaListenerAuthenticationOAuth-reference) at listener level under `authentication` field and can be set for [`custom` type authentication](https://strimzi.io/docs/operators/latest/full/configuring.html#type-KafkaListenerAuthenticationCustom-schema-reference) under `listenerConfig` field.
```
PaulRMellor, line 9:
```suggestion
The configuration is already supported at the listener level for [OAuth](https://strimzi.io/docs/operators/latest/full/configuring.html#type-KafkaListenerAuthenticationOAuth-reference) through the  `authentication` field, and for [`custom` type authentication](https://strimzi.io/docs/operators/latest/full/configuring.html#type-KafkaListenerAuthenticationCustom-schema-reference) through the `listenerConfig` field. 
```

1xx-reauth-config-for-scram-listener.md
```diff
@@ -0,0 +1,46 @@
...
+
+However, this configuration is already supported for [OAuth](https://strimzi.io/docs/operators/latest/full/configuring.html#type-KafkaListenerAuthenticationOAuth-reference) at listener level under `authentication` field and can be set for [`custom` type authentication](https://strimzi.io/docs/operators/latest/full/configuring.html#type-KafkaListenerAuthenticationCustom-schema-reference) under `listenerConfig` field. 
+
+## Motivation
+
+The `connections.max.reauth.ms` is a useful configuration for SCRAM authentications enabled on a specific listener to define the maximum lifetime of a client connection before re-authentication is required. This ensures that clients with long lived connections periodically re-authenticate in case passwords already changed and reduces the risk of having connections with expired or compromised credentials. Since this is already supported/can be set for both OAuth and `custom` type authentications, it makes sense to support it for SCRAM authentications too.
```
PaulRMellor, line 13:
```suggestion
The `connections.max.reauth.ms` is a useful configuration for SCRAM authentications enabled on a specific listener to define the maximum lifetime of a client connection before re-authentication is required. This ensures that clients with long lived connections periodically re-authenticate in case passwords already changed and reduces the risk of having connections with expired or compromised credentials. Since this configuration is already supported for OAuth and `custom` type authentications, it makes sense to support it for SCRAM authentications too.
```

1xx-reauth-config-for-scram-listener.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Proposal
+
+`maxSecondsWithoutReauthentication` field already exists for `KafkaListenerAuthenticationOAuth` class, so for consistency, the same field will be added `KafkaListenerAuthenticationScramSha512` class. User can set the configuration shown in the following example:
```
PaulRMellor, line 17:
```suggestion
The `maxSecondsWithoutReauthentication` field already exists for `KafkaListenerAuthenticationOAuth` class, so for consistency, the same field will be added to the `KafkaListenerAuthenticationScramSha512` class. Users can set the configuration as shown in the following example:
```

1xx-reauth-config-for-scram-listener.md
```diff
@@ -0,0 +1,46 @@
...
+          maxSecondsWithoutReauthentication: 3600
+```
+
+## Affected/not affected projects
+
+- strimzi-cluster-operator
+
+## Compatibility
+
+As this is a new configuration, there is no backwards compatibility issues relating to this proposal.
```
PaulRMellor, line 44:
```suggestion
As this is a new configuration, there are no backwards compatibility issues relating to this proposal.
```

1xx-reauth-config-for-scram-listener.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Rejected alternatives
```
PaulRMellor, line 46: Do we want to explain why the proposal is preferable to the alternatives suggested by Jakub?
Why we wouldn't want to per-listener options in .spec.kafka.config or type: custom?

1xx-reauth-config-for-scram-listener.md
```diff
@@ -0,0 +1,46 @@
+# Support `connections.max.reauth.ms` configuration for SCRAM listeners
```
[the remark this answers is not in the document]
PaulRMellor: Do we want a summary to lead into the proposal?

Something like: 

> This proposal adds support for configuring the Kafka `connections.max.reauth.ms` parameter at the listener level for SCRAM authentication in Strimzi.

1xx-reauth-config-for-scram-listener.md
PaulRMellor: Should we add a note on the oauth deprecation?

Something like:

> Note: OAuth configuration through `type: oauth` is planned to be deprecated. Future configurations would use `type: custom` with the appropriate OAuth settings instead.

katheris (commented):
Thanks for creating the proposal.

I'm also a bit hesitant about adding an option to `authentication` given the risk of needing to add other options in future. There are currently two other listener specific options you found, so it could just be those, but other options might be added.

Looking at how the forbidden options are handled, I wonder if we could put in some handling to allow certain listener-specific configurations without the use of regex. In [KafkaCluster.java](https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java#L353) we handle the configuration in the same method that we validate the set of listeners. So I wonder if we could build a set of prefixes that users might use and basically augment the forbidden prefix exceptions list in the KafkaConfiguration constructor. 

What do you think @scholzj @tinaselenge ?

scholzj:
> Looking at how the forbidden options are handled, I wonder if we could put in some handling to allow certain listener-specific configurations without the use of regex. In [KafkaCluster.java](https://github.com/strimzi/strimzi-kafka-operator/blob/main/cluster-operator/src/main/java/io/strimzi/operator/cluster/model/KafkaCluster.java#L353) we handle the configuration in the same method that we validate the set of listeners. So I wonder if we could build a set of prefixes that users might use and basically augment the forbidden prefix exceptions list in the KafkaConfiguration constructor.

It would be an option, yes. I'm just not sure what would be the best way to do it the best way. Keep in mind, that with many clusters having 2 or more listeners, it is lot of parsing, string concatenation, for loops, etc. Is that still "cheaper" than regular expression(s)?

Strictly speaking, I do not think we need to validate the particular listener name. It might be even bad idea as it would cause unnecessary errors when a listener is deleted etc. We just care about the pattern. So I think we can do it without the explicit listener knowledge as well. But it still seems very brute force unelegant approach ... :-/

scholzj:
> > in a (new?) proposal
> 
> Why a new one?

It seemed easier to me as it would not have much to do with the current one. But it is fine if the new one is updated as well.

tinaselenge:
Thank you for all the valuable feedback. 

I have updated the proposal with the agreed solution of allowing the per-listener configuration in the `spec.kafka.config` instead of adding a new API field for the authentication. 

I have tried implementing both of the suggested solutions from @scholzj and @katheris, regular expression and building a list of per listener config exceptions and then came to the conclusion that the latter is simpler and less expensive to implement. I have included the regular expression one to the Rejected Alternative section. 

Here is also a [POC implementation](https://github.com/strimzi/strimzi-kafka-operator/compare/main...tinaselenge:strimzi-kafka-operator:reauth?expand=1). We can improve this further, by making it build the per-listener config exceptions, only if there is any `listener.` config in the `spec.kafka.config`, otherwise carry on as before.

@scholzj , @katheris , @PaulRMellor , @ppatierno can you please have another look when you have time? Thank you!

tinaselenge:
So there seems to be 2 ways to implement this:
- Implement this config via KafkaConfiguration class (shown in the [POC](https://github.com/strimzi/strimzi-kafka-operator/compare/main...tinaselenge:strimzi-kafka-operator:reauth?expand=1)]), which modifies the FORBIDDEN_LIST in runtime. This avoids making changes to AbstractConfiguration class that is extended by other resource configuration classes. This new configuration is specific to Kafka, so I thought implementing it with KafkaConfiguration class makes sense.
- Implement it in the AbstractConfiguration class and that way KafkaConfiguration class stays general. This will require updating filterForbidden() method as well as constructor methods of AbstractConfiguration to pass in the new list of per listener configuration and doing the filter based on them. 

Overall, I think both approaches work and are similar in terms of efficiency and readability so I don't think I have a strong opinion on which one to go with. The POC shows how it is currently proposed to implement. 

Please let me know if anyone else has a strong opinion on which one should be implemented.

scholzj:
@tinaselenge Thinking about it a bit more ... I'm not sure we have any immediate usecases for other resources. And the changes are mostly limited only to the constructor in the `Kafkaconfiguration` class. While it would probably be not too hard to generalize it, it would be the Nth constructor in AbstractConfiguration and would be used only from KafkaConfiguration. So I think it is probably fine to stick with the `KafkaConfiguration` class changes only. And we can move the logic to AbstractConfiguration later when we actually see some need.

katheris (approved):
Thanks @tinaselenge I'm happy with this proposal

ppatierno (approved):
LGTM. +1 binding.

im-konge (approved):

124-reauth-config-for-scram-listener.md
```diff
@@ -0,0 +1,55 @@
...
+  kafka:
+    listeners: 
+      - name: external
+        port: 9094
+        type: route
+        tls: true
+        authentication:
+          type: scram-sha-512
+    config:
+      listener.name.external-9094.connections.max.reauth.ms: 3600000
```
PaulRMellor: Is setting to 0 allowed?

tinaselenge: Yes, it is set to 0 by default.

PaulRMellor: I was just wondering if we should mention. Maybe we don't need to

124-reauth-config-for-scram-listener.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Proposal
+
+Allow users configure `listener.name.<listener_name>.connections.max.reauth.ms` in `.spec.kafka.config` where `<listener_name>` is listener name and listener port joined by hyphen (`-`). To allow only this particular option but still forbid other listener options, another list of exceptions similar to `FORBIDDEN_PREFIX_EXCEPTIONS` will be added to `KafkaClusterSpec` class. This new list will be used to build a set of prefixes for each listener, which will be then added to the existing forbidden prefix exceptions list. The list will contain `connections.max.reauth.ms` for now but allows us to easily support more per-listener configurations in the future if we need to.
```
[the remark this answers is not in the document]
tinaselenge: Thanks Paul. I have added an explanation on this taking your suggestion. Please let me know if it still doesn't look clearer.

PR #179, opened 2025-09-24
`v1` CRD API Conversion Tool

scholzj:
This proposal covers the Conversion Tool for converting the `v1beta2` API to the `v1` API. The tool will be needed by all Strimzi users before upgrading to Strimzi 0.52 / 1.0.0 version, which removes support for the `v1beta2` and older APIs. The tool is mostly based on the `v1beta2` conversion tool we used back then. And its UX and the required upgrade workflow correspond to it as well.

### DIFF
--- a/117-v1-CRD-API-Conversion-Tool.md
+++ b/117-v1-CRD-API-Conversion-Tool.md
@@ -15,7 +15,7 @@ This proposal provides additional details.
 ### Conversion Tasks
 
 The conversion tool should be able to fulfill the following tasks:
-* Support conversion of the custom resources by modifying their structure to account for the fields that were renamed or moved:
+* Support conversion of the custom resources by modifying their structure to account for the fields that were renamed, moved, or removed:
     * Be able to do the conversion inside a Kubernetes cluster across all namespaces, in a single namespace, or for a particular custom resource only
     * Convert YAML files with the Strimzi custom resources in-place (by modifying the existing file) or by creating a new file
 * Support updating the CRDs:
@@ -27,7 +27,7 @@ Users should be able to use only some of these tasks if needed and combine the C
 
 ### Architecture
 
-The Conversion Tool will be written in Java and mostly follow the architecture of the previous tool we used last time when migrating from `v1beta1` to `v1beta2` API.
+The Conversion Tool will be written in Java and mostly follow the architecture of the [previous tool](https://github.com/strimzi/strimzi-kafka-operator/tree/0.22.0/api-conversion) we used last time when migrating from `v1beta1` to `v1beta2` API.
 Using Java would simplify the release process as it would depend directly on the Strimzi API.
 
 The Conversion Tool itself will be distributed as ZIP and TAR.GZ archives.
--- a/117-v1-CRD-API-Conversion-Tool.md
+++ b/117-v1-CRD-API-Conversion-Tool.md
@@ -1,13 +1,13 @@
 # `v1` CRD API Conversion Tool
 
-This proposal covers `v1` CRD API Conversion Tool which will be available to the users to help them with the migration from the `v1beta2` APIs to the `v1` API.
+This proposal covers the `v1` CRD API Conversion Tool, which will be available to users to help migrate from the `v1beta2` API to the `v1` API.
 
 ## Motivation
 
 The migration of the Strimzi CRs from the `v1beta2` version to the `v1` version can be done manually.
 We plan to cover the manual process as part of our documentation.
-However, given the amount of steps and in some cases large number of resources that would need to be converted, we should also provide a Conversion Tool to help with it.
-That was also approved as part of the initial [Strimzi `v1` CRD API and 1.0.0 release](./113-Strimzi-v1-CRD-API-and-1.0.0-release.md) proposal.
+However, given the number of steps and, in some cases, a large number of resources that would need to convert, we will provide a Conversion Tool to help.
+Providing a tool was approved as part of the initial [Strimzi `v1` CRD API and 1.0.0 release](./113-Strimzi-v1-CRD-API-and-1.0.0-release.md) proposal.
 This proposal provides additional details.
 
 ## Proposal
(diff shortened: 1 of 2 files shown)


117-v1-CRD-API-Conversion-Tool.md
```diff
@@ -0,0 +1,137 @@
...
+### Architecture
+
+The Conversion Tool will be written in Java and mostly follow the architecture of the previous tool we used last time when migrating from `v1beta1` to `v1beta2` API.
+Using Java would simplify the release process as it would depend directly on the Strimzi API.
+
+The Conversion Tool itself will be distributed as ZIP and TAR.GZ archives.
+The archives will contain helper scripts for running it from the command line.
+It will use Java 17 as most of the other Strimzi components.
+
+It will be also bundled in the Strimzi Operator container.
```
im-konge, line 37: Would it make sense to have simple separate image just for the conversion tool? I mean, it would be more lightweight possibly?

im-konge (commented):
The proposal looks good to me, thanks for it.
I have just one question.

117-v1-CRD-API-Conversion-Tool.md
scholzj: A separate image would be smaller. So I guess it would be more _lightweight_. But the overall release process will be more heavyweight as new image means certainly more work and time compared to another thing added to the existing image. 

Also, the conversion tool seems to have roughly ~20MB. Given the operator image has IIRC 400-500MB and most of it is the base OS, I do not think it makes sense to have a separate image.

im-konge (approved):

PaulRMellor (approved):
Looks good.
I left a few suggestions, with a couple of queries

117-v1-CRD-API-Conversion-Tool.md
```diff
@@ -0,0 +1,137 @@
...
+
+The Conversion Tool will live as an additional module in the main Strimzi Operators repository (and will be removed after the 0.52 / 1.0.0 version once it is not needed anymore).
+The module name will be `v1-api-conversion` to differentiate it from the previous version called `api-conversion`.
+That will allow it to use the same `api` module version as is currently under development.
+It will also allow us to release it as part of the regular Strimzi release process.
+
+## Documentation
+
+Documentation will include the procedure for how to use the Conversion Tool to convert the resources.
+And - eventually, in the last phase (Strimzi 0.52 / 1.0.0) - it will also cover how to upgrade the CRDs.
```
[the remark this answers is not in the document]
scholzj: I added link to the main proposal.

scholzj:
@ppatierno Do you have anything more on this proposal?

scholzj:
This proposal now has 3 binding +1 votes. It would be great to see more people vote/comment. But unless there are some more comments, I will close the vote on Thursday EOB as approved.

117-v1-CRD-API-Conversion-Tool.md
```diff
@@ -0,0 +1,135 @@
...
+  Users can also specify only specific kinds and/or a single namespace to limit the resources that will be converted.
+* The `crd-upgrade` option will _convert_ the CRDs.
+  It will first update the stored version in the CRDs to `v1`.
+  Then it will _touch_ all custom resources to make sure they are stored under the `v1` version.
+  And finally, it will remove the `v1beta2` version (and `v1alpha1` / `v1beta1` version for `KafkaTopic` and `KafkaUser` resources) from the CRD status.
+
+### Workflow
+
+Users will be able to use the `convert-file` or `convert-resource` commands at any point between the Strimzi 0.49 and 0.52 / 1.0.0 release.
+Conversion is required no later than the upgrade to Strimzi 0.52.0 / 1.0.0.
```
fvaleri, line 58:
```suggestion
Conversion must be done between 0.49.0 (included) and 0.52.0 / 1.0.0 (excluded).
```

117-v1-CRD-API-Conversion-Tool.md
```diff
@@ -0,0 +1,135 @@
...
+However, it will not be able to handle all of them (as already covered in the [Strimzi `v1` CRD API and 1.0.0 release](./113-Strimzi-v1-CRD-API-and-1.0.0-release.md) proposal).
+The conversion is not possible, for example, when it involves the handling of secrets or environment variables that must be mounted or mapped.
+
+The changes which the Conversion Tool will not be able to handle are:
+* Change from `type: oauth` authentication to `type: custom` for both Kafka brokers as well as for the client-based operands
+* Change from `type: keycloak` and `type: opa` authorization to `type: custom` authorization
+* Removal of the support for mounting Secrets as part of the `type: custom` authentication
+* Removal of the `externalConfiguration` section in the `KafkaConnect` CR
+
+However, the Conversion Tool will detect these changes, fail the conversion, and request the user to do the conversion manually before re-running the tool.
```
fvaleri, line 91: It would also be good to include a link to the relevant documentation.

117-v1-CRD-API-Conversion-Tool.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Documentation
+
+Documentation will include the procedure for how to use the Conversion Tool to convert the resources together with the documentation of the manual conversion process as well as the upgrade to the `v1` API version in the Strimzi 0.52 / 1.0.0 version.
+The migration process is described in detail in the main [_Strimzi v1 CRD API and 1.0.0 release_ proposal](https://github.com/strimzi/proposals/blob/main/113-Strimzi-v1-CRD-API-and-1.0.0-release.md).
+
+### Testing
+
+The conversion of the Strimzi custom resources will be tested using unit and integration tests in the Conversion Tool module itself.
+Additionally, System Tests will be added to test the whole lifecycle of the conversion:
```
fvaleri, line 108: I guess it's implicit that these STs will also be removed after the 0.52 / 1.0.0 version.

fvaleri (commented):
Thanks for the proposal. I just left some minor comments, but looks good.

117-v1-CRD-API-Conversion-Tool.md
```diff
@@ -0,0 +1,135 @@
...
+
+The Conversion Tool will be written in Java and mostly follow the architecture of the [previous tool](https://github.com/strimzi/strimzi-kafka-operator/tree/0.22.0/api-conversion) we used last time when migrating from `v1beta1` to `v1beta2` API.
+Using Java would simplify the release process as it would depend directly on the Strimzi API.
+
+The Conversion Tool itself will be distributed as ZIP and TAR.GZ archives.
+The archives will contain helper scripts for running it from the command line.
+It will use Java 17 as with most of the other Strimzi components.
+
+It will also be bundled in the Strimzi Operator container.
+Bundling the tool in the container enables users who cannot (or prefer not to) install the Java Runtime Environment (JRE) to run it as a Kubernetes Job or locally as a Docker container.
```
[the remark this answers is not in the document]
scholzj: I don't think so. A lot depends on what do you convert. Ultimately, running it locally is the easiest way.
* Running it from Docker means you do not need to install Java. But you would need to deal with volumes, networking, etc.
* Running it as Kubernetes Job might be an option if you are not allowed to talk with the Kubernetes server directly, and for example, have to submit things through some GitOps process. But you cannot use it to convert files, and you would need to write some RBAC files depending on what you do etc. You would also need to time the CRD Upgrade job the right way.
* If you use GitOps to manage the Strimzi resources you would need to convert the files which cannot be done as Kubernetes Job easily.
* Etc.

So, having it in the container image should open more options and this was some of the feedback from last time. But I do not think this would be a recommended or non-recommended way. It is more about the thousands of variations between our user environments.

PR #181, opened 2025-09-29
Improve template behavior in Kafka Node Pools

scholzj:
This proposal tries to address the discussion outcome from https://github.com/strimzi/strimzi-kafka-operator/issues/11791 and proposes a change to how we treat situations when the template section is configured in both `Kafka` and `KafkaNodePool` resources. This is a breaking change for users who are in this situation. But based on the previous discussion, it seems to be the preferred way going forward, and should be more understandable to users.


siegenthalerroger:
Is there any reason why the merge would only happen on the top-level properties and not all the way down as per the strategic merge patch semantics from kubernetes? 

As a user that would be what I would expect to happen, or the previous behaviour where it was all overriden. The middle ground seems like the worst in terms of _communicating_ the behaviour. I do believe that in general this is a step in the right direction though and I do prefer it to the status quo.

im-konge (approved):

PaulRMellor (approved):
Looks good
Proposal shows the benefits of allowing merge in template behavior
Clear docs will be an important aspect of the change

118-improve-template-behavior-in-Kafka-node-pools.md
```diff
@@ -0,0 +1,184 @@
...
+## Backwards compatibility
+
+**This proposal is not backwards compatible!**
+It will change how the template sections are handled for the Kafka cluster.
+Depending on the exact configuration, this change might impact the users and break or change their clusters.
+
+Users can modify their custom resources to keep their existing configuration.
+So it is important that we make sure they are aware of this change and can prepare for it.
+
+## Rejected alternatives
```
PaulRMellor, line 175: Would we want to mention why we didn't want to use deeper-level properties?

PaulRMellor:
Thanks for addressing comments.
The addition to rejected alternatives looks good. :+1:

PR #182, opened 2025-10-03
Deprecate and remove the `.spec.kafka.resources` from the `Kafka` CR

scholzj:
This proposal suggests deprecating the `.spec.kafka.resources` field of the `Kafka` custom resource and removing it in the `v1` CRD API.


im-konge (approved):
I agree with taking the second option of failing the conversion as stated in the proposal.  LGTM, thanks.

see-quick (approved):
Agree with this change. Thanks @scholzj ! 👍

119-deprecate-and-remove-kafka-spec-kafka-resources.md
```diff
@@ -0,0 +1,85 @@
...
+
+In the second phase, as part of the 0.52 / 1.0.0 release:
+* The `.spec.kafka.resources` field will be removed.
+* Its uses in the `cluster-operator` module will be removed as well.
+
+### Documentation
+
+The documentation will be updated with this change.
+We will make sure the regular documentation does not use or recommend the deprecated field.
+The conversion will be covered as part of the `v1` CRD API conversion documentation.
```
ppatierno, line 72:
```suggestion
The manual conversion will be covered as part of the `v1` CRD API conversion documentation.
```

I would just enforce that it's going to be manual, because the tool raises error and can't do much.

ppatierno (approved):
LGTM. Just one nit.

PaulRMellor (approved):
Thanks.
This change makes sense and as part of the v1 work.
Few minor suggestions

fvaleri (approved):
LGTM. Thanks.

PR #187, opened 2025-12-04
Revised Stretch Cluster Design with Pluggable Network Provider Interface

aswinayyolath:
This PR presents a revised Stretch Cluster design with a pluggable network provider interface, enabling flexible support for multiple connectivity technologies such as Submariner, Cilium and Kube primitives like NodePort, and LoadBalancer. The design retains a centralized control plane while allowing data plane resources to be realized in remote clusters via KafkaNodePool-based placement, and clearly defines the extension points for multi-cluster networking without coupling the architecture to a specific implementation.

1. Strimzi-Kafka-Operator changes: https://github.com/aswinayyolath/strimzi-kafka-operator/compare/eventstreams-stretch-cluster...aswinayyolath:strimzi-kafka-operator:48-migration
2. MCS plugin implementation: https://github.com/aswinayyolath/strimzi-stretch-mcs-plugin
3. NodePort plugin implementation: https://github.com/aswinayyolath/strimzi-stretch-nodeport-plugin
4. Loadbalancer Plugin implementation: https://github.com/aswinayyolath/strimzi-stretch-loadbalancer-plugin


124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+'cluster-east'. Please ensure Strimzi CRDs are installed in all target clusters."
+```
+
+Reconciliation fails immediately with descriptive errors for invalid configurations.
+
+
+
+### Low-Level Design and Implementation
+
+#### Garbage Collection for Remote Resources
```
katheris: I'm quite hesitant about the idea of using a ConfigMap in lieu of an owning resource, for one thing If the ConfigMap is accidentally deleted presumably it could take up to 5 minutes for the Strimzi operator to notice and put it back, which is quite a long time for the resources to be missing.

katheris (commented):
Thanks @aswinayyolath I've had an initial look. I will need to do a more thorough review but have added some initial questions.

I also had a quick look at the implementation. You seem to have the reconciliation loop duplicated in KafkaReconciler, so having to list everything twice. I'm not keen on that as the final implementation as it creates quite a maintenance problem going forwards.

124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+    // Generate advertised.listeners configuration
+    Future<String> generateAdvertisedListeners(
+        Reconciliation reconciliation,
+        String namespace,
+        String podName,
+        String clusterId,
+        Map<String, String> listeners);
+    
+    // Generate controller.quorum.voters configuration
+    Future<String> generateQuorumVoters(
```
ppatierno: A heads up here but it's worth thinking up front. Strimzi will get the support for the dynamic quorum in 2026 (I would expect in Q1) and it will be able to operate already running clusters with static quorum and new ones with dynamic quorum (also migrating from one to the other). The dynamic quorum doesn't use the `controller.quorum.voters` parameter anymore but `controller.quorum.bootstrap.servers` instead (with a different format as well). Maybe this interface should provide both since the beginning.

124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+
+##### Single Networking Provider Per Operator
+
+The networking plugin (NodePort, LoadBalancer, MCS) is loaded once at operator startup via Java ServiceLoader.
+The plugin JAR is mounted as a volume in the operator deployment.
+The plugin's initialization code runs once when the operator starts.
+
+This means all Kafka clusters managed by a single operator instance must use the same networking provider.
+Different networking providers cannot coexist in a single operator instance.
+If networking provider were configurable per Kafka CR, the operator would need to:
```
ppatierno: I think the proposal could be simplified by moving what wasn't picked as solution in the rejected alternatives section. The proposal should be just what you proposed and how it should work in your opinion. All the reasoning about other not-picked choices should not be there. Reviewers can go through the rejected alternatives to read about them. I think this is happening in more places across the proposal that could be simplified at the core this way.

ppatierno (commented):
@aswinayyolath thanks, I had a first pass and left some comments questions. I also agree with @katheris about the need to avoid any code duplication in the reconciliation process.

124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+        Reconciliation reconciliation,
+        String namespace,
+        String podName,
+        String clusterId);
+    
+    String getProviderName();
+}
+```
+
+#### Plugin Loading Mechanism
```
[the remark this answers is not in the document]
aswinayyolath: No, users can't safely change the networking plugin for an existing stretch cluster. The plugin determines fundamental networking architecture (DNS names, Service types, endpoint formats). Changing plugins would require:

1. Deleting the existing Kafka cluster (generates new certs with different SANs, new svc with different types)
2. Losing all data (Kafka state tied to pod identities and network endpoints)
3. Reconfiguring all clients (bootstrap addresses change)

The operator should detect plugin changes and reject them

124-stretch-cluster.md
aswinayyolath: ideally the operator's reconciliation loop checks for GC CM existence as the first step when reconciling any Kafka CR with stretch mode. If missing, it's recreated immediately before any other resources.

124-stretch-cluster.md
aswinayyolath: Thanks for letting us know

124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+          name: stretch-plugin-nodeport
+```
+
+The ConfigMap must be created in the same namespace where the cluster operator is deployed.
+The `mountPath` specified in `volumeMounts` must match the directory referenced in `STRIMZI_STRETCH_PLUGIN_CLASS_PATH`.
+The ConfigMap must be created before the operator pod starts, otherwise the operator will fail to load the plugin.
+
+#### Reference Implementations
+
+The Strimzi project provides reference implementations in separate repositories:
```
scholzj: We need only one reference implementation -> I would vote for node ports as they are most flexible. The rest should live outside. (of course feel free to reference them if needed to justify some parts of the plugable interface)

124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+| **ServiceExport/Routes/Ingress** | Per networking provider | Per networking provider |
+
+The central operator has write access to remote clusters via kubeconfig and creates resources directly.
+Remote operators are responsible only for reconciling StrimziPodSet resources (creating/updating pods).
+
+### Configuration Design Rationale
+
+#### Why Stretch Configuration is at Operator Level
+
+The stretch cluster configuration (`STRIMZI_CENTRAL_CLUSTER_ID`, `STRIMZI_REMOTE_KUBE_CONFIG`, plugin settings) is specified at the cluster operator level rather than at the Kafka CR or KafkaNodePool level.
```
scholzj: You did not defined these yet or explained how the look like and what they do? So it is a bit strange to use them to explain the _rationale_ for them.

124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+These credentials grant cluster wide access to Kubernetes APIs and are managed by platform administrators, not application developers.
+
+Placing these credentials at the operator level:
+
+- Aligns with Kubernetes security model where operators run with ServiceAccounts that have cluster-level permissions
+- Avoids embedding sensitive kubeconfig credentials in application level CRs (Kafka, KafkaNodePool).
+- Simplifies credential rotation by updating operator deployment once rather than every Kafka CR
+- Prevents credential sprawl across multiple Kafka instances
+
+##### Single Networking Provider Per Operator
```
scholzj: This seems pretty limiting given what the benefits are.

124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+
+Load and initialize multiple plugin JARs dynamically.
+Maintain separate ClassLoaders for each plugin to avoid conflicts.
+Handle plugin initialization failures on a per cluster basis.
+Manage plugin lifecycle (loading/unloading) during reconciliation.
+This adds significant complexity and creates potential ClassLoader isolation issues.
+
+##### Consistent Cluster Topology
+
+The `STRIMZI_CENTRAL_CLUSTER_ID` and `STRIMZI_REMOTE_KUBE_CONFIG` define the operator's view of the multi cluster topology.
```
scholzj: This likely needs some proper structure. You should not have a single kubeconfig configuration. The purpose of the central cluster ID is not clear either.

124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+The design allows the same operator to manage both stretch and non-stretch Kafka clusters:
+Operator configured with stretch settings can still manage regular (non-stretch) Kafka instances.
+Kafka CRs without `strimzi.io/enable-stretch-cluster: "true"` annotation are deployed normally in the Kubernetes cluster where the CR is applied.
+KafkaNodePools without `strimzi.io/stretch-cluster-alias` annotation are deployed to the cluster where the CR is applied.
+
+
+##### Scope for KafkaConnect and KafkaBridge
+
+These ecosystem components are Kafka clients, not part of the Kafka cluster itself.
+They must be deployed only in the central cluster where the Kafka CR resides.
```
scholzj: Why? Isn't the point of the stretch cluster that the clients can be stretched / run elsewhere as well?

124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+
+This ensures TLS hostname verification succeeds when brokers/controllers connect across clusters.
+
+#### Remote Resource Creation
+
+The central operator creates resources in remote clusters using the `RemoteResourceOperatorSupplier`:
+
+```java
+// Get operator supplier for target cluster
+ResourceOperatorSupplier supplier =
```
scholzj: I do not see any reason for different suppliers.

124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+Example topology configuration:
+
+```yaml
+spec:
+  kafka:
+    rack:
+      topologyKey: topology.kubernetes.io/zone
+```
+
+Kafka distributes replicas considering both cluster boundaries and zone labels within clusters.
```
scholzj: I think you do not necessarily want to have different racks in different clusters just by definition. This should probably rely on something like described in https://github.com/strimzi/proposals/pull/184.

124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+**Network Characteristics:**
+- Inter-cluster latency: 5-50ms
+- Requires highly stable, dedicated network links
+
+**Expected Performance:**
+- **Throughput:** 10-30% of baseline (severe degradation)
+- **Latency:** 1-10 seconds average
+- **Stability:** Sensitive to network jitter
+
+**Testing Data:** At 10ms added latency, throughput drops to 17,000 msg/sec (66% reduction). At 50ms, only 4,300 msg/sec remains (91% reduction).
```
scholzj: Why does throughput drop with latency? The general expectation would be stability issues rather than throughput drop. Were stability issues cause of the throughput drop?

124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+**Scenarios to Avoid:**
+| Deployment | Latency | Reason |
+|------------|---------|--------|
+| Regional (> 200km) | > 100ms | 96%+ throughput loss, constant publish timeouts |
+| cross region | > 150ms | System effectively unusable |
+| Intercontinental | > 250ms | Constant failures |
+
+**Alternative:** Use **MirrorMaker 2** for cross region disaster recovery. MM2 provides asynchronous replication optimized for high-latency scenarios and is the correct tool for geographic distribution.
+
+#### Summary: When to Use Stretch Clusters
```
scholzj: I think we absolutely do not want to issue any recommendations like this.

124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+
+Users must ensure kubeconfig credentials are properly secured and rotated.
+The operator validates kubeconfig expiry and reports errors before credentials expire.
+
+#### Entity Operator
+
+**Deployment:**
+- Deployed only in the central cluster
+- Manages KafkaTopic and KafkaUser resources
+- No changes required for stretch cluster support
```
scholzj: Likely needs to be properly configured to connect to the whole stretch cluster? Same with KE and CC.

124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+- Requiring strong consistency (synchronous replication)
+- Need automatic client failover without application changes
+- Organizational requirements for separate Kubernetes control planes
+- Compliance requirements for independent failure domains
+
+**❌ Use MirrorMaker 2 Instead When:**
+- Deploying across regions (> 50ms latency)
+- Geographic disaster recovery is primary goal
+- Can tolerate asynchronous replication
+- Prefer simpler operational model
```
scholzj: Mirror Maker 2 has a completely different purpose. Positioning it as an alternative does not make sense. It also provides completely different guarantees -> e.g. asynch mirroring, but much higher guarantees against cascading failures, which the stretch cluster does not provide. This is not one or the other decision.

scholzj (commented):
I left some comments. I think it is a step in the right direction. But there is lot of work left. The plugin design seems not to take into account the public versus private APIs (or did I missed it somewhere? It is easy to miss in such a long proposal). It also needs better integration into the reconciliation flow rather than just creating soem strange override points. For example lot of the stuff needs to be really plugged into the `KafkaListenersReconciler` rather than into some artificial entry points. I do not think this creates a stable design which encourages writting of the plugins.

124-stretch-cluster.md
```diff
@@ -0,0 +1,1795 @@
...
+
+This approach provides:
+- **Separation of concerns**: Core operator logic is decoupled from networking implementation details
+- **Extensibility**: Users can implement custom providers without modifying operator code
+- **Testing**: Providers can be tested independently and swapped for different environments
+
+#### Plugin JAR Deployment
+
+Users must provide the networking provider plugin JAR file to the cluster operator.
+This is accomplished by creating a ConfigMap containing the plugin JAR and mounting it as a volume in the operator deployment.
```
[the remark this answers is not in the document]
rohan-anil-kumar: I agree, we are taking a look into baking the image with the operator image. We initially thought of the ConfigMap approach to make it easier for the user to load the plugin JAR (without building a custom container image)

124-stretch-cluster.md
rohan-anil-kumar: We have used different suppliers because managing multiple remote cluster suppliers became a hassle when the codebase grew. Having a different class for managing the remote cluster suppliers made it easy to manage specific operators for specific clusters much easier. 
Earlier we had to pass multiple remote cluster operators as arguments in function calls. Now we just need to pass a single remote resource supplier class.
We're open to hear any alternate suggestions

124-stretch-cluster.md
rohan-anil-kumar: [This](https://github.com/strimzi/proposals/pull/187/files#diff-bc3052cb5e5b8e53b2248bab9b2a8132a30a54be75cd58b1c1101e7fb7294afeR478) is how we're intending to use it. 
Or am I misunderstanding your question?

124-stretch-cluster.md
rohan-anil-kumar: There is a possibility of conflicts happening between different network provider type. This would introduce additional complexity in the operator architecture.

124-stretch-cluster.md
rohan-anil-kumar: In this proposal, we were aiming to stretch kafka. Stretching kafka connect and kafka bridge are out of scope in this proposal.

124-stretch-cluster.md
rohan-anil-kumar: Can you please elaborate? As I am not sure if I understand it completely.

124-stretch-cluster.md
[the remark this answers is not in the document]
rohananilkumar: Currently there is no way for the operator to detect the plugin changes. If the user change the plugin in the deployment, the operator restarts with the new plugin. This is not recommended.

124-stretch-cluster.md
rohananilkumar: That's true. We've updated the proposal to remove MM2 as an alternative.

rohananilkumar:
@em-sav We were able to stretch Strimzi while using cilium clustermesh. Since it was hard to work with IPs directly, we wrote rules in kube DNS pods to make pod to pod communication work similar to what MCS does. We were then able to use the MCS plugin to stretch the clusters.
Please refer to [this document](https://aswinayyolath.github.io/stretch-kafka-docs/Setting-up-cilium/).

neeraj-laad:
This proposal was created to enable Kafka cluster deployments stretched across multiple low‑latency Kubernetes clusters. Over the lifetime of this proposal, we haven't seen strong interest in such deployments from other users or community members. At this point it doesn't seem justified to introduce this level of complexity to the Strimzi project until we have evidence that it will deliver value for a sufficiently broad set of users. We would therefore like to close this proposal while we validate the value proposition.

We will keep the assets created as part of this work available for others to reference and experiment with, including:

- Stretch‑cluster documentation (including Submariner and Cilium setup, DNS configuration, and test scenarios): https://aswinayyolath.github.io/strimzi-stretch-cluster

- Strimzi‑Kafka‑Operator changes: https://github.com/aswinayyolath/strimzi-kafka-operator/tree/48-migration

- MCS plugin implementation: https://github.com/aswinayyolath/strimzi-stretch-mcs-plugin

- NodePort plugin implementation: https://github.com/aswinayyolath/strimzi-stretch-nodeport-plugin

- LoadBalancer plugin implementation: https://github.com/aswinayyolath/strimzi-stretch-loadbalancer-plugin

We would like to sincerely thank all the community members and maintainers who reviewed, tested and contributed suggestions and feedback here. Your input has been extremely valuable in shaping our thinking, and we hope to build on this knowledge if and when we revisit this idea in the future.

katheris:
Thanks @neeraj-laad for the update and for all your team's work on this, I'll go ahead and close this proposal PR. If in future you want to revisit this then feel free to open a new PR.

fjarman:
It's a shame that this proposal got closed. At our company we actually considered doing similar thing and during the research stumbled on this proposal. We run on bear metal k8s cluster, that is manged by us, so it's very important to us, that there are 2 independent k8s clusters in different datacenters, managing a single HA kafka cluster, and this stretch proposal fits prefectly.

PR #191, opened 2026-01-13
Add support for broker demotion

kyguy:
### Related problem

When preparing to remove or decommission brokers from a Kafka cluster there is currently no built-in way to ensure those brokers are first demoted (no longer eligible to act as partition leaders). This can result in partitions becoming temporarily unavailable or degraded when those brokers are taken offline, especially in clusters with uneven leadership distribution. Having a mechanism to programmatically demote brokers would allow for safer, more predictable operations during maintenance or scaling events.

### Suggested solution

This feature would extend the KafkaRebalance resource to leverage Cruise Control’s `/kafkacruisecontrol/demote_broker` endpoint [1], allowing users to specify a list of brokers to be demoted. This would trigger the migration of partition leadership away from those brokers in preparation for decommissioning or maintenance.

Addresses issue discussed here [2]

[1] https://github.com/linkedin/cruise-control/wiki/REST-APIs#demote-a-list-of-brokers-from-the-kafka-cluster
[2] https://github.com/strimzi/strimzi-kafka-operator/issues/11907

### DIFF
--- a/128-broker-demotion-support.md
+++ b/128-broker-demotion-support.md
@@ -122,7 +122,7 @@ If any of these fields are specified, the rebalance request will be rejected, an
 
 ### Interaction with other rebalance modes
 
-Broker demotion is independent of other rebalance modes:
+Broker demotion is independent of other rebalance modes but can be used before or after them manually:
 
 * **add-brokers**: After new brokers are added to the cluster, broker demotion could be used to explicitly transfer partition leadership away from existing brokers to accelerate leadership adoption on newly added brokers. 
 
@@ -156,51 +156,6 @@ The new mode is opt-in and requires explicit user action.
 ### Add support for disk-level demotion
 
 Since Cruise Control's [`/demote_broker`](https://github.com/linkedin/cruise-control/wiki/REST-APIs#demote-a-list-of-brokers-from-the-kafka-cluster) endpoint includes a parameter for demoting individual disks, `brokerid_and_logdirs`, a logical follow up feature would be to add support for disk-level demotion.
-
-For disk-level demotion, partition leadership is moved off specific disk volumes on selected brokers, while replicas remain on those volumes.
-
-A `KafkaRebalance` custom resource for disk demotion would look like this:
-
-```yaml
-apiVersion: kafka.strimzi.io/v1beta2
-kind: KafkaRebalance
-metadata:
-  name: demote-disks-example
-  labels:
-    strimzi.io/cluster: my-cluster
-spec:
-  mode: demote-disks
-  moveLeadersOffVolumes:
-    - brokerId: 0
-      volumeIds: [1, 2]
-    - brokerId: 1
-      volumeIds: [0, 1]
-```
-
-This example would demote the disk volumes 1 and 2 of brokers 0 and volumes 0 and 1 of broker 1, transferring all partition leadership away from those volumes while keeping the replicas in place.
-
-#### Supported fields for `demote-disks` mode
-
-In addition to the supported fields of the `demote-brokers` mode, the `demote-disks` mode in the `KafkaRebalance` resource spec would introduce the following:
-
-| Field                            | Type    | Description
-|----------------------------------|---------|-----------------------------------------------------------------------------|
-| moveLeadersOffVolumes            | [BrokerAndVolumeIds](https://strimzi.io/docs/operators/latest/configuring.html#type-BrokerAndVolumeIds-reference) array | List of broker ID and volume pairs to be demoted in the cluster. |
-
-#### Validation and constraints
-
-The implementation would include the following validation:
-
-* The specified broker and volume IDs in the `moveLeadersOffVolumes` list must exist in the cluster. 
-Cruise Control will validate this and return an error for the `KafkaRebalance` status if invalid broker IDs or volume IDs are provided.
-
-### Add support for broker-level demotion before automatic cluster scaling
-
-A future enhancement could integrate broker-level demotion with automatic cluster scaling as described in [proposal 078](https://github.com/strimzi/proposals/blob/main/078-auto-rebalancing-cluster-scaling.md).
-In such an implementation, brokers scheduled for removal could first be automatically demoted, ensuring that partition leadership is transferred away from those brokers before scaling operations proceed. 
-This would reduce operational overhead and improve safety by minimizing leadership movement during broker removal.
-
-This enhancement would build on the existing auto-rebalancing mechanisms and could be implemented as an optional, configurable step in the cluster scaling process.
 
 ## Rejected alternatives
 
--- a/128-broker-demotion-support.md
+++ b/128-broker-demotion-support.md
@@ -31,7 +31,6 @@ The `remove-brokers` mode is too aggressive for these scenarios because it moves
 - Significant network bandwidth consumption from replica movement
 - Increased CPU and disk I/O on source and destination brokers
 - Extended time to complete the operation
-- Temporary reduction in replication factor during replica migration
 - Unnecessary disruption when the goal is only to remove leadership
 
 Broker demotion addresses these concerns by only transferring leadership, which is a lightweight operation compared to replica movement.
@@ -65,13 +64,13 @@ This example would demote brokers 3 and 4, transferring all partition leadership
 
 The supported fields, new and old, for `demote-brokers` mode in the `KafkaRebalance` resource spec are as follows:
 
-| Field                            | Type    | Description                                                                
-|----------------------------------|---------|-----------------------------------------------------------------------------|
-| brokers                          | integer array    | List of ids of broker to be demoted in the cluster.                |
-| concurrentLeaderMovements        | integer | Upper bound of ongoing leadership swaps. Default is 1000.                   |
-| skipUrpDemotion                  | boolean | Whether to skip demoting leader replicas for under-replicated partitions.   |
-| excludeFollowerDemotion          | boolean | Whether to skip demoting follower replicas on the broker to be demoted.     |
-| excludeRecentlyDemotedBrokers    | boolean | Whether to allow leader replicas to be moved to recently demoted brokers.   |
+| Field                            | Type          | Description                                                                 | Default   |
+|----------------------------------|---------------|-----------------------------------------------------------------------------|------------
+| brokers                          | integer array | List of ids of broker to be demoted in the cluster.                         | N/A       |
+| concurrentLeaderMovements        | integer       | Upper bound of ongoing leadership swaps.                                    | 1000      |
+| skipUrpDemotion                  | boolean       | Whether to skip demoting leader replicas for under-replicated partitions.   | true      |
+| excludeFollowerDemotion          | boolean       | Whether to skip demoting follower replicas on the broker to be demoted.     | true      |
+| excludeRecentlyDemotedBrokers    | boolean       | Whether to allow leader replicas to be moved to recently demoted brokers.   | false     |
 
 **NOTE**: As part of this proposal, we will also add a `excludeRecentlyDemotedBrokers` field for the `full`, `add-brokers`, and `remove-brokers` KafkaRebalance modes to give users the ability to prevent to leader replicas to be moved to recently demoted brokers.
 When `excludeRecentlyDemotedBrokers` is set to `true`, a broker is considered demoted for the duration specified by the Cruise Control `demotion.history.retention.time.ms` server configuration.
@@ -88,7 +87,7 @@ The workflow for using broker demotion follows the same pattern as other rebalan
 3. The operator transitions the `KafkaRebalance` resource to the `ProposalReady` state. 
 The proposal is stored in `status.optimizationResult` and shows which partition leadership transfers will occur.
 
-4. User reviews the proposal and approves it by annotating the resource with `strimzi.io/rebalance=approve`.
+4. If [auto-approval](https://strimzi.io/docs/operators/latest/deploying#automatically_approving_an_optimization_proposal) is not enabled, the user reviews the proposal and approves it by annotating the resource with `strimzi.io/rebalance=approve`.
 
 5. The operator executes the broker demotion via the `/demote_broker` endpoint with `dryrun=false`.
 
@@ -107,7 +106,7 @@ If any of the broker IDs are invalid, the demotion request will be rejected and 
 * When an impossible demotion operation is requested for example demoting all brokers or transferring leadership from the only in-sync replica when the KafkaRebalance `spec.skipUrpDemotion` configuration is set to `false`, the demotion request will be rejected and the error will be reported in the `KafkaRebalance` status.
 
 * If a target broker fails while leadership is being transferred to it, all demotion operations involving that broker are aborted, and the source brokers remain the leaders for the affected partitions.
-In this case, the overall demotion request continues on a best-effort basis with with the remaining proposed operations, transferring the leadership on brokers that are available.
+In this case, the overall demotion request continues on a best-effort basis with the remaining proposed operations, transferring the leadership on brokers that are available.
 
 * The following `KafkaRebalance` resource configuration fields are incompatible with, or no-ops for, broker demotion. 
 If any of these fields are specified, the rebalance request will be rejected, an error will be logged by the Strimzi Operator, and an error will be added to the KafkaRebalance status.
--- a/128-broker-demotion-support.md
+++ b/128-broker-demotion-support.md
@@ -107,6 +107,7 @@ If any of the broker IDs are invalid, the demotion request will be rejected and 
 
 * If a target broker fails while leadership is being transferred to it, all demotion operations involving that broker are aborted, and the source brokers remain the leaders for the affected partitions.
 In this case, the overall demotion request continues on a best-effort basis with the remaining proposed operations, transferring the leadership on brokers that are available.
+Also, a warning listing the affected demoted brokers that still host leader partitions will be added to the `KafkaRebalance` status.
 
 * The following `KafkaRebalance` resource configuration fields are incompatible with, or no-ops for, broker demotion. 
 If any of these fields are specified, the rebalance request will be rejected, an error will be logged by the Strimzi Operator, and an error will be added to the KafkaRebalance status.
--- a/128-broker-demotion-support.md
+++ b/128-broker-demotion-support.md
@@ -1,0 +1,225 @@
+# Broker demotion support via KafkaRebalance resource
+
+This proposal extends the `KafkaRebalance` custom resource to support broker demotion by integrating with Cruise Control's `/demote_broker` endpoint. 
+This would allow users to demote brokers, removing them from partition leadership eligibility, in preparation for maintenance, decommissioning, or other operational needs.
+
+## Current situation
+
+Currently, Strimzi's Cruise Control integration provides several [rebalancing modes](https://strimzi.io/docs/operators/latest/deploying#con-optimization-proposals-modes-str) through the `KafkaRebalance` custom resource:
+* `full`: performs a complete cluster rebalance across all brokers
+* `add-brokers`: moves replicas to newly added brokers after scaling up
+* `remove-brokers`: moves replicas off brokers before scaling down
+* `remove-disks`: performs intra-broker disk rebalancing for JBOD configurations
+
+However, there is no built-in way to demote brokers, which would remove them from partition leadership without moving replicas. 
+When preparing brokers for removal or maintenance, users must rely on the `remove-brokers` mode which moves all partition replicas off the target brokers. 
+This is more disruptive than necessary if the goal is simply to ensure the brokers are not serving as partition leaders.
+
+Cruise Control provides a dedicated [`/demote_broker`](https://github.com/linkedin/cruise-control/wiki/REST-APIs#demote-a-list-of-brokers-from-the-kafka-cluster) endpoint specifically for this use case, but Strimzi does not currently expose it.
+
+## Motivation
+
+There are a few scenarios where demoting brokers without moving replicas is beneficial:
+
+1. **Broker maintenance**: Before performing maintenance on a broker such as upgrading, patching, or restarting, operators may want to transfer leadership away to minimize impact on client traffic, while keeping the broker as a follower to maintain replication factor and availability.
+
+2. **Staged decommissioning**: In a multi-step decommissioning process, operators may first demote brokers to observe the impact on leadership distribution and client performance before committing to fully removing replicas with the `remove-brokers` mode.
+
+3. **Performance isolation**: Operators may want to reduce load on specific brokers experiencing performance issues by removing their leadership responsibilities while keeping them as followers until the issue is diagnosed and resolved.
+
+The `remove-brokers` mode is too aggressive for these scenarios because it moves all replicas off the target brokers resulting in:
+- Significant network bandwidth consumption from replica movement
+- Increased CPU and disk I/O on source and destination brokers
+- Extended time to complete the operation
+- Temporary reduction in replication factor during replica migration
+- Unnecessary disruption when the goal is only to remove leadership
+
+Broker demotion addresses these concerns by only transferring leadership, which is a lightweight operation compared to replica movement.
+
+## Proposal
+
+The proposal is to add a new `demote-brokers` mode to the existing `KafkaRebalance` custom resource, following the same pattern established by the `add-brokers` and `remove-brokers` modes introduced in [proposal 035](https://github.com/strimzi/proposals/blob/main/035-rebalance-types-scaling-brokers.md).
+
+When `spec.mode` is set to `demote-brokers` in the `KafkaRebalance` resource, paritition leadership is moved off the specified brokers while replicas remain in place.
+
+Users must provide a list of broker IDs to demote for broker-level demotion via the `spec.brokers` field.
+
+A `KafkaRebalance` custom resource for broker demotion would look like this:
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaRebalance
+metadata:
+  name: demote-brokers-example
+  labels:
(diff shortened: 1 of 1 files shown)


128-broker-demotion-support.md
```diff
@@ -0,0 +1,225 @@
...
+
+* When `demote-brokers` mode is specified, the `brokers` parameter must be provided.
+If the field is missing or empty, the operator will reject the rebalance request and report an error in the `KafkaRebalance` status.
+
+* The specified broker IDs in the `brokers` list must exist in the cluster. 
+If any of the broker IDs are invalid, the demotion request will be rejected and the error will be reported in the `KafkaRebalance` status.
+
+* When an impossible demotion operation is requested for example demoting all brokers or transfering leadship from the only in-sync replica when the KafkaRebalance `spec.skipUrpDemotion` configuration is set to `false`, the demotion request will be rejected and the error will be reported in the `KafkaRebalance` status.
+
+* The following `KafkaRebalance` resource configuration fields are incompatible with broker demotion and if specified, will cause the rebalance to be rejected, an error to logged by the Strimzi Operator, and an errorto be added to the `KafkaRebalance` status:
```
see-quick:
```suggestion
* The following `KafkaRebalance` resource configuration fields are incompatible with broker demotion and if specified, will cause the rebalance to be rejected, an error to logged by the Strimzi Operator, and an error to be added to the `KafkaRebalance` status:
```

see-quick (commented):
Thanks for the proposal, I think it looks good, just a few nits/questions... 

Out of curiosity, I didn't find it in the proposal but what happens when demotion is in progress but target broker fails in the middle of operation?

128-broker-demotion-support.md
```diff
@@ -0,0 +1,225 @@
...
+| brokers                          | integer array    | List of ids of broker to be demoted in the cluster.                |
+| concurrentLeaderMovements        | integer | Upper bound of ongoing leadership movements. Default is 1000.               |
+| skipUrpDemotion                  | boolean | Whether to skip demoting leader replicas for under-replicated partitions.   |
+| excludeFollowerDemotion          | boolean | Whether to skip demoting follower replicas on the broker to be demoted.     |
+| excludeRecentlyDemotedBrokers    | boolean | Whether to allow leader replicas to be moved to recently demoted brokers.   |
+| replicaMovementStrategies        | string array | A list of strategy class names used to determine the execution order for replica position swaps within partitions (metadata changes only, no data movement) in the generated optimization proposal. By default BaseReplicaMovementStrategy is used, which will execute the replica position swaps in the order in which they are generated. |
+| replicationThrottle             | integer | The upper bound, in bytes per second, on the bandwidth used to move replicas. There is no limit by default. |
```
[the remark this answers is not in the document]
kyguy: You are right, some of these fields are not used by Cruise Control for the `/demote-broker` endpoint even though they are accepted as valid parameters.

I just removed them from this table and updated the validation and constraints section.

kyguy:
Thanks for the review @see-quick 

> Out of curiosity, I didn't find it in the proposal but what happens when demotion is in progress but target broker fails in the middle of operation?

I just added this note to the Validation and constraints section of the proposal:

"If a target broker fails while leadership is being transferred to it, all demotion operations involving that broker are aborted, and the source brokers remain the leaders for the affected partitions.
In this case, the overall demotion request continues on a best-effort with the proposed operations, transferring the leadership on brokers that are available."

128-broker-demotion-support.md
```diff
@@ -0,0 +1,219 @@
...
+
+2. **Staged decommissioning**: In a multi-step decommissioning process, operators may first demote brokers to observe the impact on leadership distribution and client performance before committing to fully removing replicas with the `remove-brokers` mode.
+
+3. **Performance isolation**: Operators may want to reduce load on specific brokers experiencing performance issues by removing their leadership responsibilities while keeping them as followers until the issue is diagnosed and resolved.
+
+The `remove-brokers` mode is too aggressive for these scenarios because it moves all replicas off the target brokers resulting in:
+- Significant network bandwidth consumption from replica movement
+- Increased CPU and disk I/O on source and destination brokers
+- Extended time to complete the operation
+- Temporary reduction in replication factor during replica migration
```
fvaleri: I think this is incorrect. Old replicas are deleted only after the migration is completed.

128-broker-demotion-support.md
```diff
@@ -0,0 +1,219 @@
...
+The workflow for using broker demotion follows the same pattern as other rebalance modes:
+
+1. User creates a `KafkaRebalance` custom resource with `spec.mode: demote-brokers` and specifies the target broker IDs in `spec.brokers`.
+
+2. The `KafkaRebalanceAssemblyOperator` requests an optimization proposal from Cruise Control via the `/demote_broker` endpoint with `dryrun=true`.
+
+3. The operator transitions the `KafkaRebalance` resource to the `ProposalReady` state. 
+The proposal is stored in `status.optimizationResult` and shows which partition leadership transfers will occur.
+
+4. User reviews the proposal and approves it by annotating the resource with `strimzi.io/rebalance=approve`.
```
fvaleri: Does it also work with auto-approval annotation?

128-broker-demotion-support.md
```diff
@@ -0,0 +1,219 @@
...
+No structural changes to the CRD schema are needed beyond allowing the new enum value and adding new fields.
+
+* **Behavioral compatibility**: Existing rebalancing workflows are unaffected. 
+The new mode is opt-in and requires explicit user action.
+
+## Future Improvements
+
+### Add support for disk-level demotion
+
+Since Cruise Control's [`/demote_broker`](https://github.com/linkedin/cruise-control/wiki/REST-APIs#demote-a-list-of-brokers-from-the-kafka-cluster) endpoint includes a parameter for demoting individual disks, `brokerid_and_logdirs`, a logical follow up feature would be to add support for disk-level demotion.
```
fvaleri: Given that this operation is just an extension of the same CC endpoint, have you considered making this part of the proposal that will be implemented in a dedicated PR?

fvaleri (commented):
Thanks @kyguy. I left some comments.

I think it is also important to document that broker demotion does NOT prevent new partition leaders to be scheduled on the selected broker.

128-broker-demotion-support.md
```diff
@@ -0,0 +1,219 @@
...
+  * `rebalanceDisk`
+  * `excludedTopics`
+  * `concurrentPartitionMovementsPerBroker`
+  * `concurrentIntraBrokerPartitionMovements`
+  * `moveReplicasOffVolumes`
+  * `replicationThrottle`
+
+### Interaction with other rebalance modes
+
+Broker demotion is independent of other rebalance modes:
```
tinaselenge:
```suggestion
Broker demotion is independent of other rebalance modes but can be used before or after them manually:
```

128-broker-demotion-support.md
```diff
@@ -0,0 +1,219 @@
...
+spec:
+  mode: demote-disks
+  moveLeadersOffVolumes:
+    - brokerId: 0
+      volumeIds: [1, 2]
+    - brokerId: 1
+      volumeIds: [0, 1]
+```
+
+This example would demote the disk volumes 1 and 2 of brokers 0 and volumes 0 and 1 of broker 1, transferring all partition leadership away from those volumes while keeping the replicas in place.
```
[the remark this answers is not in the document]
tinaselenge: I guess, they would be moved to other disks of other brokers wherever the other followers are on? So if volume 1 of broker 2 is demoted, leaders on volume 1 will become followers making other replicas on different brokers leaders instead but leaders on volume 0 of broker 2 will remain. Is that right?

tinaselenge (commented):
Thanks for the proposal. Overall, it looks good to me. I left a few comments to clarify.

128-broker-demotion-support.md
```diff
@@ -0,0 +1,219 @@
...
+
+5. The operator executes the broker demotion via the `/demote_broker` endpoint with `dryrun=false`.
+
+6. When complete, the operator transitions the `KafkaRebalance` resource to `Ready` state.
+
+### Validation and constraints
+
+The implementation includes the following validation:
+
+* When `demote-brokers` mode is specified, the `brokers` field must be provided.
```
scholzj: Will it be validated only in the broker code or also with CEL?

128-broker-demotion-support.md
```diff
@@ -0,0 +1,219 @@
...
+spec:
+  mode: demote-brokers
+  brokers:
+    - 3
+    - 4
+```
+
+This example would demote brokers 3 and 4, transferring all partition leadership away from them while keeping the replicas in place.
+
+### Supported fields for `demote-broker` mode
```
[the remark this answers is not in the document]
kyguy: > We should try to consolidate the options and use a config section for configuring the small auxiliary options.

> But at this point I think we should discuss and work on this first and not move forward with this proposal now.

I imagine an API change of this nature would likely merit its own dedicated proposal. If you both agree, I’m happy to draft one so we can follow up on this there before moving forward with the broker demotion work. What do you think?

128-broker-demotion-support.md
kyguy: I have added a new note to the workflow section to highlight this

128-broker-demotion-support.md
```diff
@@ -0,0 +1,219 @@
...
+* When `demote-brokers` mode is specified, the `brokers` field must be provided.
+If the field is missing or empty, the operator will reject the rebalance request and report an error in the `KafkaRebalance` status.
+
+* The specified broker IDs in the `brokers` list must exist in the cluster. 
+If any of the broker IDs are invalid, the demotion request will be rejected and the error will be reported in the `KafkaRebalance` status.
+
+* When an impossible demotion operation is requested for example demoting all brokers or transferring leadership from the only in-sync replica when the KafkaRebalance `spec.skipUrpDemotion` configuration is set to `false`, the demotion request will be rejected and the error will be reported in the `KafkaRebalance` status.
+
+* If a target broker fails while leadership is being transferred to it, all demotion operations involving that broker are aborted, and the source brokers remain the leaders for the affected partitions.
+In this case, the overall demotion request continues on a best-effort basis with with the remaining proposed operations, transferring the leadership on brokers that are available.
```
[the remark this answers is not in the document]
kyguy: No plans are included for this as part of this proposal. Cruise Control moves leadership on a best effort basis and unfortunately doesn't report any errors for this that we can leverage for the `KafkaRebalance` resource. 

That being said, Cruise Control does provide a list of the leadership movements (when the `verbose` parameter is included for the [/demote_broker](https://github.com/linkedin/cruise-control/wiki/REST-APIs#demote-a-list-of-brokers-from-the-kafka-cluster) endpoint)) so we could very well detect and raise an error for this on the Strimzi side but it would add more complexity to this integration proposal/implementation. This is probably something best addressed in upstream Cruise Control first. 

Maybe it is something we can list in the "Future Improvements" section of this proposal, what do you think?

128-broker-demotion-support.md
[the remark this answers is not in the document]
kyguy: > Where do the leaders get moved? To other disks on the same broker or to other disks considering all brokers?

 To other disks on the same broker.
 
Since this proposal is meant to focus specifically on broker demotion, I will move these disk demotion–specific details into a future proposal dedicated to disk demotion.

128-broker-demotion-support.md
```diff
@@ -0,0 +1,219 @@
...
+| concurrentLeaderMovements        | integer | Upper bound of ongoing leadership swaps. Default is 1000.                   |
+| skipUrpDemotion                  | boolean | Whether to skip demoting leader replicas for under-replicated partitions.   |
+| excludeFollowerDemotion          | boolean | Whether to skip demoting follower replicas on the broker to be demoted.     |
+| excludeRecentlyDemotedBrokers    | boolean | Whether to allow leader replicas to be moved to recently demoted brokers.   |
+
+**NOTE**: As part of this proposal, we will also add a `excludeRecentlyDemotedBrokers` field for the `full`, `add-brokers`, and `remove-brokers` KafkaRebalance modes to give users the ability to prevent to leader replicas to be moved to recently demoted brokers.
+When `excludeRecentlyDemotedBrokers` is set to `true`, a broker is considered demoted for the duration specified by the Cruise Control `demotion.history.retention.time.ms` server configuration.
+By default, this value is 1209600000 milleseconds (14 days) but is configurable in the `spec.cruiseControl.config` section of the `Kafka` custom resource.
```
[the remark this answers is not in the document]
kyguy: Cruise Control stores the demotion information in memory (yes this is far from ideal). Cruise Control will make sure the ids of the demoted brokers are in the end of the replica lists and will exclude the brokers from being considered for leadership when generating optimization proposals.

kyguy:
Thanks everyone for the reviews! I am putting this proposal on hold temporarily as we address some of the API concerns raised in the thread here: https://github.com/strimzi/proposals/pull/191#discussion_r2705867586. We need devise a better path forward for the parameters that are used in conjunction with  `mode` field of the `KafkaRebalance` resource, maybe by consolidating primitive fields into a generic config section to ensure the API can support future modes cleanly. 

I am going to draft a separate proposal for those API changes and link it here. Once that proposal and its implementation is complete, we can continue this broker demotion proposal with a cleaner, more maintainable foundation.

128-broker-demotion-support.md
```diff
@@ -0,0 +1,219 @@
+# Broker demotion support via KafkaRebalance resource
+
+This proposal extends the `KafkaRebalance` custom resource to support broker demotion by integrating with Cruise Control's `/demote_broker` endpoint. 
+This would allow users to demote brokers, removing them from partition leadership eligibility, in preparation for maintenance, decommissioning, or other operational needs.
```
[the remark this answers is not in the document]
Frawless: What happens when CC pod is restarted (upgrade, migration to different node, etc).? User will have to trigger "rebalance" proposal and approval to refresh in-memory data?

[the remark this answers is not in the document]
kyguy: At this time yes, to refresh the in-memory demotion data after Cruise Control pod restart, a user would have to trigger another `KafkaRebalance` demotion request.

128-broker-demotion-support.md
```diff
@@ -0,0 +1,174 @@
...
+  * `concurrentPartitionMovementsPerBroker`
+  * `concurrentIntraBrokerPartitionMovements`
+  * `moveReplicasOffVolumes`
+  * `replicationThrottle`
+
+### Interaction with other rebalance modes
+
+Broker demotion is independent of other rebalance modes but can be used before or after them manually:
+
+* **add-brokers**: After new brokers are added to the cluster, broker demotion could be used to explicitly transfer partition leadership away from existing brokers to accelerate leadership adoption on newly added brokers.
```
ppatierno: I am not sure about this use case, isn't the auto-rebalancing (we support) already doing something like this? It's moving partitions but not demoting brokers.

ppatierno: Yes It does, thanks.

kyguy:
> I am going to draft a separate proposal for those API changes and link it here. Once that proposal and its implementation is complete, we can continue this broker demotion proposal with a cleaner, more maintainable foundation.

Opened a draft proposal for the `KafkaRebalance` API changes here https://github.com/strimzi/proposals/pull/216. We can follow up on the conversation there and return to this proposal once the proposal is reviewed and implemented.

kyguy:
This proposal has been updated with the new KafkaRebalance API design as specified in the accepted proposal: https://github.com/strimzi/proposals/pull/216), is unblocked, and ready for review!

128-broker-demotion-support.md
```diff
@@ -0,0 +1,298 @@
...
+Similarly, when preparing disks for removal, users must rely on the `remove-disks` mode which moves all partition replicas off the target disks.
+These operations are more disruptive than necessary if the goal is simply to ensure the brokers or disks are not serving as partition leaders.
+
+Cruise Control provides a dedicated [`/demote_broker`](https://github.com/linkedin/cruise-control/wiki/REST-APIs#demote-a-list-of-brokers-from-the-kafka-cluster) endpoint specifically for this use case, but Strimzi does not currently expose it via the `KafkaRebalance` resource.
+
+## Motivation
+
+There are a few scenarios where demoting brokers or disks without moving replicas is beneficial:
+
+1. **Broker or disk maintenance**: Before performing maintenance on a broker such as upgrading, patching, or restarting, operators may want to transfer leadership away to minimize impact on client traffic, while maintaining replication factor and availability.
```
scholzj, line 25: Is this really about `upgrading, patching, or restarting` brokers? If so, we do not need to add anything into the KafkaRebalance resource but into the KafkaRoller as all these tasks are handled by the operator.

128-broker-demotion-support.md
```diff
@@ -0,0 +1,298 @@
...
+* `add-brokers`:  Moves replicas to newly added brokers.
+* `remove-brokers`: Moves replicas out of brokers to be removed.
+* `remove-disks`: Moves replicas off specified JBOD volumes of specified brokers so those volumes can be removed.
+
+However, there is no built-in way to demote brokers or disks, meaning to remove them from partition leadership without moving replicas.
+When preparing brokers for removal or maintenance, users must rely on the `remove-brokers` mode which moves all partition replicas off the target brokers.
+Similarly, when preparing disks for removal, users must rely on the `remove-disks` mode which moves all partition replicas off the target disks.
+These operations are more disruptive than necessary if the goal is simply to ensure the brokers or disks are not serving as partition leaders.
+
+Cruise Control provides a dedicated [`/demote_broker`](https://github.com/linkedin/cruise-control/wiki/REST-APIs#demote-a-list-of-brokers-from-the-kafka-cluster) endpoint specifically for this use case, but Strimzi does not currently expose it via the `KafkaRebalance` resource.
```
scholzj, line 19: You present this with fancy words. But this really only moves the preferred leader replica, right? There is no cordoning etc. involved. I think you need to be a bit more clear about it here and in the next sections because it better explains what users should expect from this feature.

128-broker-demotion-support.md
```diff
@@ -0,0 +1,298 @@
...
+A `KafkaRebalance` custom resource for disk-level demotion would look like this:
+```yaml
+apiVersion: kafka.strimzi.io/v1
+kind: KafkaRebalance
+metadata:
+  name: demote-broker-disks-example
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  mode: demote-brokers
```
scholzj, line 82: Is the `demote-brokers` the right name if it can demote brokers and disks? Would `demote` make more sense?

128-broker-demotion-support.md
```diff
@@ -0,0 +1,298 @@
...
+
+  9. **Mixed broker-level and disk-level demotion in same resource (Strimzi rejects)**:
+    - **KafkaRebalance status**: `NotReady` with message: "Cannot mix broker-level and disk-level demotion. 
+      All `nodes` entries must either specify `volumeIds` or none must."
+    - **Cluster Operator log**: WARN with same message
+    - **Cruise Control log**: N/A
+
+### Interaction with other rebalance modes
+
+Broker demotion is independent of other rebalance modes but can be used before or after them manually:
```
scholzj, line 255: What does it mean that it is `independent`?

128-broker-demotion-support.md
ppatierno, line 19: Should we now use links to the new Cruise Control home as part of LF?

128-broker-demotion-support.md
```diff
@@ -0,0 +1,298 @@
...
+    - brokerId: 4
+  config:
+    concurrent_leader_movements: 10
+```
+
+This example would demote brokers 3 and 4, transferring all partition leadership away from them while keeping the replicas in place.
+
+#### Disk-level demotion
+
+For disk-level demotion, users can specify a list of volume (disk) IDs to demote via the `spec.nodes` entries following the unified targeting mechanism introduced in  [proposal 154](154-kafkarebalance-custom-resource-consolidation.md).
```
ppatierno, line 71: You should be clear about the API explaining that the list of volume/disk IDs is per broker, so you need anyway a list of brokers and for each the corresponding volumes/disks to demote.

128-broker-demotion-support.md
ppatierno: In general, we always tried to match the CC endpoint (even if we have already an exception with "full" mode) so I would agree with Jakub on this.

128-broker-demotion-support.md
```diff
@@ -0,0 +1,298 @@
...
+### Implementation Strategy
+
+1. **Validation**
+
+    - **Mode-specific operand validation**:
+      - `nodes` is required and non-empty for `demote-brokers` mode.
+      - For `demote-brokers` mode, `volumeIds` on `nodes` entries is optional.
+        When provided, only partitions on the specified volumes are demoted.
+        When omitted, all partitions on the broker are demoted.
+      - Broker-level and disk-level demotion cannot be mixed in a single `KafkaRebalance` resource.
```
ppatierno, line 170: Why? Can't you use the `/demote_brokers` entry point this way?

```shell
/kafkacruisecontrol/demote_broker
    ?brokerid=1
    &brokerid_and_logdirs=2-/var/kafka/data0,2-/var/kafka/data1
```

by having a `KafkaRebalance` as ...

```yaml
kind: KafkaRebalance
metadata:
  name: demote-brokers-example
  labels:
    strimzi.io/cluster: my-cluster
spec:
  mode: demote-brokers
  nodes:
    - brokerId: 1
    - brokerId: 2
          volumeIds:
          - 0
          - 1
  config:
    concurrent_leader_movements: 10
```

or did I get what you meant wrong?

katheris (commented):
Thanks @kyguy, I'm generally happy with the approach described here. My only questions are around the comment Paolo added for whether the `brokerid` and `brokerid_and_logdirs` query parameters can be used in the same request. If that is the case then I would be in favour of us keeping a single new mode called `demote-brokers` to line up with the Cruise Control naming.

However, if the user really can only demote brokers or demote specific disks, and not mix and match I wonder if we should consider having two modes, like we do for `remove-brokers` and `remove-disks`.

Also, am I correct in thinking that this would only demote the brokers and not perform a preferred leadership election? That wasn't clear to me from the proposal, but that is my understanding of what was described.

128-broker-demotion-support.md
```diff
@@ -0,0 +1,298 @@
...
+        All `nodes` entries must either specify `volumeIds` (disk-level) or none must (broker-level).
+      - The specified broker IDs in the `nodes` list must exist in the cluster.
+      - Impossible demotion operations are rejected, for example demoting all brokers or transferring leadership from the only in-sync replica when `skip_urp_demotion` is set to `"false"` in `spec.config`.
```
tinaselenge, line 173: how is this validated? Are we checking against all the impossible situations, such as checking in sync replicas first etc.?

128-broker-demotion-support.md
```diff
@@ -0,0 +1,298 @@
...
+
+4. If [auto-approval](https://strimzi.io/docs/operators/latest/deploying#automatically_approving_an_optimization_proposal) is not enabled, the user reviews the proposal and approves it by annotating the resource with `strimzi.io/rebalance=approve`.
+
+5. The operator executes the broker demotion via the `/demote_broker` endpoint with `dryrun=false`.
+
+6. When complete, the operator transitions the `KafkaRebalance` resource to `Ready` state.
+
+### Implementation Strategy
+
+1. **Validation**
```
tinaselenge, line 163: Are these validations all done by the operator, or some by CC?

128-broker-demotion-support.md
```diff
@@ -0,0 +1,298 @@
...
+    - **Mode-specific operand validation**:
+      - `nodes` is required and non-empty for `demote-brokers` mode.
+      - For `demote-brokers` mode, `volumeIds` on `nodes` entries is optional.
+        When provided, only partitions on the specified volumes are demoted.
+        When omitted, all partitions on the broker are demoted.
+      - Broker-level and disk-level demotion cannot be mixed in a single `KafkaRebalance` resource.
+        All `nodes` entries must either specify `volumeIds` (disk-level) or none must (broker-level).
+      - The specified broker IDs in the `nodes` list must exist in the cluster.
+      - Impossible demotion operations are rejected, for example demoting all brokers or transferring leadership from the only in-sync replica when `skip_urp_demotion` is set to `"false"` in `spec.config`.
+      - If a target broker fails during leadership transfer, demotion operations involving that broker are aborted and the remaining operations continue on a best-effort basis.
```
tinaselenge, line 174: Is this how CC handles failure rather than a validation?

128-broker-demotion-support.md
```diff
@@ -0,0 +1,298 @@
...
+
+##### Supported parameters
+
+Following the `spec.config` approach from [proposal 154](154-kafkarebalance-custom-resource-consolidation.md), the parameters supported by Cruise Control's [`/demote_broker`](https://github.com/cruise-control-for-kafka/cruise-control/wiki/REST-APIs#demote-a-list-of-brokers-from-the-kafka-cluster) endpoint are passed as key-value pairs in `spec.config`:
+
+| Config Key                          | Description                                                                         | Default   |
+|-------------------------------------|-------------------------------------------------------------------------------------|-----------|
+| `concurrent_leader_movements`       | Upper bound of ongoing leadership movements.                                        | N/A       |
+| `skip_urp_demotion`                 | Whether to skip demoting leader replicas for under-replicated partitions.           | `"true"`  |
+| `exclude_follower_demotion`         | Whether to skip demoting follower replicas on the broker to be demoted.             | `"true"`  |
```
tinaselenge, line 118:
```suggestion
| `exclude_follower_demotion`         | When true, only partitions where the broker is currently the leader are affected. When false, replica lists are also reordered for partitions where the broker is a follower, pushing its ID to the end so it becomes the least preferred candidate for future leader elections on those partitions.            | `"true"`  |
```

PR #204, opened 2026-03-05
add proposal for multi namespace topic operator

KyriosGN0:
This PR aims to introduce a multi namespace topic operator, its based on the proposal for UTO, and uses a different approach as opposed to https://github.com/strimzi/strimzi-kafka-operator/issues/1206

### DIFF
--- a/133-multi-namespace-topic-operator.md
+++ b/133-multi-namespace-topic-operator.md
@@ -60,7 +60,7 @@ The existing `spec.entityOperator.topicOperator.watchedNamespace` field is depre
 
 **Interaction rules:** When both `watchedNamespace` and `watchedNamespaces` are set, `watchedNamespaces` takes precedence and a warning is logged indicating that the singular field is being ignored. When only the singular field is set, the TO behaves exactly as it does today — watching the Kafka cluster's own namespace plus the single additional namespace specified.
 
-**Label impact:** Migrating from the singular to the plural field does **not** require any changes to `strimzi.io/cluster` labels on existing `KafkaTopic` resources. The TO accepts both the unqualified and qualified label forms regardless of which configuration field is used. However, the unqualified form is deprecated (see below) and should be migrated to the qualified `<namespace>/<name>` form.
+**Label impact:** Migrating from the singular to the plural field does **not** require any changes to `strimzi.io/cluster` labels on existing `KafkaTopic` resources. The TO accepts both the unqualified and qualified label forms regardless of which configuration field is used. However, the unqualified form is deprecated (see below) and should be migrated to the qualified `<namespace>/<name>` form, the TO will also log a warning that there is a possiblty of multiple TO watching over the topic unless its labels is updated. 
 
 #### Deprecation of unqualified `strimzi.io/cluster` label
 
@@ -333,4 +333,3 @@ Not affected:
 2. **Forces naming convention**: Encoding the namespace in the topic name forces a specific naming convention (`<namespace>.<topic>`) that may not match existing topic naming patterns or be desirable for all users.
 3. **Dual CRD confusion**: Having both `KafkaTopic` and `KafkaNamespaceTopic` with different sync semantics (the original proposal suggested unidirectional sync for the new CRD) creates confusion about which to use when.
 4. **Unnecessary complexity**: Extending the existing UTO to watch multiple namespaces achieves the same goal without introducing a new resource type, by building on the conflict resolution and lifecycle mechanisms the UTO already has.
-
--- a/133-multi-namespace-topic-operator.md
+++ b/133-multi-namespace-topic-operator.md
@@ -1,0 +1,312 @@
+# Multi-namespace watching for the Topic Operator
+
+Issue [#1206](https://github.com/strimzi/strimzi-kafka-operator/issues/1206)
+
+## Current situation
+
+The Unidirectional Topic Operator (UTO, [proposal 051](https://github.com/strimzi/proposals/blob/main/051-unidirectional-topic-operator.md)) watches `KafkaTopic` resources in a single namespace — typically the namespace where the `Kafka` resource is deployed. The UTO reconciles topic state unidirectionally from Kubernetes to Kafka, uses `creationTimestamp`-based conflict resolution when multiple `KafkaTopic` CRs reference the same Kafka topic, and employs finalizer-based deletion.
+
+The UTO is deployed as a container within the Entity Operator pod, which the Cluster Operator manages. The `STRIMZI_NAMESPACE` environment variable controls which namespace the UTO watches.
+
+## Motivation
+
+In Kubernetes clusters, it is common practice to isolate applications in separate namespaces. When a Kafka cluster is shared among multiple applications or teams, each application typically lives in its own namespace. Currently, all `KafkaTopic` resources must be created in the Kafka cluster's namespace, which creates several problems:
+
+- **Access control**: Teams that need to create topics must have RBAC permissions in the Kafka namespace, which may also grant access to other infrastructure resources they should not touch.
+- **Resource management**: Kubernetes namespace-level quotas, and RBAC cannot be used to separate topic ownership per team.
+- **GitOps workflows**: Application teams want to declare their Kafka topics alongside their application manifests in their own namespace, not in a shared infrastructure namespace.
+- **Namespace hygiene**: Hundreds of `KafkaTopic` resources from different applications co-mingle in a single namespace, making management and auditing harder.
+
+Allowing the Topic Operator to watch `KafkaTopic` resources across multiple namespaces solves these problems while maintaining the existing UTO semantics.
+
+## Proposal
+
+This proposal extends the UTO to watch `KafkaTopic` resources in multiple namespaces. It builds on the UTO's existing mechanisms for conflict resolution, deletion, and status reporting.
+
+### Configuration
+
+A new field `watchedNamespaces` is added to the Topic Operator configuration in the `Kafka` custom resource:
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+  namespace: kafka
+spec:
+  entityOperator:
+    topicOperator:
+      watchedNamespaces:
+        - team-a
+        - team-b
+        - team-c
+```
+
+The `watchedNamespaces` field accepts:
+- A list of specific namespace names (e.g., `["team-a", "team-b"]`)
+- A wildcard `["*"]` to watch all namespaces
+
+The Topic Operator's own namespace (where the `Kafka` resource lives) is **always** watched implicitly. It does not need to be listed in `watchedNamespaces`. When `watchedNamespaces` is not set or is empty, the behavior is unchanged — the TO watches only its own namespace.
+
+For the standalone Topic Operator deployment, the existing `STRIMZI_NAMESPACE` environment variable is extended to accept a comma-separated list of namespaces or `*`, consistent with how the Cluster Operator already supports multi-namespace watching.
+
+### RBAC
+
+Watching additional namespaces requires the Topic Operator to have Kubernetes RBAC permissions in those namespaces. The Cluster Operator is responsible for creating and managing these RBAC resources.
+
+**For explicit namespace list:**
+
+The Cluster Operator creates a `Role` and `RoleBinding` in each watched namespace, granting the Entity Operator's `ServiceAccount` permissions to watch, list, get, and update `KafkaTopic` resources and their status.
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: strimzi-topic-operator
+  namespace: team-a
+  labels:
+    strimzi.io/cluster: my-cluster
+rules:
+  - apiGroups: ["kafka.strimzi.io"]
+    resources: ["kafkatopics", "kafkatopics/status"]
+    verbs: ["get", "list", "watch", "create", "patch", "update", "delete"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: strimzi-topic-operator
+  namespace: team-a
+  labels:
+    strimzi.io/cluster: my-cluster
+subjects:
+  - kind: ServiceAccount
+    name: my-cluster-entity-operator
+    namespace: kafka
+roleRef:
+  kind: Role
+  name: strimzi-topic-operator
+  apiGroup: rbac.authorization.k8s.io
+```
+
+**For wildcard (`*`):**
+
+The Cluster Operator creates a `ClusterRole` and `ClusterRoleBinding` instead:
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: strimzi-topic-operator-my-cluster
+  labels:
+    strimzi.io/cluster: my-cluster
+rules:
+  - apiGroups: ["kafka.strimzi.io"]
+    resources: ["kafkatopics", "kafkatopics/status"]
+    verbs: ["get", "list", "watch", "create", "patch", "update", "delete"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: strimzi-topic-operator-my-cluster
+  labels:
+    strimzi.io/cluster: my-cluster
+subjects:
+  - kind: ServiceAccount
+    name: my-cluster-entity-operator
+    namespace: kafka
+roleRef:
+  kind: ClusterRole
+  name: strimzi-topic-operator-my-cluster
+  apiGroup: rbac.authorization.k8s.io
+```
+
+When `watchedNamespaces` is updated (namespaces added or removed), the Cluster Operator reconciles the RBAC resources: creating new `Role`/`RoleBinding` pairs for added namespaces and deleting them for removed namespaces. RBAC resources created by the Cluster Operator are labeled with `strimzi.io/cluster` so they can be identified and cleaned up.
+
+### Cross-namespace conflict resolution
+
+The UTO already maintains an in-memory map of *resolved topic name* → (`namespace`, `name`) pairs for conflict detection within a single namespace. This proposal extends that map to span all watched namespaces.
+
+The **resolved topic name** is `spec.topicName` if set, otherwise `metadata.name`. This is critical — two `KafkaTopic` CRs in different namespaces can reference the same Kafka topic through different mechanisms:
+
+- Namespace A: `KafkaTopic` with `metadata.name: my-topic` (no `spec.topicName`, resolves to `my-topic`)
+- Namespace B: `KafkaTopic` with `metadata.name: app-topic`, `spec.topicName: my-topic` (resolves to `my-topic`)
+
+Both reference the same Kafka topic. The conflict resolution rules are:
+
+1. The `KafkaTopic` with the **oldest** `metadata.creationTimestamp` is the owner, regardless of which namespace it is in.
+2. All other `KafkaTopic` CRs referencing the same Kafka topic get:
+
+```yaml
+status:
+  conditions:
+  - type: Ready
+    status: "False"
+    reason: ResourceConflict
+    message: "Managed by team-a/my-topic"
+    lastTransitionTime: "2024-11-24T17:00:00Z"
+```
+
+This is identical to how the UTO handles conflicts within a single namespace today, extended across namespace boundaries.
+
+### Deletion semantics
+
+The UTO uses Kubernetes finalizers to ensure that deleting a `KafkaTopic` also deletes the corresponding Kafka topic. This behavior is unchanged in the multi-namespace case.
+
+When the **owning** `KafkaTopic` (the oldest one) is deleted:
+1. The finalizer logic deletes the Kafka topic from the Kafka cluster.
+2. Any other `KafkaTopic` CRs in other namespaces that referenced the same Kafka topic will have their status updated to reflect that the topic no longer exists. They can then be cleaned up by their respective namespace owners, or if left in place, they will attempt to recreate the topic on their next reconciliation (becoming the new owner since no conflict exists).
+
+When a **non-owning** `KafkaTopic` (one with `ResourceConflict` status) is deleted:
+1. The finalizer is removed without deleting the Kafka topic (same as existing UTO behavior for conflicting resources).
+2. No impact on the owning `KafkaTopic` or the Kafka topic itself.
+
+The existing `strimzi.io/managed: false` annotation continues to work as before — it allows a `KafkaTopic` to be deleted without deleting the Kafka topic, enabling ownership transfers.
+
+### Preventing overlapping Topic Operator instances
+
+When multiple Kafka clusters exist in the same Kubernetes cluster, each has its own Topic Operator instance. If two TO instances are configured to watch overlapping namespaces, conflicts can arise.
+
+To prevent this, the TO sets an annotation on each `KafkaTopic` it reconciles:
+
+```yaml
+metadata:
+  annotations:
+    strimzi.io/managed-by: kafka/my-cluster
+```
+
+The annotation value follows the format `<namespace>/<kafka-cr-name>`, using `/` as a separator (not `-`, which would be ambiguous).
+
+When a TO encounters a `KafkaTopic` with a `strimzi.io/managed-by` annotation that does not match its own identity, it skips the resource and updates its status:
+
+```yaml
+status:
+  conditions:
+  - type: Ready
+    status: "False"
+    reason: ManagedByOtherInstance
+    message: "This KafkaTopic is managed by kafka/other-cluster"
+    lastTransitionTime: "2024-11-24T17:00:00Z"
+```
+
+This is a best-effort mechanism — there is a brief race window during initial reconciliation where two TOs could both attempt to claim a resource. The `strimzi.io/managed-by` annotation should be set atomically during the first reconciliation to minimize this window.
+
+### Namespace lifecycle considerations
+
+**Adding a namespace to `watchedNamespaces`:**
+The Cluster Operator creates RBAC resources in the new namespace and restarts/reconfigures the TO. The TO starts watching the new namespace and reconciles any existing `KafkaTopic` resources in it.
+
+**Removing a namespace from `watchedNamespaces`:**
+The Cluster Operator should remove the `strimzi.io/topic-operator` finalizer from all `KafkaTopic` resources in the namespace being removed, then delete the RBAC resources. This prevents `KafkaTopic` resources with finalizers from blocking namespace deletion after the TO stops watching.
+
+**Watched namespace is deleted:**
+If the TO is running, it processes the finalizers normally — deleting Kafka topics as appropriate. If the TO is not running (or the namespace was already removed from the watch list), `KafkaTopic` resources with finalizers will block namespace deletion. In this case, manual finalizer removal is necessary (e.g., `kubectl patch kafkatopic <name> -n <ns> --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'`). This is an existing limitation of the UTO that applies equally to the single-namespace case, and should be documented as operational guidance.
+
+**TO restart:**
+On startup, the TO scans all watched namespaces to rebuild its in-memory conflict resolution map before beginning reconciliation. This is the same process used today for a single namespace, extended to multiple namespaces. Startup time increases linearly with the number of watched namespaces and `KafkaTopic` resources.
+
+### Example
+
+A Kafka cluster in namespace `kafka`, with topics managed by teams in `team-a` and `team-b`:
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+  namespace: kafka
+spec:
+  entityOperator:
+    topicOperator:
+      watchedNamespaces:
+        - team-a
+        - team-b
+```
+
+Team A creates a topic in their namespace:
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaTopic
+metadata:
+  name: orders
+  namespace: team-a
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  partitions: 12
+  replicas: 3
+  config:
+    retention.ms: "604800000"
+```
+
+The TO reconciles this topic, creates it in Kafka, and sets the annotation:
+
+```yaml
+metadata:
+  name: orders
+  namespace: team-a
+  annotations:
+    strimzi.io/managed-by: kafka/my-cluster
+  finalizers:
+    - strimzi.io/topic-operator
+status:
+  topicName: orders
+  conditions:
+  - type: Ready
+    status: "True"
+    lastTransitionTime: "2024-11-24T17:00:00Z"
+```
+
+If Team B creates a `KafkaTopic` also named `orders` in their namespace, it receives a conflict error:
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaTopic
+metadata:
+  name: orders
+  namespace: team-b
+status:
+  conditions:
+  - type: Ready
+    status: "False"
+    reason: ResourceConflict
+    message: "Managed by team-a/orders"
+    lastTransitionTime: "2024-11-24T17:05:00Z"
+```
+
+## Affected/not affected projects
+
+Affected:
+- **Topic Operator**: Extended to watch multiple namespaces, cross-namespace conflict resolution
+- **Cluster Operator**: Creates RBAC resources for watched namespaces, passes `watchedNamespaces` configuration to the Entity Operator
+
+Not affected:
+- User Operator, Kafka Connect operator, Kafka MirrorMaker operator
+- KafkaTopic CRD schema (no changes to spec fields)
+
+## Compatibility
+
+- The `watchedNamespaces` field is optional with an empty default, preserving existing single-namespace behavior.
+- No changes to the `KafkaTopic` CRD spec schema.
+- The `strimzi.io/managed-by` annotation is new but does not affect existing single-namespace deployments (it will simply be set to the TO's own namespace/cluster, which is a no-op for conflict detection with a single TO instance).
+- Existing `KafkaTopic` resources continue to work without modification.
+
+## Future work
+
+- Multi-namespace support for the User Operator (`KafkaUser` resources) — this will require a separate proposal addressing the security implications of cross-namespace ACL management (see [PR #137 discussion](https://github.com/strimzi/proposals/pull/137)).
+- Namespace-level policies to restrict which topic configurations can be set from a given namespace (e.g., allowing infra teams to retain control over certain topic-level configs).
+
+## Rejected alternatives
+
+### New `KafkaNamespaceTopic` CRD
+
+[Issue #1206](https://github.com/strimzi/strimzi-kafka-operator/issues/1206) proposed a `KafkaNamespaceTopic` CRD that would encode the Kubernetes namespace in the Kafka topic name (e.g., `team-a.orders`). This approach was rejected because:
+
+1. **CRD maintenance cost**: Each custom resource is expensive to maintain. A new CRD requires its own controller logic, status management, validation webhooks, documentation, and migration tooling.
+2. **Forces naming convention**: Encoding the namespace in the topic name forces a specific naming convention (`<namespace>.<topic>`) that may not match existing topic naming patterns or be desirable for all users.
+3. **Dual CRD confusion**: Having both `KafkaTopic` and `KafkaNamespaceTopic` with different sync semantics (the original proposal suggested unidirectional sync for the new CRD) creates confusion about which to use when.
+4. **Unnecessary complexity**: Extending the existing UTO to watch multiple namespaces achieves the same goal without introducing a new resource type, by building on the conflict resolution and lifecycle mechanisms the UTO already has.
+
+### Using `strimzi.io/cluster` label alone for conflict detection
+
+The `strimzi.io/cluster` label identifies which Kafka cluster a `KafkaTopic` belongs to, but since `Kafka` CRs are namespaced, two Kafka clusters in different namespaces can have the same name. A `KafkaTopic` with `strimzi.io/cluster: my-cluster` in namespace `team-a` could ambiguously refer to either `kafka/my-cluster` or `staging/my-cluster`. The `strimzi.io/managed-by` annotation with the format `<namespace>/<name>` eliminates this ambiguity.
--- a/133-multi-namespace-topic-operator.md
+++ b/133-multi-namespace-topic-operator.md
@@ -50,6 +50,8 @@ The Topic Operator's own namespace (where the `Kafka` resource lives) is **alway
 
 For the standalone Topic Operator deployment, the existing `STRIMZI_NAMESPACE` environment variable is extended to accept a comma-separated list of namespaces or `*`, consistent with how the Cluster Operator already supports multi-namespace watching.
 
+**Deprecation of `watchedNamespace` (singular):** The existing `spec.entityOperator.topicOperator.watchedNamespace` field is deprecated in favor of `watchedNamespaces` (plural). When `watchedNamespaces` is set, `watchedNamespace` is ignored and a deprecation warning is logged. Users should migrate by converting `watchedNamespace: team-a` to `watchedNamespaces: [team-a]`.
+
 ### RBAC
 
 Watching additional namespaces requires the Topic Operator to have Kubernetes RBAC permissions in those namespaces. The Cluster Operator is responsible for creating and managing these RBAC resources.
@@ -58,11 +60,13 @@ Watching additional namespaces requires the Topic Operator to have Kubernetes RB
 
 The Cluster Operator creates a `Role` and `RoleBinding` in each watched namespace, granting the Entity Operator's `ServiceAccount` permissions to watch, list, get, and update `KafkaTopic` resources and their status.
 
+The Role and RoleBinding names include the Kafka cluster's namespace and name to ensure uniqueness: `strimzi-topic-operator-<kafka-namespace>-<kafka-name>`. This prevents two Kafka clusters watching the same tenant namespace from overwriting each other's RBAC resources.
+
 ```yaml
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
-  name: strimzi-topic-operator
+  name: strimzi-topic-operator-kafka-my-cluster
   namespace: team-a
   labels:
     strimzi.io/cluster: my-cluster
@@ -74,7 +78,7 @@ rules:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
-  name: strimzi-topic-operator
+  name: strimzi-topic-operator-kafka-my-cluster
   namespace: team-a
   labels:
     strimzi.io/cluster: my-cluster
@@ -84,7 +88,7 @@ subjects:
     namespace: kafka
 roleRef:
   kind: Role
-  name: strimzi-topic-operator
+  name: strimzi-topic-operator-kafka-my-cluster
   apiGroup: rbac.authorization.k8s.io
 ```
 
@@ -162,33 +166,29 @@ When a **non-owning** `KafkaTopic` (one with `ResourceConflict` status) is delet
 
 The existing `strimzi.io/managed: false` annotation continues to work as before — it allows a `KafkaTopic` to be deleted without deleting the Kafka topic, enabling ownership transfers.
 
-### Preventing overlapping Topic Operator instances
+### Identifying ownership via `strimzi.io/cluster`
 
-When multiple Kafka clusters exist in the same Kubernetes cluster, each has its own Topic Operator instance. If two TO instances are configured to watch overlapping namespaces, conflicts can arise.
+The existing `strimzi.io/cluster` label already identifies which Kafka cluster a `KafkaTopic` belongs to. This proposal extends it to eliminate ambiguity when multiple Kafka clusters share the same name across different namespaces.
 
-To prevent this, the TO sets an annotation on each `KafkaTopic` it reconciles:
-
-```yaml
-metadata:
-  annotations:
-    strimzi.io/managed-by: kafka/my-cluster
-```
-
-The annotation value follows the format `<namespace>/<kafka-cr-name>`, using `/` as a separator (not `-`, which would be ambiguous).
-
-When a TO encounters a `KafkaTopic` with a `strimzi.io/managed-by` annotation that does not match its own identity, it skips the resource and updates its status:
+**Single-namespace mode (unchanged):** Users continue to label topics with the Kafka cluster name:
 
 ```yaml
-status:
-  conditions:
-  - type: Ready
-    status: "False"
-    reason: ManagedByOtherInstance
-    message: "This KafkaTopic is managed by kafka/other-cluster"
-    lastTransitionTime: "2024-11-24T17:00:00Z"
+labels:
+  strimzi.io/cluster: my-cluster
 ```
 
-This is a best-effort mechanism — there is a brief race window during initial reconciliation where two TOs could both attempt to claim a resource. The `strimzi.io/managed-by` annotation should be set atomically during the first reconciliation to minimize this window.
+**Multi-namespace mode (new):** Users label topics with the fully-qualified `<namespace>/<name>` form:
+
+```yaml
+labels:
+  strimzi.io/cluster: kafka/my-cluster
+```
+
+The TO filters its watch using this label. In multi-namespace mode, the TO only reconciles resources whose `strimzi.io/cluster` label matches its own `<namespace>/<name>` identity (e.g., `kafka/my-cluster`). It also accepts the unqualified form `my-cluster` for backward compatibility when there is no ambiguity — but the qualified form `kafka/my-cluster` is required when multiple Kafka clusters share the same name across namespaces.
+
+Because each TO instance selects only topics carrying its own qualified identity, two TOs watching overlapping namespaces do not conflict: `kafka/my-cluster` and `staging/my-cluster` are distinct label values and each TO ignores resources labeled for the other.
+
+If a `KafkaTopic` is in scope for a TO but lacks the expected `strimzi.io/cluster` label entirely (e.g., due to a misconfiguration), the TO skips it silently and logs a WARN. No status writes are performed on resources the TO does not own, which prevents reconciliation fights between competing instances.
 
 ### Namespace lifecycle considerations
 
@@ -196,7 +196,13 @@ This is a best-effort mechanism — there is a brief race window during initial 
 The Cluster Operator creates RBAC resources in the new namespace and restarts/reconfigures the TO. The TO starts watching the new namespace and reconciles any existing `KafkaTopic` resources in it.
 
 **Removing a namespace from `watchedNamespaces`:**
-The Cluster Operator should remove the `strimzi.io/topic-operator` finalizer from all `KafkaTopic` resources in the namespace being removed, then delete the RBAC resources. This prevents `KafkaTopic` resources with finalizers from blocking namespace deletion after the TO stops watching.
+The Cluster Operator must follow this sequence to ensure finalizers can be cleaned up before RBAC is revoked:
+
+1. While RBAC still exists, the CO removes the `strimzi.io/topic-operator` finalizer from all `KafkaTopic` resources in the namespace being removed.
+2. The CO updates the TO configuration to remove the namespace from `watchedNamespaces`.
+3. The CO deletes the RBAC resources (`Role`/`RoleBinding`) for that namespace.
+
+This ordering is critical: deleting RBAC before removing finalizers would leave the TO unable to reach the `KafkaTopic` resources, causing finalizers to block any subsequent namespace deletion.
 
 **Watched namespace is deleted:**
 If the TO is running, it processes the finalizers normally — deleting Kafka topics as appropriate. If the TO is not running (or the namespace was already removed from the watch list), `KafkaTopic` resources with finalizers will block namespace deletion. In this case, manual finalizer removal is necessary (e.g., `kubectl patch kafkatopic <name> -n <ns> --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'`). This is an existing limitation of the UTO that applies equally to the single-namespace case, and should be documented as operational guidance.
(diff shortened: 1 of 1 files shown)


scholzj (commented):
Thanks for the proposal. I left some comments about things that would not work or would be blockers for me in terms of backwards compatiblity etc.

But I do not think that this is the right direction. While it might create something that somehow works, I think it would be a hacky solution with a bad design and user experience. I think if we want to seriously go into the multi-namespace world, we might need to start from scratch and create a better match/respect for the Kubernetes namespace model.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,312 @@
...
+```
+
+This is a best-effort mechanism — there is a brief race window during initial reconciliation where two TOs could both attempt to claim a resource. The `strimzi.io/managed-by` annotation should be set atomically during the first reconciliation to minimize this window.
+
+### Namespace lifecycle considerations
+
+**Adding a namespace to `watchedNamespaces`:**
+The Cluster Operator creates RBAC resources in the new namespace and restarts/reconfigures the TO. The TO starts watching the new namespace and reconciles any existing `KafkaTopic` resources in it.
+
+**Removing a namespace from `watchedNamespaces`:**
```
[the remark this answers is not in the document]
scholzj: Sorry, maybe my comment was too confusing.
* Cluster Operator does not touch any topics. It should not remove any finalizers from them. (not 100% sure the CO is typo here or intention)
* Topic Operator would never know that some namespace was removed as it is removed through restart. So it starts with the new namespace configuration and has no idea a namespace was removed.
* And even if it knew about it, it would not have the RBAC rights to do anything with the topics because the RBAC rights are managed independently through CO, and there is no synchronization point.
* Plus we in general do not delete the user-facing custom resources.

I do not think you can really fix it to be honest. All what can be done here is that this is user's responsibility to either delete the topics first or manually later.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,312 @@
...
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: Kafka
+metadata:
+  name: my-cluster
+  namespace: kafka
+spec:
+  entityOperator:
+    topicOperator:
+      watchedNamespaces:
```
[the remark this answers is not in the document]
scholzj: Deprecation is fine. But we are not abandoning fields from one day to another. We do things in backwards compatible way.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,312 @@
...
+1. The finalizer logic deletes the Kafka topic from the Kafka cluster.
+2. Any other `KafkaTopic` CRs in other namespaces that referenced the same Kafka topic will have their status updated to reflect that the topic no longer exists. They can then be cleaned up by their respective namespace owners, or if left in place, they will attempt to recreate the topic on their next reconciliation (becoming the new owner since no conflict exists).
+
+When a **non-owning** `KafkaTopic` (one with `ResourceConflict` status) is deleted:
+1. The finalizer is removed without deleting the Kafka topic (same as existing UTO behavior for conflicting resources).
+2. No impact on the owning `KafkaTopic` or the Kafka topic itself.
+
+The existing `strimzi.io/managed: false` annotation continues to work as before — it allows a `KafkaTopic` to be deleted without deleting the Kafka topic, enabling ownership transfers.
+
+### Preventing overlapping Topic Operator instances
```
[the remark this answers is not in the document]
KyriosGN0: >  How does the operator know it is in single namespace or multi namespace mode? 

according to the `watchedNamespaces` field, if its the plural its multi namespace, if its the singular, its single namespace 

> How does it know if there are more clusters (more TOs watching the same namespace) or not?

if its multi namespace it uses the strimzi.io/cluster = <namespace_of_to>/<kafka_name> label, so if you have 2 TO watching the same namespace. so if we create a topic in namespace a, its owned by TO of the namespace kafka, (due to multi namespace mode), then another TO tries to own it, it can't cuz its already managed by another TO (due to the label)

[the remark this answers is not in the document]
KyriosGN0: OK, i had a misunderstanding on my end about the `strimzi.io/cluster` label (i mistakenly thought its controlled by the operator) 
since its controlled by the user, i updated the proposal to deprecate the short name (just cluster) and when switching from the singular to plural users should also migrate from the short to a fully name (namespace/kafka_name) 
the TO will accept both of them, but it log a warning (or we can more forceful and also skip reconciliation when this situation occurs) that there is a possibility of a conflict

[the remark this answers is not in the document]
scholzj: I'm not sure it is so simple. If you just log a warning, how do you know which Topic Operator should manage it? And if you instead throw an error, you break backwards compatibility.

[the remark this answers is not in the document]
KyriosGN0: it should be the user's responsibility to know which topic relates to which Kafka during the migration

KyriosGN0:
Hey @scholzj, is there anything else i need to address? do i need to request other maintainers as reviewers for this proposal? (saw it on other proposals)

scholzj:
> Hey @scholzj, is there anything else i need to address? do i need to request other maintainers as reviewers for this proposal? (saw it on other proposals)

They should have notifications. But I requested their reviews for it explicitely.

ppatierno:
I added @fvaleri as SME on the topic operator.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
+# Multi-namespace watching for the Topic Operator
+
+Issue [#1206](https://github.com/strimzi/strimzi-kafka-operator/issues/1206)
+
+## Current situation
+
+The Unidirectional Topic Operator (UTO, [proposal 051](https://github.com/strimzi/proposals/blob/main/051-unidirectional-topic-operator.md)) watches `KafkaTopic` resources in a single namespace — typically the namespace where the `Kafka` resource is deployed. The UTO reconciles topic state unidirectionally from Kubernetes to Kafka, uses `creationTimestamp`-based conflict resolution when multiple `KafkaTopic` CRs reference the same Kafka topic, and employs finalizer-based deletion.
+
+The UTO is deployed as a container within the Entity Operator pod, which the Cluster Operator manages. The `STRIMZI_NAMESPACE` environment variable controls which namespace the UTO watches.
```
im-konge: I know that you talk about standalone deployment, but maybe it would be worth mentioning both modes here - so that UTO is deployed either as part of Entity Operator or it can be deployed standalone?

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+
+**Migration steps:** Convert `watchedNamespace: team-a` to `watchedNamespaces: [team-a]`. No other changes are required — no label changes, no RBAC changes, and no modifications to existing `KafkaTopic` resources.
+
+**Interaction rules:** When both `watchedNamespace` and `watchedNamespaces` are set, `watchedNamespaces` takes precedence and a warning is logged indicating that the singular field is being ignored. When only the singular field is set, the TO behaves exactly as it does today — watching the Kafka cluster's own namespace plus the single additional namespace specified.
+
+**Label impact:** Migrating from the singular to the plural field does **not** require any changes to `strimzi.io/cluster` labels on existing `KafkaTopic` resources. The TO accepts both the unqualified and qualified label forms regardless of which configuration field is used. However, the unqualified form is deprecated (see below) and should be migrated to the qualified `<namespace>/<name>` form, the TO will also log a warning that there is a possiblty of multiple TO watching over the topic unless its labels is updated.
```
im-konge:
```suggestion
**Label impact:** Migrating from the singular to the plural field does **not** require any changes to `strimzi.io/cluster` labels on existing `KafkaTopic` resources. The TO accepts both the unqualified and qualified label forms regardless of which configuration field is used. However, the unqualified form is deprecated (see below) and should be migrated to the qualified `<namespace>/<name>` form, the TO will also log a warning that there is a possibility of multiple TO watching over the topic unless its labels are updated. 
```

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+
+**Interaction rules:** When both `watchedNamespace` and `watchedNamespaces` are set, `watchedNamespaces` takes precedence and a warning is logged indicating that the singular field is being ignored. When only the singular field is set, the TO behaves exactly as it does today — watching the Kafka cluster's own namespace plus the single additional namespace specified.
+
+**Label impact:** Migrating from the singular to the plural field does **not** require any changes to `strimzi.io/cluster` labels on existing `KafkaTopic` resources. The TO accepts both the unqualified and qualified label forms regardless of which configuration field is used. However, the unqualified form is deprecated (see below) and should be migrated to the qualified `<namespace>/<name>` form, the TO will also log a warning that there is a possiblty of multiple TO watching over the topic unless its labels is updated. 
+
+#### Deprecation of unqualified `strimzi.io/cluster` label
+
+The unqualified `strimzi.io/cluster` label (e.g., `my-cluster`) is deprecated in favor of the qualified `<namespace>/<name>` form (e.g., `kafka/my-cluster`).
```
im-konge: So when UTO will have `watchedNamespaces: ["kafka1", "kafka2"]`, the name of the Kafka cluster will by `my-cluster` and I will create KafkaTopics with label `strimzi.io/cluster: my-cluster` in both Namespaces, will they be reconciled by UTO? Or UTO will just take the KafkaTopics with such label from the Namespace it is running in only?

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+
+**Scope:** This deprecation applies only to `KafkaTopic` resources managed by the Topic Operator. Other resources that use the `strimzi.io/cluster` label (e.g., `KafkaUser` resources managed by the User Operator) are not affected and continue to use the unqualified form.
+
+### RBAC
+
+Watching additional namespaces requires the Topic Operator to have Kubernetes RBAC permissions in those namespaces. The Cluster Operator is responsible for creating and managing these RBAC resources.
+
+**For explicit namespace list:**
+
+The Cluster Operator creates a `Role` and `RoleBinding` in each watched namespace, granting the Entity Operator's `ServiceAccount` permissions to watch, list, get, and update `KafkaTopic` resources and their status.
```
im-konge: This is correct for the mode when it is running in EntityOperator, but in standalone it should be created by user. So I guess we should change the files we are providing for standalone UTO, right?

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+The Role and RoleBinding names include the Kafka cluster's namespace and name to ensure uniqueness: `strimzi-topic-operator-<kafka-namespace>-<kafka-name>`. This prevents two Kafka clusters watching the same tenant namespace from overwriting each other's RBAC resources.
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: strimzi-topic-operator-kafka-my-cluster
+  namespace: team-a
+  labels:
+    strimzi.io/cluster: my-cluster
```
im-konge, line 127: Maybe it's a stupid question, but should the label here follow the `namespace/kafka-cluster-name` pattern?

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+```
+
+## Affected/not affected projects
+
+Affected:
+- **Topic Operator**: Extended to watch multiple namespaces, cross-namespace conflict resolution
+- **Cluster Operator**: Creates RBAC resources for watched namespaces, passes `watchedNamespaces` configuration to the Entity Operator
+
+Not affected:
+- User Operator, Kafka Connect operator, Kafka MirrorMaker operator
```
im-konge: Maybe instead of this list you can just say that `strimzi.io/cluster` remains the same for any other resource?

im-konge (commented):
Thanks for the proposal, I left few comments.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+
+## Current situation
+
+The Unidirectional Topic Operator (UTO, [proposal 051](https://github.com/strimzi/proposals/blob/main/051-unidirectional-topic-operator.md)) watches `KafkaTopic` resources in a single namespace — typically the namespace where the `Kafka` resource is deployed. The UTO reconciles topic state unidirectionally from Kubernetes to Kafka, uses `creationTimestamp`-based conflict resolution when multiple `KafkaTopic` CRs reference the same Kafka topic, and employs finalizer-based deletion.
+
+The UTO is deployed as a container within the Entity Operator pod, which the Cluster Operator manages. The `STRIMZI_NAMESPACE` environment variable controls which namespace the UTO watches.
+
+## Motivation
+
+In Kubernetes clusters, it is common practice to isolate applications in separate namespaces. When a Kafka cluster is shared among multiple applications or teams, each application typically lives in its own namespace. Currently, all `KafkaTopic` resources must be created in the Kafka cluster's namespace, which creates several problems:
```
fvaleri: This is not correct. The watched namespace configuration allows the TO to watch a different namespace from where the Kafka cluster is deployed. The Kafka cluster's namespace is simply the default value.

I would suggest to rephrase to something like this: 
> Currently, the Topic Operator watches a single namespace, either the Kafka cluster's own namespace (default) or one alternative namespace configured via `watchedNamespace` field or `STRIMZI_NAMESPACE` environment variable. This is inconvenient for the following reasons:

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+
+- **Access control**: Teams that need to create topics must have RBAC permissions in the Kafka namespace, which may also grant access to other infrastructure resources they should not touch.
+- **Resource management**: Kubernetes namespace-level quotas, and RBAC cannot be used to separate topic ownership per team.
+- **GitOps workflows**: Application teams want to declare their Kafka topics alongside their application manifests in their own namespace, not in a shared infrastructure namespace.
+- **Namespace hygiene**: Hundreds of `KafkaTopic` resources from different applications co-mingle in a single namespace, making management and auditing harder.
+
+Allowing the Topic Operator to watch `KafkaTopic` resources across multiple namespaces solves these problems while maintaining the existing UTO semantics.
```
fvaleri: There is no need to refer to the Topic Operator as UTO, as this is the only supported implementation at the moment.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+        - team-a
+        - team-b
+        - team-c
+```
+
+The `watchedNamespaces` field accepts:
+- A list of specific namespace names (e.g., `["team-a", "team-b"]`)
+- A wildcard `["*"]` to watch all namespaces
+
+The Topic Operator's own namespace (where the `Kafka` resource lives) is **always** watched implicitly. It does not need to be listed in `watchedNamespaces`. When `watchedNamespaces` is not set or is empty, the behavior is unchanged — the TO watches only its own namespace.
```
fvaleri: Why do we always add TO's namespace implicitly? In order to be backwards compatible we only need to default to that, but if the user sets a custom list, then we only watch namespaces in that list.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+
+#### Deprecation of `watchedNamespace` (singular)
+
+The existing `spec.entityOperator.topicOperator.watchedNamespace` field is deprecated in favor of `watchedNamespaces` (plural).
+
+**Deprecation timeline:** `watchedNamespace` continues to work in its current form but logs a deprecation warning on every Cluster Operator reconciliation. It will be removed in a future major version.
+
+**Migration steps:** Convert `watchedNamespace: team-a` to `watchedNamespaces: [team-a]`. No other changes are required — no label changes, no RBAC changes, and no modifications to existing `KafkaTopic` resources.
+
+**Interaction rules:** When both `watchedNamespace` and `watchedNamespaces` are set, `watchedNamespaces` takes precedence and a warning is logged indicating that the singular field is being ignored. When only the singular field is set, the TO behaves exactly as it does today — watching the Kafka cluster's own namespace plus the single additional namespace specified.
```
fvaleri: I would suggest to use CEL validation to reject the resource at admission time when both fields are set, making the conflict a hard error rather than a runtime precedence game.

> When only the singular field is set, the TO behaves exactly as it does today — watching the Kafka cluster's own namespace plus the single additional namespace specified.

Not correct, as explained in another comment.

133-multi-namespace-topic-operator.md
fvaleri: Yes, we should also update the doc with the various deployment modes, similar to what we have for the Cluster Operator.

133-multi-namespace-topic-operator.md
KyriosGN0: yes, they will both be reconciled (and if they have different configuration they will lead a cycle of conflicts), its the user's responsibility to make sure this doesn't happen, This is why we log a warning about using the unqualified form

133-multi-namespace-topic-operator.md
KyriosGN0: Yes, it sounds like the best way to avoid confusion (but we would need to announce their removal, and a docs upgrade)

133-multi-namespace-topic-operator.md
KyriosGN0: you are correct this is my miss

133-multi-namespace-topic-operator.md
KyriosGN0: fixed

133-multi-namespace-topic-operator.md
KyriosGN0: @im-konge updated the proposal accordingly

133-multi-namespace-topic-operator.md
im-konge: Yeah, I guess it should be mentioned in the proposal as here you are talking just about the EntityOperator configuration.

133-multi-namespace-topic-operator.md
im-konge: Yeah, cool. I didn't mean same KafkaTopics (like same topic names), but different topic names in different Namespaces without label having the `<kafka-namespace>/<cluster-name>`. But I think it's fine if they will be reconciled. Thanks

Maybe it would be worth mentioning? I don't remember if it was mentioned somewhere (possibly later in the proposal), but I think it would be worth to mention it here as well (something similar was in the previous section about `watchNamespaces`).

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: strimzi-topic-operator-kafka-my-cluster
+  namespace: team-a
+  labels:
+    strimzi.io/cluster: team-a/my-cluster
+subjects:
+  - kind: ServiceAccount
+    name: my-cluster-entity-operator
```
fvaleri, line 152: In packaging we have `strimzi-topic-operator` as SA name for standalone mode.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+1. The `KafkaTopic` with the **oldest** `metadata.creationTimestamp` is the owner, regardless of which namespace it is in.
+2. All other `KafkaTopic` CRs referencing the same Kafka topic get:
+
+```yaml
+status:
+  conditions:
+  - type: Ready
+    status: "False"
+    reason: ResourceConflict
+    message: "Managed by team-a/my-topic"
```
fvaleri: Isn't this leaking information about team-a?

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+
+### Namespace lifecycle considerations
+
+**Adding a namespace to `watchedNamespaces`:**
+The Cluster Operator creates RBAC resources in the new namespace and restarts/reconfigures the TO. The TO starts watching the new namespace and reconciles any existing `KafkaTopic` resources in it.
+
+**Removing a namespace from `watchedNamespaces`:**
+The Cluster Operator must follow this sequence to ensure finalizers can be cleaned up before RBAC is revoked:
+
+1. While RBAC still exists, the CO removes the `strimzi.io/topic-operator` finalizer from all `KafkaTopic` resources in the namespace being removed.
```
fvaleri: KafkaTopic finalizers are owned by the Topic Operator, not Cluster Operator.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+2. The CO updates the TO configuration to remove the namespace from `watchedNamespaces`.
+3. The CO deletes the RBAC resources (`Role`/`RoleBinding`) for that namespace.
+
+This ordering is critical: deleting RBAC before removing finalizers would leave the TO unable to reach the `KafkaTopic` resources, causing finalizers to block any subsequent namespace deletion.
+
+**Watched namespace is deleted:**
+If the TO is running, it processes the finalizers normally — deleting Kafka topics as appropriate. If the TO is not running (or the namespace was already removed from the watch list), `KafkaTopic` resources with finalizers will block namespace deletion. In this case, manual finalizer removal is necessary (e.g., `kubectl patch kafkatopic <name> -n <ns> --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'`). This is an existing limitation of the UTO that applies equally to the single-namespace case, and should be documented as operational guidance.
```
fvaleri: This is not a limitation, but it is working as designed. You should also mention than the user has the option to disable KafkaTopic finalizers.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+
+**TO restart:**
+On startup, the TO scans all watched namespaces to rebuild its in-memory conflict resolution map before beginning reconciliation. This is the same process used today for a single namespace, extended to multiple namespaces. Startup time increases linearly with the number of watched namespaces and `KafkaTopic` resources.
```
fvaleri: Let's say we have a single Topic Operator instance reconciling multiple or all namespaces. One of them could generate a lot of KT events starving the others. How do you plan to cope with that?

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+- **Topic Operator**: Extended to watch multiple namespaces, cross-namespace conflict resolution
+- **Cluster Operator**: Creates RBAC resources for watched namespaces, passes `watchedNamespaces` configuration to the Entity Operator
+
+Not affected:
+- `strimzi.io/cluster` remains the same for all other resources
+- KafkaTopic CRD schema (no changes to spec fields)
+
+## Compatibility
+
+- The `watchedNamespaces` field is optional with an empty default, preserving existing single-namespace behavior.
```
fvaleri: The single-namespace behavior is to default to the Kafka cluster's namespace when deployed by the CO, or the TO's namespace when standalone.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+- Namespace-level policies to restrict which topic configurations can be set from a given namespace (e.g., allowing infra teams to retain control over certain topic-level configs).
+
+## Rejected alternatives
+
+### New `KafkaNamespaceTopic` CRD
+
+[Issue #1206](https://github.com/strimzi/strimzi-kafka-operator/issues/1206) proposed a `KafkaNamespaceTopic` CRD that would encode the Kubernetes namespace in the Kafka topic name (e.g., `team-a.orders`). This approach was rejected because:
+
+1. **CRD maintenance cost**: Each custom resource is expensive to maintain. A new CRD requires its own controller logic, status management, validation webhooks, documentation, and migration tooling.
+2. **Forces naming convention**: Encoding the namespace in the topic name forces a specific naming convention (`<namespace>.<topic>`) that may not match existing topic naming patterns or be desirable for all users.
```
fvaleri: That would be just a default, but the user should be able to customize the prefix.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+- `watchedNamespace` (singular) is deprecated. When `watchedNamespaces` (plural) is set, the singular field is ignored with a deprecation warning logged. The migration path is: `watchedNamespace: team-a` → `watchedNamespaces: [team-a]` — no label, RBAC, or KafkaTopic changes needed.
+- Existing `KafkaTopic` resources continue to work without modification.
+
+## Future work
+
+- Multi-namespace support for the User Operator (`KafkaUser` resources) — this will require a separate proposal addressing the security implications of cross-namespace ACL management (see [PR #137 discussion](https://github.com/strimzi/proposals/pull/137)).
+- Namespace-level policies to restrict which topic configurations can be set from a given namespace (e.g., allowing infra teams to retain control over certain topic-level configs).
```
fvaleri: If we claim multi-namespace support, users would not expect to be limited by topic names from other teams. In my mind, this is a prerequisite for the current proposal.

I imagine something like this:

```sh
kind: KafkaTopicPolicy
metadata:
  name: my-cluster-policy
  namespace: team-a
  labels:
    strimzi.io/cluster: kafka/my-cluster
spec:
  topicNamePrefix: "team-a."
  allowedConfigs:
    - retention.ms
    - cleanup.policy
  maxPartitions: 24
  maxReplicas: 3
  defaults:
    retention.ms: "604800000"
```

Platform teams manage policies per namespace, application teams manage their own topics. When there is no policy, the Topic Operator would apply some sensible default (e.g. namespace prefix). Of course the are many more details to discuss, so I agree that this should be a dedicated proposal.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
...
+
+**Migration steps:** Convert `watchedNamespace: team-a` to `watchedNamespaces: [team-a]`. No other changes are required — no label changes, no RBAC changes, and no modifications to existing `KafkaTopic` resources.
+
+**Interaction rules:** When both `watchedNamespace` and `watchedNamespaces` are set, `watchedNamespaces` takes precedence and a warning is logged indicating that the singular field is being ignored. When only the singular field is set, the TO behaves exactly as it does today — watching the Kafka cluster's own namespace plus the single additional namespace specified.
+
+**Label impact:** Migrating from the singular to the plural field does **not** require any changes to `strimzi.io/cluster` labels on existing `KafkaTopic` resources. The TO accepts both the unqualified and qualified label forms regardless of which configuration field is used. However, the unqualified form is deprecated (see below) and should be migrated to the qualified `<namespace>/<name>` form, the TO will also log a warning that there is a possibility of multiple TO watching over the topic unless its labels are updated.
```
fvaleri: It would be good to describe what happens with overlapping namespaces. The edge cases are: two or more TOs pointing to the same cluster or different clusters with the same name.

---

Another approach to solve this problem would be to leverage `STRIMZI_RESOURCE_LABELS`. When `watchedNamespaces` is used, the CO could add a label automatically:

```yaml
  labels:                                                                                                                                                                                                                                                                                  
    strimzi.io/cluster: my-cluster
    strimzi.io/cluster-namespace: kafka # new
```

There are multiple benefits:

- The `strimzi.io/cluster` label stays unqualified. No deprecation, no migration of existing label values.
- Two TOs pointing to `my-cluster` in namespaces `kafka` and `staging` watching the same tenant namespace would have distinct selectors and never clash.
- Single-namespace deployments that don't set `watchedNamespaces` could keep working with just `strimzi.io/cluster=my-cluster` (no CR changes).
- Logs would not be flooded by unqualified label warnings.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,335 @@
+# Multi-namespace watching for the Topic Operator
+
+Issue [#1206](https://github.com/strimzi/strimzi-kafka-operator/issues/1206)
```
fvaleri, line 3: UTO is the only supported implementation, so we can just refer to it as Topic Operator or TO.

fvaleri (commented):
Hello @KyriosGN0, thanks for creating this proposal.

Kafka support for multi-tenancy is not great, and this is probably the main reason why we don't have multi-namespace support. The elephant in the room is the topic name mapping that you are not addressing here.

133-multi-namespace-topic-operator.md
KyriosGN0: so in your mind, a proposal about a KafkaTopicPolicy (or a proposal about central management of topic config) is a blocker for this proposal ?

KyriosGN0:
Hey, @fvaleri @im-konge @scholzj, i have changed the solution in the proposal following https://github.com/strimzi/proposals/pull/204#discussion_r3035633529 (as i like it more). 
Thanks in Advance and sorry for the ping

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,519 @@
...
+
+**Event starvation:** A single Topic Operator instance reconciling events across many namespaces may be starved if one namespace generates a very high volume of events, delaying reconciliation for other namespaces. In high-traffic environments, operators should consider running separate Topic Operator instances per team or per namespace group, rather than a single shared instance, to bound the blast radius of event load. Future work may introduce per-namespace reconciliation queues to mitigate this.
+
+### Deletion semantics
+
+The Topic Operator uses Kubernetes finalizers to ensure that deleting a `KafkaTopic` also deletes the corresponding Kafka topic. The `strimzi.io/topic-operator` finalizer is added and managed exclusively by the Topic Operator — not the Cluster Operator. Users who do not want finalizer-based topic deletion can opt out by setting `spec.topicDeletion: false` on the `KafkaTopic` resource, in which case deleting the CR leaves the Kafka topic intact. This behavior is unchanged in the multi-namespace case.
```
fvaleri: I think your coding agent hallucinated here. There is no `topicDeletion` field.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,519 @@
...
+
+When the **owning** `KafkaTopic` (the oldest one) is deleted:
+1. The finalizer logic deletes the Kafka topic from the Kafka cluster.
+2. Any other `KafkaTopic` CRs in other namespaces that referenced the same Kafka topic will have their status updated to reflect that the topic no longer exists. They can then be cleaned up by their respective namespace owners, or if left in place, they will attempt to recreate the topic on their next reconciliation (becoming the new owner since no conflict exists).
```
fvaleri: > the same Kafka topic will have their status updated to reflect that the topic no longer exists.

This is also made up. The conflict status remains unchanged until the next periodic reconciliation.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,519 @@
...
+
+**Deprecation timeline:** `watchedNamespace` continues to work in its current form but logs a deprecation warning on every Cluster Operator reconciliation. It will be removed in a future major version.
+
+**Migration steps:** Convert `watchedNamespace: team-a` to `watchedNamespaces: [team-a]`. Because switching to `watchedNamespaces` enables multi-namespace mode, the `STRIMZI_RESOURCE_LABELS` selector will now require both `strimzi.io/cluster` and `strimzi.io/cluster-namespace` labels. Before making this change, add the `strimzi.io/cluster-namespace` label to all existing `KafkaTopic` resources (see the migration example below). No RBAC changes are required — the Cluster Operator handles RBAC automatically.
+
+**Interaction rules:** Setting both `watchedNamespace` and `watchedNamespaces` simultaneously is rejected at admission time via a CEL validation rule on the `Kafka` CRD. Users must migrate to `watchedNamespaces` before adding additional namespaces. When only the singular field is set, the TO behaves exactly as it does today — watching the Kafka cluster's own namespace plus the single additional namespace specified.
```
fvaleri: > the TO behaves exactly as it does today — watching the Kafka cluster's own namespace plus the single additional namespace specified.

Again. This is not how it works today.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,519 @@
...
+**Kafka-managed deployment:** When the Cluster Operator deploys the Topic Operator with `watchedNamespaces` configured, it automatically sets `STRIMZI_RESOURCE_LABELS` to `strimzi.io/cluster=<cluster-name>,strimzi.io/cluster-namespace=<kafka-namespace>`. The TO then requires both labels to match before reconciling a `KafkaTopic`. This ensures that two TOs watching the same namespace for different clusters with the same name (e.g., `my-cluster` in namespace `kafka` and `my-cluster` in namespace `staging`) never clash — their `strimzi.io/cluster-namespace` values differ.
+
+**Standalone deployment:** The operator administrator sets both labels in `STRIMZI_RESOURCE_LABELS`:
+
+```yaml
+- name: STRIMZI_RESOURCE_LABELS
+  value: "strimzi.io/cluster=my-cluster,strimzi.io/cluster-namespace=kafka"
+```
+
+**Behaviour without the label:** In multi-namespace mode, a `KafkaTopic` that has `strimzi.io/cluster` but lacks `strimzi.io/cluster-namespace` will **not** be reconciled by the TO — it will not match the label selector. The TO logs a WARN for such resources to help users identify missing labels. This prevents accidental misrouting when multiple TOs watch overlapping namespaces.
```
fvaleri, line 102: What about my suggestion of adding the new label automatically? Did you reject that? Why? That would avoid the need for logging (which anyway should be in batches, not for every single KT).

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,519 @@
...
+kind: ClusterRole
+metadata:
+  name: strimzi-topic-operator-kafka-my-cluster
+  labels:
+    strimzi.io/cluster: my-cluster
+    strimzi.io/cluster-namespace: kafka
+rules:
+  - apiGroups: ["kafka.strimzi.io"]
+    resources: ["kafkatopics", "kafkatopics/status"]
+    verbs: ["get", "list", "watch", "create", "patch", "update", "delete"]
```
fvaleri, line 175: This role definition is not correct and there is some resource missing. We follow the principle of least privilege.

133-multi-namespace-topic-operator.md
fvaleri: What about this? It is a security concern.

133-multi-namespace-topic-operator.md
fvaleri: What about this?

133-multi-namespace-topic-operator.md
fvaleri: What about this?

133-multi-namespace-topic-operator.md
fvaleri: What about this?

133-multi-namespace-topic-operator.md
fvaleri: What about this?

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,519 @@
...
+- `watchedNamespace` (singular) is deprecated. Setting both `watchedNamespace` and `watchedNamespaces` simultaneously is rejected at admission time via CEL validation on the `Kafka` CRD. The migration path is: `watchedNamespace: team-a` → `watchedNamespaces: [team-a]` — no label, RBAC, or KafkaTopic changes needed.
+- Existing `KafkaTopic` resources continue to work without modification in single-namespace mode. When enabling multi-namespace mode (`watchedNamespaces`), existing topics must be labeled with `strimzi.io/cluster-namespace` before the switch (see the migration example in the proposal).
+
+## Future work
+
+- Multi-namespace support for the User Operator (`KafkaUser` resources) — this will require a separate proposal addressing the security implications of cross-namespace ACL management (see [PR #137 discussion](https://github.com/strimzi/proposals/pull/137)).
+- **Per-namespace reconciliation queues:** To address the event starvation risk described above, future work may introduce separate reconciliation queues per namespace so that a high-traffic namespace cannot starve reconciliation for others in the same Topic Operator instance.
```
fvaleri: This is the noisy neighbor problem that any multi-tenant solution should address.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,519 @@
...
+
+1. While RBAC still exists, the CO removes the `strimzi.io/topic-operator` finalizer from all `KafkaTopic` resources in the namespace being removed.
+2. The CO updates the TO configuration to remove the namespace from `watchedNamespaces`.
+3. The CO deletes the RBAC resources (`Role`/`RoleBinding`) for that namespace.
+
+This ordering is critical: deleting RBAC before removing finalizers would leave the TO unable to reach the `KafkaTopic` resources, causing finalizers to block any subsequent namespace deletion.
+
+**Watched namespace is deleted:**
+If the TO is running, it processes the finalizers normally — deleting Kafka topics as appropriate. If the TO is not running (or the namespace was already removed from the watch list), `KafkaTopic` resources with finalizers will block namespace deletion. In this case, manual finalizer removal is necessary (e.g., `kubectl patch kafkatopic <name> -n <ns> --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'`). This is an existing limitation of the Topic Operator that applies equally to the single-namespace case, and should be documented as operational guidance.
```
fvaleri: This is working as expected and well documented.

133-multi-namespace-topic-operator.md
fvaleri: > o in your mind, a proposal about a KafkaTopicPolicy (or a proposal about central management of topic config) is a blocker for this proposal ?

This proposal would enable multi-tenancy on a shared Kafka cluster, so topic namespacing is critical IMO. Multi-tenancy is a much broader topic than multi-namespace support, but this is a fundamental piece.

133-multi-namespace-topic-operator.md
KyriosGN0: Will change this to a more "ambiguous" message "This Topic is already managed by another topic operator)

133-multi-namespace-topic-operator.md
KyriosGN0: Will fix

133-multi-namespace-topic-operator.md
KyriosGN0: Will correct it

133-multi-namespace-topic-operator.md
KyriosGN0: In my mind the watchedNamespaces list is not an override to the default behaviour but an extension of it, it doesn't make sense to me to remove the kafka cluster namespace when we omit it from the list

133-multi-namespace-topic-operator.md
KyriosGN0: Will rephrase

133-multi-namespace-topic-operator.md
KyriosGN0: This is an example for how the CO does creates the role binding, not the standalone manifests

133-multi-namespace-topic-operator.md
KyriosGN0: @fvaleri you mean when the topic is managed via the `strimzi.io/cluster=my-cluster` but still doesn't have the `strimzi.io/cluster-namespace` label?

133-multi-namespace-topic-operator.md
KyriosGN0: Added the missing resources, but i don't understand why its not correct, (it doesn't include the user and secrets by design, since its only for topic CR)

133-multi-namespace-topic-operator.md
KyriosGN0: removed

133-multi-namespace-topic-operator.md
KyriosGN0: added the Per-namespace reconciliation queues as the solution to the event starvation problem

133-multi-namespace-topic-operator.md
KyriosGN0: fixed

133-multi-namespace-topic-operator.md
KyriosGN0: fixed

133-multi-namespace-topic-operator.md
KyriosGN0: fixed

133-multi-namespace-topic-operator.md
KyriosGN0: ended up removing this section

133-multi-namespace-topic-operator.md
fvaleri: In most standard Kubernetes resources, the user's input completely replaces the default rather than extending it. If you want to do it differently, you should provide a clear motivation.

133-multi-namespace-topic-operator.md
fvaleri: Yes, when `watchedNamespaces` is used, the CO should automatically add `strimzi.io/cluster-namespace=<kafka-namespace>` to the `STRIMZI_RESOURCE_LABELS` environment variable passed to the TO.

Main benefits:
1. The `strimzi.io/cluster` label stays unqualified. No deprecation, no migration of existing label values.
2. Single-namespace deployments that don't set `watchedNamespaces` could keep working with just `strimzi.io/cluster=my-cluster` (no CR changes). Logs would not be flooded by unqualified label warnings.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,519 @@
...
+    reason: ResourceConflict
+    message: "Already managed by another topic operator"
+    lastTransitionTime: "2024-11-24T17:00:00Z"
+```
+
+This is identical to how the Topic Operator handles conflicts within a single namespace today, extended across namespace boundaries.
+
+**Topic name isolation:** Conflict detection is the mechanism for preventing two teams from accidentally managing the same Kafka topic. It is intentionally reactive rather than preventive — teams *can* share a topic across namespaces (one namespace owns it, others receive a `ResourceConflict` status). For use cases that require strict naming isolation between teams, the recommended approach is to run separate Topic Operator instances per team rather than a shared one.
+
+**Event starvation:** A single Topic Operator instance reconciling events across many namespaces may be starved if one namespace generates a very high volume of events, delaying reconciliation for other namespaces, to address this we would also implement er-namespace reconciliation queues
```
fvaleri: This is another critical point for me (in addition to topic namespacing support), so we need more details on how this would work for the Topic Operator. The CO uses a thread-per-namespace approach where each thread has its own watch, and that helps to mitigate the noisy neighbor problem (a namespace producing a huge amount of events starving the others).

fvaleri (changes requested):
@KyriosGN0 hi, sorry for the late reply.

There are a few critical points that I think still need to be addressed before we can move forward with the proposal. I tried to clarify them, but feel free to ask me if anything is not clear.

133-multi-namespace-topic-operator.md
KyriosGN0: > In most standard Kubernetes resources, the user's input completely replaces the default rather than extending it. If you want to do it differently, you should provide a clear motivation.

I Think in terms of User's expectation its really beneficial to always include the Kafka Cluster namespace, if it was me i would be surprised to find out if the kafka cluster namespace is not watched. 

   
re: the 2nd comment (im sorry i just can't seem to comment) Agreed, i will update the proposal 

> This is another critical point for me (in addition to topic namespacing support), so we need more details on how this would work for the Topic Operator. The CO uses a thread-per-namespace approach where each thread has its own watch, and that helps to mitigate the noisy neighbor problem (a namespace producing a huge amount of events starving the others).

I tend to agree with the thread-per-namespace approach, but i also think that topics will be modifed/created/deleted in higher frequency, and with the amount of possible namespace, we possibly could reach some limitation around the number of threads (this is not an area i am familiar with, so i might be completely wrong)

thanks @fvaleri

KyriosGN0: @fvaleri gentle reminder 🙏

fvaleri (changes requested):
Thanks for the updates @KyriosGN0.

I went through the latest revision and there are still a few points from my earlier comments that I think need to be addressed before we can move forward. I would also wait for other maintainer's reviews, as you need at least two positive maintainer votes.

1. **Topic namespacing**. I consider this a prerequisite for the current proposal, not future work. If we claim multi-namespace support, users will expect topic name isolation between teams. The current answer of "run separate TO instances per team" undermines the purpose of shared multi-namespace watching. This does not mean we need to design KafkaTopicPolicy here, but the proposal should at least define a minimal default behavior (e.g. namespace prefix) and acknowledge it as a hard dependency rather than deferring it entirely.

2. **Implicit namespace watching**. The proposal still says the TO's own namespace is "always watched implicitly." In standard Kubernetes resources, user input replaces the default rather than extending it. If we want to deviate from that convention, we need a clear technical justification in the proposal. Defaulting to the TO's namespace when watchedNamespaces is empty is fine for backwards compatibility, but an explicit list should mean exactly that list.

3. **Noisy neighbor mitigation**. This is another critical point that currently has only a one-liner. The CO uses a thread-per-namespace approach where each thread has its own watch, which helps prevent a single namespace from starving the others. The proposal needs to describe how the TO would handle this concretely: architecture, threading/queuing model, and any resource isolation guarantees.

4. **Information leakage in conflict messages**. The proposal is currently self-contradictory. The conflict resolution section uses "Already managed by another topic operator", but the example section still shows "Managed by team-a/orders". The security section also describes leaking the winner's namespace/name as expected behavior. We wuldn't have this problem if point 1 wuole be properly addressed.

5. **RBAC definitions**. A few issues remain: The Role for explicit namespaces groups kafkatopics and kafkatopics/status with all verbs including delete on status, which is overly broad. We follow the principle of least privilege. The Role is missing kafkatopics/finalizers, which is present in the ClusterRole for wildcard mode. These should be consistent. There is a typo in the ClusterRole: [""kafkatopics/status"] (double opening quote).

6. **Standalone ServiceAccount in examples**. The text correctly mentions strimzi-topic-operator for standalone mode, but the RoleBinding example still uses my-cluster-entity-operator. It would help to either show both modes or add a note clarifying this is the Kafka-managed case only.

fvaleri:
@KyriosGN0 it's been some time since you posted an update on this proposal. Are you still willing to contribute this feature? 

@strimzi/maintainers do you agree or disagree with my comments as summarized in https://github.com/strimzi/proposals/pull/204#pullrequestreview-4219309861?

KyriosGN0:
Hey @fvaleri I would like to continue working on this proposal however I would like to get some opinion from you around my comment for point 3 
> I tend to agree with the thread-per-namespace approach, but i also think that topics will be modifed/created/deleted in higher frequency, and with the amount of possible namespace, we possibly could reach some limitation around the number of threads (this is not an area i am familiar with, so i might be completely wrong)

for point 1: understood, i will add namespace prefixing as the current default and that a future policy type CR is needed (should an issue be opened about it?) 

about 2,4,5,6 i will update the proposal accodingly (probably until sunday)
Thanks!

scholzj:
I shared my take on this on the beginning.

But I think one thing to take carefully into account would be the impact of [CVE-2026-55225](https://github.com/strimzi/strimzi-kafka-operator/security/advisories/GHSA-mw9r-p8xp-wx96) which directly relates to this feature.

fvaleri:
@KyriosGN0 happy to hear you're continuing. Take your time, no rush.

Point 3: This is challenging and it may take some time to settle in. I would suggest to start from the `DESIGN.md` document that you find in `src/main/java`. Then, move on with making yourself familiar with the existing code before thinking about any solution. We also have some performance tests in `systemtest` module that would be useful to compare the new architecture with the old one. 

> I tend to agree with the thread-per-namespace approach, but i also think that topics will be modifed/created/deleted in higher frequency, and with the amount of possible namespace, we possibly could reach some limitation around the number of threads (this is not an area i am familiar with, so i might be completely wrong)

Note that it's not 1 thread per KafkaTopic event, but 1 thread per namespace. I imagine one shared informer dispatching to per-namespace queues for the "all cluster namespaces" use case (cluster-scoped RBAC), and one informer and queue per namespace for the "namespace list" use case (namespace-scoped RBAC). This would give you structural isolation similar to how the Cluster Operator works, but with much lighter threads. That said, there will always be a limit on how many namespaces a single operator can handle, and that's fine. In extreme cases, as you say, standalone mode naturally supports horizontal scaling by deploying multiple independent instances.

Point 1: Yes, please open an issue for the full policy CR so we can track it separately. But the proposal must still define namespace prefix as the built-in default behavior for topic name resolution in multi-namespace mode.

CVE-2026-55225: The core attack vector (Secret exfiltration via cross-namespace RoleBindings) does not apply here because the TO only needs RBAC on KafkaTopics resources in watched namespaces, not on Secrets. The proposal should call this out explicitly as a security boundary. That said, the broader principle from the fix still applies: `watchedNamespaces` must be gated by the existing `STRIMZI_ENTITY_OPERATOR_WATCHED_NAMESPACE_ENABLED` environment variable. The proposal should state that when the flag is disabled (default), `watchedNamespaces` is rejected, consistent with how the singular `watchedNamespace` is handled after the fix.

I hope this clears up all the open doubts, otherwise let me know.

scholzj:
> [CVE-2026-55225](https://github.com/advisories/GHSA-mw9r-p8xp-wx96): The core attack vector (Secret exfiltration via cross-namespace RoleBindings) does not apply here because the TO only needs RBAC on KafkaTopics resources in watched namespaces, not on Secrets. The proposal should call this out explicitly as a security boundary. That said, the broader principle from the fix still applies: `watchedNamespaces` must be gated by the existing `STRIMZI_ENTITY_OPERATOR_WATCHED_NAMESPACE_ENABLED` environment variable. The proposal should state that when the flag is disabled (default), `watchedNamespaces` is rejected, consistent with how the singular `watchedNamespace` is handled after the fix.

It absolutely does apply here. The principle of the CVE does apply to any resources, even non-secrets. In many cases, things such as Kafka topic names are considered confidential. One could also interfere with other cluster topics through these RBAC rights. In addition, the general assumption also is that we aim for symmetry and while there might be separate proposals, there is no point in having TO watch all namespaces if we would plan to block it for UO. So this has to play a major part in considering this feature and how it integrates. It does not mean it has to be a blocker. But maybe it for example should not be available through the `Kafka` CR but only as a standalone deployment etc.

fvaleri:
> It absolutely does apply here. 

Yes, I actually meant the example described in the CVE description, but I expressed that poorly. Anyway I did say "the broader principle from the fix still applies".

KyriosGN0:
Hey @fvaleri, i updated the proposal according to our discussion i will open the issue for policy CR tomorrow morning

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,653 @@
...
+
+```
+<namespace>.<metadata.name>
+```
+
+For example, a `KafkaTopic` named `orders` in namespace `team-a` is materialized in Kafka as `team-a.orders`, while an `orders` in `team-b` becomes `team-b.orders`. The two never collide, so each namespace gets an isolated topic namespace without any per-team configuration.
+
+**Single-namespace mode is unchanged:** When `watchedNamespaces` is not set (single-namespace mode), no prefix is applied — `metadata.name` (or `spec.topicName`) resolves exactly as it does today. This preserves all existing topic names and requires no migration for current deployments.
+
+**Opt-out via `spec.topicName`:** Setting `spec.topicName` explicitly overrides the prefix and uses the given name verbatim, with no namespace prefixing. This is the deliberate escape hatch for teams that need to address a topic by an exact name — for example to share a single Kafka topic across namespaces, or to adopt a topic created under a different convention. It is the only path that can produce a cross-namespace name collision, which is handled by conflict resolution below.
```
fvaleri, line 218: You can't "share a single Kafka topic across namespaces" using spec.topicName, as that would produce a conflicts. Instead, topic adoption use case is good.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,653 @@
...
+
+**Removing a namespace from `watchedNamespaces`:**
+The Cluster Operator must follow this sequence to ensure finalizers can be cleaned up before RBAC is revoked:
+
+1. While RBAC still exists, the Topic operator removes the `strimzi.io/topic-operator` finalizer from all `KafkaTopic` resources in the namespace being removed.
+2. The CO updates the TO configuration to remove the namespace from `watchedNamespaces`.
+3. The CO deletes the RBAC resources (`Role`/`RoleBinding`) for that namespace.
+
+**Topic Operator restart:**
+On startup, the Topic Operator scans all watched namespaces to rebuild its in-memory conflict resolution map before beginning reconciliation. This is the same process used today for a single namespace, extended to multiple namespaces. Startup time increases linearly with the number of watched namespaces and `KafkaTopic` resources.
```
fvaleri, line 354: The map is a deduplication optimization for batching, not a conflict resolution mechanism. It ensures the same Kafka topic doesn't get reconciled multiple times within a single batch cycle. Conflict detection itself remains reactive, happening during reconciliation against the broker.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,653 @@
...
+  name: orders
+  namespace: kafka
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  partitions: 12
+  replicas: 3
+```
+
+> **⚠️ Migration is not transparent — existing topic names change unless pinned.** Multi-namespace mode prefixes the resolved Kafka topic name with the namespace for **every** watched namespace, including the Kafka CR's own namespace. A topic `orders` in namespace `kafka` resolves to `kafka.orders` once multi-namespace mode is enabled. Because Kafka topics cannot be renamed, the operator would create a new empty `kafka.orders` topic and stop managing the original `orders` topic (orphaning its data). **Before enabling multi-namespace mode, you must pin every existing topic's current name via `spec.topicName` (Step 1).**
```
fvaleri, line 476: An alternative would be to stop clients, pre-create `kafka.orders` topic in Kafka and migrate all data from the original `orders` topic. Then they can update the KT to start managing it.

133-multi-namespace-topic-operator.md
```diff
@@ -0,0 +1,653 @@
...
+
+#### Exposure model
+
+This proposal makes multi-namespace watching available **both** via the `Kafka` CR (`spec.entityOperator.topicOperator.watchedNamespaces`, Cluster-Operator–managed) **and** via the standalone deployment, in both cases behind the gate above.
+
+There is a competing view (raised in review) that, for symmetry with the User Operator and to keep the Cluster Operator from minting cross-namespace RBAC on a tenant's behalf, multi-namespace watching should be available **only** in standalone mode and not through the `Kafka` CR. This proposal does not adopt that restriction — gating the feature behind a cluster-admin-controlled flag already keeps the privileged decision out of tenant hands, and exposing it through the `Kafka` CR preserves the managed-RBAC convenience that is the main reason to use the Cluster Operator at all. The standalone-only alternative is recorded under [Rejected alternatives](#standalone-only-exposure). Symmetry with a future multi-namespace User Operator is acknowledged as a cross-cutting design concern and is tracked with that work (see [Future work](#future-work)).
```
fvaleri, line 584: We don't need to mention review, just update the proposal with the agreed design.

fvaleri (commented):
@KyriosGN0 thanks for the updates. It looks mostly good.

I see you list multiple possible solutions and then shared-nothing one (option C). That's fine, but I would move the other solutions to the rejected alternatives.

Can you address the comment from Jakub? We would need to enable multi-namespace support only in standalone mode until the user operator is also updated. You can move that content to future work.

PR #210, opened 2026-03-20
Gateway API-based `type: tlsroute` listener

scholzj:
This proposal introduces a new listener type `tlsroute` based on the Gateway API and on the `TLSRoutes` resources. It closely mirrors the similar Ingress and Route listeners, just following the Gateway API.


136-tls-route-listener.md
```diff
@@ -0,0 +1,231 @@
...
+It might, however, be addressed in a future proposal.
+
+## Affected projects
+
+This proposal affects only the Strimzi Cluster Operator.
+
+## Backwards compatibility
+
+This proposal is fully backwards compatible.
+The existing listeners are not affected in any way by the new listener.
```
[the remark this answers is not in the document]
scholzj: I do not plan any migration guide. Migration from Ingress to Gateway API with the same addresses without outage would heavily depend on the user's infrastructure and likely also on the specific Ingress controller and Gateway API implementation used. In general, I would expect users to simply create a new listener, reconfigure all clients, delete the old listener. And that IMHO does not need any speicfic guide. I added this to the out-of-scope section.

134-tls-route-listener.md
```diff
@@ -0,0 +1,231 @@
...
+  rules:
+    - backendRefs:
+        - name: my-cluster-aston-1000
+          port: 9094
+```
+
+After creating the `TLSRoute` resources, Strimzi will wait for the `.status` section to be updated and contain at least one parent reference.
+This indicates that the `TLSRoute` was accepted by the gateway.
+Strimzi will not perform a detailed check of which gateway is among the parent references.
+This will help provide flexibility and simplify support for different implementations.
```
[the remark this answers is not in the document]
scholzj: > is the operator going to update the `Kafka` status if there is any error reported in some `TLSRoute` status? 

I do not think we should do any detailed parsing of any warnings or errors beyond checking the parent references. I think that could be unpredictable how it would work with different implementations. I added it to the proposal.

> What's going to be reported when Gateway doesn't exist?

We will timeout and report an error. I added it to the proposal.

136-tls-route-listener.md
```diff
@@ -0,0 +1,231 @@
...
+      kind: Gateway
+      name: kafka-gateway
+      sectionName: kafka
+  rules:
+    - backendRefs:
+        - name: my-cluster-aston-1000
+          port: 9094
+```
+
+After creating the `TLSRoute` resources, Strimzi will wait for the `.status` section to be updated and contain at least one parent reference.
```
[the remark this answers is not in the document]
scholzj: > what does "at least one" mean here? I can image that a user could have more than one Gateway implementations installed but can they be used for handling one TLSRoute at the same time? AFAICS the `parentRefs` is plural and hosting an array so it sounds to be possible but how is it going to work? Is it something to allow HA?

"At least one" means that Strimzi will be happy as long as there is at least one parent reference.

It is hard to guess when users might want/need to use multiple parents. I imagine situations when you have two Gateways that each expose the Kafka cluster into completely different networks but using the same DNS names. But I guess the exact use-cases remain to be seen 🤷 .

In anycase, Gateway API seems to support it and did not wanted to deviate from it in our API as that might be hard to fix later.

136-tls-route-listener.md
```diff
@@ -0,0 +1,231 @@
...
+
+`TLSRoute` resources support TLS passthrough as well as TLS termination mode.
+TLS passthrough means that the TLS connection is forwarded to Strimzi as-is, including encryption.
+In this case, the TLS connection will use the TLS certificate from the Kafka brokers and can use mTLS authentication as well.
+TLS termination means that the TLS connection will be terminated in the gateway.
+The Kafka client will connect with encryption, but the connection will not be encrypted when it reaches the Apache Kafka broker.
+In this case, the connection will use the Gateway's TLS certificate and mTLS authentication would not be available.
+Support for TLS termination mode is new, and I have not found any Gateway API implementation supporting it yet.
+But I expect it will eventually be supported.
+To support both modes, Strimzi will support `type: tlsroute` listeners both with and without TLS encryption enabled in the Strimzi listener.
```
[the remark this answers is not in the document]
scholzj: I tried to explain it a bit better and added some diagrams.

Essentially, with TLS Passthrough, the TLS encrytion is applied to the whole path from the Kafka client to the Kafka broker. In this case, the TLS server certificate has to be configured in the Kafka broker.

With TLS Termination, the TLS encryption is applied between the Kafka client and the Gateway. The connection between Gateway and the Kafka broker will be a plaintext connection only. The TLS server certificate has to be configured by the user on the Gateway.

The reasons for using the TLS Termination might be performance (i.e. if the Gateway has faster TLS processing then Java), certificate management (You want to configure TLS certs centrally in the Gateway and not in the brokers - a common question we get for Ingress, but where this is not supported), or for example integration through other software such as service mesh that will secure the path within the Kubernetes cluster.

Using TCP route is not exactly the same as TLSRoute with TLS Termination. With TCPRoute, you do not nee to use encrpytion even between the Kafka client and the Gateway (which is required for TLSRoutes). 

----

Please let me know if this helped to explain it.

scholzj:
@ppatierno Thanks for your comments. I either incorporated them or replied to them (or I tried to do both).

136-tls-route-listener.md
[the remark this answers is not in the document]
ppatierno: So taking what you say here:

> From a Strimzi perspective, the connection will not use TLS encryption.

If I got it correctly, when you use `type: tlsroute` ... with `tls: true` it's configured to use TLS passthrough but if you set `tls: false` it's using TLS termination? Did I get it right? This is what I don't get it, how you distinguish in the listener configuration which kind of TLS mode you are using.

134-tls-route-listener.md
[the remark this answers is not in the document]
ppatierno: So if I got it correctly, if the Gateway exists, the `TLSRoute` has a correct parentRef configured to point to it but for any reasons it's not working and it reports some error conditions in the status, it won't be detected by Strimzi and the user will realize only when clients are not working or looking at `TLSRoute`?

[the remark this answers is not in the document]
scholzj: Yes, correct. But with multiple implementations, it is hard to predict what errors could we expect and how to react to them. At the end, that was the main limitation of the `type: ingress` listener that it expected things in the `.status` that many Ingress controllers did not use.

I do not see this as a problem. At the end it is "bring your own Gateway" model. So we have only limited knowledge if the Gateway works properly, if it routes the correct addresses etc. So the possible error surface is pretty huge and this would anyway cover only small part of it even if we took the compatibility risk and decided to handle it.

136-tls-route-listener.md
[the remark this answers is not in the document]
scholzj: The TLS termination or passthrough is a Gateway property (or rather Listener property, which is normally configured in the Gateway): https://gateway-api.sigs.k8s.io/reference/1.5/spec/#listenertlsconfig. So the `tls` flag tells Strimzi whether to configure the Kafka cluster to expect TLs encryption or not. But it does not configure anything in the `TLSRoute` resources for this.

134-tls-route-listener.md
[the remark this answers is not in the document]
ppatierno: Ok, let's leave as it is then but I would expected the Gateway implementation to report any warning/error into the `TLSRoute` status. As Gateway API is an ... API ... it would be great to have some sort of "standardization" about it. I can understand the Gateway implementation is different across different ... implementations but when it comes to API (which are users facing) then I would expect some standard way across different Gateways.

136-tls-route-listener.md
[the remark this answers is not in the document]
ppatierno: Ah cool! Now I got it. Would you mind to add this link to the proposal to make it clearer? (at least it would be useful for me who as first reader didn't get it).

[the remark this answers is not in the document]
scholzj: Ok, updated. 👍

ppatierno (approved):
LGTM.

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+## Gateway API
+
+[Gateway API](https://gateway-api.sigs.k8s.io/) is an API specification only.
+It does not provide any implementation on its own.
+It provides only the Custom Resource Definitions for the API.
+The API is then supported/implemented by other projects.
+A list of implementations can be found on the [Gateway API website](https://gateway-api.sigs.k8s.io/implementations/#gateway-controller-implementation-status).
+
+The Gateway API specification is versioned.
+Its versioning and release cycle is fully independent on the Kubernetes version.
```
PaulRMellor:
```suggestion
Its versioning and release cycle are fully independent of the Kubernetes version.
```

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+## Motivation
+
+[Gateway API](https://gateway-api.sigs.k8s.io/) is an official Kubernetes project that focuses on L4 and L7 routing.
+It represents the next generation of Kubernetes Ingress.
+It supports L7 routing of HTTP and gRPC traffic.
+It also supports L4 routing of TCP traffic.
+Gateway API aims to address both north-south traffic (to/from the Kubernetes cluster) and east-west traffic (within the Kubernetes cluster through a service mesh).
+For this proposal, the focus is north-south traffic routing and providing Apache Kafka access to Kafka clients running outside of the Kubernetes cluster.
+
+Our users have already been able to use Gateway API with Strimzi for some time.
```
PaulRMellor:
```suggestion
Strimzi users have already been able to use Gateway API for some time.
```

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+They can, for example, use the `type: cluster-ip` listener designed for access within the Kubernetes cluster and manually manage the Gateway API resources and advertised hosts and ports.
+More details can be found in the [Accessing Kafka with Gateway API](https://strimzi.io/blog/2024/08/16/accessing-kafka-with-gateway-api/) blog post on the Strimzi blog.
+
+But with Gateway API becoming the _next_ Kubernetes routing standard, with the growing number of compatible Gateway API implementations, and with the deprecation of `type: ingress` due to Ingress NGINX Controller retirement, it makes sense for Strimzi to add direct support for Gateway API.
```
PaulRMellor:
```suggestion
But with Gateway API emerging as the next Kubernetes routing standard, compatible implementations continuing to grow, and `type: ingress` being deprecated following the retirement of the Ingress NGINX Controller, it’s a natural step for Strimzi to add direct Gateway API support.
```

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+The APIs that are relevant to the Kafka protocol (`TLSRoute` and `TCPRoute`) are also finally maturing and moving to the Standard API channel.
+Direct support will simplify the configuration of Gateway API resources, as Strimzi will manage them automatically and users will no longer need to do it manually.
+It will also make it easier to combine Gateway API with other Strimzi features, such as horizontal autoscaling, which would otherwise be complicated (as users would need to create Gateway API resources when new brokers are added).
+
+## Proposal
+
+Strimzi will implement support for Gateway API using its `TLSRoute` resources.
```
PaulRMellor:
```suggestion
Strimzi will implement support for Gateway API using `TLSRoute` resources.
```

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+Strimzi currently supports four types of external listeners:
+* Load balancers (`type: loadbalancer`)
+* Node ports (`type: nodeport`)
+* OpenShift Routes (OpenShift only) (`type: route`)
+* Ingress with TLS passthrough (based on the Ingress NGINX Controller for Kubernetes) (`type: ingress`)
+
+The Ingress-based listener (`type: ingress`) is currently deprecated.
+The main motivation for the deprecation is the [retirement](https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/) of the [Ingress NGINX Controller for Kubernetes](https://github.com/kubernetes/ingress-nginx), on which this listener type was based.
+
+## Gateway API
```
PaulRMellor:
```suggestion
## Kubernetes Gateway API
```

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+
+[Gateway API](https://gateway-api.sigs.k8s.io/) is an API specification only.
```
PaulRMellor:
```suggestion
Kubernetes [Gateway API](https://gateway-api.sigs.k8s.io/) is an official Kubernetes project that defines API specifications for L4 and L7 routing.
```

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+It does not provide any implementation on its own.
```
PaulRMellor:
```suggestion
It does not include an implementation.
```

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+
+The Gateway API specification is versioned.
+Its versioning and release cycle is fully independent on the Kubernetes version.
+Each version has two release channels: Standard and Experimental.
+The Standard channel contains stable APIs.
+The Experimental channel includes everything from the Standard channel plus additional experimental changes and APIs that might still change or be removed in the future.
+
+## Motivation
+
+[Gateway API](https://gateway-api.sigs.k8s.io/) is an official Kubernetes project that focuses on L4 and L7 routing.
```
PaulRMellor:
```suggestion
The [Gateway API](https://gateway-api.sigs.k8s.io/) represents the next generation of Kubernetes Ingress.
```

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+It represents the next generation of Kubernetes Ingress.
```
PaulRMellor:
```suggestion

```

134-tls-route-listener.md
PaulRMellor: I thought it would be better to get a single, clear definition introduced in this section first, rather than saying it's an API spec here and then saying it's an an official Kubernetes project in the next section

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+
+But with Gateway API becoming the _next_ Kubernetes routing standard, with the growing number of compatible Gateway API implementations, and with the deprecation of `type: ingress` due to Ingress NGINX Controller retirement, it makes sense for Strimzi to add direct support for Gateway API.
+The APIs that are relevant to the Kafka protocol (`TLSRoute` and `TCPRoute`) are also finally maturing and moving to the Standard API channel.
+Direct support will simplify the configuration of Gateway API resources, as Strimzi will manage them automatically and users will no longer need to do it manually.
+It will also make it easier to combine Gateway API with other Strimzi features, such as horizontal autoscaling, which would otherwise be complicated (as users would need to create Gateway API resources when new brokers are added).
+
+## Proposal
+
+Strimzi will implement support for Gateway API using its `TLSRoute` resources.
+In the Strimzi API, it will be added as `type: tlsroute`.
```
PaulRMellor:
```suggestion
In the Strimzi API, this listener will be added as `type: tlsroute`.
```

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+      parentRefs:
+        - name: kafka-gateway
+          sectionName: kafka
+      bootstrap:
+        host: kafka.192.168.1.221.sslip.io
+      advertisedHostTemplate: kafka-{nodeId}-broker.192.168.1.221.sslip.io
+      hostTemplate: kafka-{nodeId}-broker.192.168.1.221.sslip.io
+```
+
+`TLSRoute` resources support TLS passthrough as well as TLS termination mode.
```
PaulRMellor:
```suggestion
`TLSRoute` resources support TLS passthrough and TLS termination.
```

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+While the TLS mode applies to the `TLSRoute` resources, it is configured in the [Gateway Listener TLS configuration](https://gateway-api.sigs.k8s.io/reference/1.5/spec/#listenertlsconfig) in the Gateway or in ListenerSet resources.
```
PaulRMellor:
```suggestion
While the TLS mode applies to the `TLSRoute` resources, it is configured in the [Gateway Listener TLS configuration](https://gateway-api.sigs.k8s.io/reference/1.5/spec/#listenertlsconfig) in the `Gateway` or in `ListenerSet` resources.
```

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+TLS passthrough means that the TLS connection is forwarded to Strimzi as-is, including encryption.
+In this case, the TLS connection will use the TLS certificate from the Kafka brokers and can use mTLS authentication as well.
+
+![TLS Passthrough](./images/134-TLS-Passthrough.svg)
+
+TLS termination means that the TLS connection will be terminated in the gateway.
+The Kafka client will connect with encryption, but the connection will be decrypted by the Gateway.
+And from the Gateway it will continue in an unencrypted work to the Apache Kafka brokers.
+In this case, the connection will use the Gateway's TLS certificate and mTLS authentication would not be available.
```
PaulRMellor:
```suggestion
In this case, the connection will use the Gateway's TLS certificate, and mTLS authentication would not be available.
```

136-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+
+![TLS Passthrough](./images/134-TLS-Passthrough.svg)
+
+TLS termination means that the TLS connection will be terminated in the gateway.
+The Kafka client will connect with encryption, but the connection will be decrypted by the Gateway.
+And from the Gateway it will continue in an unencrypted work to the Apache Kafka brokers.
+In this case, the connection will use the Gateway's TLS certificate and mTLS authentication would not be available.
+From a Strimzi perspective, the connection will not use TLS encryption.
+Support for TLS termination mode is new, and I have not found any Gateway API implementation supporting it yet.
+But I expect it will eventually be supported.
```
PaulRMellor, line 116: would we flag it as dev preview/experimental?

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+From a Strimzi perspective, the connection will not use TLS encryption.
+Support for TLS termination mode is new, and I have not found any Gateway API implementation supporting it yet.
+But I expect it will eventually be supported.
+
+![TLS Termination](./images/134-TLS-Termination.svg)
+
+To support both modes, Strimzi will support `type: tlsroute` listeners both with and without TLS encryption enabled in the Strimzi listener.
+The TLS encryption being enabled or disabled will not have any impact on how Strimzi configures the `TLSRoute` resources, as the TLs mode is configured in the Gateway.
```
PaulRMellor:
```suggestion
TLS encryption being enabled or disabled will not have any impact on how Strimzi configures the `TLSRoute` resources, as the TLs mode is configured in the Gateway.
```

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+It will only impact on whether Strimzi configures the Kafka brokers for TLS encryption or not.
```
PaulRMellor:
```suggestion
It affects only whether Strimzi configures Kafka brokers to use TLS encryption.
```

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+  rules:
+    - backendRefs:
+        - name: my-cluster-aston-1000
+          port: 9094
+```
+
+After creating the `TLSRoute` resources, Strimzi will wait for the `.status` section to be updated and contain at least one parent reference.
+This indicates that the `TLSRoute` was accepted by the gateway.
+Strimzi will not perform a detailed check of which gateway is among the parent references.
+It will also not check for any other conditions and try to detect any other warnings, errors, or failures.
```
PaulRMellor:
```suggestion
It will not check for other conditions or try to detect any other warnings, errors, or failures.
```

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+In case the `TLSRoute` does not have any parent references after the configured time limit (the Cluster Operator _reconciliation timeout_), the reconciliation will fail with a corresponding error.
+This will help provide flexibility and simplify support for different implementations, that might behave differently when it comes to exact conditions, warnings, and errors..
```
PaulRMellor:
```suggestion
This will help provide flexibility and simplify support for different implementations that handle conditions, warnings, and errors differently.
```

134-tls-route-listener.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Migration from `type: ingress` listener
+
+The exact process for migrating from the `type: ingress` listener is out of scope for this proposal as it depends on the specific Ingress and Gateway API implementation and on user's infrastructure.
+Strimzi will not provide any documentation documenting such a migration.
+
+However, users can always:
+
+1. Add a new `type: tlsroute` listener.
+2. Reconfigure their Kafka clients to use the the new listener.
```
PaulRMellor:
```suggestion
2. Reconfigure Kafka clients to use the new listener.
```

PaulRMellor (approved):
Looks good. The proposal introduces useful support through configuration. I left a few suggestions to improve readability and clarity.

136-tls-route-listener.md
scholzj: I don't think so, given how little it impacts Strimzi itself. It is mostly abotu the Gateway implementations.

136-tls-route-listener.md
```diff
@@ -0,0 +1,281 @@
...
+But Strimzi will not be involved in `Gateway` management.
+Users will bring their own `Gateway` and only reference it in the Strimzi configuration.
+
+`TLSRoute` resources moved to the _Standard_ API in version 1.4 of the Gateway API specification.
+In version 1.5, the `v1` version of the `TLSRoute` API was introduced.
+While the `v1` version is relatively new, it should provide a stable and future-proof API.
+The `v1` API is expected to be supported in version 7.7.0 of the Fabric8 Kubernetes Client as well.
+This proposal suggests building `tlsroute` listener support on the `v1` API.
+
+The `type: tlsroute` listener will work on the same principle as the existing `type: ingress` and `type: route` listeners.
```
see-quick, line 64: Can we somehow reference this principle or add one sentence to explaining it easy way?

134-tls-route-listener.md
```diff
@@ -0,0 +1,281 @@
...
+      advertisedHostTemplate: kafka-{nodeId}-broker.192.168.1.221.sslip.io
+      hostTemplate: kafka-{nodeId}-broker.192.168.1.221.sslip.io
+```
+
+`TLSRoute` resources support TLS passthrough and TLS termination.
+While the TLS mode applies to the `TLSRoute` resources, it is configured in the [Gateway Listener TLS configuration](https://gateway-api.sigs.k8s.io/reference/1.5/spec/#listenertlsconfig) in the `Gateway` or in `ListenerSet` resources.
+TLS passthrough means that the TLS connection is forwarded to Strimzi as-is, including encryption.
+In this case, the TLS connection will use the TLS certificate from the Kafka brokers and can use mTLS authentication as well.
+
+![TLS Passthrough](./images/134-TLS-Passthrough.svg)
```
see-quick: nice diagram! 👍

136-tls-route-listener.md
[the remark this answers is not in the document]
see-quick: Okay I have got this also after this discussion ...  (so thanks for it! 👍 )

> The reasons for using the TLS Termination might be performance (i.e. if the Gateway has faster TLS processing then Java)

Is this really true? I mean you can imagine that depends on the Gateway implementation ... or? 
But yeah as you mentioned further there a benefits to it.

136-tls-route-listener.md
```diff
@@ -0,0 +1,281 @@
...
+Strimzi will not perform a detailed check of which gateway is among the parent references.
+It will not check for other conditions or try to detect any other warnings, errors, or failures.
+In case the `TLSRoute` does not have any parent references after the configured time limit (the Cluster Operator _reconciliation timeout_), the reconciliation will fail with a corresponding error.
+This will help provide flexibility and simplify support for different implementations that handle conditions, warnings, and errors differently.
+
+### Testing strategy
+
+While system tests for `type: tlsroute` might be added in the future, they are not covered by this proposal.
+We already have other listener types that rely fully on unit and manual testing only (e.g., Ingress).
+The value of system tests is also diminished in scenarios with multiple different implementations that might behave differently.
```
see-quick, line 231: Maybe we could investigate it further and make a cost/benefit analysis (i.e., how much it is needed to do from the ST POV)? But yeah we can do it after the proposal ...

see-quick (approved):
Thanks for the proposal! 👍 LGTM.

im-konge (approved):

136-tls-route-listener.md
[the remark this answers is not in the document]
scholzj: That will surely depend on the implementation. I assume something like Envoy might be faster. But I do not know for sure ... and probably don't care. What matters is talso what the users believe 😉.

Also, as mentioned above, there is also the centralized certificate management benefit as well.

136-tls-route-listener.md
scholzj: Good point ... added.

scholzj:
This proposal now has 4 binding +1 votes. Ig there are no new comments, I will close the vote on Monday EOB as approved.

scholzj:
Approved with 4 binding (+ my own implicit) +1 votes.

PR #212, opened 2026-03-30
Replace in-process ConnectCluster with StrimziConnectCluster

see-quick:
This proposal aims to replace it with the containerized `StrimziConnectCluster` from [test-container](https://github.com/strimzi/test-container), removing three Kafka Connect dependencies from the operator.

### DIFF
--- a/136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
+++ b/136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
@@ -29,20 +29,22 @@ Other already mentioned point is that also shrinks the dependency tree and elimi
 
 ## Proposal
 
-### Changes in test-container
+### New repository: strimzi/test-connectors
 
-Add a built-in testing connector to the test-container project so consumers can use it without depending on Kafka Connect libraries.
+Create a new repository under the Strimzi organization [`strimzi/test-connectors`](https://github.com/strimzi/test-connectors) to host Kafka Connect connectors used for testing purposes.
+This repository (i.) own independent release lifecycle (ii.) produces connector JARs as release artifacts and 
+(iii.) can host multiple test connectors in the future (e.g., source and sink connectors with different testing behaviors).
 
 #### StrimziTestingConnector
 
-A source connector (moved from the operators `TestingConnector`) with configurable behavior:
+The initial connector is a source connector (moved from the operators `TestingConnector`) with configurable behavior:
 
 - Configurable startup/shutdown delays (`start.time.ms`, `stop.time.ms`, `task.start.time.ms`, `task.stop.time.ms`)
 - Configurable task poll behavior (`task.poll.time.ms`, `task.poll.records`)
 - Configurable failures (`fail.on.start`, `task.fail.on.start`)
 - Configurable topic and partition count (`topic.name`, `num.partitions`)
 
-This class depends on `connect-api` (with `provided` scope).
+This class depends on `org.apache.kafka:connect-api`.
 
 #### StrimziTestingConnectorConfig
 
@@ -50,36 +52,50 @@ A constants-only class that exposes the connectors configuration keys and class 
 
 ```java
 public final class StrimziTestingConnectorConfig {
-    public static final String CONNECTOR_CLASS_NAME = "io.strimzi.test.container.StrimziTestingConnector";
+    public static final String CONNECTOR_CLASS_NAME = "io.strimzi.test.connectors.StrimziTestingConnector";
     public static final String FAIL_ON_START = "fail.on.start";
     public static final String TASK_FAIL_ON_START = "task.fail.on.start";
     public static final String START_TIME_MS = "start.time.ms";
     public static final String STOP_TIME_MS = "stop.time.ms";
-    // ... 
+    // ...
 }
 ```
 
 This allows consumers (like the operator) to reference connector configuration without adding `connect-api` to their classpath.
 
+### Changes in test-container-images
+
+The built connector JARs from `strimzi/test-connectors` are bundled into the Kafka Connect test container images at build time.
+The JAR is placed at `/opt/kafka/plugins/<name>/` inside the image.
+
+### Changes in test-container
+
 #### StrimziConnectCluster builder addition
 
-A new builder method enables the testing connector:
+Two new builder methods enable testing connectors:
 
 ```java
 public StrimziConnectClusterBuilder withTestingConnector() { }
+public StrimziConnectClusterBuilder withTestingConnector(String name) { }
 ```
 
-When enabled, the connector JAR is automatically built and copied into the container at `/opt/kafka/plugins/strimzi-testing-connector/`, and `plugin.path` is configured accordingly.
+The no-arg overload defaults to the `StrimziTestingConnector` (i.e., source connector).
+The `String name` overload allows selecting a specific connector by name, which supports adding more connectors to `strimzi/test-connectors` in the future (e.g., a sink connector) without changing the builder API.
+
+When enabled, the `plugin.path` is configured to include `/opt/kafka/plugins/<connector-name>/`, where the pre-built connector JAR is already present in the image.
 
 ### Changes in strimzi-kafka-operator
 
-The main changes are removal of `ConnectCluster.java` and `TestingConnector.java`, which would be moved to the test-container.
-With that we would remove `org.apache.kafka:connect-runtime`, `org.apache.kafka:connect-file` and `org.apache.kafka:connect-api` from the operator repo.
-And eventually we update integration test where we replace in-process with new one (i.e., inside test-container).
+These changes happen after the test-connectors repo, test-container-images, and test-container changes are in place:
+
+1. Remove `ConnectCluster.java` and `TestingConnector.java` from the operator.
+2. Remove `org.apache.kafka:connect-runtime`, `org.apache.kafka:connect-file`, and `org.apache.kafka:connect-api` dependencies.
+3. Add a dependency on `strimzi/test-connectors` (just for `StrimziTestingConnectorConfig` and there will be no `connect-api` transitive dependency).
+4. Update `KafkaConnectApiIT` and `KafkaConnectorIT` to use `StrimziConnectCluster` with `.withTestingConnector()`.
 
 ## Compatibility
 
-This adds new public API surface to test-container (`StrimziTestingConnector`, `StrimziTestingConnectorConfig`, and `withTestingConnector()` builder method), but it is strictly opt-in.
+This adds new public API surface to test-container (the `withTestingConnector()` builder method), but it is strictly opt-in.
 The testing connector is only included when the user explicitly calls `.withTestingConnector()` on the builder (i.e., the default behavior of `StrimziConnectCluster` remains unchanged).
 
 On the operator side, the changes are test-only and do not affect any public APIs.
@@ -90,8 +106,23 @@ On the operator side, the changes are test-only and do not affect any public API
 
 The current approach works, but it introduces unnecessary dependencies that have already caused issues with CVEs.
 
+### Build the connector JAR at runtime
+
+An approach where the connector classes are extracted and assembled into a JAR at container startup (inside `containerIsStarting`) was prototyped.
+While functional, this requires hardcoding class file paths or implementing dynamic class matching rules, making it fragile and harder to maintain.
+Pre-building the JAR as part of a proper release lifecycle and bundling it into the image is cleaner.
+
+### Adopt the echo-sink connector
+
+The existing [echo-sink connector](https://github.com/scholzj/echo-sink) could be moved under the Strimzi organization.
+However, it would need delay and failure simulation features added to match the current `TestingConnector` capabilities.
+
+### Embed connector code in test-container
+
+Placing the connector source code directly in the test-container project would require building the connector JAR at runtime, since test-container does not produce container images itself.
+It is cleaner to keep the connector in a separate repository and bundle the pre-built JAR during the test-container-images build process instead.
+
 ### Add connect-api as a compile dependency in test-container
 
-The `StrimziTestingConnector` needs the Connect API to implement the `SourceConnector` interface.
-Making it a compile dependency would force all test-container consumers to pull in the Connect API even if they don't use it.
-Using `provided` scope keeps it optional, and `StrimziTestingConnectorConfig` gives consumers access to configuration constants without the dependency.
+Making `connect-api` a compile dependency would force all test-container consumers to pull in the Connect API even if they don't use it.
+A separate repository avoids this entirely.
--- a/136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
+++ b/136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
@@ -24,7 +24,7 @@ The operator can use this directly instead of maintaining its own in-process imp
 ## Motivation
 
 The main motivation is that it removes unnecessary dependencies (i.e., three Connect libraries are only used by two test classes).
-And with that it also reduce overall maintenance for us having both in-memory nad container variant.
+And with that it also reduce overall maintenance for us having both in-memory and container variant.
 Other already mentioned point is that also shrinks the dependency tree and eliminates a recurring source of CVE-related upgrade blockers.
 
 ## Proposal
@@ -38,7 +38,7 @@ This repository:
 - Produces connector JARs as release artifacts
 - Can host multiple test connectors in the future (e.g., source and sink connectors with different testing behaviors)
 
-#### StrimziTestSourceConnector
+#### StrimziFaultInjectionSourceConnector
 
 The initial connector is a source connector (moved from the operator's `TestingConnector`) with configurable fault-injection behavior:
 
@@ -49,13 +49,13 @@ The initial connector is a source connector (moved from the operator's `TestingC
 
 This class depends on `org.apache.kafka:connect-api`.
 
-#### StrimziTestSourceConnectorConfig
+#### StrimziFaultInjectionSourceConnectorConfig
 
 A constants-only class that exposes the connector's configuration keys and class name **without** depending on the Kafka Connect API:
 
 ```java
-public final class StrimziTestSourceConnectorConfig {
-    public static final String CONNECTOR_CLASS_NAME = "io.strimzi.test.connectors.StrimziTestSourceConnector";
+public final class StrimziFaultInjectionSourceConnectorConfig {
+    public static final String CONNECTOR_CLASS_NAME = "io.strimzi.test.connectors.StrimziFaultInjectionSourceConnector";
     public static final String FAIL_ON_START = "fail.on.start";
     public static final String TASK_FAIL_ON_START = "task.fail.on.start";
     public static final String START_TIME_MS = "start.time.ms";
@@ -78,13 +78,13 @@ The JAR is placed at `/opt/kafka/plugins/<connector-name>/` inside the image.
 A dedicated builder method for each test connector:
 
 ```java
-public StrimziConnectClusterBuilder withStrimziTestSourceConnector() { }
+public StrimziConnectClusterBuilder withStrimziFaultInjectionSourceConnector() { }
 ```
 
 When a sink test connector is added in the future, a corresponding method is added:
 
 ```java
-public StrimziConnectClusterBuilder withStrimziTestSinkConnector() { }
+public StrimziConnectClusterBuilder withStrimziFaultInjectionSinkConnector() { }
 ```
 
 When enabled, the `plugin.path` is configured to include `/opt/kafka/plugins/<connector-name>/`, where the pre-built connector JAR is already present in the image.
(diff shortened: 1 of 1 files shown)


136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
```diff
@@ -0,0 +1,96 @@
...
+And with that it also reduce overall maintenance for us having both in-memory nad container variant.
+Other already mentioned point is that also shrinks the dependency tree and eliminates a recurring source of CVE-related upgrade blockers.
+
+## Proposal
+
+### Changes in test-container
+
+Add a built-in testing connector to the test-container project so consumers can use it without depending on Kafka Connect libraries.
+
+#### StrimziTestingConnector
```
scholzj: Doesn't this need to live somewhere in the Test Container images?

136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
```diff
@@ -0,0 +1,96 @@
...
+
+#### StrimziConnectCluster builder addition
+
+A new builder method enables the testing connector:
+
+```java
+public StrimziConnectClusterBuilder withTestingConnector() { }
+```
+
+When enabled, the connector JAR is automatically built and copied into the container at `/opt/kafka/plugins/strimzi-testing-connector/`, and `plugin.path` is configured accordingly.
```
scholzj: Following up on the previous comment - are you really building it at runtime? That seems strange. I would expect it to be part of the test container images and the `withTestingConnector()` method just modify the plugin path.

Maybe you can elaborate on how this will be actually done?

136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Changes in strimzi-kafka-operator
+
+The main changes are removal of `ConnectCluster.java` and `TestingConnector.java`, which would be moved to the test-container.
+With that we would remove `org.apache.kafka:connect-runtime`, `org.apache.kafka:connect-file` and `org.apache.kafka:connect-api` from the operator repo.
+And eventually we update integration test where we replace in-process with new one (i.e., inside test-container).
```
scholzj: I guess the other way around? You first need to add it to the test container, replace it in the integration tests and only then remove it from Strimzi Operators?

136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
```diff
@@ -0,0 +1,96 @@
...
+(i.) **`ConnectCluster`**, which starts an in-process Kafka Connect cluster using `org.apache.kafka.connect.cli.ConnectDistributed`;
+(ii.) **`TestingConnector`**, a source connector with configurable delays and failures, used by `KafkaConnectApiIT` and `KafkaConnectorIT`.
+
+These classes require three test-scoped dependencies:
+- `org.apache.kafka:connect-runtime`
+- `org.apache.kafka:connect-file`
+- `org.apache.kafka:connect-api`
+
+The `connect-runtime` dependency, in particular, pulls in a large transitive dependency tree.
+This has repeatedly caused issues within CVEs (one of them might be https://github.com/strimzi/strimzi-kafka-operator/pull/12494).
```
scholzj: Maybe you can elaborate here that it is not so much the Connect as test dependency causing the CVE issues. But it is dependency conflicts for things such as Jetty it pulls in.

scholzj (commented):

136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
see-quick: Yeah, that was my approach which I checked (i.e., build it basically runtime and copy with to the containers once during startup ) but I didn't think about having it in the test-container-images as there is no java code :/ and wasn't sure ... but thinking about it again it would be maybe better I guess also when we would add more test connectors in the future... I guess?

see-quick: Like approach with having it directly inject it via images would be:
1.  having some parent `plugins/pom.xml` in the test-container-images which would follow the versioning of test-container-images as we have it today
2. then you would have child (i.e., `plugins/strimzi-testing-connector/pom.xml` which would inherit whole stuff and name it `strimzi-testing-connector` (for our testing connector)
3. and also the whole `StrimziTestingConnector.java` would be moved to the `test-container-images`
4. and eventually do some make target (i.e., `build_plugins`) which would build and copy it into `images/plugins/<connector-name>/` possibly?
5. and when building the images we would just copy those jars into image?

That's how you imagine it? Because having code within `test-container` (i.e., StrimziTestingConnector.java) and reference it during building time within `test-container-images` would I think mean we have circular depedency AFAIK?

scholzj: I think having the Java code directly in the Test Container Images repo would be probably bad idea as there is nothing else Java. So I agree with that. My thought was more about including the JAR there into the container images. However ...

* I think the idea to _copy it to the container_ at runtime is hard to understand for me. How do you make sure it is in the container before you start it? Does it have some init containers? But if you can do it, I think that is fine. So I would be curious how it looks like.
* But the idea to build the Java code at runtime seems pretty wrong. Not sure if I misunderstood it or not. I could for example imagine that you have the JAR built at release and just copy it to the container. Not really building it on the fly.
* Obviously, we can consider a separate repo as well for the connector. It can have its own release cycle and be bundled in the test container images as I suggested. Or use a different connector. AFAIK we currently use my `echo-sink` connector in STs in operators. We can use that one (I'm happy to move it to Strimzi if desired). Or we can create a brand new Strimzi one and use it also in STs.

see-quick: >  How do you make sure it is in the container before you start it? Does it have some init containers? But if you can do it, I think that is fine.

You basically copy the built connector of the jar within the `containerIsStarting` testcontainer method i.e., 
```java
JarOutputStream jos = new JarOutputStream(baos)) {
...
            // just to show ... (but this one should be dynamically obtained with some rules ... (which we define)
            String[] classFiles = {
                "io/strimzi/test/container/StrimziTestingConnector.class",
                "io/strimzi/test/container/StrimziTestingConnector$TestingTask.class"
            };

            for (String classFile : classFiles) {
                try (InputStream is = getClass().getClassLoader().getResourceAsStream(classFile)) {
                    if (is != null) {
                        jos.putNextEntry(new JarEntry(classFile));
                        is.transferTo(jos);
                        jos.closeEntry();
                    }
                }
            }

            jos.finish();
            ...
}
```
Obviously this is a bit bad as it's hardcoded :/ but maybe having some dynamic matching would be acceptable? (as we would have define some rules that okay all connector related classes must contain `*Connector*` ... ? 

And of course another bad problem is that every time when test starts it has to build it ( ... but on the happy note when I tested it took like `3 ms` so I think I can live with that ... :)

> But the idea to build the Java code at runtime seems pretty wrong. Not sure if I misunderstood it or not. I could for example imagine that you have the JAR built at release and just copy it to the container. Not really building it on the fly.

This is a bit more tricky ... what I tried in my first appraoches was to get location (i.e., `StrimziConnectContainer.class.getProtectionDomain().getCodeSource().getLocation();`) which would resolve as `~/.m2/.../io/strimzi/strimzi-test-container/0.115.0/strimzi-test-container-0.115.0.jar `and that would eventually copy whole jar (which also copies other classes not just StrimziTestingConnector). I also then re-fine my approach to just allow specific classes (i.e., "io/strimzi/test/container/StrimziTestingConnector.class", "io/strimzi/test/container/StrimziTestingConnector$TestingTask.class" ) to avoid some classloading magic or coflicts with Kafka transitive classes. Note that copying whole jar with also other classes in during compile time ... and having those specific classes (road which I went) is runtime ... 

> Obviously, we can consider a separate repo as well for the connector. It can have its own release cycle and be bundled in the test container images as I suggested. Or use a different connector. AFAIK we currently use my echo-sink connector in STs in operators. We can use that one (I'm happy to move it to Strimzi if desired). Or we can create a brand new Strimzi one and use it also in STs.

Yeah might be the option I was considering this at the start but it feel like a too much effort ... while the approach with runtime building worked well.

scholzj: > > Obviously, we can consider a separate repo as well for the connector. It can have its own release cycle and be bundled in the test container images as I suggested. Or use a different connector. AFAIK we currently use my echo-sink connector in STs in operators. We can use that one (I'm happy to move it to Strimzi if desired). Or we can create a brand new Strimzi one and use it also in STs.
> 
> Yeah might be the option I was considering this at the start but it feel like a too much effort ... while the approach with runtime building worked well.

I think we just need to decide how deep we want to go ... from just pulling the JARs of my connector up to creating our own. But any of that would be a much cleaner approach than extracting and copying some compiled classes.

see-quick: In the operator we need `SourceConnector`, but yeah I think looking at all possible options; if we don't want to go with a bit hackish style (i.e., extracting and copying some compiled classes) then my take would be having separate repo for connector write our own with all functionality as it is currently in `StrimziTestingConnector` (i.e., connector/test start-up failure, connector start/stop delay, task start/stop/poll delay and there are also some more which I don't remember exactly but you get the point). So I am more open to do it in separate repo :)) that also opens doors for multiple connectors in the future, which we could potentionally use (and eventually also using Sink connector within test-container)?

see-quick: It would be also great others point on this @strimzi/maintainers ? WDYT ^^

scholzj: Refresh my memory please ... what does the source produce?

im-konge: I think having something "clearer" like separate repo for the testing connector(s) would be better. We are anyway using the echo-sink connector from Jakub in STs to test the auto-restarts IIRC. Having something like that in our org would be good, we can enhance it as we want.

Also FMPOV having something like this in the code (I mean the stuff around `JarOutputStream jos = new JarOutputStream(baos))`) would be tricky when you come back to it after some time. But maybe it's just my view.

But I'm +1 for having separate repo for one (or even more) connectors and bundling it in the images directly.

im-konge (commented):
Thanks for the proposal, it looks good to me - but I agree with the comments from Jakub + the one comment around adding the connector JAR to the image is something that needs to be resolved.

136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
see-quick: > Refresh my memory please ... what does the source produce?

Just long variable which is increasing i.e.,
```java
for (int i = 0; i < taskPollRecords; i++) {
                records.add(new SourceRecord(Collections.singletonMap("", ""),
                        Collections.singletonMap("", ""),
                        topicName, (int) record % numPartitions,
                        null, null,
                        valueSchema, record++));
            }
``` 
but not sure we somehow test this values within `KafkaConnectorIT`(I think that integration tests are mainly for reconciliation behaviour not specific data ...). So in the end I think type of connector doesn't matter => what matters is just support for configuring those delay/failure properties within connector.

136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
see-quick: I think this one would be anyway updated once decide in the https://github.com/strimzi/proposals/pull/212#discussion_r3044264326 conversation and then I will update it.

136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
see-quick: updated

136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
see-quick: So to sum up this discussion for others we have two options:
1. Separate Strimzi repo for testing connector(s) = this would mean adding new repo under strimzi (i.e., `strimzi/test-connectors`), basically copy our testing connector, it would own release lifecycle and bundle JARs into `test-container-images`. In the future we can have multiple connectors to test.  
2. Adopt [echo-sink connector](https://github.com/scholzj/echo-sink) = this mean move into Strimzi, but would need to add delay/failure simulation simulation features.
 
@strimzi/maintainers WDYT?

I am more inclined for option 1 (also @im-konge is for 1. I assume that from previous comments).

ppatierno: Catching up with the discussion, I am for option 1 as well.

see-quick: Okay so I will goahead and update the proposal with first approach. Thanks.

see-quick: Updated so please take a look.

136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
```diff
@@ -0,0 +1,128 @@
...
+And with that it also reduce overall maintenance for us having both in-memory nad container variant.
+Other already mentioned point is that also shrinks the dependency tree and eliminates a recurring source of CVE-related upgrade blockers.
+
+## Proposal
+
+### New repository: strimzi/test-connectors
+
+Create a new repository under the Strimzi organization [`strimzi/test-connectors`](https://github.com/strimzi/test-connectors) to host Kafka Connect connectors used for testing purposes.
+This repository (i.) own independent release lifecycle (ii.) produces connector JARs as release artifacts and 
+(iii.) can host multiple test connectors in the future (e.g., source and sink connectors with different testing behaviors).
```
scholzj: Whould this be more readable as a simple bullet points?

136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
```diff
@@ -0,0 +1,128 @@
...
+
+#### StrimziTestingConnector
```
scholzj: Is this the name? Or what should be the name? If you expect multiple connectors in the future, it should probably at leats include the Source / Sink in the name.

scholzj: Is this the name? Or what should be the name? If you expect multiple connectors in the future, it should probably at leats include the Source / Sink in the name. That should be reflected in all the naming I guess.

see-quick: Yeah, you are right. What about `StrimziFaultSourceConnector` or `StrimziTestFaultSourceConnector` maybe as that connector allows use to configure delays and failures?

scholzj: Maybe `StrimziTestSourceConnector` would be enough? I think the Source / SInk in the name would be good.

see-quick: Sure, StrimziTestSourceConnector is also good.

136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Changes in test-container
+
+#### StrimziConnectCluster builder addition
+
+Two new builder methods enable testing connectors:
+
+```java
+public StrimziConnectClusterBuilder withTestConnector() { }
+public StrimziConnectClusterBuilder withTestConnector(String name) { }
```
see-quick: I am thinking about this part ... should we have only
```java
public StrimziConnectClusterBuilder withTestConnector(String name) { }
```
as part of the public API. And if so is just String a way? I was wondering if having some enum i.e.,
```java
public enum StrimziTestConnectors {
     NAME_OF_CONNECTOR1,
     NAME_OF_CONNECTOR2 
     // other connectors possibly added into future 
  }
```
would be good?

in user code it would look like:
```java
StrimziConnectCluster connectCluster = new StrimziConnectCluster.StrimziConnectClusterBuilder()
       .withGroupId("my-cluster")
       .withKafkaCluster(kafkaCluster)
       .withTestConnector(StrimziTestConnectors.NAME_OF_CONNECTOR1)
       .build();
```



In practice if we would add any connector to the `test-connectors` repo that would mean adding it into image eventually. That enum at least would give you clear point of what connectors you can use as end user ... instead of using some String.

@strimzi/maintainers ?

scholzj: I do not think you will have that many connectors. So I would just use the full connector name in the builder. So `withStrimziTestSourceConnector()`. We might have some reason to add some Sink test connector and that would be it, or? So you would have 2 builder methods.

Alternatively, if you expect users to also use their own connectors, you can use the generic part. But in that case you should just use Strings and not any enums.

see-quick: > I do not think you will have that many connectors. So I would just use the full connector name in the builder. So withStrimziTestSourceConnector(). We might have some reason to add some Sink test connector and that would be it, or? So you would have 2 builder methods.

Yeah, that make sense. Thanks

> Alternatively, if you expect users to also use their own connectors, you can use the generic part. But in that case you should just use Strings and not any enums.

Yeah, true. I think that can be added in the future if we see demand for it. I will stick with just `withStrimziTestSourceConnector()` and with potentional to also add `withStrimziTestSinkConnector() in the future then.

scholzj (approved):
LGTM. Thanks. +1.

136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
```diff
@@ -0,0 +1,134 @@
...
+These transitive dependencies can conflict with the versions used by the operator itself, blocking CVE fixes.
+For example, [updating Jetty to address CVEs](https://github.com/strimzi/strimzi-kafka-operator/pull/12494) was blocked because the operator Jetty upgrade clashed with the version pulled in by `connect-runtime`, and there was no simple override to resolve it.
+
+Since [Strimzi Proposal #91](091-add-connect-to-test-container.md), the test-container project already provides `StrimziConnectCluster`, a containerized Kafka Connect cluster.
+The operator can use this directly instead of maintaining its own in-process implementation.
+
+## Motivation
+
+The main motivation is that it removes unnecessary dependencies (i.e., three Connect libraries are only used by two test classes).
+And with that it also reduce overall maintenance for us having both in-memory nad container variant.
```
ppatierno:
```suggestion
And with that it also reduce overall maintenance for us having both in-memory and container variant.
```

136-replace-in-process-connect-cluster-with-strimzi-connect-cluster.md
```diff
@@ -0,0 +1,134 @@
...
+### New repository: strimzi/test-connectors
+
+Create a new repository under the Strimzi organization [`strimzi/test-connectors`](https://github.com/strimzi/test-connectors) to host Kafka Connect connectors used for testing purposes.
+This repository:
+
+- Owns an independent release lifecycle
+- Produces connector JARs as release artifacts
+- Can host multiple test connectors in the future (e.g., source and sink connectors with different testing behaviors)
+
+#### StrimziTestSourceConnector
```
ppatierno: Should we have a description about what actually this connector does? Isn't it a too generic name if in the future we are going to add more connectors? I mean, for example, `FileSourceConnector` (within Kafka) is something more specific. AFAICS, the `TestingConnector` doesn't have a real source but generates incrementing numbers so some "synthetic/fake" data. Maybe we can describe what it does here and call it `SyntheticDataSrouceConnector` ? Or something similar?

scholzj: I don't think the synthetic data are the main purpose of the connector. The main purpose is the artificially raised errors and timing options. So I think the Testing name corresponds to this. Whereas synthetic data does not, because I do not think we have any intention to attract more Prs for more sophisticated data, etc. to this connector.

ppatierno: I am fine with having a different name than "Synthetic" but just "Testing" doesn't mean anything and means everything at the same time. Even if the goal is not about synthetic data, the current naming it's totally unclear imho. I can't understand from naming something like "Testing" that it's about raising error and timing issues.

scholzj: That is fine. All I was trying to say is that it has way more to do with Testing (emulating various test conditions) than with Synthetic data.

see-quick: Hmm ... if we really want to be specific in naming maybe `FaultTestSourceConnector` would be the choice maybe even with Strimzi prefix (i.e., StrimziFaultTestSourceConnector)? WDYT?

scholzj: What about `StrimziFaultInjectionSourceConnector`?

ppatierno: +1 for `StrimziFaultInjectionSourceConnector`

see-quick: Updated :)

ppatierno (approved):

im-konge (approved):

see-quick:
This proposal has 3 + 1 approvals. I plan to merge it by the end of the week unless there are any objections. Thanks!

tinaselenge (approved):
LGTM

PR #214, opened 2026-04-03
Add proposal for configuring mTLS validityDays and renewalDays per Ka…

sebastiangaiser:
…fkaUser

This proposal suggests the solution for https://github.com/strimzi/strimzi-kafka-operator/issues/12336.

### DIFF
--- a/138-configurable-validitydays-and-renewaldays-per-kafkauser.md
+++ b/138-configurable-validitydays-and-renewaldays-per-kafkauser.md
@@ -43,10 +43,13 @@ spec:
 When set, these values override the cluster-level defaults from the `Kafka` resource for that specific user.
 When omitted, the cluster-level defaults continue to apply.
 
+If `validityDays` is set or reduced on an existing `KafkaUser` whose current certificate would already be expired or would exceed the new validity period, the User Operator must renew the certificate immediately on the next reconciliation rather than waiting for the natural renewal window.
+
 The User Operator must enforce that `renewalDays` < `validityDays`.
 If this constraint is violated, the `KafkaUser` reconciliation must fail with a descriptive status condition.
 
-These fields only apply when using `spec.authentication.type: tls` and must be rejected by schema validation for other authentication types as this indicates a misconfiguration.
+These fields only apply when using `spec.authentication.type: tls`.
+Since the OpenAPI schema cannot enforce cross-field constraints, a CEL validation rule will be used to reject these fields when `type` is not `tls`, ensuring a misconfiguration is caught at admission time.
 
 ## Affected/not affected projects
 
@@ -71,8 +74,8 @@ metadata:
   name: my-user
   namespace: my-kafka
   annotations:
-    strimzi.io/validityDays: 7d
-    strimzi.io/renewalDays: 36h
+    strimzi.io/validityDays: 7
+    strimzi.io/renewalDays: 2
 spec:
   ...
 ```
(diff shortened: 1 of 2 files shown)


im-konge (approved):
Thanks for the proposal :) LGTM.

138-configurable-validitydays-and-renewaldays-per-kafkauser.md
```diff
@@ -0,0 +1,81 @@
...
+    renewalDays: 2
+```
+
+When set, these values override the cluster-level defaults from the `Kafka` resource for that specific user.
+When omitted, the cluster-level defaults continue to apply.
+
+The User Operator must enforce that `renewalDays` < `validityDays`.
+If this constraint is violated, the `KafkaUser` reconciliation must fail with a descriptive status condition.
+
+These fields only apply when using `spec.authentication.type: tls` and must be rejected by schema validation for other authentication types as this indicates a misconfiguration.
```
scholzj: You would need to use CEL validation for that as the schema itself cannot enforce that. It might be worth mentioning it here in the proposal.

scholzj (approved):
Thanks for the proposal. I left one nit, but otherwise it looks good to me.

138-configurable-validitydays-and-renewaldays-per-kafkauser.md
```diff
@@ -0,0 +1,81 @@
...
+
+```yaml
+apiVersion: kafka.strimzi.io/v1
+kind: KafkaUser
+metadata:
+  name: my-user
+  namespace: my-kafka
+  annotations:
+    strimzi.io/validityDays: 7d
+    strimzi.io/renewalDays: 36h
```
see-quick: It's a bit inconsitent while in proposal you mean integers here I can see `validityDays: 7d`  or `renewalDays: 36h`. I assume here should be also just integers?

138-configurable-validitydays-and-renewaldays-per-kafkauser.md
```diff
@@ -0,0 +1,81 @@
...
+  name: my-user
+  namespace: my-kafka
+spec:
+  authentication:
+    type: tls
+    validityDays: 7
+    renewalDays: 2
+```
+
+When set, these values override the cluster-level defaults from the `Kafka` resource for that specific user.
```
see-quick, line 43: So if someone adds `validityDays: 7` to existing KafkaUser, whose current certificate has 300 days remaining, would it be immediately renewed or only when the existing cert would naturally eneter its renew window? I assume it would be `immediately renewed`? But correct me if I am wrong.

see-quick (approved):
LGTM 👍, just some nits to consider. Thanks.

138-configurable-validitydays-and-renewaldays-per-kafkauser.md
sebastiangaiser: Yes, added it to the proposal.

138-configurable-validitydays-and-renewaldays-per-kafkauser.md
sebastiangaiser: Yes, changed it.

138-configurable-validitydays-and-renewaldays-per-kafkauser.md
sebastiangaiser: Yes, added it to the proposal.

scholzj:
This proposal now has 3 binding +1 votes. If there are no more comments by Monday 13th EOB, I will close it as approved.

ppatierno (approved):
LGTM.

scholzj:

Thanks for the proposal @sebastiangaiser.

PR #215, opened 2026-04-06
Add proposal for allow background deletion propagation with Strimzi is rolling pods

guancioul:
This proposal suggests a solution for https://github.com/strimzi/strimzi-kafka-operator/issues/12212.
Introduces a new feature gate to enable background deletion propagation for Pods, ensuring network connectivity is maintained during graceful shutdowns in environments using CNIs like Cilium.


139-background-deletion-propagation.md
```diff
@@ -0,0 +1,73 @@
...
+This change is primarily implemented in the `PodOperator` used by the `KafkaRoller`, `ManualPodCleaner`, and other components. Note that the `StrimziPodSetController` already uses `Background` deletion and is not affected by this change.
+While `Foreground` deletion is generally safer for cascading cleanup of complex resources, `Background` deletion is the appropriate choice for Pods in CNI environments where endpoint lifecycle must match the container lifecycle.
+
+## Affected/not affected projects
+
+The only affected project is the `strimzi-kafka-operator` repository, particularly the following classes:
+
+- `FeatureGates`
+- `PodOperator`
+- `KafkaRoller`
```
scholzj, line 55: You said `KafkaConnectRoller` uses `Orphan` strategy. Should that be aligned as well?

137-background-deletion-propagation.md
```diff
@@ -0,0 +1,73 @@
...
+- `FeatureGates`
+- `PodOperator`
+- `KafkaRoller`
+
+Other projects in the Strimzi organization are not affected.
+
+## Compatibility
+
+This proposal is fully backwards compatible.
+The default behavior of the operator will remain unchanged (using `Foreground` deletion) unless the `UseBackgroundPodDeletion` feature gate is explicitly enabled by the user.
```
scholzj: Well, I do not think this is how we see backwards compatibility. The feature gate would eventually transition to beta and be enabled by default. So if the feature gate being disabled is what ensures the compatibility, then you just do not break it when introducing it but 2 releases later.

I think this whle proposal has to be based on the fact that the background deeltion:
* Helps the Cilium users
* Does not break things for anyone else

And while we do not know this is the case for sure, we expect that it will be the case and we use the feature gate to confirm it.

So I would phrase this in that direction.

137-background-deletion-propagation.md
```diff
@@ -0,0 +1,73 @@
...
+Switching to `Background` deletion propagation allows the Pod object to be removed from the API while leaving its dependents (like the network endpoint) intact until the Pod's containers have fully exited.
+
+## Proposal
+
+The use of background deletion propagation for Pods will be gated behind a feature gate called `UseBackgroundPodDeletion`. 
+
+| Phase | Strimzi versions       | Default state                                          |
+|:------|:-----------------------|:-------------------------------------------------------|
+| Alpha | 1.1.0 - 1.2.0          | Disabled by default                                    |
+| Beta  | 1.3.0 - 1.4.0          | Enabled by default                                     |
```
scholzj: I would probably go for longer here as many users do not upgrade to every single release, and this might be easy to skip. So I think 1.1-1.2 are fine for alpha. I would probably make beta at least 3, but rather 4 releases (1.3-1.6) and GA in 1.7.

scholzj (commented):
Thanks for the proposal. I left some comments and requested a review from Strimzi maintainers. But I think it looks pretty good already.

137-background-deletion-propagation.md
guancioul: Thanks for the reminder! I've updated the table with the new versions for the Beta and GA phases to ensure a smoother transition for users.

139-background-deletion-propagation.md
guancioul: I have updated the proposal to include `KafkaConnectRoller` for consistency. Thanks!

137-background-deletion-propagation.md
guancioul: Thanks for the clarification on the feature gate lifecycle and the expectations for backward compatibility. I’ve updated the "Compatibility" section to focus on the inherent safety of the change rather than just its default state.

scholzj (approved):
Thanks. LGTM.

im-konge (approved):

ppatierno (approved):

scholzj:
This proposal now has 3 binding +1 votes. If there are no more comments by Thursday EOB, it will be merged as approved.

Frawless (approved):

PaulRMellor (approved):

scholzj:
Approved with 5 binding +1 votes..

PR #218, opened 2026-04-22
Per broker annotation and label template

pkotak:
Proposal for adding listener-level annotation and label templates for per-broker Services, Routes, and Ingresses.


140-per-broker-annotation-and-label-templates.md
```diff
@@ -0,0 +1,352 @@
...
+That is especially visible in KRaft clusters using node pools, where users often need to think in terms of node IDs and pod names rather than replica ordinals.
+
+## Motivation
+
+In most real deployments, per-broker annotations and labels follow a predictable pattern.
+Examples include:
+
+* `external-dns.alpha.kubernetes.io/hostname: kafka-{nodeId}.example.com.`
+* `my-company.io/pod-name: {nodePodName}`
+* `topology.strimzi.io/broker-id: {nodeId}`
```
scholzj: Here as well as elsewhere -> any `strimzi.io` labels and annotations should be set by Strimzi it self. Not by the user. So I would not use them as the example. Just use some custom ones as the one above for example, or some real live examples such as the External DNS one.

140-per-broker-annotation-and-label-templates.md
```diff
@@ -0,0 +1,352 @@
...
+
+Strimzi already solved this problem for hostnames through `hostTemplate` and `advertisedHostTemplate`.
+The same approach fits per-broker annotations and labels well.
+
+## Proposal
+
+This proposal suggests adding two new listener-level fields under `spec.kafka.listeners[].configuration`:
+
+* `annotationsTemplate`
+* `labelsTemplate`
```
scholzj: Should this be `perBrokerAnnotations` and `perBrokerLabels`? To make it clear it does not appyl to bootstrap resources. The existing templates are not prefixed that wy, but there it is a bit more obvious hows they are used.

140-per-broker-annotation-and-label-templates.md
```diff
@@ -0,0 +1,352 @@
...
+        port: 9094
+        type: loadbalancer
+        tls: true
+        configuration:
+          annotationsTemplate:
+            external-dns.alpha.kubernetes.io/hostname: kafka-{nodeId}.example.com.
+            my-company.io/pod-name: "{nodePodName}"
+          labelsTemplate:
+            strimzi.io/node-id: "{nodeId}"
+            strimzi.io/pod-name: "{nodePodName}"
```
scholzj: As mentioned above -> we should not mix `strimzi.io` labels here.

140-per-broker-annotation-and-label-templates.md
```diff
@@ -0,0 +1,352 @@
...
+Bootstrap metadata remains configured explicitly through `configuration.bootstrap.annotations` and `configuration.bootstrap.labels`.
+
+### Conflict handling and precedence
+
+Users should be able to set a default pattern and still override specific brokers.
+So the precedence should be:
+
+1. render `annotationsTemplate` / `labelsTemplate`
+2. overlay `configuration.brokers[].annotations` / `configuration.brokers[].labels`
+3. use the merged result on the generated resource
```
scholzj: How would users _unset_ a label or annotation in this model? I think it would be better to instead of merging the labels and annotations to do it the way that:
* If `configuration.brokers[].annotations` / `configuration.brokers[].labels` are set, they will be used
* When they are unset, `annotationsTemplate` / `labelsTemplate` will be used

I think it makes more sense that way and gives more flexibility (users who already set `configuration.brokers[].annotations` / `configuration.brokers[].labels` for given broker can easily set there everything)

140-per-broker-annotation-and-label-templates.md
```diff
@@ -0,0 +1,352 @@
...
+```yaml
+annotations:
+  external-dns.alpha.kubernetes.io/hostname: special-broker.example.com.
+  external-dns.alpha.kubernetes.io/ttl: "60"
+labels:
+  dns.zone: secondary
+  strimzi.io/node-id: "1"
+```
+
+This behavior is more flexible than using the template only when the per-broker map is completely absent.
```
scholzj: As explained above, it is not because you cannot unset values.

140-per-broker-annotation-and-label-templates.md
```diff
@@ -0,0 +1,352 @@
...
+#### Kafka cluster generation changes
+
+The per-broker resource generation in `KafkaCluster` already routes all per-broker listener metadata through `ListenersUtils`.
+Therefore the feature can be implemented centrally without duplicating logic in each resource builder.
+
+The affected generated resources are:
+
+* per-broker external services
+* per-broker routes
+* per-broker ingresses
```
scholzj: The TLSRoutes resources would be affected as well (not implemented yet, but already approved as a proposal). I guess it should also apply to `cluster-ip` type per-broker services in `type: cluster-ip` listeners?

140-per-broker-annotation-and-label-templates.md
```diff
@@ -0,0 +1,352 @@
...
+* The change is additive and introduces no breaking API behavior.
+* Users can adopt the templates gradually.
+
+## Rejected alternatives
+
+### Apply templates only when `brokers[].annotations` or `brokers[].labels` are absent
+
+This was rejected because it makes partial overrides awkward.
+Users often want a common default plus a targeted override for one broker.
+Merging template values first and then overlaying explicit broker metadata is more useful.
```
scholzj: Sorry, I would argue that the partial overrides are awkward 😉

scholzj (commented):
Thanks for the proposal. I left some comments. But I think it looks mostly good.

140-per-broker-annotation-and-label-templates.md
pkotak: Agreed. How about using `perBrokerAnnotationsTemplate` and `perBrokerLabelsTemplate`. That way its consistent with the other template naming conventions e.g. `advertisedHostTemplate` while preserving the explicit nature of the variable name

140-per-broker-annotation-and-label-templates.md
pkotak: Good suggestion. It would indeed give more flexibility.

140-per-broker-annotation-and-label-templates.md
pkotak: `cluster-ip` yes.
Do you prefer i leave a small note that it will support TLSRoutes when implemented?

140-per-broker-annotation-and-label-templates.md
pkotak: From my perspective, I was thinking about the merge approach because its great for _defaults + small broker-specific additions_ but weak for the clarity of hidden changes that can happen use case. 

I don't have strong opinions on this one but i do agree the unset behavior is cleaner when choosing only between the template or specific additions.

140-per-broker-annotation-and-label-templates.md
scholzj: That sounds good. Not sure why I left the template out.

scholzj (approved):
Thanks. I think this looks good to me.

pkotak:
Thanks for the approval! I don’t have permission to merge this PR. @scholzj , would you be able to merge it or do we need some more buy ins? 
I did not update the README.md table index on purpose to avoid any serial number conflict if another proposal gets merged before this one.

scholzj:
> Thanks for the approval! I don’t have permission to merge this PR. @scholzj , would you be able to merge it or do we need some more buy ins? I did not update the README.md table index on purpose to avoid any serial number conflict if another proposal gets merged before this one.

No, Strimzi proposals need at least 3 +1 votes from Strimzi maintainers to be approved. So we need to wait for more people to review it. I requested their reviews, so hopefully they will get to it.

im-konge (approved):
LGTM, thanks for the proposal.

140-per-broker-annotation-and-label-templates.md
```diff
@@ -0,0 +1,363 @@
...
+## Motivation
+
+In most real deployments, per-broker annotations and labels follow a predictable pattern.
+Examples include:
+
+* `external-dns.alpha.kubernetes.io/hostname: kafka-{nodeId}.example.com.`
+* `my-company.io/pod-name: {nodePodName}`
+* `my-company.io/broker-id: {nodeId}`
+
+Requiring users to repeat these values broker by broker has several drawbacks:
```
PaulRMellor:
```suggestion
Requiring users to repeat these values for each broker has several drawbacks:
```

140-per-broker-annotation-and-label-templates.md
```diff
@@ -0,0 +1,363 @@
...
+Bootstrap metadata remains configured explicitly through `configuration.bootstrap.annotations` and `configuration.bootstrap.labels`.
+
+### Conflict handling and precedence
+
+Users should be able to define a common default while also having a simple way to fully customize specific brokers.
+To keep the behavior predictable and to allow users to omit values coming from the template, the precedence should be:
+
+1. if `configuration.brokers[].annotations` is set for a broker, use it as the full annotation map for that broker
+2. otherwise, if `configuration.perBrokerAnnotationsTemplate` is set, render it for that broker
+3. if neither is set, use no broker annotations
```
PaulRMellor:
```suggestion
3. if neither is set, do not apply broker annotations
```

140-per-broker-annotation-and-label-templates.md
```diff
@@ -0,0 +1,363 @@
...
+* Topic Operator behavior
+* User Operator behavior
+* Kafka protocol behavior itself
+* bootstrap listener metadata behavior
+
+## Compatibility
+
+This proposal is fully backwards compatible.
+
+* Existing configurations using only `brokers[].annotations` and `brokers[].labels` continue to work unchanged.
```
PaulRMellor:
```suggestion
* Existing configurations using `brokers[].annotations` and `brokers[].labels` continue to work unchanged.
```

PaulRMellor (approved):
LGTM.
Just a couple of minor suggestions as I read through

tinaselenge (approved):
LGTM. Thank you for the proposal.

scholzj:
This now has 4 binding +1 votes. If there are no more comments by Tuesday EOB, it will be merged as approved.

ppatierno (approved):
LGTM. Thanks.

scholzj:

@pkotak Sorry, I forgot about this during the week. But merging now. Thanks for the proposal. Feel free to open the implementation PR when you are ready.

PR #220, opened 2026-05-03
Strimzi Gatekeeper plugin system

scholzj:
Inspired by Kubernetes Admission webhooks and other Kubernetes tools, the Strimzi Gatekeeper plugins suggested by this proposal provide a strong abstraction that should help us to better structure our own code base, remove complexity, split it into separate layers, and improve tests. But it should also be useful for all kinds of external use cases presented by Strimzi users and vendors.


141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
+# Strimzi _Gatekeeper_ plugin system
+
+This proposal suggests introducing a new plugin system called _Gatekeeper_.
+
+## Current situation
+
+Strimzi currently supports only one type of plugin: Pod Security Providers.
```
PaulRMellor:
```suggestion
Strimzi currently supports only one type of plugin: Pod Security Provider plugins.
```

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+The Pod Security Provider plugins can be used to inject security context for the various containers deployed by the Strimzi Cluster Operator.
```
PaulRMellor:
```suggestion
The Pod Security Provider plugins can be used to inject security contexts into the containers deployed by the Strimzi Cluster Operator.
```

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+The Pod security providers are called directly from the Cluster Operator methods that generate the containers.
```
PaulRMellor:
```suggestion
The Pod Security Provider plugins are called directly from the Cluster Operator methods that generate containers.
```

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+The first is the internal code structure, including its complexity and testability.
+The second is the ability to add new custom features that might or might not be part of the core Strimzi project.
+
+### Strimzi code structure, complexity, and testability
+
+While the Pod Security Provider plugins are useful, their design is not practical, especially in the way they integrate with the Cluster Operator.
+Because they integrate directly into the methods that generate the containers, they lead to complex internal logic within the core operator methods.
+The operator has to:
+* Take into account the security context configured by the user in the custom resource
+* Pass it through the Pod security provider to get the actual context that should be used
```
PaulRMellor:
```suggestion
* Pass it through the Pod Security Provider plugin to get the security context that should be used
```

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+* Set it in the container
+
+This increases the complexity of the operator code.
+It also increases the testing requirements, because you need to test the various combinations within the core operator.
+
+And while the only plugin type we have today is the Pod Security Provider, we repeat this pattern in other places as well.
```
PaulRMellor:
```suggestion
Although Pod Security Provider plugins are currently the only supported plugin type, similar patterns are repeated elsewhere in the code base.
```

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+
+None of these options is ideal.
+While the Gatekeeper plugin system described below does not solve every user requirement, it could help in at least some cases.
+One of the later sections describes examples of Gatekeeper plugins that users might implement to address these issues.
+
+## Proposal
+
+Gatekeeper plugins would be invoked at the beginning and at the end of the reconciliation, controlling entry to and exit from that reconciliation.
+When reconciliation starts, the custom resource being reconciled would be passed to the _entry_ method of the Gatekeeper plugin.
+When reconciliation completes, whether successfully or with an error, the result and the updated custom resource(s), including status, would be passed to the _exit_ method of the Gatekeeper plugins.
```
PaulRMellor: Would exit plugins run even if reconciliation fails before resource generation or status updates occur?

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+
+There would be two basic types of Gatekeeper plugins:
```
PaulRMellor:
```suggestion
There would be two types of Gatekeeper plugins:
```

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+* Validating plugins
+* Mutating plugins
+
+The validating plugins would receive the custom resource(s) and would be able to fail reconciliation by raising an error.
+But they would not be able to modify the custom resource(s) and/or their status.
+
+The mutating plugins would be able to fail the reconciliation by raising an error.
+They would also be able to modify the custom resource(s) before the reconciliation starts or their status when the reconciliation ends.
```
PaulRMellor:
```suggestion
They would also be able to modify the custom resource(s) before the reconciliation starts or modify their status when the reconciliation ends.
```

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+These modifications would of course need to follow the Strimzi API.
+The plugins would not be able to add any new fields.
+Any modifications made by the plugin would be internal only.
+They would not be persisted into the Kubernetes API.
```
PaulRMellor: I'm not clear on persisted resource state and reconciliation-scoped mutations.
Maybe an example would help.

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+They would only be used within a given reconciliation, similarly to how we used _virtual node pools_ when migrating to node pools or how we modified the Connect and MirrorMaker2 custom resources during the migration from `v1beta2` to `v1` APIs.
```
PaulRMellor:
```suggestion
They would only be used within a given reconciliation, similarly to how virtual node pools were used during migration to node pools or how Connect and MirrorMaker2 custom resources were modified internally during migration from the `v1beta2` to `v1` APIs.
```

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+### Implementation
+
+### Configuring Gatekeeper plugins
+
+There would be three different groups of Gatekeeper plugins:
+* Mandatory plugins would be hardcoded in the Strimzi source code and would not be configurable.
+  These might be the plugins that Strimzi requires to work.
+* Default plugins would have a default configuration hardcoded in the Strimzi source code.
+  These would be the plugins Strimzi provides and _recommends_.
+  But the default plugins could be changed by the users who can configure their own default plugins.
```
PaulRMellor:
```suggestion
  But the default plugins could be changed by users who wish to configure their own default plugins.
```

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+* Custom plugins would be unset by default, and users would need to configure them explicitly.
+  Custom plugins would be the plugins provided by Strimzi but disabled by default.
+  Or actually custom plugins brought in by the user.
```
PaulRMellor:
```suggestion
  Or custom plugins brought in by the user.
```

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+So mirroring the configuration might lead to errors.
+Users can use the `template` section of `.spec.entityOperator` to configure the Gatekeeper plugins for User and Topic operators.
+When using standalone Topic or User operators, users can set these variables in their standalone deployments.
+
+#### Per-resource configuration
+
+The plugins will be configured only on a per-operator basis.
+They will not support any direct configuration on a per-custom resource basis.
+However, plugins can easily skip some resources based on an annotation.
+For example, a custom resource with the annotation `my-gatekeeper-plugin.scholz.cz/skip=true` can be used to tell the plugin to skip this custom resource.
```
PaulRMellor: Are there any security or policy considerations around skipping resources?

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+* Mandatory plugins
+
+Within each plugin group, the plugins will be invoked in the order specified by the user or by the Strimzi source code.
+
+At reconciliation _exit_, they will always be called in the following order:
+* Mandatory plugins
+* Default plugins
+* Custom plugins
+
+Within each plugin group, the plugins will be invoked in the reverse order specified by the user or by the Strimzi source code.
```
PaulRMellor, line 242: What is the rationale for reversing the order during exit processing? Should we say more on this behaviour?

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+
+The Gatekeeper plugins would be defined through a series of interfaces.
+
+The first interface, `GatekeeperPlugin`, would define the plugins.
+It would contain a single method, `configure(...)`, which would be called once at startup when the plugin is instantiated.
+As a parameter, it would accept a `GatekeeperPluginConfigurationContext` record.
+Initially, the context would contain an instance of the Fabric8 Kubernetes client and an instance of the Strimzi `PlatformFeatures` class describing the environment in which the operator is running.
+The `configure(...)` method would return `void`.
+
+There would also be two additional interfaces per assembly operator:
```
PaulRMellor: Is it clear what we mean by _assembly operator_? Should we give some context?

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+Initially, the context would contain an instance of the Fabric8 Kubernetes client and an instance of the Strimzi `PlatformFeatures` class describing the environment in which the operator is running.
+The `configure(...)` method would return `void`.
+
+There would also be two additional interfaces per assembly operator:
+* Mutating plugin interface named `GatekeeperMutating<Type>Plugin` (e.g. `GatekeeperMutatingKafkaPlugin` or `GatekeeperMutatingKafkaBridgePlugin`)
+* Validating plugin interface named `GatekeeperValidating<Type>Plugin` (e.g. `GatekeeperValidatingKafkaPlugin` or `GatekeeperValidatingKafkaBridgePlugin`)
+
+Plugins would exist for the following types:
+* `Kafka` (handles both `Kafka` and `KafkaNodePool` resources)
+* `KafkaConnect` (handles both `KafkaConnect` and `KafkaConnector` resources)
```
PaulRMellor, line 296: Would a plugin handling more than one resource introduce any ambiguities in the process?

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+Plugins would exist for the following types:
+* `Kafka` (handles both `Kafka` and `KafkaNodePool` resources)
+* `KafkaConnect` (handles both `KafkaConnect` and `KafkaConnector` resources)
+* `KafkaMirrorMaker2`
+* `KafkaBridge`
+* `KafkaRebalance`
+* `KafkaUser`
+* `KafkaTopic`
+
+At least initially, there would be no plugin for `StrimziPodSet` resources.
```
PaulRMellor: Why are we not including this first time around? Maybe mention?

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+For Kafka and Kafka Connect, it would return a record with the primary custom resource and a collection of the secondary custom resources.
+Only the updates to the `.status` section of these resources would take any effect.
+Validating plugins would return a _void_ completion stage.
+
+#### Loading and calling the plugins
+
+The same mechanism currently used for Pod Security Provider plugins would also be used for Gatekeeper plugins.
+The plugins would be loaded at operator startup through a `GatekeeperPluginFactory` singleton class.
+The factory will take the configured plugins and load them through the Java Service Loader mechanism based on the `GatekeeperPlugin` interface.
+When some configured plugins are missing, it will report the error and fail.
```
PaulRMellor:
```suggestion
If any configured plugins are missing, the operator will report an error and fail to start.
```

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+#### Internal use cases
+
+The motivation section already described several internal use cases where Gatekeeper plugins might allow us to simplify our code base and tests.
+
+Let's take labels as a detailed example.
+Strimzi could provide two different Gatekeeper plugins:
+* 3scale discovery labels plugin
+* Custom environment variable-based labels plugin
+
+The _3scale discovery labels plugin_ would be a mutating `KafkaBridge` Gatekeeper plugin.
```
PaulRMellor:
```suggestion
The _3scale discovery labels plugin_ would be a mutating Gatekeeper plugin for `KafkaBridge` resources.
```

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+* Security context (possibly replacing the Pod Security Providers in the future by injecting the desired security context into the template section directly)
+
+#### Other use cases
+
+There are various use cases that Strimzi might want to provide and that could be implemented as Gatekeeper plugins.
+For example:
+
+* **Security enforcement:** Enforce particular security requirements through Gatekeeper plugins.
+  Similar to Pod Security Providers managing Kubernetes security contexts, we could have additional plugins for other security-related concerns.
+  For example, they could enforce TLSv1.3 or quantum-safe cipher suites in Apache Kafka configuration.
```
PaulRMellor, line 468: Would this or similar enforcement require Kafka version awareness?

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+  This could be done in either a validating or mutating fashion.
+* **Pre/Post update hooks:** Execute custom tasks before or after the operator or operand is upgraded.
+  While we currently do not have any specific use case, such a feature might be useful in the future.
+* **Access Operator alternative:** A validating Gatekeeper plugin for `Kafka` and `KafkaUser` could be used to copy cluster or user information into different namespaces or clusters.
```
PaulRMellor, line 472: should this be an example of a mutating plugin?

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+  A Gatekeeper plugin could load an external username and inject it into the custom resource.
+* **Manual cluster ID configuration:** Configure the desired cluster ID in an annotation, and a Mutating Gatekeeper plugin will inject it into the status when the reconciliation starts.
+
+### Supportability
+
+_With great power comes great responsibility._
+Gatekeeper plugins provide a powerful abstraction.
+However, when used incorrectly, they could also break many things.
+For example, this could happen if a mutating plugin makes an invalid change to a custom resource,
+or if a plugin takes too long while making external calls.
```
PaulRMellor, line 519: Could this be mitigated through execution timeouts?

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+* What the plugin does and what libraries are used for it
+* What dependency updates and API changes occurred in a given Strimzi release.
+
+For example, if the plugin does some custom Kubernetes integration based on the Fabric8 library, it might need to be updated when the Fabric8 library updates in Strimzi.
+
+### Initial plugins
+
+The initial implementation would include the following plugins:
+* **Custom resource logging plugins:** used as an example of a validating plugin.
+  It would log the full YAML of the custom resource(s).
```
PaulRMellor, line 548: Any danger in what this can expose?

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,377 @@
...
+  It would be shipped as part of Strimzi, but would not be enabled by default.
+* **Labels from environment variables plugin:** would read labels from the existing environment variables and inject them into the custom resource.
+  This plugin will be enabled by default as a _default_ plugin and will replace the related Cluster Operator functionality.
```
PaulRMellor, line 551: Any upgrade risks during migration from the current implementation?

PaulRMellor (commented):
The motivation and intended benefits are clearly explained, especially around reducing reconciliation complexity and improving extensibility.

I think the implementation details could be expanded further for such a significant change. For example, sequence diagrams or example flows for validating and mutating plugins to show execution order, mutation behaviour, and success/failure handling.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: No ...
* It would be optional and disabled by default, mainly serving as example
* The custom resources should not contain anything super secret anyway
* They are already logged by the Kubernetes client at lower log levels. So the purpose here is more example rather than thinking it is somehow super useful.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: I don't think so. Obviously, there might always be some bugs. But I think these plugins should be pretty straight forward to implement.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: We could do that in theory. But it is hard to decide what would be the right timeout and how it might affect the possible ways this feature might be used. There are also other way how to _break things_. So while this could address one issue many would remain. So I think the trust that the plugin authors and users can be responsible is a better way to go.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: It would actually be validating plugin only ... it would just need the information about the resource and it would copy the information somewhere else. So it would never modify the custom resource it self.

Also, It is meant here mainly as an example of what kind of integration you could do through this mechanism. Replacing Access Operator with it might be considered in the future, but would require a dedicated proposal etc.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: I think these configurations are pretty stable in Kafka. But it could happen that they change in the future, in which case it might be relevant. It is hard to predict what some other plugin would need.

However, the plugins will have pretty much the same information about the Kafka version as the operator has. They can read it from the Kafka CR `.spec` or `.status` sections if it is present. Or determine the default version if it is not. So I think this is something the plugins could work with if needed.

141-Strimzi-Gatekeeper-plugin-system.md
scholzj: `StrimziPodSet` resources are internal resources only and are not exposed to Strimzi users.
They also serve a very simple purpose - to create the Pods.
And if needed, any modifications to the Pods can be easily done also through Kubernetes's own mechanisms.
However, this can be reconsidered in the future if we see some demand and use-cases for it.

I will add it to the proposal.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: I'm not sure I fully follow the question. Do you mean a plugin that support for example both Kafka/KafkaNodePool and KafkaBridge? Or do you mean the plugins that combine Kafka + KafkaNodePool and/or KafkaConnect + KafkaConnector?

----

The first one is just about making it easier to write and distribute the plugins. My general expectation is that there would be two types of plugins:
* Plugins doing something special for one particular operand
* Plugins handling all resources

I would not necessarily expect some exotic combinations. But in any case, if you have a plugin that should handle all resources, it is easier if you need to include one JAR and not 7 of them. Also, technically for configuring, loading, and invoking the plugins it is easier for us to have a single stack of plugins and just check which resource they implement then configure, discover, and maintain 7 independent plugin stacks.

----

As for the Kafka + KafkaNodePool and/or KafkaConnect + KafkaConnector handling, the idea is to handle the full operand. For Kafka, you need the node pools to have information abotu the number of nodes, their types, storage, resources, etc. That might be important for the tasks the plugins might have. Similarly for Connect and Connectors.

141-Strimzi-Gatekeeper-plugin-system.md
scholzj: I reworded it. The _assembly operator_ term should be known to Strimzi developers. But not to regular users. The way it is meant here is basically _operand type_.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: I improved this a bit in the proposal text and added diagram that might make it more obvious.

To be honest, I'm not sure if I have some super strong explanation for it. It just seems completely natural. I see the plugins as _wrappers_ with their _entry_ and _exit_ methods. So the reverse order helps to naturally wrap it. When you were the plugin closest to the reconciliation at entry, you will be the closest at exit. And the other way around.

141-Strimzi-Gatekeeper-plugin-system.md
scholzj: To be clear, there will be no general way how to skip any plugin. If the plugin authors think skipping the plugin can be useful, they could add support for it. I made it more clear in the proposal text.

So for example, if you write a plugin that enforces some security settings, it is up to the plugin to decide if you want to have it skippable or not. There will be no built-in way that will get Strimzi to skip it.

141-Strimzi-Gatekeeper-plugin-system.md
scholzj: Yes. An error raised by the plugin in its _entry_ method will be treated as any other reconciliation failure. The exact error will be added to the status section of the custom resource and reported in the operator logs. It would also trigger the _exit_ methods of all the Gatekeeper plugins. This applies to both unexpected errors due to some bug or issue as well as to the resource being rejected by a (validating) plugin.

I updated the text.

141-Strimzi-Gatekeeper-plugin-system.md
scholzj: I added some examples and bit more explanation.

----

The modifications done to the custom resource in the _entry_ method will be internal only and will be used only within the same reconciliation.
The updated resources would not be updated in the Kubernetes API and the changes done by the plugin would not be stored anywhere.
The changes will also not be visible to other applications through the Kubernetes API.
As a result, the plugins would need to apply the changes to the custom resource at every invocation.
_(This is similar to how virtual node pools were used during migration to node pools or how Connect and MirrorMaker2 custom resources were modified internally during migration from the `v1beta2` to `v1` APIs.)_

For example:
* User creates a `Kafka` CR with the following configuration:
  ```yaml
  kind: Kafka
  metadata:
    name: my-cluster
  spec:
    kafka:
      # ...
  ```
* A TLSv1.3 enforcer implemented as mutating plugin would inject into it TLSv1.3 configuration in its _entry_ method:
  ```yaml
  kind: Kafka
  metadata:
    name: my-cluster
  spec:
    kafka:
      config:
        ssl.protocol: TLSv1.3
        ssl.enabled.protocols: TLSv1.3
      # ...
  ```
* This modified resource would be used for the reconciliation.
  However, it would never be stored to the Kubernetes API in this form and other users using the Kubernetes API would never see it in this form.

Modifications done by the _exit_ methods to the `.status` section of the custom resource would be persisted to the Kubernetes API and visible to users and other applications.
However, Strimzi cleans up the `.status` section during the reconciliation.
So the plugins might need to update the status in the same way in every reconciliation as well.

For example:
* A `Kafka` CR was reconciled and is Ready:
  ```yaml
  kind: Kafka
  metadata:
    name: my-cluster
  spec:
    # ...
  status:
    conditions: 
      - lastTransitionTime: "2019-08-20T11:37:00.706Z"
        status: "True"
        type: Ready
  ```
* A Security helper implemented as mutating plugin would inject a warning condition in its _exit_ method:
  ```yaml
  kind: Kafka
  metadata:
    name: my-cluster
  spec:
    # ...
  status:
    conditions: 
      - lastTransitionTime: "2019-08-20T11:37:00.706Z"
        status: "True"
        type: Ready
      - lastTransitionTime: "2019-08-20T11:37:00.706Z"
        status: "True"
        type: Warning
        message: "Listener plain does not have TLS encryption enabled"
  ```
* This modified resource (its `.status` section) would be stored in the Kubernetes API.
  Users and other tools querying the Kubernetes API would see the warning message.

scholzj:
@PaulRMellor Thanks for the review. I tried to address your comments and also added some more diagrams and examples.

144-Strimzi-Gatekeeper-plugin-system.md
PaulRMellor: Right. I'm not sure which I meant! But your explanation makes sense.

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,497 @@
...
+
+#### Ordering
+
+The Gatekeeper plugins will be ordered.
+At reconciliation _entry_, they will always be called in the following order:
+* Custom plugins
+* Default plugins
+* Mandatory plugins
+
+Within each plugin group, the plugins will be invoked in the order specified by the user or by the Strimzi source code.
```
im-konge, line 233: So the order in which user specifies the plugins inside the `STRIMZI_GATEKEEPER_CUSTOM_PLUGINS` env variable will be taken and used in the execution, right? (sorry, just asking to be sure I understand it fully).

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,497 @@
...
+
+The first interface, `GatekeeperPlugin`, would define the plugins.
+It would contain a single method, `configure(...)`, which would be called once at startup when the plugin is instantiated.
+As a parameter, it would accept a `GatekeeperPluginConfigurationContext` record.
+Initially, the context would contain an instance of the Fabric8 Kubernetes client and an instance of the Strimzi `PlatformFeatures` class describing the environment in which the operator is running.
+The `configure(...)` method would return `void`.
+
+There would also be two additional interfaces per each supported operand type:
+* Mutating plugin interface named `GatekeeperMutating<OperandType>Plugin` (e.g. `GatekeeperMutatingKafkaPlugin` or `GatekeeperMutatingKafkaBridgePlugin`)
+* Validating plugin interface named `GatekeeperValidating<OperandType>Plugin` (e.g. `GatekeeperValidatingKafkaPlugin` or `GatekeeperValidatingKafkaBridgePlugin`)
```
im-konge, line 292: Maybe completely stupid question, but could this be more generic? So you don't have two classes per operand, but just two classes for handling all the stuff?
I mean, I think it's because of the handling of the things per each resource without need of some type conversion or something, but I'm curious if you thought about it.

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,497 @@
...
+
+* **Security enforcement:** Enforce particular security requirements through Gatekeeper plugins.
+  Similar to Pod Security Providers managing Kubernetes security contexts, we could have additional plugins for other security-related concerns.
+  For example, they could enforce TLSv1.3 or quantum-safe cipher suites in Apache Kafka configuration.
+  This could be done in either a validating or mutating fashion.
+* **Pre/Post update hooks:** Execute custom tasks before or after the operator or operand is upgraded.
+  While we currently do not have any specific use case, such a feature might be useful in the future.
+* **Access Operator alternative:** A validating Gatekeeper plugin for `Kafka` and `KafkaUser` could be used to copy cluster or user information into different namespaces or clusters.
+  The plugin would wait for the reconciliation to complete and in the _exit_ method it would distribute the credentials or cluster coordinates to places defined in annotation.
+  _(Included as an example only - the actual replacement of Access Operator by a Gatekeeper plugin would require a separate proposal.)_
```
im-konge, line 474: Should we start thinking about it now instead giving more effort to Access Operator?

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,497 @@
...
+* **3scale discovery labels plugin:** would inject the 3scale discovery labels into the `KafkaBridge` resource.
+  This plugin will be enabled by default as a _default_ plugin and will replace the related Cluster Operator functionality.
+
+The existing functionality these plugin would replace is relatively simple and moving it to the plugins should constitute a minimal risk.
+
+## Future plans
+
+Gatekeeper plugins would provide a powerful extension point for Strimzi.
+However, they would not solve all use cases.
+Additional extension points might be considered in the future, such as pluggable rollers for Kafka and Kafka Connect or pre- and post-rolling-update plugins for them.
```
im-konge, line 583: When I was reading through this, I was wondering if something similar can be done even for the Kafka Roller v2 - I mean we discussed in the past that it might be better to have it pluggable, with default roller being provided by us. Maybe this is the way to go there? WDYT?

im-konge (commented):
Thanks for the proposal, it LGTM. I have just few questions.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: I think having the roller pluggable or at least having some pluggable hooks like this would make sense. But TBH, I think we need to take things step by step and that is something to deal with next time we try to tackle the roller.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: I'm happy to think about it. But:
* It would be completely different model (I think that better, but who knows)
* This proposal has 2 reviews and 0 approvals - so it would be premature to think it replaces anything
* If we do that, it should likely be a separate proposal -> Different interest groups, different topic, managing the scope, etc.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: These would be just interfaces that you implement. So, for example:
* The `Custom resource logging plugin` I mention in the proposal would be a single class that implements **all** 7 validating interfaces 
* The `Labels from environment variables plugin` would be one class implementing 4 mutating interfaces
* The `3scale discovery labels plugin` would be one class that implements just the `GatekeeperMutatingKafkaBridgePlugin` interface

You could of course consider having just 2 or even one interface with 14 or 28 methods instead of 14 interfaces with 2 methods. But I think the granularity works well here and makes it more clear which plugin actually implements what.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: Yes, correct.

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,497 @@
...
+  Users and other tools querying the Kubernetes API would see the warning message.
+
+All modifications - regardless whether in the _entry_ or _exit_ methods - would of course also need to follow the Strimzi API.
+The plugins would not be able to add any new fields.
+
+Each plugin would be allowed to implement the _entry_ and _exit_ methods for one or more custom resources.
+That would allow us to write plugins that, for example, apply only to `Kafka` resources.
+Or plugins that want to gate all custom resources.
+
+
```
PaulRMellor: The Error handling title isn't showing properly with these blank lines

PaulRMellor (approved):
Thanks for addressing my comments.
The example flow and diagrams are a great help

144-Strimzi-Gatekeeper-plugin-system.md
im-konge: I know it sounds like it's straightforward, but maybe it would be worth mentioning in the sentence?

144-Strimzi-Gatekeeper-plugin-system.md
im-konge: yeah, that makes sense, thanks :)

144-Strimzi-Gatekeeper-plugin-system.md
im-konge: Yeah I know it is premature to think about it and something for different proposal etc., but I just wanted to mention it, as we already discussed to have the Access Operator work a bit differently with different model.
So maybe we should think about this now before we will plan for some more changes in the Access Operator repo - CC @katheris

144-Strimzi-Gatekeeper-plugin-system.md
im-konge: Makes sense, thanks.

im-konge (approved):
The proposal LGTM, thanks a lot.

Frawless (approved):

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,503 @@
...
+For example, if the plugin does some custom Kubernetes integration based on the Fabric8 library, it might need to be updated when the Fabric8 library updates in Strimzi.
+
+### Initial plugins
+
+The initial implementation would include the following plugins:
+* **Custom resource logging plugins:** used as an example of a validating plugin.
+  It would log the full YAML of the custom resource(s).
+  It would be shipped as part of Strimzi, but would not be enabled by default.
+* **Labels from environment variables plugin:** would read labels from the existing environment variables and inject them into the custom resource.
+  This plugin will be enabled by default as a _default_ plugin and will replace the related Cluster Operator functionality.
```
tinaselenge, line 551: Should some of these initial plugins such as this one be provided as mandatory plugins rather than default since they are replacing existing cluster operator functionality? Thinking about one of the examples you gave earlier with  `STRIMZI_CUSTOM_KAFKA_BRIDGE_LABELS`,  if user has it set,  they could accidentally disable the plugin and the labels are no longer injected as expected?

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,503 @@
...
+  A Gatekeeper plugin could load an external username and inject it into the custom resource.
+* **Manual cluster ID configuration:** Configure the desired cluster ID in an annotation, and a Mutating Gatekeeper plugin will inject it into the status when the reconciliation starts.
+
+### Supportability
+
+_With great power comes great responsibility._
+Gatekeeper plugins provide a powerful abstraction.
+However, when used incorrectly, they could also break many things.
+For example, this could happen if a mutating plugin makes an invalid change to a custom resource,
+or if a plugin takes too long while making external calls.
```
tinaselenge, line 519: I guess each plugin will take certain amount of time, adding to the reconciliation duration. Would that be something we should include in the doc too?

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,503 @@
...
+#### Example
+
+The following diagram shows the Strimzi reconciliation flow with the Gatekeeper plugins:
+
+1. When a custom resource is created or updated, Kubernetes triggers Strimzi operator through a watch and sends it the Custom Resource A (CR)
+2. Strimzi reconciliation starts with the invocation of the Gatekeeper plugins
+3. The CR (A) is passed to first plugin, which in this case is mutating plugin
+4. The plugin modifies the CR (A) and returns it back to Strimzi as CR (B)
+5. Strimzi takes the CR (B) and passes it to the next plugin
+6. In this case a validating plugin which returns only a success or an error after validating the CR (B)
```
tinaselenge, line 268: Could there be case of validating plugins that can be executed at the same time, without waiting for each? Would they always be executed sequentially?

scholzj: I would expect some validating plugins to be able to run in parallel. But they need to fit in between the mutating plugins (you might want to do some validations only after some mutation for example). Also, validating plugins might do things such as register with other systems or platforms. In that case, it might be desired to run the validating plugin only after other validating plugins (when you are sure the resource is otherwise valid).

So we would need to have some configurability to specify what should run in parallel and what not. And that would compromise the UX. On the other hand, if everything runs serially, then:
* The UX is much better
* I do not imagine anything where you would require the plugins to run in parallel. So you can still do the same things with it.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: These labeling plugins are not in any way required for Strimzi to work. In fact, they were added based on some Red Hat specific use-cases and it is not clear if anyone is using them upstream (or even downstream at Red Hat I guess). I personally think we should in the long term consider having them not enabled by default at all and possibly remove them completely. However, for the time being, I do not think there is anything wrong with giving the users an option to disable these if they want to. So this basically maintains the status quo by default while giving the users an option to disable if they want to for example safe some resources etc.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: Good point. I added it to the list of potential issues to document. Thanks.

144-Strimzi-Gatekeeper-plugin-system.md
tinaselenge: I see, I didn't know the history behind these functionalities so was wondering why they would change from non optional to optional. Thanks for explaining.

scholzj:
@tinaselenge Do you have anything more on this? Thanks.

tinaselenge (approved):
LGTM, thanks Jakub.

scholzj:
This has now 4 binding +1 votes. If there are no new votes or comments, I will close this proposal as approved on Friday EOB. (If you need more time, I'm happy to discuss it, but please let me know.)

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,505 @@
...
+7. The CR (B) is passed to the regular Strimzi reconciliaiton process which manages given operand accordingly
+8. At the end, when the core Strimzi reconciliation is completed, Strimzi prepares the new `.status` section and together with the rest of the custom resource invokes with it the plugins as CR (C)
+9. It is first passed to the plugin 2
+10. Validating plugin 2 does not change it and only returns success or failure
+11. Next it passes it to the plugin 1
+12. Mutating plugin modifies the `.status` section (e.g. add a warning condition) and return CR (D)
+13. Strimzi uses the `.status` section from CR (D) and updates it in the Kubernetes API server
+
+![Strimzi reconciliation flow](./images/141-Strimzi-Gatekeeper-plugins-system-invocation.svg)
```
ppatierno: Not sure about why the validation and mutating plugin are called in the reverse order on exit. Couldn't it be possible that you need to mutate the CR exiting from the Strimzi reconciliation and then validating it? What's exactly the use case or the rationale to do the other way around?

Also, while mutating + validating makes totally sense on enter, on the exit side could the reverse order break stuff? I mean, you are first validating and then mutating but maybe the mutation can break something (which won't be validated anymore).

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,505 @@
...
+
+#### Plugin interfaces
+
+The Gatekeeper plugins would be defined through a series of interfaces.
+
+The first interface, `GatekeeperPlugin`, would define the plugins.
+It would contain a single method, `configure(...)`, which would be called once at startup when the plugin is instantiated.
+As a parameter, it would accept a `GatekeeperPluginConfigurationContext` record.
+Initially, the context would contain an instance of the Fabric8 Kubernetes client and an instance of the Strimzi `PlatformFeatures` class describing the environment in which the operator is running.
+The `configure(...)` method would return `void`.
```
ppatierno, line 287: What happens if the configuration fails somehow and the `configure` method throws an exception? Is the cluster operator ignoring and proceeding or failing the reconciliation?

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,505 @@
...
+
+The _entry_ method would have two or three parameters:
+* `Gatekeeper<Type>EntryContext` would be a record containing additional information passed to the plugin.
+Initially, it would be an empty record without any fields.
+However, it might be used in the future.
+* The custom resource which the plugin should evaluate.
+* In case of Kafka and Kafka Connect plugins, a collection with the secondary resources (Kafka node pools or connectors).
+* For mutating plugins, the method would return a completion stage containing the updated custom resources.
+For Kafka and Kafka Connect, it would return a record with the primary custom resource and a collection of the secondary custom resources.
+Validating plugins would return a _void_ completion stage.
```
ppatierno, line 329: If it returns void, how a failed validation is returned? by throwing an exception?

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,505 @@
...
+  A Gatekeeper plugin could load an external username and inject it into the custom resource.
+* **Manual cluster ID configuration:** Configure the desired cluster ID in an annotation, and a Mutating Gatekeeper plugin will inject it into the status when the reconciliation starts.
+
+### Supportability
+
+_With great power comes great responsibility._
+Gatekeeper plugins provide a powerful abstraction.
+However, when used incorrectly, they could also break many things.
+For example, this could happen if a mutating plugin makes an invalid change to a custom resource,
+or if a plugin takes too long while making external calls.
```
ppatierno, line 519: I would assume that any enter/exit method call within the plugin will be done in a dedicated "blocking" Vert.x call and not on the event loop right?

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,505 @@
...
+It would also trigger the _exit_ methods of all the Gatekeeper plugins.
+This applies to both unexpected errors due to some bug or issue as well as to the resource being rejected by a (validating) plugin.
+
+### Implementation
+
+### Configuring Gatekeeper plugins
+
+There would be three different groups of Gatekeeper plugins:
+* Mandatory plugins would be hardcoded in the Strimzi source code and would not be configurable.
+  These might be the plugins that Strimzi requires to work.
```
ppatierno, line 189: I would just add a sentence saying that users can't disable these plugins.

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,505 @@
...
+While it means that we will have 14 different interfaces, it provides much better overview of what the plugin actually implements compared to having for example just 2 big interfaces with many methods.
+The plugin implementations can choose which and how many interfaces will be implemented in their plugin class.
+It also allows us to better log which plugins are actually called, because we will know what custom resources are supported by given plugin.
+And we will not be needed to make unnecessary calls to some default methods that would do nothing.
+
+Each interface would have two methods:
+* `<Type>Entry` (e.g. `kafkaEntry`) called at the beginning of the reconciliation
+* `<Type>Exit` (e.g. `kafkaExit`) called at the end of the reconciliation
+
+The interfaces would provide default implementations of these methods that _do nothing_.
```
ppatierno, line 319: I would provide an example, as Java snippet, of one of the interfaces to make it clearer the naming, the input/output parameters, and the default implementations.

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,505 @@
...
+The mandatory and default (when `STRIMZI_GATEKEEPER_DEFAULT_PLUGINS` is not set) plugins will be called in the order as specified by Strimzi.
+
+At reconciliation _exit_, they will always be called in the following order:
+* Mandatory plugins
+* Default plugins
+* Custom plugins
+
+Within each plugin group, the plugins will be invoked in the reverse order specified by the user or by the Strimzi source code.
+
+![Plugins wrapping the reconciliation](./images/141-Strimzi-Gatekeeper-plugins-system-onion-wrapping.svg)
```
ppatierno: Should we have some explanation about what happens if a plugin crashes/raises exception within the chain? I would assume Strimzi default/mandatory will work without issues :-P ... but if the custom plugin crashes at the beginning (on enter) or at the end (on exit), what's going to happen?

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: It is Gatekeeper ... so imagine it as doors or gates. When you walk in, you first pass through the building doors, then through the doors to the stairs, walk up and through the doors of your room walk in. And when you go out, the order is the opposite. First the doors of your room, down the stairs, the stairwell doors and last the doors of the building. Or imagine it as a Java wrapper classes ... you call the wrapper and that calls the real class. Whatever the real class produces is returned back to the wrapper who returns it to you.

Would there be some cases where this would not fit? Possibly. But the idea of the plugins wrapping around the core operator as layers seems most natural to me and would IMHO work best in most use-cases.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: Yes. But please keep in mind that the plugins themselves would not use Vert.x to make sure they are ready for when we remove it.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: It would failt the reocnciliation. Added to the proposal.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: Added.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: Yes. By throwing the exception. Added to the proposal.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: The operator will exit when the `configure(...)` method fails. Added to the proposal.

Given the possible nature of the plugins (security validation etc.), we cannot ignore the errors. However:
* I would expect errors in our plugin configuration should be worthy the crashloop as they would indicate some major environment issue
* Errors in custom plugins might be the same. Or they might be some stupid mistakes from the plugin authors. But that is a responsibility of the user who installs such a plugin.

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,535 @@
...
+
+The following diagram shows the Strimzi reconciliation flow with the Gatekeeper plugins:
+
+1. When a custom resource is created or updated, Kubernetes triggers Strimzi operator through a watch and sends it the Custom Resource A (CR)
+2. Strimzi reconciliation starts with the invocation of the Gatekeeper plugins
+3. The CR (A) is passed to first plugin, which in this case is mutating plugin
+4. The plugin modifies the CR (A) and returns it back to Strimzi as CR (B)
+5. Strimzi takes the CR (B) and passes it to the next plugin
+6. In this case a validating plugin which returns only a success or an error after validating the CR (B)
+7. The CR (B) is passed to the regular Strimzi reconciliaiton process which manages given operand accordingly
```
katheris, line 269: I'm unclear about where this slots into the existing reconciliation flow. For example when we reconcile Kafka the KafkaAssemblyOperator reconciles not only Kafka but also the CAs, Cruise Control, KafkaExporter etc, so would the plugins be called before any of those, or would the plugins be called just before the KafkaReconciler is invoked?

144-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,535 @@
...
+### Cross-version support
+
+The Gatekeeper plugins should be able to work across multiple Strimzi versions without updates.
+However, it will always, to some extent, depend on:
+* What the plugin does and what libraries are used for it
+* What dependency updates and API changes occurred in a given Strimzi release.
+
+For example, if the plugin does some custom Kubernetes integration based on the Fabric8 library, it might need to be updated when the Fabric8 library updates in Strimzi.
+
+### Initial plugins
```
katheris, line 544: Where would these plugins live? Within the cluster-operator repo or as separate repos. Also above you talked about cross-version support, for the Strimzi supplied plugins would we strictly version them for a specific operator version or version them separately?

katheris (commented):
I can definitely see the opportunities that this mechanism provides.

I had some clarifying questions to better understand how you see this fitting into the Kafka reconciliation logic, and how the versioning of mandatory and default plugins would work.

I think the supportability section highlights some useful mitigations for the risk this feature introduces. The initial plugin set focuses on simple cases where either the operand is augmented or validation is done so those would be straight forward to debug hopefully. However the proposal does mention more complex use cases and I can imagine people writing plugins that for example make modifications to the Kubernetes cluster or make calls to Kafka. In these cases I do worry whether this feature will result in lots of bugs being raised due to subtleties in the way things are handled in Strimzi. Some of the reconciliation logic in Strimzi (particularly around certificates and rolling) relies on a very specific order of events, sometimes over multiple reconciliations and it's not always clear which parts of the overall reconciliation are coupled. Perhaps this is something that could/should be improved in parallel with supporting this new feature, but while I am happy with the initial set of suggested plugins I would be very hesitant to encourage users to implement more complex plugins.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: It is about the custom resources. So it will be called at the start of the `Kafka` CR reconciliation and at the end of it. I.e. the entry method would be called before the CA reconciliation. I added it to the proposal. The reconcilers inside the `KafkaAssemblyOperator` are just internal implementation details that help to structure the code in a more convenient way. But this internal structure will not be exposed to the plugins.

144-Strimzi-Gatekeeper-plugin-system.md
scholzj: The initial ones would all live in the Strimzi Operators repo directly ad be released together with it. For the future ones, I think it is up to discussion how we do it. There are many technical alternatives that might suit best for different plugins.

I added a new section on this topic.

scholzj:
> I think the supportability section highlights some useful mitigations for the risk this feature introduces. The initial plugin set focuses on simple cases where either the operand is augmented or validation is done so those would be straight forward to debug hopefully. However the proposal does mention more complex use cases and I can imagine people writing plugins that for example make modifications to the Kubernetes cluster or make calls to Kafka. In these cases I do worry whether this feature will result in lots of bugs being raised due to subtleties in the way things are handled in Strimzi. Some of the reconciliation logic in Strimzi (particularly around certificates and rolling) relies on a very specific order of events, sometimes over multiple reconciliations and it's not always clear which parts of the overall reconciliation are coupled. Perhaps this is something that could/should be improved in parallel with supporting this new feature, but while I am happy with the initial set of suggested plugins I would be very hesitant to encourage users to implement more complex plugins.

@katheris To be honest, I do not think the plugins can cause any major issues to the certificate management. All they do is that they validate or modify the custom resource. I do not think they can in any special way interfere with the CA processes.

Obviously, user might use a plugin that is badly written and is failing every reconciliation right at the beginning for a year while the certificates expire. But that is clearly user's responsibility and problem. And that can actually happen already today (and we saw it happen) for example due to broken RBAC, missing resources (v1beta1 Ingress), etc. So I do not think this is entirely new.

As for the CA actions that take multiple reconciliations, they anyway need to be able to survive any errors and crashes. We dealt with it already in the past and hopefully made it resilient enough. If we have such issues, it is a problem anyway, regardless the plugins. But I do not think the plugins would interfere with this in any way.

katheris (approved):
Thanks @scholzj for adding those extra parts, they make sense to me and particularly for the Kafka reconciliation I think the proposal of running it before and after the Assembly operator, rather than before the KafkaReconciler addresses some of my concerns about the impact to the reconciliation so I'm happy to approve this.

scholzj:
Thanks @katheris. @ppatierno Do you have anything more on this? Thanks.

144-Strimzi-Gatekeeper-plugin-system.md
ppatierno: I understand you are trying to make a comparison with the concept of wrapper (or even stack) but I am still not sure it applies in this specific case of validation + mutation. I still think that if on exit you validate first, then the mutation can break something which is not validate anymore. 
I do agree to reverse the order for Custom plugins, Default plugins and Mandatory plugins instead.

scholzj: Sorry, I do not think I follow the last comment and whether you are fine with it or not.

Keep in mind that the order is not dependent on whether it is mutating or validating plugin. Users can mix them up in the configuration. (Also, TBH, I do not think the validating plugin would really validate anything in the exit method as there is not much to gain when you throw an error when the reconciliation is done. So the purpose of the exit method is more when you need to call some hook etc. but do not want to modify the status in any way).

ppatierno: > Sorry, I do not think I follow the last comment and whether you are fine with it or not.

I was just saying that reversing the order across custom, default and mandatory plugins as explained in the proposal makes sense to me versus reversing order on exit for validating and mutating.

> Keep in mind that the order is not dependent on whether it is mutating or validating plugin. Users can mix them up in the configuration.

This makes me wonder if I am missing something on the definition of validating|mutating within a Strimzi plugin. Can a custom plugin can be validating, mutating or both? This is what I have got from the picture, that it can be both. So given a custom plugin written by a user, on entry, validation happens before mutation but your proposal is about calling them in the inverse order on exit.

scholzj: The validating plugin is a subset of the mutating plugin.

The validating plugin:
* Gets a copy of the custom resource(s)
* Can validate it:
    * Approve it by doing nothing
    * Reject it by throwing an exception
* Can in theory interact with 3rd-party system. E.g. register it to your monitoring stack or into your management console, something like that

The mutating plugin can all of this. But in addition to that, it can also modify the custom resource(s).

So having a single plugin implement both makes no sense. If you want to do mutation and validation, you can just use mutating plugin. For example something like this:
* Mutating TLS 1.3 enforcer:
    * If the Kafka CR has TLS v1.2 or TLSv1 explicitly configured by the user, just fail the validation and the reconciliation
    * If it does not have anything configured and relies on the default, inject `TLSv1.3` into the configuration

So you do not necessarily need two plugins for that.

The main reason why I propose validating and mutating plugins as separate entities is the _trust_. If you use some _third party_ plugin, validating plugin has less potential to cause harm as all it can do is failing the validation. Mutating plugin can in theory for example disable authentication or authorization. So to deploy a mutating plugin, you should be sure it can be trusted. For validating, you should still be sure, but a little bit less.

Does this help as an explanation?

----

As for the ordering, the way I imagine it, there will be a single list of plugins mixing the mutating and validating plugins as the user configured them. So the ordering will always apply to all of them equally.

ppatierno: Sorry if I am dumb but I am trying to understand in more details how it's supposed to work.

> The validating plugin is a subset of the mutating plugin.

Can we make it clearer in the proposal.
I would also add the description you have here (about what validating and mutating plugins can do) to the proposal. It's more detailed than what is available in the proposal and it helped me more.

Let's come back to the ordering now :-)
So my understanding is that validating/mutating vs custom/default/mandatory plugins classifications are orthogonal. Am I right? So you can write a custom plugin which is a validating plugin (or a mutating plugin, or both but it could not make any sense). 
If my understanding is correct, let's consider the following example with the following plugins implementing corresponding validating and mutating interfaces:

Custom plugins: [Custom_MutatingPlugin1, Custom_ValidatingPlugin2]
Default plugins: [Default_ValidatingPlugin3]
Mandatory plugins: [Mandatory_MutatingPlugin4]

Following the order you describe about how custom/default/mandatory plugins are called, we should have ...

Entry order:
1. Custom_MutatingPlugin1
2. Custom_ValidatingPlugin2
3. Default_ValidatingPlugin3
4. Mandatory_MutatingPlugin4
5. Core reconciliation

Exit order (reversed):
1. Core reconciliation
2. Mandatory_MutatingPlugin4
3. Default_ValidatingPlugin3
4. Custom_ValidatingPlugin2
5. Custom_MutatingPlugin1

Notice that on exit, we have Custom_ValidatingPlugin2 running before Custom_MutatingPlugin1 and this is where my concern applies: the validating plugin runs before the mutating plugin on exit, so any mutations by Custom_MutatingPlugin1 won't be validated by Custom_ValidatingPlugin2.
Also, how the order is enforced across plugins of the same type (custom, default or mandatory)? At this point, does it really make sense thinking about an order between validating and mutating? I am just thinking aloud to grasp how it's going to work and why :-/

scholzj: > Can we make it clearer in the proposal.
I would also add the description you have here (about what validating and mutating plugins can do) to the proposal. It's more detailed than what is available in the proposal and it helped me more.

Ok, I tried to add it there.

> Notice that on exit, we have Custom_ValidatingPlugin2 running before Custom_MutatingPlugin1 and this is where my concern applies: the validating plugin runs before the mutating plugin on exit, so any mutations by Custom_MutatingPlugin1 won't be validated by Custom_ValidatingPlugin2.
Also, how the order is enforced across plugins of the same type (custom, default or mandatory)? At this point, does it really make sense thinking about an order between validating and mutating? I am just thinking aloud to grasp how it's going to work and why :-/

I think you are too fixated on the idea that the validating plugin validates the mutating plugin changes. That is not the case. The mutating plugins might modify the custom resource. So their order might matter. For example running a validating plugin which checks that every listener is secured before a mutating plugin injects an insecure listener in the next step would matter. But this is mainly aspect of the entry methods. The exit method of a mutating plugin modifies only the status and does not impact the reconciliation or the operand anymore. So I do not think that this is a problem.

ppatierno: > The exit method of a mutating plugin modifies only the status and does not impact the reconciliation or the operand anymore.

Ok this makes sense but then, should we try to enforce it through the method interface by passing the status class and not the entire model? So in the following example having `KafkaBridgeStatus` instead of `KafkaBridge`?


```java
default CompletionStage<KafkaBridge> KafkaBridgeExit(GatekeeperMitatingKafkaBridgeExitContext context, KafkaBridge kafkaBridge) {
        return CompletableFuture.completedFuture(kafkaBridge);
    }
```

scholzj: I think that in many cases you would need to have the whole resource as a context. For example to be able to pair the entry and exit calls. So I think the choice is to have `KafkaBridge` and `KafkaBridgeStatus` separately. Or just have a KafkaBridge as a mix of both. But I do not think you can have only the status, that might impact what you can do in the plugin.

Maybe we can use that? Provide the whole KafkaBridge with the _original_ status as one field and the new status as a separate field which could be modified in mutating plugins. E.g:

```java
    default CompletionStage<KafkaBridgeStatus> KafkaBridgeExit(GatekeeperMitatingKafkaBridgeExitContext context, KafkaBridge kafkaBridge, KafkaBridgeStatus newKafkaBridgeStatus) {
        return CompletableFuture.completedFuture(newKafkaBridgeStatus);
    }
```

WDYT?

ppatierno: That could make more sense. It's a pity we can't make the `KafkaBridge` instance immutable easily. Even just using `final` doesn't help because it would make the reference immutable but not the content. I think the best we can do is documenting and explaining that any changes to the `KafkaBridge` instance is ignored.

scholzj: Ok, I will update the proposal.

> It's a pity we can't make the KafkaBridge instance immutable easily. Even just using final doesn't help because it would make the reference immutable but not the content. I think the best we can do is documenting and explaining that any changes to the KafkaBridge instance is ignored.

So, we will be passing a copy of the KafkaBridge instance here (as well as to all the validating plugins) and throwing it away right after we cal the plugin. So while the object itself would not be really immutable (as in compiler rejecting any changes), if the plugin does any changes to it it will be thrown out anyway and not show up anywhere.

ppatierno: Ok, lgtm then. Looking forward at the changes.

scholzj: Actually, it is not so simple. It works easily enough for Bridge. But not for Connect or Kafka where we have the main resource as well as the secondary resources (Connectors and Node Pool). For these, we cannot just pass around the statuses alone because it would not be clear where do they belong to. So I think I need to think a bit more about this. 🤔

scholzj: Ok, thinking about it a bit more and digging a bit into the code. What if we:
* Keep the single entry methods as they are
* Split the exit methods for Kafka and Kafka Connect into separate exit methods for the main resource (Kafka / KafkaConnect) and secondaty resources (KafkaNodePool / KafkaConnector)
    * I think that better corresponds to the code where the statuses for these methods are updated separately. 
    * We could use the separate Status parameter as discussed earlier
    * I do not think it hurts the expected tasks the plugins would do. The secondary statuses will be passed through the plugin, but only the main resource status is called with the actual reconciliation end. I think the plugins can handle this whatever they do.

Any thoughts on this idea @ppatierno?

ppatierno: so if I got it right, the two exit methods will be called at different times during the reconciliation because status update on a secondary resource happen before while the status update on the main resource happens at the end of the reconciliation.
I think it's feasible and makes sense to me.

scholzj: Yes, you got it right. I will update the proposal.

scholzj: Ok, the changes should be now in the PR.

141-Strimzi-Gatekeeper-plugin-system.md
```diff
@@ -0,0 +1,612 @@
...
+This would fail the reconciliation and the exception (its error) will be used in the custom resource `.status` section.
+
+The _exit_ method would have three or four parameters:
+* `Gatekeeper<Type>ExitContext` would be a record containing additional information passed to the plugin.
+Initially, it would be an empty record without any fields.
+However, it might be used in the future.
+* The original custom resource which the plugin should evaluate including the original status section.
+* In case of Kafka and Kafka Connect, a collection with the secondary resources (Kafka node pools or connectors).
+* The new custom resource status section, which the plugin should evaluate.
+* In case of Kafka and Kafka Connect plugins, there will be a secondary _exit_ method for handling the sattus of the secondary resource.
```
ppatierno:
```suggestion
* In case of Kafka and Kafka Connect plugins, there will be a secondary _exit_ method for handling the status of the secondary resource.
```

ppatierno (approved):
LGTM. +1 binding. Thanks for the patience and the good discussion but now changes look good to me.

scholzj:
> LGTM. +1 binding. Thanks for the patience and the good discussion but now changes look good to me.

Thanks for the approval and the review.

scholzj:
This proposal has no 6 binding +1 votes. If there are no more comments, I will merge it tonight.

PR #211, opened 2026-03-30
Added proposal for auto-rebalance on imbalanced cluster feature in operator

ShubhamRwt:
This PR aims to introduce the self-healing feature in Strimzi. This proposal contains all the comments and suggestion left on the old proposal [](https://github.com/strimzi/proposals/pull/161). This proposal aim to utilize the auto-rebalancing feature of Strimzi to introduce the self healing.

Here is an example that you can try -> https://github.com/ShubhamRwt/auto-rebalance-on-imbalance-examples
The metrics and alerts are something which still need to be implemented in the image but the above image will give a gist around how the feature work. 

### Checklist

- [x] AI assistance was used to create this PR (see the [Strimzi AI policy](https://github.com/strimzi/governance/blob/main/AI_POLICY.md))
- [x] The proposal uses Markdown format with one sentence per line
- [ ] BEFORE MERGING: The next free sequence number was used for the proposal and all its assets (images, etc.)
- [ ] BEFORE MERGING: The proposal index in README.md was updated

PR #221, opened 2026-05-08
Proposal 142: Automatic detection and recovery of offline log directories

acgtun:
This proposal adds automatic detection and recovery of offline log directories to the Strimzi cluster operator.

**Key capabilities:**
- Detection via `describeLogDirs()` Admin API during reconciliation
- ISR-aware rolling restarts with existing `KafkaAvailability` safety checks
- Weaker availability check that excludes partitions on offline dirs (they're already degraded) — restricted to broker-only nodes to preserve KRaft quorum safety
- Configurable cooldown (default 30 min) to prevent infinite restart loops on permanent hardware failures
- Restart precedence over dynamic reconfiguration

**Related PR:** https://github.com/strimzi/strimzi-kafka-operator/pull/12726 (implementation)


acgtun:
covered by [proposal 211](https://github.com/strimzi/proposals/pull/211).

PR #223, opened 2026-05-09
Add support for SSL in MQTT Bridge

antonio-pedro99:
This is a proposal to implement support for TLS/SSL as requested by https://github.com/strimzi/strimzi-mqtt-bridge/issues/84


142-add-support-to-ssl-mqtt-bridge
```diff
@@ -0,0 +1,207 @@
...
+This will help to protect sensitive data and prevent unauthorized access.
+
+## Proposal
+
+Since the most ideal use case for the MQTT bridge is to connect IoT devices to Kafka, and some of these devices may be under constraints environments, I propose to implement a simple authentication mechanism by enabling secure MQTT connections(also known as MQTT over TLS) and using client certificates for authentication.
+This approach is widely supported by MQTT clients and provides a secure way to authenticate devices without the need for complex authentication protocols.
+
+Since our bridge is not a full MQTT broker, we will not implement a full authorization mechanism.
+Instead, we can implement a simple topic-based access control mechanism that allows devices to specify which topics they are allowed to publish or subscribe to.
+This can be achieved by using client certificates to identify devices and associating them with specific topic permissions.
```
tinaselenge: Will this be a part of a follow up proposal for the authorization? Because the rest of the proposal does not talk about associating the client certificate to topic permissions. If this is case, I think this sentence should be moved to line 187 where you mentioned that you will open a follow up proposal for authorization part, which I agree with.

tinaselenge (commented):
Thank you @antonio-pedro99 for the well written proposal. I don't have an experience with MQTT Bridge, therefore I can't offer in-depth feedback for this proposal. However, what has been proposed generally makes sense and sounds good to me but as I said, I can't say much on the actual implementation part that you explained. 

I've just left one small comment.

142-add-support-to-ssl-mqtt-bridge
```diff
@@ -0,0 +1,207 @@
...
+## Technical implementation
+
+### Configuration
+
+We will begin by adding new configuration options to the MQTT Bridge.
+These options will allow users to enable secure MQTT connections and specify the necessary SSL/TLS configurations.
+It will look something like this:
+
+```application.properties
+mqtt.server.tls.port=8883
```
katheris: What's the reason for adding an extra config here? In components like the Bridge we just change the default port when TLS is enabled, can we do the same here? Presumably the other port (1883 I think) won't be available once this feature is enabled?

142-add-support-to-ssl-mqtt-bridge
```diff
@@ -0,0 +1,207 @@
...
+mqtt.server.ssl.enabled=true
```
katheris: Small tweak suggested to match the naming we used in Bridge for consistency
```suggestion
mqtt.server.ssl.enable=true
```

142-add-support-to-ssl-mqtt-bridge
```diff
@@ -0,0 +1,207 @@
...
+mqtt.server.ssl.need-client-auth=true
+mqtt.server.ssl.trusted.certificates.location=path/to/client-ca.pem
+mqtt.server.ssl.trusted.certificates=--BEGIN CERTIFICATE--\n...\n--END CERTIFICATE--
+```
+
+Optionally, we can also support the configuration for specifying the protocols and cipher suites:
+
+```application.properties
+mqtt.server.ssl.protocols=TLSv1.2,TLSv1.3
+mqtt.server.ssl.ciphers=TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384
```
katheris: Suggested tweak to match naming elsewhere
```suggestion
mqtt.server.ssl.enabled.protocols=TLSv1.2,TLSv1.3
mqtt.server.ssl.enabled.ciphers=TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384
```

142-add-support-to-ssl-mqtt-bridge
```diff
@@ -0,0 +1,207 @@
...
+```
+
+### Handling authorization
+
+A very common way to implement authorization in MQTT is to make use of ACLs(Access Control Lists).
+Our MQTT Bridge is not a full MQTT broker and it's stateless, meaning that we do not persist any client session information, only the mapping rules. 
+
+I will submit a separate proposal to implement the authorization.
+This will make this proposal more focused and easier to review.
+I need a clear understanding ot the authorization mechanism alternatives before I can propose a specific implementation for the authorization support in the MQTT Bridge.If it's okay to the Strimzi maintainers, I will submit the authorization proposal after we have implemented the authentication support in this proposal.
```
katheris:
```suggestion
I need a clear understanding of the authorization mechanism alternatives before I can propose a specific implementation for the authorization support in the MQTT Bridge. If it's okay to the Strimzi maintainers, I will submit the authorization proposal after we have implemented the authentication support in this proposal.
```

142-add-support-to-ssl-mqtt-bridge
antonio-pedro99: > Will this be a part of a follow up proposal for the authorization?

yes, authorization is out of the scope for this proposal

katheris (commented):
Hey @antonio-pedro99 I took a quick look and have suggested some changes to better align with config we have elsewhere. However this proposal was discussed on the most recent community call and @ppatierno is going to take a look and see if he thinks this feature should be prioritised right now or not. So I've added the comments incase we decide to go ahead but there's no need to address these comments now as we can wait until Paolo has taken a look first.

im-konge:
I'm removing myself from the reviewers, as I don't have needed expertise for this and should be reviewed by @kyguy and @ppatierno mainly as experts.

ppatierno:
I had a quick look but I think the proposal is mixing authN, authZ and TLS support.
I think we should address them separately:

* enable TLS only for encrypting the connection between the client and the bridge
* having some authN mechanism that could be mTLS only at the beginning
* set up an authZ mechanism to authorize clients to access topics

Comparing what we have today within the HTTP bridge, only the TLS support would be "easy" to add and providing a minimum of security.
For the rest, I don't think they are priorities given the low demand we have for the MQTT bridge (we would have higher priority to add authN and authZ in the HTTP bridge instead). Also authN and authZ would need deep thoughts about all the possibilities to implement them.

katheris:
Hey @antonio-pedro99 based on Paolo's comments can you update the proposal to focus only on enabling TLS and remove the configuration that is specific to authentication and any other parts that are auth related?

antonio-pedro99:
> Hey @antonio-pedro99 based on Paolo's comments can you update the proposal to focus only on enabling TLS and remove the configuration that is specific to authentication and any other parts that are auth related?

Hi @katheris, yes, I will update it once I have some time.
Thanks for the reminder.

142-add-support-to-ssl-mqtt-bridge
antonio-pedro99: @katheris , it was my initial assumption: separate the configs. But makes sense to stick with whatever already exists in other components.

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,181 @@
...
+```application.properties
+mqtt.server.ssl.enabled.protocols=TLSv1.2,TLSv1.3
+mqtt.server.ssl.enabled.ciphers=TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384
+```
+
+> If these options are not specified, the MQTT Bridge will use the default SSL/TLS configurations provided by the Java runtime.
+
+After this, we are going to create a new configuration wrapper class to load and manage these new config options, say `MqttSslConfig`.
+This class will then be part of the existing `MqttConfig`.
+The SSL/TLS state will be derived from the configured MQTT server port rather than from a separate enable flag; for example, `MqttSslConfig.isEnabled()` can return true when the configured port is 8883.
```
kyguy, line 58: I understand that 8883 is the standardized port for MQTT over TLS but what happens when a user configures TLS on a another port? What if `mqtt.server.port=8883` but TLS is not configured? Would it be more reliable to use the the `mqtt.server.ssl.*` to check whether TLS is enabled?

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,181 @@
...
+
+Our `MqttServerInitializer` will then use this `MqttSslContextProvider` to create the SSL context and add the SSL handler to the pipeline.
+It would look something like this:
+
+```java
+ @Override
+    protected void initChannel(SocketChannel ch) {
+        if (sslConfig.isEnabled()) {
+            MqttSslContextProvider sslContextProvider = new MqttSslContextProvider(sslConfig);
+            SslContext sslContext = sslContextProvider.createSslContext();
```
kyguy, line 153: Do we need to call `createSslContext()` for every new client connection? Couldn't we call this once in `MqttServerInitializer` and reuse it?

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,181 @@
...
+We will not add a separate TLS port configuration.
+The MQTT Bridge will continue to use a single MQTT listener, and MQTT over TLS will use the existing MQTT server port configuration with the standard MQTT over TLS port, 8883.
+It will look something like this:
+
+```application.properties
+mqtt.server.port=8883
+mqtt.server.ssl.certificate.location=path/to/server-cert.pem
+mqtt.server.ssl.key.location=path/to/server-key.pem
+mqtt.server.ssl.certificate=--BEGIN CERTIFICATE--\n...\n--END CERTIFICATE--
+mqtt.server.ssl.key=--BEGIN PRIVATE KEY--\n...\n--END PRIVATE KEY--
```
kyguy, line 41: I imagine we would want some configurations for a truststore as well right? (assuming we want to support MTLS)

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,181 @@
...
+It does not change MQTT topic handling, Kafka mapping behavior, or MQTT client session semantics.
+
+## Technical implementation
+
+### Configuration
+
+We will begin by adding new configuration options to the MQTT Bridge.
+These options will allow users to enable secure MQTT connections and specify the necessary SSL/TLS configurations.
+We will not add a separate TLS port configuration.
+The MQTT Bridge will continue to use a single MQTT listener, and MQTT over TLS will use the existing MQTT server port configuration with the standard MQTT over TLS port, 8883.
```
kyguy, line 33: So the bridge would only allow either all connections over TLS or all plaintext, it won't support both TLS and non-TLS connections simultaneously, correct? If so, is that something maybe worth mentioning explicitly here? 

Is the purpose of doing this to simplify the iimplementation? If so, maybe it is a potential "future work" item we could add at the end of the proposal if it is not going to be supported initially. WDYT?

kyguy (commented):
Hi @antonio-pedro99 so sorry I am late to the party. Great start on the proposal, I think it is awesome you are driving this! I left a couple of comments to start. I'll help you flesh this out with periodic reviews if you would like to help get this over the line.

.DS_Store
scholzj, line 1: This should not be committed I guess.

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,181 @@
+# Add SSL/TLS support to MQTT Bridge
```
scholzj, line 1:
```suggestion
# Add SSL/TLS encryption support to MQTT Bridge
```

Maybe this way it would be a bit more clear what we focus on here?

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,181 @@
...
+
+> Note: All configuration options that support both inline and file-based values are mutually exclusive. 
+Inline values take precedence over file-based values.
+
+Optionally, we can also support the configuration for specifying the protocols and cipher suites:
+
+```application.properties
+mqtt.server.ssl.enabled.protocols=TLSv1.2,TLSv1.3
+mqtt.server.ssl.enabled.ciphers=TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384
+```
```
scholzj, line 52: You mean ... these configurations will be supported, but they will be optional only for the user I guess?

```suggestion
We will also support the configuration for specifying the enabled protocols and cipher suites.
These options will be optional for the users and when not set, the Netty / JRE default values will be used.

```application.properties
mqtt.server.ssl.enabled.protocols=TLSv1.2,TLSv1.3
mqtt.server.ssl.enabled.ciphers=TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384
```
```

File: 142-add-support-to-ssl-mqtt-bridge.md [hunk 12]
```diff
@@ -0,0 +1,181 @@
...
+
+> If these options are not specified, the MQTT Bridge will use the default SSL/TLS configurations provided by the Java runtime.
+
+After this, we are going to create a new configuration wrapper class to load and manage these new config options, say `MqttSslConfig`.
```
scholzj, line 56: If you mention things such as class names etc., it should be clear that that would be the name used. Or you should not mention the name at all and leave it for the PR. E.g.

```suggestion
A new class named `MqttSslConfig` will be used as a new configuration wrapper class to load and manage these new config options.
```

The way you fraze sounds like it is one of the many options, but that is not how the proposal should sound.

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,181 @@
...
+This class will then be part of the existing `MqttConfig`.
+The SSL/TLS state will be derived from the configured MQTT server port rather than from a separate enable flag; for example, `MqttSslConfig.isEnabled()` can return true when the configured port is 8883.
+It would look something like this:
```
scholzj, line 59: Again, the proposal needs to be exact in the things it covers. So it should not looks `something like this`.


```suggestion
It would look like this:
```

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,181 @@
...
+
+    // other methods...
+}
+```
+
+### Handling SSL/TLS in the server
+
+To handle SSL/TLS connections, we will need to add the Netty SSL handler to the pipeline of the server.
+This will require us to modify the existing `MqttServerInitializer` class to include the new handler.
+The final look of the `MqttServerInitializer` class will be something like this:
```
scholzj, line 85: Same as above ... leave out the `something`.

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,181 @@
...
+        ch.pipeline().addLast("encoder", MqttEncoder.INSTANCE);
+        ch.pipeline().addLast("handler", this.mqttServerHandler);
+    }
+}
+``` 
+
+Thankfully, Netty provides APIs to easily add SSL/TLS support, so we can leverage those APIs without having to implement the SSL/TLS logic from scratch. 
+
+We will introduce a new component to encapsulate the logic for creating the Netty `SslContext` before adding the SSL handler to the pipeline.
+This component will look something like this:
```
scholzj, line 118: Same as above ... I will not repeat this comment anymore. But please fix it in other places where you use this as well.

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,181 @@
...
+
+We will need to implement both unit and integration tests to verify the functionality of SSL/TLS support in the MQTT Bridge.
+
+We can make use of the Netty `EmbeddedChannel` to test the server pipeline initialization in isolation.
+This will allow us to verify that the SSL handler is added before the MQTT decoder when SSL/TLS is configured.
+Integration tests should verify that MQTT clients can connect successfully over TLS using the configured server certificate and key.
+
+## Affected/not affected projects
+
+This proposal will primarily affect the MQTT Bridge component of Strimzi.
```
scholzj, line 172:
```suggestion
This proposal will affect only the MQTT Bridge component of Strimzi.
```

I don't think it will affect anything else, or?

scholzj (commented):
I left some nits, mainly about the formal side of the proposal. You should make sure the proposal sounds _authoritative_ as it describes how it will be done and how it _might_ be done.

Apart from that, I'm not a subject matter expert for the MQTT Bridge. But the technical aspects seemed reasonable.

142-add-support-to-ssl-mqtt-bridge.md
scholzj: It is just an encryption. Truststore would be needed for mTLS authentication only. Or are there some internal connections between instances of the MQTT bridge etc.? I do not think so, or?

kyguy: > It is just an encryption.

Ah yes, you are right, I forgot this proposal is just for encryption. The mTLS can be addressed in the future proposal for authentication.

@antonio-pedro99 ignore this thread!

142-add-support-to-ssl-mqtt-bridge.md
ppatierno: I agree with Kyle we should try to follow as much as possible the same approach we have in the HTTP bridge so that people using the two will be comfortable. So yes, I think that a flag like `mqtt.server.ssl.enable=true` would be beneficial (as HTTP bridge has `http.ssl.enable=true`).

142-add-support-to-ssl-mqtt-bridge.md
ppatierno: Yes let's focus on encryption only as I requested. We'll move with authN and authZ (by also adding mTLS) in the future.

142-add-support-to-ssl-mqtt-bridge.md
ppatierno: I would stick with plan or encrypted but not supporting them at the same time. It's the same behavior as HTTP bridge and I would stick with consistency.

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+> Note: All configuration options that support both inline and file-based values are mutually exclusive. 
+Inline values take precedence over file-based values.
+
+Optionally, we can also support the configuration for specifying the protocols and cipher suites:
+
+```application.properties
+mqtt.server.ssl.enabled.protocols=TLSv1.2,TLSv1.3
+mqtt.server.ssl.enabled.ciphers=TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384
```
ppatierno, line 51:
```suggestion
mqtt.server.ssl.enabled.cipher.suites=TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384
```
I would follow the same parameter name pattern as the HTTP bridge.

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,181 @@
+# Add SSL/TLS support to MQTT Bridge
+
+This proposal is to add support for SSL/TLS to the MQTT Bridge component of Strimzi.
+This would allow MQTT clients to connect to the MQTT Bridge over an encrypted MQTT over TLS connection.
```
ppatierno, line 4:
```suggestion
This would allow MQTT clients to connect to the MQTT Bridge over an TLS encrypted connection.
```

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Configuration
+
+We will begin by adding new configuration options to the MQTT Bridge.
+These options will allow users to enable secure MQTT connections and specify the necessary SSL/TLS configurations.
+We will not add a separate TLS port configuration.
+The MQTT Bridge will continue to use a single MQTT listener, and MQTT over TLS will use the existing MQTT server port configuration with the standard MQTT over TLS port, 8883.
+It will look something like this:
+
+```application.properties
```
ppatierno, line 36: As already mentioned above we should have a dedicated `mqtt.server.ssl.enable` configuration property.

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Motivation
+
+The motivation for adding SSL/TLS support to the MQTT Bridge is to enhance the transport security of MQTT Bridge deployments.
+Since a common use case for the MQTT Bridge is to connect IoT devices to Kafka, MQTT traffic might pass through networks where encryption is required.
+Supporting MQTT over TLS allows users to protect data in transit without changing the MQTT protocol version or the bridge mapping behavior.
+
+## Proposal
+
+This proposal adds support for MQTT over TLS by allowing users to configure a server certificate and private key for the MQTT listener.
```
ppatierno, line 20: We should say explicitly that, like the HTTP bridge, it supports certificate and key in PEM format and not PKCS12 for example.

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,181 @@
...
+Inline values take precedence over file-based values.
+
+Optionally, we can also support the configuration for specifying the protocols and cipher suites:
+
+```application.properties
+mqtt.server.ssl.enabled.protocols=TLSv1.2,TLSv1.3
+mqtt.server.ssl.enabled.ciphers=TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384
+```
+
+> If these options are not specified, the MQTT Bridge will use the default SSL/TLS configurations provided by the Java runtime.
```
ppatierno, line 54: Let's do the same as the bridge which defaults to `TLSv1.2,TLSv1.3`

ppatierno (changes requested):
@antonio-pedro99 Thanks for the changes, I have added additional comments.

142-add-support-to-ssl-mqtt-bridge.md
antonio-pedro99: > I agree with Kyle we should try to follow as much as possible the same approach we have in the HTTP bridge so that people using the two will be comfortable. So yes, I think that a flag like `mqtt.server.ssl.enable=true` would be beneficial (as HTTP bridge has `http.ssl.enable=true`).

the initial proposal already included these, config keys, I guess there were a [comment](https://github.com/strimzi/proposals/pull/223#discussion_r3490647427) suggesting to match the config with the HTTP bridge, and I have changed the keys to do so. 
But yes, I agree with both of you.
 
@katheris does it makes sense?

142-add-support-to-ssl-mqtt-bridge.md
antonio-pedro99: Yes, this is an interesting take, this is a left over from the initial proposal where clients would also need to identify themselves using a cert, mTLS.

142-add-support-to-ssl-mqtt-bridge.md
antonio-pedro99: @katheris @ppatierno @kyguy my take is to leave the `mqtt.server.port=8883` but add others `mqtt.server.ssl.`, like `mqtt.server.ssl.enable=true`

142-add-support-to-ssl-mqtt-bridge.md
antonio-pedro99: maybe I did not get, but it was already mentioned

`mqtt.server.ssl.enabled.protocols=TLSv1.2,TLSv1.3`

ppatierno: The sentence says it defaults to the ones from Java runtime which is not true for the bridge which defines:

```java
/** Default SSL enabled protocols to be used when SSL is enabled */
public static final String DEFAULT_SSL_ENABLED_PROTOCOLS = "TLSv1.2,TLSv1.3";
```

So in the bridge we are using the above as default which may or may not match the default ones from Java runtime (depending on the runtime).

scholzj (approved):
Not really an expert on the MQTT Bridge. But this looks good to me. Thanks @antonio-pedro99.

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -114,49 +116,52 @@ public class MqttServerInitializer extends ChannelInitializer<SocketChannel> {
...
-    private KeyManagerFactory createKeyManagerFactory() {
+    private static KeyManagerFactory createKeyManagerFactory(MqttSslConfig sslConfig) {
         // logic to create KeyManagerFactory using the provided keystore configuration
     }
 }
 ```
 
-Our `MqttServerInitializer` will then use this `MqttSslContextProvider` to create the SSL context and add the SSL handler to the pipeline.
-It would look something like this:
+The server bootstrap will then load this `MqttSslContextProvider` before constructing the `MqttServerInitializer`.
```
ppatierno, line 154: Is this `MqttSslContextProvider` similar or something like the `SSLContextFactory` you are developing here https://github.com/strimzi/metrics-reporter/pull/130?

antonio-pedro99: Yes, but simplier since Netty already provides built-in support for SSL/TLS encryption.

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,188 @@
...
+mqtt.server.ssl.certificate=--BEGIN CERTIFICATE--\n...\n--END CERTIFICATE--
+mqtt.server.ssl.key=--BEGIN PRIVATE KEY--\n...\n--END PRIVATE KEY--
+```
+
+> Note: All SSL certificate and key should be in PEM format, it will not support format like PKCS12.
+> Note: All configuration options that support both inline and file-based values are mutually exclusive. 
+Inline values take precedence over file-based values.
+
+We will also support the configuration for specifying the enabled protocols and cipher suites.
+These options will be optional for the users and when not set, the Netty / JRE default values will be used.
```
ppatierno, line 49: I would remove this sentence because it's true for `mqtt.server.ssl.enabled.cipher.suites` but not for `mqtt.server.ssl.enabled.protocols` as you stated in the next few lines. I would just leave what we have on line 56 and 58.

142-add-support-to-ssl-mqtt-bridge.md
ppatierno: @antonio-pedro99 so this sentence "The SSL/TLS state will be derived from the configured MQTT server port rather than from a separate enable flag" is still wrong here. Should it be removed?

142-add-support-to-ssl-mqtt-bridge.md
```diff
@@ -0,0 +1,188 @@
...
+This proposal will affect only the MQTT Bridge component of Strimzi.
+
+## Compatibility
+
+We are currently using MQTT version 3.1.1.
+SSL/TLS is transport-level security and does not depend on MQTT protocol features, so this proposal should be compatible with both MQTT 3.1.1 and a future MQTT 5.0 upgrade.
+
+## Rejected alternatives
+
+Handling authentication and authorization is out of the scope for this proposal. They should be addressed separately.
```
ppatierno, line 188: This is not a "rejected alternatives" because it's not a different solution to implement SSL (so reaching the same goal). AuthN and AuthZ are just out of the scope, so no goals for this proposal.
I think you should just remove this section and maybe move this sentence somewhere in the proposal (in the "Proposal" section?).

ppatierno (changes requested):

142-add-support-to-ssl-mqtt-bridge.md
antonio-pedro99: Makes sense

antonio-pedro99:
> @antonio-pedro99 so this sentence "The SSL/TLS state will be derived from the configured MQTT server port rather than from a separate enable flag" is still wrong here. Should it be removed?

i do not think it's wrong, but it's not necessary to mention it since the config keys are enough to explain the behavior

ppatierno:
> i do not think it's wrong, but it's not necessary to mention it since the config keys are enough to explain the behavior

To me it sounds wrong because this is not how it's going to work. You are not deriving SSL enabled/disabled from the port but we agreed on a specific configuration parameter. The sentence says the exact opposite which is not the truth.

142-add-support-to-ssl-mqtt-bridge.md
ppatierno: You removed the line from here but didn't move where it could make more sense as suggested.
I would put the same sentence right after:

> This proposal is limited to SSL/TLS encryption for MQTT client connections.

In the "Proposal" section.

ppatierno (approved):
LGTM. Thanks!

tinaselenge (approved):
LGTM, thank you @antonio-pedro99 .

ppatierno:
with 3 +1 (binding) votes this proposal is now approved. I will wait until tomorrow EOD for additional feedback before merging.
@katheris let me know if you have something more to this. Thanks!

kyguy (approved):
Nice work @antonio-pedro99!

ppatierno:
@antonio-pedro99 can you update the README file with the next index assigned to the proposal so I can merge it? Thanks!

antonio-pedro99:
@ppatierno I haven't received feedback on updating the MQTT config prefix in this proposal, and I believe it's appropriate to do so.

ppatierno:
With 3 +1 (binding) and 1 +1 (non binding) votes this proposal is now approved. Thanks Antonio!

PR #224, opened 2026-05-17
Add proposal for alerting on stuck reconciliations

nikita-kibitkin:
This proposal suggests a solution for https://github.com/strimzi/strimzi-kafka-operator/issues/11634.

Introduces a per-resource gauge that exports the start time of the in-progress reconciliation, so users can alert on
reconciliations that started but never completed without scanning operator logs.

### DIFF
--- a/144-stuck-reconciliation-alerting.md
+++ b/144-stuck-reconciliation-alerting.md
@@ -7,8 +7,9 @@ It follows the discussion in [strimzi-kafka-operator#11634](https://github.com/s
 
 Strimzi operators already expose metrics about reconciliations, including counters for started, successful, failed, and locked reconciliations, and a timer for completed reconciliation duration.
 
-The operators also log progress while a reconciliation is running.
-For example, the Cluster Operator starts a periodic progress warning in `AbstractOperator.withLock(...)`, and the Topic and User Operator controller loops do the same in `AbstractControllerLoop.reconcileWrapper(...)`.
+The Cluster Operator and User Operator also log progress while a reconciliation is running.
+For example, the Cluster Operator starts a periodic progress warning in `AbstractOperator.withLock(...)`, and the User Operator does the same through `AbstractControllerLoop.reconcileWrapper(...)`.
+The Topic Operator uses a different batching loop, but it also has per-topic reconciliation timer boundaries in `BatchingTopicController`.
 These warnings are useful when somebody is already looking at the logs.
 
 But if a reconciliation gets stuck and never completes, it does not become a failed reconciliation.
@@ -20,17 +21,17 @@ This can also block important maintenance work such as certificate renewal.
 
 Users should be able to alert on reconciliations which have been running for too long without having to search logs for repeated `Reconciliation is in progress` messages.
 
-The proposed metric is intended to:
+This proposal is intended to give Prometheus rules enough information to:
 
 * Identify the affected resource by kind, namespace, and name.
-* Expose how long the reconciliation has been running.
-* Avoid false alerts after completed reconciliations, including deletion reconciliations.
-* Keep deployment-specific thresholds out of the operator itself.
+* Calculate how long the reconciliation has been running from the start timestamp.
+* Stop matching completed reconciliations, including completed deletion reconciliations.
+* Leave the alert threshold to Prometheus rules instead of adding a fixed operator timeout.
 * Work with Prometheus alerting rules.
 
 ## Proposal
 
-The operators will expose a new gauge metric for active reconciliations:
+The Strimzi Cluster Operator, Topic Operator, and User Operator will expose a new gauge metric for active reconciliations:
 
 ```text
 strimzi_reconciliations_in_progress_start_time_seconds{kind="Kafka",namespace="myproject",name="my-cluster"} 1778630400
@@ -48,8 +49,10 @@ strimzi.reconciliations.in.progress.start.time.seconds
 ### Metric lifecycle
 
 The metric exists only while a reconciliation is running.
-It is set after the reconciliation lock is acquired and removed from the local `MeterRegistry` when the reconciliation completes.
-If the operator fails to acquire the lock, the resource is not actively being reconciled by that execution path, so the metric is not created.
+It is set when the resource is accepted for reconciliation and removed from the local `MeterRegistry` when the reconciliation completes.
+For the Cluster Operator and User Operator, this means after the reconciliation lock is acquired.
+For the Topic Operator, this means after the topic is accepted into the in-flight batch.
+If the resource is not accepted for reconciliation, for example because the lock is not acquired or the topic is already in flight, the metric is not created.
 
 If the reconciliation gets stuck and the completion path is never reached, the metric remains exported with the original start timestamp.
 Prometheus can then alert based on elapsed time:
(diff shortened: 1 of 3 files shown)


144-stuck-reconciliation-alerting.md
```diff
@@ -0,0 +1,129 @@
...
+## Motivation
+
+Users should be able to alert on reconciliations which have been running for too long without having to search logs for repeated `Reconciliation is in progress` messages.
+
+The proposed metric is intended to:
+
+* Identify the affected resource by kind, namespace, and name.
+* Expose how long the reconciliation has been running.
+* Avoid false alerts after completed reconciliations, including deletion reconciliations.
+* Keep deployment-specific thresholds out of the operator itself.
```
scholzj: What does this mean?

144-stuck-reconciliation-alerting.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Motivation
+
+Users should be able to alert on reconciliations which have been running for too long without having to search logs for repeated `Reconciliation is in progress` messages.
+
+The proposed metric is intended to:
+
+* Identify the affected resource by kind, namespace, and name.
+* Expose how long the reconciliation has been running.
+* Avoid false alerts after completed reconciliations, including deletion reconciliations.
```
scholzj: What do you mean with deletion reconciliations? In some cases, we do not really have those.

144-stuck-reconciliation-alerting.md
```diff
@@ -0,0 +1,129 @@
...
+
+If the reconciliation gets stuck and the completion path is never reached, the metric remains exported with the original start timestamp. Prometheus can then alert based on elapsed time:
+
+```promql
+time() - strimzi_reconciliations_in_progress_start_time_seconds > 6 * 60 * 60
+```
+
+The operator will not define a built-in timeout. Different Strimzi deployments can have different reconciliation durations, so the threshold should be part of the user's Prometheus alerting rule.
+
+Removing the local meter also avoids tombstone values such as `-1`. Once Prometheus observes that the series is no longer exported, the series becomes stale and instant-vector alerts stop matching the old value. This is the same cleanup model already used by per-resource metrics such as `strimzi.resource.state`, where the operator removes local meters when they no longer apply.
```
scholzj: For someone who is not a Prometheus expert ... can you provide more details on when/how will Prometheus mark it stale? It matters here because it defines how reliable the alerting would be.

144-stuck-reconciliation-alerting.md
```diff
@@ -0,0 +1,129 @@
...
+
+* Identify the affected resource by kind, namespace, and name.
+* Expose how long the reconciliation has been running.
+* Avoid false alerts after completed reconciliations, including deletion reconciliations.
+* Keep deployment-specific thresholds out of the operator itself.
+* Work with Prometheus alerting rules.
+
+## Proposal
+
+The operators will expose a new gauge metric for active reconciliations:
```
scholzj: Maybe it would be good to be more specific here ... does `operators` here mean all 3 operators? Cluster / Topic / User? Or does it mean the internal operators inside the Cluster Operator (e.g. Kafka, KafkaConnect, ...)? You touch on it later. But it might be good to clarify it right at the beginning.

144-stuck-reconciliation-alerting.md
```diff
@@ -0,0 +1,129 @@
...
+
+* `kind`
+* `namespace`
+* `name`
+
+Both existing reconciliation paths already have the required identity and lifecycle hooks:
+
+* `Reconciliation` exposes `kind()`, `namespace()`, and `name()`.
+* In the Cluster Operator, `AbstractOperator.withLock(...)` starts the progress-warning timer after the lock is acquired and cancels it when the asynchronous reconciliation completes.
+* In the Topic and User Operator controller loop, `AbstractControllerLoop.reconcileWrapper(...)` starts the progress-warning task before calling `reconcile(...)` and cancels it in the `finally` block.
```
scholzj: Do these have access to the metrics objects?

scholzj (commented):
Thanks for the proposal and sorry it took so long to review. I left some clarification comments, but I think it looks mostly good otherwise.

ppatierno:
Overall it looks ok. I will wait for answers to the questions asked by Jakub. Also, could you please split the paragraph with one sentence per line please? It helps on reviewing to make comments on every specific line. Thanks!

144-stuck-reconciliation-alerting.md
nikita-kibitkin: Reworded this section. The bullets now describe what information the metric gives to Prometheus rules: identify the resource, calculate elapsed time from the start timestamp, stop matching completed reconciliations after cleanup, and keep the threshold in the alert rule instead of baking a timeout into the operator.

144-stuck-reconciliation-alerting.md
nikita-kibitkin: Expanded the lifecycle section. The start time is only the gauge value, not an explicit Prometheus sample timestamp. After the local meter is removed, the next successful scrape of the same endpoint no longer returns the series, Prometheus marks it stale, and instant-vector selectors stop returning it after that stale marker.

144-stuck-reconciliation-alerting.md
nikita-kibitkin: Made this explicit in the proposal section: this means the Cluster Operator, Topic Operator, and User Operator. I also clarified below that the Topic Operator uses a different path (`BatchingTopicController`) from the Cluster/User Operator paths.

144-stuck-reconciliation-alerting.md
nikita-kibitkin: Yes, clarified in the implementation section. `AbstractOperator` has an `OperatorMetricsHolder` through `metrics()`, `AbstractControllerLoop` exposes a `ControllerMetricsHolder`, and `BatchingTopicController` already has a `TopicOperatorMetricsHolder` field. The helper can therefore live on the existing metrics holders instead of passing a new metrics dependency into the wrappers.

nikita-kibitkin:
Thanks for the review.

I pushed two commits:
* a mechanical reflow commit applying one sentence per line;
* a clarification commit addressing the review comments.

The proposal now clarifies the Prometheus rule goals, the Cluster / Topic / User Operator scope, Prometheus staleness behavior and where the implementation can access the existing metrics holders.

scholzj (approved):
Thanks for the updates. This looks good to me. +1 binding.

145-stuck-reconciliation-alerting.md
```diff
@@ -0,0 +1,183 @@
...
+If the resource is not accepted for reconciliation, no in-progress metric should be created because that execution path is not actively reconciling the resource.
+
+The implementation can reuse the existing `removeMetric(...)` pattern used by other per-resource metrics.
+This is important for deleted resources because the desired terminal state is no exported series, not a special value.
+
+Tests should cover the main lifecycle cases:
+
+* The metric is created while a reconciliation is running.
+* The metric is removed after success or failure.
+* The metric is not created when the resource is not accepted for reconciliation.
```
ppatierno, line 127: I would add also "when it's skipped". We can think of pause reconciliation applied to a Kafka custom resource for example.

ppatierno (approved):
LGTM. +1 (binding). I just left a nit.

145-stuck-reconciliation-alerting.md
nikita-kibitkin: Added coverage for skipped reconciliations in the lifecycle, implementation and tests sections.

im-konge (approved):
LGTM, thanks.

im-konge:
This proposal now has 3 +1 binding votes, if there are no comments until Friday 19th June EOB, I'm going to close this proposal as approved.

im-konge:
Sorry, I just totally forgot about merging this one 🤦 I will merge it today. Thanks

im-konge:
Proposal approved with 3 +1 binding votes. Thanks for the proposal @nikita-kibitkin

PR #227, opened 2026-06-10
Configurable Security for Internal Kafka Cluster Communication

scholzj:
### Type of Change

- New Proposal

### Description

This proposal introduces _Configurable Security for Internal Kafka Cluster Communication_. It allows users to disable TLS encryption on all internal communication and switch from mTLS to Service Account based authentication.

A prototype (not configurable, but with disabled TLS and SA-based auth) can be found here: https://github.com/strimzi/strimzi-kafka-operator/compare/main...scholzj:istio-support?expand=1

### Checklist


scholzj:
@strimzi/maintainers Please have a looooook 🙏 🙏 🙏 🙏 🙏

scholzj:
@strimzi/maintainers It would be great if you can find some time to review this proposal. We need to be able to maintain the project velocity and show the ability to add new features if we want to move towards graduation.

Frawless (approved):

145-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,405 @@
...
+#### Cruise Control Authentication
+
+Cruise Control will use the Service Account tokens when connecting to the Kafka cluster to collect its metrics.
+However, it uses its own API key-based authentication for its HTTP interface.
+This will remain unchanged for the time being, and the new cluster configuration will only be used to enable or disable TLS encryption.
+
+In the long term, we should aim to add support for Service Account-based authentication to the Cruise Control HTTP interface as well.
+That would also help improve some of the limitations of the current authentication model, such as lack of credential rotation.
+However, the Cruise Control community currently seems dysfunctional.
+This should be revisited later depending on the future of Cruise Control.
```
im-konge: Should we instead of providing the option to disable or enable the TLS encryption, have it enabled and keep it enabled for the time being, and only when it could be improved on CC side, we can implement the same thing here? I mean, maybe it's completely stupid question, but I'm thinking about the case when the behavior can be weird for someone in case that for most parts we will use SA Auth and CC will not use it.
Also, how it would be disabled? In case that the annotation will not contain the `strimzi-tls` and `strimzi-mtls`?

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,405 @@
...
+                                  "type": "strimzi-tls"
+                              },
+                          "authentication":
+                              {
+                                  "type": "strimzi-mtls"
+                              }
+                      }
+  spec:
+    # ...
+```
```
im-konge, line 263: I suppose that missing annotation would mean that TLS and mTLS is used, right? So the default as we have today - to not break anything.

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,405 @@
...
+Therefore, we will have built-in protection to prevent that.
+The current security configuration will be tracked in the `Kafka` CR status.
+At the start of every reconciliation, Strimzi will compare the current and desired configuration.
+If the desired configuration contains invalid changes, Strimzi will reject it and continue using the old configuration.
+This is similar to the existing storage configuration protection mechanism.
+
+When the current status is not present, the operator will accept the desired configuration and use it.
+This should happen only for:
+* Newly deployed clusters
+* When the user executes the configuration change procedure (described in the previous section)
```
im-konge, line 340: Will there be some warning to the users that they are trying to configure something which is forbidden for running cluster - when they just set the annotation, they don't pause the reconciliation and all those steps you mentioned above?

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,405 @@
...
+It will be instantiated at the beginning of the Kafka reconciliation in the `KafkaAssemblyOperator`.
+It will be passed through the various reconcilers and model classes.
+The `KafkaClusterSecurity` object will hold the information about the security configuration.
+It will be used to decide how to configure the different security aspects, such as client or broker configuration.
+
+### Testing Strategy
+
+This feature will be covered by system tests in several different ways:
+
+* A basic test covering the transition between configurations:
```
im-konge, line 396: I think it would be also good to have tests covering the scenarios when users will change the configuration without doing the steps they need.
Given that the test-case you already mentioned will bring it back to mTLS there will be no need to deploy another Kafka cluster - so it will not take much time I guess. But that is just an idea.

Also as prat of the checks, is there some possibility to ensure we are really using something else as auth? Or it would be complicated to test on ST level?

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,405 @@
...
+    * Deploy a regular cluster with TLS+mTLS and check that it works
+    * Stop the cluster
+    * Reconfigure it to no encryption and Service Account authentication
+    * Check that the cluster works
+    * Stop the cluster
+    * Reconfigure it back to TLS+mTLS
+    * Check that the cluster works
+* Upgrade tests using no encryption and Service Account authentication
+* Additionally, we would add a way to run all tests with no encryption and Service Account authentication
+    * Either as a new pipeline
```
im-konge, line 407: It depends on how often we are planning to release. If you want the most easiest way, we can have just new pipeline which will trigger running the tests based on some env var, then the automation in the tests would just add the annotation to the Kafka cluster. 

Or if you expect to run it just for releases, we can have it as argument.

Actually, I was thinking about having some "nightly" pipelines using GHA - to move this from our Jenkins instance - and we can have this "new pipeline" running there.

im-konge (commented):
Thanks for the proposal, I left some comments - but I'm not expert in this area. So maybe they will sound like ... the answer is completely obvious :)

150-configurable-security-of-internal-communication.md
scholzj: I think it is pretty similar to the feature gates pipeline. You probably don't need to run it for every PR. But you do not want to run it only when releasing either.

145-configurable-security-of-internal-communication.md
scholzj: Sorry, not sure I follow the comment. It will not be weirder than today when everything used mTLS but Cruise Control HTTP interface uses its own API keys. So I do not think this changes anything in that respect.

150-configurable-security-of-internal-communication.md
scholzj: Yes, that is covered on the line 206:
```
When the annotation is not set, Strimzi will default to the current state with TLS encryption and mTLS authentication.
```

150-configurable-security-of-internal-communication.md
scholzj: I added it there. I think it can be pretty easily checked by unit tests, but we can have an ST for it as well.

150-configurable-security-of-internal-communication.md
scholzj: I updated this to make it fail as I think that provides a better protection and less undefined situations when users would try to edit the status and so on.

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,405 @@
...
+
+### Configuring the Internal Cluster Security
+
+With the ongoing work on cert-manager support, we currently have multiple parallel streams of work related to cluster security that might lead to API changes.
+This proposal therefore suggests using only an annotation for now to configure cluster security.
+Using an annotation means:
+* We are not binding to a specific API layout just yet by adding it to the custom resource spec.
+* We indicate that this is an _early access_ feature.
+
+Once both the work on this proposal and the cert-manager integration are complete, we can try to come up with a permanent API.
```
tinaselenge, line 225: SEP-100 (cert-manager integration) has API changes already defined and approved, so why can't there be a proper API design?

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,405 @@
...
+                      {
+                          "encryption":
+                              {
+                                  "type": "strimzi-tls"
+                              },
+                          "authentication":
+                              {
+                                  "type": "strimzi-mtls"
+                              }
+                      }
```
tinaselenge, line 260: Has Strimzi ever used annotations for feature configuration like this before? because it seems like an unusual pattern to me.

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,405 @@
...
+### Protecting the Running Clusters
+
+Changing the security configuration for an existing cluster could easily break the cluster.
+Therefore, we will have built-in protection to prevent that.
+The current security configuration will be tracked in the `Kafka` CR status.
+At the start of every reconciliation, Strimzi will compare the current and desired configuration.
+If the desired configuration contains invalid changes, Strimzi will reject it and continue using the old configuration.
+This is similar to the existing storage configuration protection mechanism.
+
+When the current status is not present, the operator will accept the desired configuration and use it.
```
tinaselenge, line 337: Should we add a more safety checks like strimzi.io/pause-reconciliation: true is set and no broker Kafka pods exist if user is changing this configuration?

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,405 @@
...
+These tokens can be issued in two different ways:
+* Through [projected volumes](https://kubernetes.io/docs/concepts/storage/projected-volumes/#serviceaccounttoken), which allow us to mount the token as a file into a Pod.
+  This mechanism will be used to project the tokens into the Pods that belong to the Kafka operand (Kafka nodes, Entity Operator, Cruise Control, Kafka Exporter).
+  These components will use their existing Service Accounts.
+* Using the [`TokenRequest` Kubernetes API](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/).
+  This will be used by the Cluster Operator.
+  While the Cluster Operator has its own Service Account for which the projected volume can be used, this Service Account is bound to the Cluster Operator namespace.
+  Its identity would change when you deploy a new Cluster Operator into another namespace (for example, as part of a phased upgrade).
+  The identity would also be shared between all Kafka operands managed by this Cluster Operator.
+  Instead, we will create a new Service Account for each Kafka cluster.
```
tinaselenge, line 121: Does this mean, we are adding runtime dependency of Kubernetes API? When not using SA token, if Kubernetes API goes down, while broker pods are already running, they would continue to run until they are restarted. Whereas when using SA token, in this case, broker pods will stop working as soon as tokens expire right?

tinaselenge (commented):
Thank you for the proposal. I've a few questions in the comments, otherwise mostly looks good to me.

150-configurable-security-of-internal-communication.md
scholzj: In a way yes. Short-lived credentials are more secure but also more sensitive to any outages. I added two paragraphs describing this risk.

That said, I do not think a long API server outage without any worker node outages is very likely. So I think that in most cases such events will impact the mTLS clusters as well.

150-configurable-security-of-internal-communication.md
scholzj: I added paragraph about this as well. But essentially:
* We cannot check that the paused annotation is there. When it is paused, the operator does not do anything. It will do something only once it is removed. So it cannot be checked.
* Checking for Pods not being present creates many edge situations such as when the operator fails in the middle of the reconciliation etc.

So I think checking the status is sufficient. It clearly indicates a new cluster or user intent. And when the user intentionally does this, they should be able to take over responsibility for the Pods being shutdown etc.

150-configurable-security-of-internal-communication.md
scholzj: Not exactly like this. But we had many similar annotations. For example to enable/disable KRaft and Kafka Node Pool. Those were only booleans. We have also the Node Pool ID assignment annotations which are JSON Arrays. The Connector Offsets in Connect and MM2 also use annotations with complex values rather then just booleans.

But I do not think we had a JSON object like this. The Gatekeeper proposal expects something similar for the Gatekeeper plugin configuration (https://github.com/strimzi/proposals/blob/main/144-Strimzi-Gatekeeper-plugin-system.md#per-resource-configuration). However, that would never move later to the Strimzi API. That will remain an annotation forever.

150-configurable-security-of-internal-communication.md
scholzj: SEP-100 happened almost year and a half ago. While I raised the need to do this for several years, I did not had any detailed ideas how to so it back then. And I did not know that I will have motivation and time to work on this in during summer of 2026. So yes, SEP-100 approved the API. That said, we have this proposal now and assuming it would be approved in some shape and form, I'm ready to implement it.

So I think it makes absolute sense to rethink the `.spec.clusterCa` and `.spec.clientsCa` API as well. And have for example some _umbrella_ API section for the cluster-wide security things.

However, I'm myself not 100% sure how exactly should such an API look like. And if I came up with it right now, it would have major impact on the cert-manager work as well. I also do not know what future requirements might arise if/when we have this proposal implemented. So using the annotation:
* Buys us time to come up with the proper _cluster-wide security umbrella API_
* Buys us time to finish the cert-manager implementation without a new disruption into it
* Allows us to make changes to what this proposal suggests without breaking API changes, leaving us with APIs deprecated after 2 releases etc. It allows us to be a bit more agile.

----

I think thinking about that makes sense also for you and Kate as part of the cert-manager implementation. That said, the `.spec.clusterCa` and `.spec.clientsCa` APIs exist. And while I think they are not the long term future, we will anyway need to deprecated them and replace them. So if you add the stuff from SEP-100 to them does not does as much damage as if I create some brand new API for this proposal and we have to rework it completely a month later.

----

So I think using the annotation for the time being gives us the flexibility we need right now. One other alternative I considered was something used for example by the Gateway API where they have _stable_ and _experimental_ CRDs with different APIs. And maybe one day I would propose it independently on this proposal. But that solution would require a lot more effort then using an annotation for _few months_ before we figure out and stabilize the API.

scholzj:
Thanks for the reviews @tinaselenge @im-konge. Your comments should be either incorporated or replied to. Please have a look.

scholzj:
And thanks for your approval as well @Frawless

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,425 @@
...
+Instead of managing its own certificates, Strimzi would rely on cert-manager (or, in the future, possibly other certificate managers) to provide the TLS certificates.
+This direction has been approved in the [SEP-100 proposal](https://github.com/strimzi/proposals/blob/main/100-external-certificate-manager.md), and the implementation is currently in progress.
+
+## Motivation
+
+While cert-manager support would be a major step forward, it does not change the fact that TLS encryption and mTLS authentication will still be hardcoded and required for any Strimzi-based Kafka cluster.
+There are various reasons why hardcoded TLS encryption and mTLS authentication are not optimal.
+Strimzi users seem to request this for two main reasons.
+
+The first reason is performance and cost.
```
ppatierno, line 52: What's the cost reason here?

145-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,425 @@
...
+
+This proposal focuses on two tasks:
+* Allowing users to disable TLS encryption for internal communication
+* Providing an alternative authentication mechanism
+
+The proposed alternative authentication mechanism is based on Kubernetes Service Accounts and their tokens.
+The Service Account tokens can be used in an OAuth-like fashion and are supported by Apache Kafka's `OAUTHBEARER` SASL mechanism.
+They work on most Kubernetes clusters without requiring additional dependencies such as Keycloak.
+
+While the main purpose of Service Account-based authentication is to provide an alternative authentication mechanism when TLS encryption is disabled, users can also use it with TLS encryption enabled.
```
ppatierno: So does it mean that the proposal allows to keep TLS encryption but disable mTLS? Did I lost it across the proposal or it should be clarified?

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,425 @@
...
+This annotation would contain JSON with the configuration.
+When the annotation is not set, Strimzi will default to the current state with TLS encryption and mTLS authentication.
+
+It will initially contain two sections:
+* `encryption`
+* `authentication`
+
+Both sections will have a `type` field.
+Both will support the `none` type to indicate no encryption or authentication.
+The `encryption` section will have an additional type `strimzi-tls`.
```
ppatierno, line 241: Any additional future values you see here?

145-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,425 @@
...
+Changing the security configuration for an existing cluster could easily break the cluster.
+Therefore, we will have built-in protection to prevent that.
+The current security configuration will be tracked in the `Kafka` CR status.
+At the start of every reconciliation, Strimzi will compare the current and desired configuration.
+If the desired configuration contains invalid changes, Strimzi will reject it, and fail the reconciliation without applying any changes to the actual Kafka cluster.
+The user will be required to either revert the changes or to follow the correct procedure.
+
+When the current status is not present, the operator will accept the desired configuration and use it.
+This should happen only for:
+* Newly deployed clusters
```
ppatierno: How are you detecting newly created cluster? I remember that in the past having such distinction was not accepted because of the idempotency the operator should have.

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,425 @@
...
+In reality, we expect changes to the internal communication security configuration to be a very rare event.
+In most cases, users should decide at the beginning what kind of security they want to use depending on their environment.
+Instead of providing a complex process for changing the configuration while maintaining cluster availability, we will require cluster downtime.
+
+To change the cluster security configuration, users will use the following procedure:
+* Pause the reconciliation of the `Kafka` CR
+* Stop the cluster (delete the `StrimziPodSets`, `Pods`, and `Deployments` with `kubectl` or a similar tool)
+* Update the desired security configuration in the `Kafka` CR annotation (later in the `.spec` section)
+* Edit the `.status` section of the `Kafka` CR and remove the current security configuration from it (see next section for more details)
+* Unpause the reconciliation of the `Kafka` CR
```
ppatierno, line 328: This procedure seems fragile or similar to what users do with clusterId recovery which were considered to be banned as a bad practice.

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,425 @@
...
+Currently, there is no plan to cover things such as service mesh integration in the system tests.
+
+### Documentation
+
+This feature will initially be documented as _early access_.
+
+### Existing Limitations
+
+Kafka Exporter currently does not support reading the tokens from a file.
+A PR will be opened to add this feature.
```
ppatierno, line 421: Do you mean that one of the goal of this proposal is also fixing the missing thing in the Kafka Exporter project?

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,425 @@
...
+
+However, that means that Kafka clusters with disabled TLS would still depend on the CA management.
+Certificate or CA renewals would cause unnecessary rolling updates and might be a source of bugs.
+We should also aim to allow users to disable the Cluster CA completely.
+However, to keep the scope of this proposal under control, this is not part of this proposal and should be the subject of a future follow-up proposal.
+
+## Out of Scope
+
+This proposal focuses only on the Apache Kafka clusters managed through the `Kafka` and `KafkaNodePool` custom resources.
+While it might serve as an inspiration for securing Connect or MirrorMaker 2 clusters in the future, they are out of scope for this proposal.
```
ppatierno, line 469: So it means that if users want to disable TLS encryption and enable service account authN for the Kafka cluster, the Strimzi operator will continue to generate CA certs for enabling TLS and mTLS on Connect and MM2? Why doesn't the proposal cover them as well?

scholzj: The Strimzi CAs are covered in the `Future Work -> Strimzi Cluster CA` section. As explained there, the main reason is to keep the scope under control. The Clients CA has actually no relationship to this proposal. The Cluster CA is used not only for the internal cluster communication, but also as the default server certificate for regular user listeners that are not directly tied to this proposal. So the ability to disable the Cluster CA would increase the scope of this proposal. It would also increase the implementation code conflicts of this work and the cert-manager work.

So I do expect that I will write a proposal for having possibility to completely disable them. But I do not think it makes sense to pull it directly into this proposal.

150-configurable-security-of-internal-communication.md
scholzj: Yes and no. Kafka Exporter is not part of Strimzi. So while it is needed for the complete functionality, it will be just opened as a contribution to the Kafka Exporter. (I actually have the code ready)

150-configurable-security-of-internal-communication.md
scholzj: Which part do you think is fragile or bad practice? Editing the status?

The problem with the cluster ID recovery is that:
* The Cluster ID is typically unknown, and people need to first deploy the cluster to find out from logs what the real ID is and then set it. I think that is a bigger issue then editing the status itself. Due to thet I think it is a bad UX rather then a bad practice.
* The recovery is often involuntary process users need to go through after some disaster. In this proposal, editing the status section is part of an intentional and planned configuration change from which the user benefits.

As far as this proposal is concerned, we want something that captures and confirms the user intent -> clearly shows us that the user read the configuration procedure, understands how it works and the consequences. Requiring the user to edit status is IMHO good way how to do it. You cannot do it with regular edit, you need a _special_ command for it etc. So I think it is good solution with minimal impact on the code.

If you have any better alternatives, I would be happy to discuss them.

150-configurable-security-of-internal-communication.md
scholzj: I do not know really. Right now I work with two types: `none` and `strimzi-tls`. But I can imagine that one day in the final API it would split the Strimzi CAs and CAs managed by other mechanisms (cert-manager, Vault, Cloud Provider KMS services, etc.).

The problem is that we today default to TLS encryption and mTLS authentication. So we cannot easily treat the  absence of the configuration as it being disabled. That is why the `type` is needed and needs to have the `none` type as well as some type representing the current state.

145-configurable-security-of-internal-communication.md
scholzj: Yes, that is exactly what the sentence says. We have the options and making it possible in the code would be trivial, I do not see why would we enforce it. The motivation for that might be for example to use short-lived credentials (SA tokens) for authentication rather than long-lived certificates.

Could you please clarify what do you consider unclear on this? What and where should be clarified? I'm not really sure about is as the sentence says exactly what you ask about. Thanks.

150-configurable-security-of-internal-communication.md
scholzj: TLS costs a lot of CPU. So by disabling TLS, you reduce the CPU consumption and this reduce the costs. I added more detailed explanation to the proposal.

145-configurable-security-of-internal-communication.md
scholzj: There is no special detection of new clusters. New clusters do not have any status set with the current security settings. So they would automatically accept whatever security settings the user uses. I tried to make it more clear in the proposal.

150-configurable-security-of-internal-communication.md
ppatierno: My question was more why not adding the support you are proposing also to Connect and MM2 since the beginning.

150-configurable-security-of-internal-communication.md
ppatierno: It will be interesting to see how it is going to work if the Kafka Exporter project won't release a new versions with such PR. Last release seems to be February 2025. Without it, the Strimzi implementation won't be complete.

150-configurable-security-of-internal-communication.md
scholzj: Sorry, support for what exactly? Connect (and MM2 based on Connect) do not use any security right now. So:
* Proposal for disabling TLS / mTLS does not make any sense for them as they do not use it
* Including securing Connect and MM2 with Service Account tokens would significantly increase the size and scope of this proposal

150-configurable-security-of-internal-communication.md
scholzj: I agree. I think for graduation we would need to also pass some audits of end-of-life dependencies. So if Kafka Exporter is not able to release and continue, we might need to decide what next about it (fork, replacement, something of our own, etc.) regardless of this proposal. This feature might be a nice test 🤷.

Sadly, the same applies to Cruise Control as well, which is even significantly more dead than Kafka Exporter which is at least merging some PRs.

145-configurable-security-of-internal-communication.md
ppatierno: I am not a native English speaker but a straight forward sentence like the following would have been simpler:

> The Service Account-based authentication can be used when TLS is disabled or not.

Anyway I can live with yours, I got the meaning.

145-configurable-security-of-internal-communication.md
ppatierno: I am trying to understand what happens at reconciliation level. Is the operator detecting that the status is empty so it's a newly deployed cluster or not?

150-configurable-security-of-internal-communication.md
ppatierno: > Including securing Connect and MM2 with Service Account tokens would significantly increase the size and scope of this proposal

That was the support I was talking about.

150-configurable-security-of-internal-communication.md
ppatierno: > the same applies to Cruise Control as well, which is even significantly more dead than Kafka Exporter

I have more hopes for Cruise Control ;-)

145-configurable-security-of-internal-communication.md
scholzj: Yes. The reconciliation will:
* Construct the desired configuration from the annotation (or use the current defaults if the annotation is missing)
* Check if the current configuration is stored in the `.status`
* If the current configuration is missing (for a new cluster or for a cluster where the user removed it manually as part of the _migration_), the desired changes will be used
* If the current configuration is present, it will compare the desired and current configurations and:
    * Fail the reconciliation if they differ (=> someone changed the configuration for an existing cluster)
    * Proceed wih the reconciliation if they match

I tried to make it more clear in the proposal

145-configurable-security-of-internal-communication.md
scholzj: Ok, thanks for the explanation. I tried to reword it to make it more clear.

150-configurable-security-of-internal-communication.md
scholzj: Ok, I added the explanation to the Out of Scope section to make it more clear there. But I think including it would significantly increase the scope of this proposal and its size and thus make reviews harder. And given the goal is a bit different (disabling TLS / mTLS versus adding authentication) I do not think there would be any real synergies.

145-configurable-security-of-internal-communication.md
ppatierno: Perfect, thanks!

145-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,436 @@
...
+This proposal focuses on two tasks:
+* Allowing users to disable TLS encryption for internal communication
+* Providing an alternative authentication mechanism
+
+The proposed alternative authentication mechanism is based on Kubernetes Service Accounts and their tokens.
+The Service Account tokens can be used in an OAuth-like fashion and are supported by Apache Kafka's `OAUTHBEARER` SASL mechanism.
+They work on most Kubernetes clusters without requiring additional dependencies such as Keycloak.
+
+The main purpose of Service Account-based authentication is to provide an alternative authentication mechanism when TLS encryption is disabled.
+However, users will be able to use the Service Account-based authentication can be used when TLS is disabled or not.
```
katheris:
```suggestion
However, users will be able to use the Service Account-based authentication regardless of whether TLS is disabled or not.
```

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,436 @@
...
+  Its identity would change when you deploy a new Cluster Operator into another namespace (for example, as part of a phased upgrade).
+  The identity would also be shared between all Kafka operands managed by this Cluster Operator.
+  Instead, we will create a new Service Account for each Kafka cluster.
+  This Service Account will live in the Kafka cluster's namespace and will be named `<cluster-name>-cluster-operator`.
+  The Cluster Operator will use the `TokenRequest` API to get the token for this account and use it to authenticate with the given Kafka cluster.
+
+The audience used by these tokens will be based on the name of the Kafka cluster and its namespace to ensure uniqueness.
+The default expiration for the tokens will be 3600 seconds, which is the recommended value by Kubernetes.
+
+![Service Account-based authentication](./images/145-configurable-security-of-internal-communication.svg)
```
katheris, line 128: The diagram seems to be missing a step 2, but also I guess it's just showing which components have a projected token vs using a token request? I'm not sure it necessarily adds much to the proposal so I wouldn't mind if it was removed.

145-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,436 @@
...
+The tokens have default expiration of 1 hour and typically do not use longer expiration then 24 hours.
+So when the Kubernetes API is not available when the token needs to be refreshed, the authentication would fail.
+Today, we also rely on the Kubernetes API to load the certificates.
+But because they are long-lived, a running Kafka cluster might work for much longer before it is impacted by Kubernetes API server outage.
+
+However, this is acceptable trade-of:
+* We are using more secure short-lived credentials which always carry this risk
+* I do not think outages when the Kubernetes API server is down but everything else in the Kubernetes cluster is stable and available is rare.
+  In many situation, Kubernetes API outages would be accompanied by outages of the worker nodes, Pod restarts, etc.
+  In which case even the operands reliant on mTLS would be affected.
```
katheris: If I'm understanding correctly what you intended here I think this sentence should be:
```suggestion
* I think it is rare to have outages when the Kubernetes API server is down but everything else in the Kubernetes cluster is stable.
  In many situation, Kubernetes API outages would be accompanied by outages of the worker nodes, Pod restarts, etc.
  In which case even the operands reliant on mTLS would be affected.
```

Assuming my understanding of the above is correct - What are you basing this assumption on? My understanding is that the API server runs in the control plane, whereas the workloads are in the data plane. And according to the Kubernetes docs typically all control plane components are run on the same machine, whereas the node components run on every node. So to me it seems somewhat likely that clusters might experience control plane problems that do not affect all or some of the nodes. So while I think that it's reasonable to say that this an opt-in feature and therefore users are accepting the risk, using the word "rare" here seems a bit strong to me.

145-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,436 @@
...
+#### Cruise Control Authentication
+
+Cruise Control will use the Service Account tokens when connecting to the Kafka cluster to collect its metrics.
+However, it uses its own API key-based authentication for its HTTP interface.
+This will remain unchanged for the time being, and the new cluster configuration will only be used to enable or disable TLS encryption.
+
+In the long term, we should aim to add support for Service Account-based authentication to the Cruise Control HTTP interface as well.
+That would also help improve some of the limitations of the current authentication model, such as lack of credential rotation.
+However, the Cruise Control community currently seems dysfunctional.
+This should be revisited later depending on the future of Cruise Control.
```
katheris: This sentence makes the current state unclear, so suggesting linking to the latest update from the project itself:
```suggestion
However, the [development and maintenance of the Cruise Control project seems to be paused while LinkedIn discusses the potential of donating it to an open source foundation](https://github.com/linkedin/cruise-control/issues/2346#issuecomment-4547334366).
This should be revisited later depending on the future of Cruise Control.
```

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,436 @@
...
+This might seem obvious for the Client CA, which has no role in internal communication and is used only for external users.
+TLS-based internal communication uses the Cluster CA as both the server and user certificate issuer.
+
+The Cluster CA, on the other hand, plays a key role in securing internal TLS-based communication.
+However, even when the internal communication does not use TLS, the Cluster CA is still used to provide the default server certificates for the user-configurable listeners.
+It therefore cannot be simply ignored when TLS is disabled for internal communication.
+
+However, that means that Kafka clusters with disabled TLS would still depend on the CA management.
+Certificate or CA renewals would cause unnecessary rolling updates and might be a source of bugs.
+We should also aim to allow users to disable the Cluster CA completely.
```
katheris, line 445: Does this mean when implementing this proposal you do not intend to change any of the behaviour around the rolling updates triggered from the CaReconciler, despite the fact that the nodes will not have their truststores configured and therefore do not need to be rolled in this case?

When you say "disable the Cluster CA completely" what do you mean here? Allowing people who also don't want to have Strimzi issuing listener certificates to configure Strimzi so it doesn't do anything Cluster CA related? As to me that is something different from simply skipping rolling updates so I'm not clear why adding the logic to skip unnecessary rolling updates would massively change the scope of this proposal

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,436 @@
...
+However, to keep the scope of this proposal under control, this is not part of this proposal and should be the subject of a future follow-up proposal.
```
katheris, line 446: Is this a future proposal you also intend to raise? Also would this be added before we marked the feature as GA?

145-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,436 @@
...
+With the ongoing work on cert-manager support, we currently have multiple parallel streams of work related to cluster security that might lead to API changes.
+This proposal therefore suggests using only an annotation for now to configure cluster security.
+Using an annotation means:
+* We are not binding to a specific API layout just yet by adding it to the custom resource spec.
+* We indicate that this is an _early access_ feature.
+
+Once both the work on this proposal and the cert-manager integration are complete, we can try to come up with a permanent API.
+Given that we just released version 1.0.0 and finalized the `v1` API, we should have enough time to finalize it before the next API version change.
+
+The cluster security would be configured in the `strimzi.io/cluster-security` annotation.
```
katheris: Should this have the word `internal` somewhere? Since my understanding is that this would never be used to replace the existing listener specific configuration we have already, and to me `cluster-security` makes me think of securing external access to the cluster.

150-configurable-security-of-internal-communication.md
katheris: Within the cert-manager work the part around the API has actually been the code that is the simplest and is the commit that has been there for the longest time. The hard part has been determining how to integrate the code with the existing security code that we have. Since opening my initial PR back in December last year I have been working on refactoring the internals to simplify the execution flow within the codebase when cert-manager is incorporated.

My understanding of this PR is it is adding the feature of changing the internal communication but in the quickest and simplest way possible, so without changing the API and without changing the internals much beyond changing the config of the components (and of course the new part around the service account authentication). Then you are proposing to do this later once this and the cert-manager code lands. To me that doesn't make sense.

On the API side ultimately we can never plan for every single possible future usecase, but to me it seems like we have a good idea of the requirements and we know what config is needed for cert-manager so we could at least discuss a potential API.

On the code side the main thing that has been blocking the cert-manager code is finding an implementation that works within the current requirements. Here you are suggesting adding a new security related class that will be passed through the reconciler and model classes but no mention of how that should relate to the existing Ca classes. So if this proposal is approved we have on one side the cert-manager work which is blocked behind refactoring and then this work which would likely land but still impact the refactoring being done for cert-manager. To me this seems an indication that either the cert-manager work needs to be delivered ahead of some of the refactoring, or we need to deliver this work keeping in mind the intended structure.

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,436 @@
...
+  # ...
+```
+
+### Further Implementation Details
+
+A new model class called `KafkaClusterSecurity` will be introduced.
+It will be instantiated at the beginning of the Kafka reconciliation in the `KafkaAssemblyOperator`.
+It will be passed through the various reconcilers and model classes.
+The `KafkaClusterSecurity` object will hold the information about the security configuration.
+It will be used to decide how to configure the different security aspects, such as client or broker configuration.
```
katheris, line 382: Can you expand more on the expected internal changes for this? For example in the case that no TLS is required presumably the step creating the certificate Secrets would be skipped?

150-configurable-security-of-internal-communication.md
katheris: Given this isn't intended to be the final API why not have multiple annotations e.g. `strimzi.io/cluster-internal-encryption-type`, `strimzi.io/cluster-internal-authentication-type` and `strimzi.io/cluster-internal-authentication-expiration` rather than a JSON object that needs to be parsed?

katheris (commented):
Thanks for the proposal @scholzj.

Overall I think I understand the intent of the feature, but I added some comments around the way you intend to implement it and the impact on cert-manager and the wider security refactoring I've been working on with Tina.

145-configurable-security-of-internal-communication.md
scholzj: As far as I know, the discussions are pretty much stalled and not moving forward at this point. So I prefer to not pull this into the proposal. Either Cruise Control gets this moving forward again at some point. Or we will need to act regardless of this proposal.

scholzj: (I'm happy to replace the word `dysfunctional` with something more politically correct if that is the point. But I think it describes the current state.)

150-configurable-security-of-internal-communication.md
scholzj: > Does this mean when implementing this proposal you do not intend to change any of the behaviour around the rolling updates triggered from the CaReconciler, despite the fact that the nodes will not have their truststores configured and therefore do not need to be rolled in this case?

No, I do not plan to touch it within this proposal. Keep in mind that Cluster CA might be still used for the user-defined listeners. So doing this would be pretty major impact:
* Full multistage rolling update when mTLS is used internally
* Limited single stage rolling update when TLS encryption is used internally or on the user-defined listeners
* Nothing when it is not used for anything?

Also, any change in this would increase the conflict surface between my work and the cert-manager work. That is why I think it makes sense to touch it now. Users who do not need it can for the time being simply set the validity of the CAs and certificates to a very long time as a workaround.

> When you say "disable the Cluster CA completely" what do you mean here? Allowing people who also don't want to have Strimzi issuing listener certificates to configure Strimzi so it doesn't do anything Cluster CA related? As to me that is something different from simply skipping rolling updates so I'm not clear why adding the logic to skip unnecessary rolling updates would massively change the scope of this proposal

What I thought about is completely disabling the CA(s). Not just the rolling updates. If the CA is not used, what is the point of having it there. However, there is a reason why it is not included in this proposal and it is _subject to change_ 😉.

150-configurable-security-of-internal-communication.md
scholzj: I'm not sure if I would raise it right after this proposal. That might not be the case. But I might raise it eventually. I do not think marking this feature as GA is in any way tied to this.

145-configurable-security-of-internal-communication.md
scholzj: It is not intended to replace the current listener configuration. But if we at some point add the CA on/off switch etc., it would be a bit more than just internal cluster security. So I prefer to stick with the current name. The name of the future API, I think that remains up to discussion.

145-configurable-security-of-internal-communication.md
katheris: I think saying something like `However, the development and maintenance of the Cruise Control project seems to be paused for the moment.` would suffice if you don't want to link off anywhere

150-configurable-security-of-internal-communication.md
scholzj: It would be in general series of if-else conditions. E.g. `if (tls) ... else ...` or `if (mTls) ... else if (sa) ... else ...`. They will be used in the places such as:
* In the `DefaultKafkaAgentClientProvider` and `KafkaAgentClient` classes to access the Kafka cluster with the correct security configuration
* In the `KafkaBrokerConfigurationBuilder` and `KafkaCluster` classes to configure the Apache Kafka brokers and controllers
* In Entity Topic and User Operator model classes to configure the User and Topic Operators
* In the Kafka Exporter model class to configure the Kafka Exporter
* In the Cruise Control model class to configure Cruise Control

Basically, where we configure the security in the operands or where we configure the clients to acccess the operands.

(Added to the proposal as well)

145-configurable-security-of-internal-communication.md
scholzj: Ok, changed.

150-configurable-security-of-internal-communication.md
katheris: The multi stage rolling is currently handled by CaReconciler doing the first rolling update and then the relevant component reconcilers doing the second. In CaReconciler there is a function called `verifyClusterCaFullyTrustedAndUsed` which is used to determine whether to roll the pods, and for cert-manager since we are hooking into the existing mechanisms for determining if a rolling update is needed I'm not expecting us to change that method. So to me having a simple check there would allow you to skip the rolling update without any major conflicts (or perhaps any conflicts at all). Then the rolling update on the component side again for cert-manager it links into the existing code so I wouldn't expect conflicts.

I agree that completely disabling the CA(s) should be out of scope for this proposal.

145-configurable-security-of-internal-communication.md
scholzj: I removed the `rare` word as it is not really quantifiable.

But the Kubernetes control plane is normally highly available in production-grade clusters. So you would normally need to have multiple things happening at the same time in order to have some significant outage. So I do think it is pretty likely it would be accompanied with outages to other components. Keep in mind that we do load the TLS certificates from the Secrets. So any Pod or Container restart while the Kubernetes API is not available would cause issues as well. So you do not need much of additional disruption to have a problem even without the Service Accounts.

150-configurable-security-of-internal-communication.md
scholzj: Good catch. I updated the diagram.

145-configurable-security-of-internal-communication.md
katheris: My assumption is that if we updated to have the CA on/off switch we would fully define the API, so if we go ahead with the annotations option do we need a name that could be applies more widely?

150-configurable-security-of-internal-communication.md
scholzj: The JSON object allows us to add more options should they turn out to be necessary. such as expiration or the JWKS endpoint address. So I think it is better than using separate annotations.

150-configurable-security-of-internal-communication.md
katheris: Thanks, the other area I was wondering about was the reconcilers, would we skip the step to create the certificate Secret(s) if no encryption is enabled, or since we need the certificates for the listeners would we create the certificates Secret still with all the certificates but just not use them in the config, or only request the certificate if the user hasn't specified custom certs for every listener?

150-configurable-security-of-internal-communication.md
scholzj: I do not think we have the complete idea TBH. If nothing else the part about the disabling of the CAs is not really figured out properly. For this proposal, We do not know exactly what will be the requirements different users would raise etc. So I think it is something we need to plan.

I currently do not expect any major conflicts between this proposal and the cert-manager work. That would ultimately arise only from the next steps. That is one of the (multiple) reasons for the split.

150-configurable-security-of-internal-communication.md
scholzj: I do not think it is super complicated. But it is yet another thing to cover, test, another set of conflicts with the cert-manager work. That is why this proposal doesn't deal with it.

145-configurable-security-of-internal-communication.md
scholzj: I'm not sure if it would be that simple. But sure, if it helps to get this approved, I can rename it to `strimzi.io/internal-cluster-security`. Its just an annotation. 🤷

150-configurable-security-of-internal-communication.md
scholzj: This proposal does not propose any changes to that. So the certificates would be created as before and might be used for the user-configured listeners depending on their configuration. If not, they will be unused.

Again, that limits the scope of the proposal, of the implementation, and the conflict surface with the other work.

145-configurable-security-of-internal-communication.md
im-konge: Maybe I just misread it. I thought that with this proposal there will be option to just enable/disable the TLS encryption. That's why I said that "should we just keep it as is until we know the future of CC and we are able to do anything there".

scholzj: Cruise Control has two different types of connections:
* The Kafka connections is today using TLS/mTLS and would follow the same logic as almost everything else in the proposal. You would be able to enable/disable TLS and choose the authentication (mTLS or Service Accounts). There is no problem with that.
* The second interface is the HTTP interface. That uses Cruise Control's custom API keys for authentication. Not mTLS. So the extent to which proposal applies to this interface will be limited. You would be able to disable / enable TLS. But the authentication will remain as the CC's API keys. Making it possible to disable TLS is important, because it allows for example Istio integration.

I tried to make it more clear in the proposal.

scholzj:
@ppatierno @katheris @tinaselenge @im-konge Do you have anything more? I think I either incorporated or replied to all your comments. Thanks.

145-configurable-security-of-internal-communication.md
im-konge: Ah okay, thanks for the changes :) sorry for stupid questions.

150-configurable-security-of-internal-communication.md
im-konge: Yeah I think having a new pipeline for this would be nice. So we don't have to remember switching some params and I guess it will be even better in terms of visibility - in the ST actions chart.

im-konge (approved):
LGTM, thanks for the changes.

scholzj:
@ppatierno @katheris @tinaselenge Anything more on this proposal? Thanks.

150-configurable-security-of-internal-communication.md
katheris: Ok, thanks makes sense

katheris:
Thanks for addressing my comments on this @scholzj. I had another look over and to me there are still some areas where I have concerns.

While I understand that you are aiming to reduce conflicts with the cert-manager code while adding this feature to me it feels like the result is that the proposal feels unfinished. Specifically around preventing additional rolling updates and not requesting certificates which are not needed. If the cert-manager feature wasn't currently being worked on presumably both these would be included in this proposal.

On the subject of cert-manager can you add something about how you expect this to integrate with cert-manager please? From what you've described I expect this feature should be useable whether Strimzi is managing the certificates or cert-manager (once that feature lands), but it would be useful to make sure we are on the same page about whether this would be useable with cert-manager certificates.

Finally I think this proposal is lacking a clear outline of the intended next steps or timeline for this feature. The only parts that seem to indicate the intention are:
```
Given that we just released version 1.0.0 and finalized the `v1` API, we should have enough time to finalize it before the next API version change.
```
and
```
This feature will initially be documented as _early access_.
```
Neither of these is very specific on what the expected timeline is for defining a proper API rather than using annotations or what would be needed in order to consider this feature a GA feature.

In order to address both the parts that conflict with cert-manager and the timeline part, what if you followed a similar approach to the [KRaft liveness/readiness proposal](https://github.com/strimzi/proposals/blob/main/046-kraft-liveness-readiness.md#proposed-order-of-tasks-to-complete-this-proposal)? There we proposed a phased approach which I think could work here. This proposal could describe the intended behaviour for things like the rolling but propose that those parts that might cause conflicts be delivered separately once cert-manager has landed. That way as a reviewer I can understand the full intended behaviour for this feature without needing it to be blocked behind the cert-manager implementation.

@ppatierno @tinaselenge also curious to see what you think of this suggestion

scholzj:
> While I understand that you are aiming to reduce conflicts with the cert-manager code while adding this feature to me it feels like the result is that the proposal feels unfinished. Specifically around preventing additional rolling updates and not requesting certificates which are not needed. If the cert-manager feature wasn't currently being worked on presumably both these would be included in this proposal.

The proposal is not unfinished. The proposal just focuses on a single specific feature. The next features would be covered by the future proposals.

I do not think they would be included in any case. I think this proposal is already big enough even without that and a lot of maintainers seem to struggle with reviewing even smaller scope proposal. So I think splitting the things makes sense in any case.

These things have also pretty simple workaround such as setting the renewal period to 99 years etc. Then you do not have any renewals and rolling updates because of them.

> On the subject of cert-manager can you add something about how you expect this to integrate with cert-manager please? From what you've described I expect this feature should be useable whether Strimzi is managing the certificates or cert-manager (once that feature lands), but it would be useful to make sure we are on the same page about whether this would be useable with cert-manager certificates.

This proposal is about disabling TLS. So it does not integrate with cert-manager in any way, nor change in any way how TLS is used when enabled. It just disables/enables TLS and disables/enables mTLS in the respective configurations of the various servers and clients used by Strimzi.

If you have any specific concerns, please share them and I can include them in the proposal.

> Finally I think this proposal is lacking a clear outline of the intended next steps or timeline for this feature. The only parts that seem to indicate the intention are:
> 
> ```
> Given that we just released version 1.0.0 and finalized the `v1` API, we should have enough time to finalize it before the next API version change.
> ```
> 
> and
> 
> ```
> This feature will initially be documented as _early access_.
> ```
> 
> Neither of these is very specific on what the expected timeline is for defining a proper API rather than using annotations or what would be needed in order to consider this feature a GA feature.

Early access is a term used pretty generically for many new features in Strimzi without any specific timelines to indicate new feature with limited experience. But I can remove it from the proposal. Or add a timeline. 🤷 

_Update: I added a new Timeline section that covers the timeline for removing the early-access label._

Doing the design and planning of any API changes carefully and in a way that leads to stable API is a feature not a bug. I'm happy to discuss alternatives such as having two sets of CRDs etc. But I think using an annotation is a good way forward that provides a good mix of simplicity and flexibility. So I suggest you give it a chance. But as I said, I'm happy to discuss the alternatives. So feel free to suggest better ideas.

> In order to address both the parts that conflict with cert-manager and the timeline part, what if you followed a similar approach to the [KRaft liveness/readiness proposal](https://github.com/strimzi/proposals/blob/main/046-kraft-liveness-readiness.md#proposed-order-of-tasks-to-complete-this-proposal)? There we proposed a phased approach which I think could work here. This proposal could describe the intended behaviour for things like the rolling but propose that those parts that might cause conflicts be delivered separately once cert-manager has landed. That way as a reviewer I can understand the full intended behaviour for this feature without needing it to be blocked behind the cert-manager implementation.

No, I do not want to write 1000s of lines long proposals. I think the proposal covers sufficiently my future intent:
* Cluster CA is still needed depending on the configuration even with disabled TLS
* Any disabling of the Cluster CA is subject of a future proposal

This proposal delivers a complete solution for https://github.com/strimzi/strimzi-kafka-operator/issues/2266. That is its focus.

150-configurable-security-of-internal-communication.md
tinaselenge: When we have a concrete API, would that need a migration? Would we need to clean up the annotation? Or since it's an early access feature, we would just tell users to remove the annotation and use the API instead?

tinaselenge (commented):
Apologies for my delayed response! Thank you for addressing the comments. 

I had similar thoughts to @katheris's comment, the proposal does feel like it leaves some open questions about how the follow up work fits together. That said, I understand the reasoning for keeping the scope focused. Large proposals tend to get stuck or take a long time to review, so for me, a smaller proposal that has self-contained scope is good. I think the concern could be addressed by being more explicit, for example, in a Future Improvements section or joined with the Timeline section you've just added, about what follow up proposals are planned and roughly what they'll cover. I know you've touched on this in the proposal by mentioning what's not included, and in the comments about future proposals. However, consolidating this into a dedicated section could make the intended end-state clearer. I understand that the end state still needs to be worked out. But we already know some of the pieces — defining the API, introducing a mechanism to skip unnecessary rolling updates, disabling the CA completely, etc.

I initially thought the annotation-based configuration is a bit unusual, but given this is early access and you want to avoid committing to an API layout before the cert-manager work lands, I agree that it's acceptable for now.

Looking at the POC and the discussion, it doesn't look like this will conflict with the cert-manager integration work, which is good. 

You probably already plan on this but just in case, want to mention that when we document this feature, I think we should include the operational trade-offs users will have when enabling this feature:

  - Unnecessary rolling updates: With TLS disabled, CA renewals will still trigger rolling updates. Users should be advised to set the renewal period to a long value (e.g. 99 years) as a workaround until a follow up feature is delivered disabling the Cluster CA entirely.
  - Kubernetes API availability: Short-lived SA tokens rely on the kubelet reaching the API server for rotation. If the API server is unavailable, components could lose authentication. This is a meaningful operational consideration compared to mTLS with longer lived certificates, and users should be aware of it before switching.

Even though this is an early access feature, documenting the workarounds and trade-offs now might help users. 

Overall, I am happy with the proposal. I've just left one minor question.

150-configurable-security-of-internal-communication.md
scholzj: I added the way we would migrate from the annotation to the final API to the proposal. It is something like this:

```
Eventually - once we settle on the final `Kafka` API for configuring the cluster security (see the "Future Work" section below) - users will have to migrate from the `strimzi.io/internal-cluster-security` annotation to the new API.
This will be done in several steps.

* In the Phase 0 (this proposal), all users will rely on the annotation
* In the Phase 1, we will introduce the new API in the `Kafka` CR.
  In this phase, the annotation will still be supported, but it would internally convert to the new API (e.g. using an internal mutating Gatekeeper plugin).
  This phase will take several releases and users will have time to migrate from the annotation to the `Kafka` CR API during this time window.
* In the Phase 2, we will drop the support for the annotation and only the `Kafka` CR API will be used.
  The `strimzi.io/internal-cluster-security` annotation will be ignored by the Cluster Operator.
  Any clusters that were not migrated to the new API will not be broken, but their reconciliation will be failing until the resource is updated to use the new API.
```

scholzj:
Thanks for the feedback @tinaselenge. I tried to incorporate your points into the proposal. Please have a look. Thanks.

----

> I had similar thoughts to @katheris's comment, the proposal does feel like it leaves some open questions about how the follow up work fits together. That said, I understand the reasoning for keeping the scope focused. Large proposals tend to get stuck or take a long time to review, so for me, a smaller proposal that has self-contained scope is good. I think the concern could be addressed by being more explicit, for example, in a Future Improvements section or joined with the Timeline section you've just added, about what follow up proposals are planned and roughly what they'll cover. I know you've touched on this in the proposal by mentioning what's not included, and in the comments about future proposals. However, consolidating this into a dedicated section could make the intended end-state clearer. I understand that the end state still needs to be worked out. But we already know some of the pieces — defining the API, introducing a mechanism to skip unnecessary rolling updates, disabling the CA completely, etc.

The Future Improvements section is already there. I added the final API there and I added a clear intent where I plan to work on the follow-up proposals and where.

> You probably already plan on this but just in case, want to mention that when we document this feature, I think we should include the operational trade-offs users will have when enabling this feature:
> 
> * Unnecessary rolling updates: With TLS disabled, CA renewals will still trigger rolling updates. Users should be advised to set the renewal period to a long value (e.g. 99 years) as a workaround until a follow up feature is delivered disabling the Cluster CA entirely.
> * Kubernetes API availability: Short-lived SA tokens rely on the kubelet reaching the API server for rotation. If the API server is unavailable, components could lose authentication. This is a meaningful operational consideration compared to mTLS with longer lived certificates, and users should be aware of it before switching.
>
> Even though this is an early access feature, documenting the workarounds and trade-offs now might help users.

Yeah, that sounds reasonable in general - regardless the early access. I added this to the Documentation section.

scholzj:
@tinaselenge @ppatierno Anything more on this?

katheris:
Thanks for the updates @scholzj it helps me to better understand the intentions here. For me there are two outstanding questions:
1. Will this feature be potentially marked GA before agreeing the final CR API, or will we expect to do the migration as described before marking it as GA?
2. If a user decides to turn mTLS off for internal listeners, but is still using the Cluster CA to issue client facing listener certificates then the additional rolling updates for a new key are not needed, however the Cluster CA is still being used. Will this scenario be included in the proposal around disabling Cluster CA?

scholzj:
@katheris Please see my answers below.

> 1. Will this feature be potentially marked GA before agreeing the final CR API, or will we expect to do the migration as described before marking it as GA?

My take would be that yes, the feature could be GA even without the final API. However, if you prefer to make it GA only after the final API, I'm open to update the proposal accordingly. Please let me know.

> 2. If a user decides to turn mTLS off for internal listeners, but is still using the Cluster CA to issue client facing listener certificates then the additional rolling updates for a new key are not needed, however the Cluster CA is still being used. Will this scenario be included in the proposal around disabling Cluster CA?

I do not think disabling mTLS is enough to simplify the rolling updates. Is it?

Even with simple TLS, the Kafka nodes connect with each other and still need to trust each other in the regular server authentication handshake. So you might still need to roll out the trust first and only then roll out the new server keys?

Imagine you are rolling out the new CA in a cluster with 3 nodes:
* Nodes 0 and 1 have the old CA and the old server certificate
* Node 2 is rolled to use the new CA and the new server certificate
* => When node 2 tries to connect the nodes 0 or 2, it will need to trust the old CA to connect
* => when node 0 or 1 tries to connect to node 2, they will need to trust the new CA to connect

So I think even with TLS only without mTLS, the additional CA rolling updates are still needed?

My plan was to not add further complexity to the CA management by having two ways how the CA is managed rolled out. When TLS is disabled and all you need is a listener certificate, it should be easy enough to get one and use one as the listener certificate. So if the user disables TLS and want to keep the Cluster CA, it would be rolled still in the same way. Again, if you absolutely want to have that, I can add it. But I think that it would add complexity without any obvious benefits.

katheris:
> My take would be that yes, the feature could be GA even without the final API. However, if you prefer to make it GA only after the final API, I'm open to update the proposal accordingly. Please let me know.

I would prefer if we stated intent to only make it GA after the final API. I would be open to us revisiting this decision in the future, but I think we should aim to keep the annotation only when it's an early access feature.

> So I think even with TLS only without mTLS, the additional CA rolling updates are still needed?

You are right, I was thinking that the updated broker will still have the old CA in their truststore, but hadn't considered the old brokers connecting to the new one. In that case, I'm happy with the separation we have here between this proposal and future ones.

scholzj:
> > My take would be that yes, the feature could be GA even without the final API. However, if you prefer to make it GA only after the final API, I'm open to update the proposal accordingly. Please let me know.
> 
> I would prefer if we stated intent to only make it GA after the final API. I would be open to us revisiting this decision in the future, but I think we should aim to keep the annotation only when it's an early access feature.

@katheris Ok, I updated the timeline to say that it will be early access until the final API is available. Thanks. Anything else? :-)

tinaselenge (approved):
Thanks @scholzj, I agree about marking it GA only after the final API is added. With that being added to the proposal, I have nothing else to add.

katheris (approved):
Thanks @scholzj, I'm happy with this proposal now

scholzj:
Thanks @tinaselenge @katheris. I appreciate it.

scholzj:
This has now 4 binding +1 votes (plus mine implicit). If there are no more comment till Wednesday EOB, I will merge it as approved.

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,486 @@
...
+                                  "expirationSeconds": 1800
+                              }
+                      }
+  spec:
+    # ...
+```
+
+At the start of the reconciliation, the Cluster Operator will validate the settings:
+* Ensure the configuration is valid JSON with a valid layout
+* Ensure the configuration makes sense (e.g. no mTLS authentication with disabled encryption)
```
ppatierno, line 294: what does the operator do if the configuration is not valid? The reconciliation fails? The default configuration is used?

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,486 @@
...
+
+This ensures that a random change to the security configuration does not break the Kafka cluster.
+
+#### `Kafka` CR `.status` Changes
+
+We cannot track the status in an annotation, so we have to make an actual change in the `.status` section of the `Kafka` CR.
+To support future schema changes, we can use an opaque object as the type.
+That way, if we need to change the schema in the future, we do not need to break CRD compatibility; we only need to handle the changes in our code.
+
+The status would be tracked in `.status.clusterSecurity` and follow a schema similar to the current annotation:
```
ppatierno, line 363: should we use the same as the annotation like `internalClusterSecurity`?

150-configurable-security-of-internal-communication.md
```diff
@@ -0,0 +1,486 @@
...
+
+Kafka Exporter currently does not support reading the tokens from a file.
+A PR will be opened to add this feature.
+Until it is merged and released, Kafka Exporter might be unavailable with Service Account-based authentication.
+
+### Timeline
+
+This feature will be initially released as _early access_ to collect some early feedback from Strimzi users.
+The _early access_ state of this feature will be communicated as part of the documentation and release notes (`CHANGELOG.md`).
+The _early access_ label will be removed once the final `Kafka` custom resource API is available to the Strimzi users.
```
ppatierno, line 428: I was just wondering why are we moving from a "feature gated" model to an "early access" model. We have been using feature gates as way to provide an early access for features to users and validate them before GA. What's the rational behind not using feature gates?

scholzj: We have never used feature gates for things gated by API. I.e. if user needs to opt-in into the future by changing some configuration in the custom resources (in this case in the annotation in the initial phase). Another current example of this is the auto-rebalancing feature, cert-manager, etc.

Feature gates are something we normally used for things that had no other configuration and were transitioning in the background. E.g. `UseConnectBuildWithBuildah` -> there is no flag to choose Kaniko or Buildah in the `KafkaConnect` CR. It is configured only through the feature gate. Same for `ServerSideApplyPhase1`, `UseStrimziPodSets`, etc.

There only feature gates that in some way combined both were `UseKRaft` and `KafkaNodePools` feature gates. But as they involved also different custom resources, upgradability, etc., I'm not sure they are fully comparable to this.

150-configurable-security-of-internal-communication.md
scholzj: I do not think so. This will eventually cover the whole cluster security.

150-configurable-security-of-internal-communication.md
scholzj: It will fail the reconciliation. I added it to the proposal.

150-configurable-security-of-internal-communication.md
ppatierno: I see the goal to cover the whole cluster security but AFAIU we don't want to set a final API with this proposal. So I was wondering, because it's focused on internal cluster security using an annotation it should be the same for the status even if using a field. It's also true that removing status or changing it is simpler than changing spec fields across releases.

scholzj: I cannot use an annotation because this is set by the operator. If I set the status in annotation, some other tooling will wipe it out. So I have to use the status. My take is that at the end, the `clusterSecurity` has better chance to survive then `internalClusterSecurity`. If you insist on it , I can rename the status to `internalClusterSecurity` as well. But I think there is 90%+ chance that it will end up deprecated and removed eventually.

Please let me know.

ppatierno: No Sorry, I didn't mean to use an annotation, we can't on the status I know. My wording was misleading (again :-)). Regarding the status field name I was just wondering if we should use a potential final name (I agree that clusterSecurity has better chance to survive) while this solution is still temporary. In the end we will use such field only for the internal security for now until the new and definitive API will be decided. Anyway, I can live with that.

ppatierno (approved):

scholzj:
Thanks @ppatierno

scholzj:

PR #228, opened 2026-06-12
Add proposal for cross-namespace credential push in KafkaAccess

guancioul:
### Type of Change

- New Proposal

### Description

This proposal addresses a security vulnerability reported in https://github.com/strimzi/kafka-access-operator/issues/105 .

Any user with write access to a namespace can obtain Kafka credentials by creating a `KafkaAccess` CR referencing a `KafkaUser` in another namespace, without authorization from the credential owner.

The proposed solution changes the operator to a push-based model: `KafkaAccess` CRs must be created in the same namespace as the referenced `Kafka` and `KafkaUser` resources, and a new `spec.targetNamespace` field specifies where the credential `Secret` should be delivered. When a `KafkaUser` is referenced, its owner must explicitly annotate it with a list of permitted target namespaces.

### Checklist

### DIFF
--- a/145-cross-namespace-credential-push-for-kafkaaccess.md
+++ b/145-cross-namespace-credential-push-for-kafkaaccess.md
@@ -1,0 +1,239 @@
+# Cross-namespace Credential Distribution for KafkaAccess
+
+This proposal introduces a push-based model for the Kafka Access Operator to securely deliver Kafka credentials across namespaces, replacing the current pull-based design that allows any user with namespace access to obtain Kafka credentials without explicit authorization from the credential owner.
+
+## Current situation
+
+The Kafka Access Operator currently uses a **pull-based model**.
+A user creates a `KafkaAccess` CR in their own namespace (e.g. `dev`) and references `Kafka` and `KafkaUser` resources that reside in a different namespace (e.g. `kafka`).
+The operator then creates a `Secret` in `dev` — the namespace of the `KafkaAccess` CR — containing all the Kafka connection details and credentials.
+
+```yaml
+apiVersion: access.strimzi.io/v1alpha1
+kind: KafkaAccess
+metadata:
+  name: kafka-access
+  namespace: dev                 # KafkaAccess lives here; Secret will be created here too
+spec:
+  kafka:
+    name: kafka-cluster
+    namespace: kafka             # cross-namespace reference
+    listener: plain
+  user:
+    apiGroup: kafka.strimzi.io
+    kind: KafkaUser
+    name: kafka-user
+    namespace: kafka             # cross-namespace reference
+```
+
+The Kafka Access Operator is granted a `ClusterRole` that allows it to read `KafkaUser` and `Kafka` resources and their associated secrets across all namespaces, and to create/update secrets in any namespace where a `KafkaAccess` might be placed.
+
+## Motivation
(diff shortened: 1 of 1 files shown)


im-konge:
Hi @guancioul , thanks for the proposal, I really appreciate it.

Sorry for leaving this without review for 2 weeks, I didn't have time to look at it yet. But I have a few points to this and the whole Access Operator project that I wanted to share with you.

I was thinking a bit about future of Access Operator. When Jakub created proposal about Gatekeeper and plugin system - https://github.com/strimzi/proposals/blob/main/144-Strimzi-Gatekeeper-plugin-system.md - I was thinking about having a plugin instead of the operator, which will make some things much easier. But I'm waiting for the implementation of the plugin system to be in place before starting some PoC and proposal. I guess it would make usage of the "Kafka Access" a bit easier to users that are already using Strimzi, because anyway Access Operator needs CRDs that are used by Strimzi.

Also, I think we need to be a bit careful about the cross-namespace stuff, as we had CVE in operators and I think that the Access Operator might have access to Secrets across Namespaces, that it shouldn't.

Finally, which a little bit circles back to first point, it seems that we don't have enough capacity to work on the Access Operator - there were some tries, but we all are busy with different things, which means that, at least from my side, in case that this will be approved and merged, I will not have enough time to guide you through the implementation in the repository, and I guess same applies to @katheris. And I would not like to make you feel that we don't care. Because we do, we are just busy with other things.

Based on these points, I would like to maybe close this proposal for now - at least until we decide what to do with the project and what are the next steps.

guancioul:
Thanks for the context, @im-konge — that all makes sense, and I appreciate you being upfront about capacity instead of just leaving it hanging.

Given that the architecture direction isn't settled yet (Gatekeeper plugin system vs. continuing with the current operator) and the team doesn't have bandwidth to guide an implementation right now, I agree it's better to close this for now rather than risk the work being thrown away once that direction is decided.

One ask though: could we leave a note on strimzi/kafka-access-operator#105 pointing out that the underlying vulnerability is still open, and that it'll likely need to be addressed as part of whatever comes next — whether that's a Gatekeeper plugin replacement (as mentioned in proposal #144) or continued work on the current operator? I just don't want the security issue itself to get lost along with this proposal.

im-konge:
We should definitely do that, thanks for raising this. And also thanks for the proposal. I know that you spend time on this and I'm sorry that it didn't lead to any conclusion or implementation PR.
Let's see what we will do next and in case that we will have some decision how we will go further, we will let you know - maybe there will be this or other work in case you will be still interested on contributing.

Also, is there any other issue - for example in the operators repo - which would be interesting for you? 

Thanks

scholzj:
TBH, I think the cross-namespace access here is different as the Access Operator is a separately deployed tool. So if you can deploy it with access to multiple namespaces, you already have the access yourself. So it is in the position of the Cluster Operator rather than the position of User or Topic operators here.

(I agree with the need to rethink what the future of Access Operator is here. Just do not see any particular concerns with regards to the recent CVE.)

im-konge:
Yeah you are right that it's a different case in comparison with User or Topic operators, but I think that we anyway need to have in mind to give just needed rights and nothing else. I think that we discussed that on community call, I'm not sure.

PR #231, opened 2026-06-15
Add proposal for safe rollout of listener connectivity changes

acgtun:
### Type of Change

- New Proposal

### Description

Proposal 148: make connectivity-affecting listener changes (type change, port change, adding or removing a listener) safe to roll out and roll back, for all listener types that provision per-broker Kubernetes resources (`nodeport`, `loadbalancer`, `cluster-ip`, `route`, `ingress`).

Today `KafkaReconciler` reconciles every per-broker listener resource up front (`listeners()`) and then rolls the pods one at a time. That leaves a window where a not-yet-rolled broker's `Service`/`Route`/`Ingress` already points at a listener the broker is not yet serving, so clients fail for the whole roll. On `nodeport`, a type-change rollback also re-allocates random node ports, which can move a port onto a different listener and cause SASL-mechanism mismatches that do not self-heal.

The proposal reconciles each broker's listener resources in lockstep with that broker's own restart, so each broker's resources always match the listener it is serving. It covers the `KafkaRoller` hook point, a per-broker convergence guard for idempotency/resumability, bootstrap-last handling, the internal-listener precondition, the KRaft-only correctness argument, an implementation sketch, a feature gate, testing, and rejected alternatives.

AI assistance was used to help structure and edit this proposal; the problem, design, trade-offs, and reasoning are my own (the work comes from a real listener-rollout incident).

### Checklist

### DIFF
--- a/148-safe-listener-connectivity-rollout.md
+++ b/148-safe-listener-connectivity-rollout.md
@@ -1,0 +1,275 @@
+# Safe rollout of listener connectivity changes (per-broker listener-resource reconciliation + deterministic node ports)
+
+Make connectivity-affecting listener changes (listener type changes, port changes, adding/removing a listener) safe to roll out and roll back. This applies to **all listener types that provision per-broker Kubernetes resources** — `nodeport`, `loadbalancer`, `cluster-ip`, `route`, and `ingress` — not just `nodeport`. The proposal has two complementary parts:
+
+- **Per-broker listener-resource reconciliation aligned with the rolling update (the general fix).** Reconcile each broker's listener resources (its `Service`, and `Route`/`Ingress` where applicable) in lockstep with that broker's own restart, instead of replacing all per-broker resources for the whole cluster up front while the brokers restart one by one. This removes the availability window — present for **every** per-broker listener type — where a not-yet-rolled broker's resources already point at a listener the broker is not yet serving.
+- **Deterministic node ports (a `nodeport`-specific hardening).** Give `nodeport` listeners stable, formula-derived node ports so deployments and rollbacks never reshuffle ports between listeners. This removes the most severe failure from the incident below (clients hitting the *wrong SASL mechanism*), with a small, low-risk change and no change to the rolling-update machinery.
+
+The two are independent and can ship separately. The rest of this document refers to them as **Phase 1 (deterministic node ports)** and **Phase 2 (per-broker reconciliation)**, numbered by expected delivery order (Phase 1 is the smaller change and is expected to land first), not by importance — Phase 2 is the general fix.
+
+`loadbalancer` listeners have an address-churn problem analogous to the `nodeport` port reshuffle (the cloud may reassign the external IP/hostname when the `Service` is recreated); equivalent pinning mitigations exist but are cloud-provider-specific and are out of scope here. The general remedy for `loadbalancer` (and all other types) is Phase 2.
+
+> **Scope: KRaft only.** This proposal assumes KRaft-based Kafka clusters and does not consider ZooKeeper-based clusters (ZooKeeper support has been removed from current Strimzi). The Phase 2 partial-roll correctness argument relies on KRaft broker-registration semantics (see [KRaft assumption](#kraft-assumption)).
+
+## Current situation
+
+The Kafka reconciliation pipeline reconciles listeners and rolls the pods in two separate, sequential phases:
+
+```java
+// KafkaReconciler.reconcile()
+.compose(i -> listeners())                       // (1) reconcile ALL Services + compute advertised.listeners
+...
+.compose(i -> brokerConfigurationConfigMaps())   // (2) write broker configs (incl. advertised.listeners)
+...
+.compose(i -> podSet())
+.compose(podSetDiffs -> rollingUpdate(podSetDiffs)) // (3) restart pods ONE BY ONE
+```
+
+In phase (1), `listeners()` reconciles the shared bootstrap resources and **every** per-broker listener resource at once. Each external listener provisions per-broker Kubernetes objects — a `Service` for `nodeport`/`loadbalancer`/`cluster-ip`, plus a `Route` (`route`) or `Ingress` (`ingress`) — and they are all reconciled cluster-wide before any pod rolls.
+
+The brokers, however, only pick up the new listener configuration when they restart, which happens **one pod at a time** in phase (3). This gap produces two failure modes.
+
+#### Failure mode A — resource/broker mismatch window (all per-broker listener types)
+
+When the listener **type** changes (for example `nodeport` → `cluster-ip` or a rollback back to `nodeport`), a `port` changes, or a listener is added/removed, all per-broker resources are switched/replaced immediately, but each broker only starts serving the new listener when it restarts. Brokers that have **not yet** restarted still serve the **old** listener, but their resources have **already** been switched. Clients connecting to those brokers fail (connection refused / unreachable / wrong endpoint) for the whole duration of the roll, not just briefly during a restart. This applies to **every** per-broker listener type, regardless of how addresses are assigned.
+
+#### Failure mode B — address churn on resource recreate (`nodeport` and `loadbalancer`)
+
+Some listener types do not have a stable externally-visible address across a resource recreate:
+
+- For `nodeport`, the `spec.ports[].nodePort` is normally assigned randomly by Kubernetes from the service node-port range. `ServiceOperator.patchNodePorts` only preserves it when the current and desired `Service` are **both** already `NodePort` (or both `LoadBalancer`):
+
+```java
+// ServiceOperator.internalUpdate(...)
+if (("NodePort".equals(current.getSpec().getType()) && "NodePort".equals(desired.getSpec().getType()))
+        || ("LoadBalancer".equals(current.getSpec().getType()) && "LoadBalancer".equals(desired.getSpec().getType())))   {
+    patchNodePorts(current, desired);
+```
+
+So across a **type change** the old port cannot be preserved and Kubernetes allocates fresh random ports on the way back to `NodePort`. The severe consequence is **cross-listener collision**: random reallocation can move a port that previously fronted one listener onto a *different* listener, so a client with cached metadata sends, say, a SCRAM handshake to a port now serving the Kerberos listener — a hard authentication-protocol mismatch that does not self-heal.
+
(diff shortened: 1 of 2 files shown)


ppatierno:
Please split the proposal by having one line per sentence and also change the description if you leveraged AI help to write it (which is pretty clear to me). Thanks!

148-safe-listener-connectivity-rollout.md
```diff
@@ -0,0 +1,301 @@
...
+
+The proposal reconciles each broker's listener resources in lockstep with that broker's own restart.
+Those resources are the broker's `Service`, plus its `Route` or `Ingress` where applicable.
+Today the operator replaces every per-broker resource for the whole cluster up front and then restarts the brokers one by one.
+Reconciling per broker removes the availability window where a not-yet-rolled broker's resources already point at a listener the broker is not yet serving.
+That window exists for every per-broker listener type.
+
+> **Scope: KRaft only.**
+> This proposal assumes KRaft-based Kafka clusters and does not cover ZooKeeper-based clusters, which have already been removed from current Strimzi.
+> The partial-roll correctness argument relies on KRaft broker-registration semantics (see [KRaft assumption](#kraft-assumption)).
```
scholzj: We do not support ZooKeeper for a long time. So this seems pointless.

148-safe-listener-connectivity-rollout.md
```diff
@@ -0,0 +1,301 @@
...
+
+This proposal affects the Strimzi Cluster Operator only: `KafkaReconciler` (reconcile ordering), `KafkaListenersReconciler` (split into shared, per-broker, and bootstrap reconciliation), and `KafkaRoller` (a strictly-placed, idempotent pre-restart hook and just-in-time advertised host and port supply).
+
+No CRD or API changes are required.
+No other Strimzi projects are affected.
+Only KRaft-based clusters are in scope (see [KRaft assumption](#kraft-assumption)).
+
+## Compatibility
+
+- **Gated** behind a feature gate that is disabled by default initially (see [Feature gate and rollout](#feature-gate-and-rollout)), so it ships with zero behaviour change until explicitly enabled.
```
scholzj: I think the scale and spread of the changes needed for something like this cannot be practically gated.

148-safe-listener-connectivity-rollout.md
```diff
@@ -0,0 +1,301 @@
...
+- **Behaviour is unchanged** for non-connectivity-affecting reconciliations, and for connectivity changes on internal listeners, which fall back to the current behaviour.
+- **No API or CRD change.**
+- **Reconciliation takes longer** for connectivity-affecting changes, because per-broker resource reconciliation (and waiting for an asynchronously-assigned address) is interleaved with the roll.
+  This is most pronounced for `loadbalancer`, where each per-broker LB is provisioned by the cloud (often minutes) and is now serialized across the roll; for large clusters (200+ brokers) the total can be very large.
+  Per-broker waits are bounded by `operationTimeoutMs`, but the overall reconcile may approach or exceed the reconciliation interval, so overlap and lock behaviour and progress logging must be considered.
+  The migration is resumable (per-broker convergence guard), so an aborted reconcile continues on the next pass rather than restarting from scratch.
```
scholzj: I think these things have possibly big negative impact on other users who struggle with the rolling updates length already. So we need to be careful about it. (But I appreciate the honesty at mentioning it)

I'm not sure what exactly you mean with the restarting from scratch. That is general principle the operator tries to follow in all cases.

148-safe-listener-connectivity-rollout.md
```diff
@@ -0,0 +1,301 @@
...
+- **Listener status** is fully populated only after the roll completes; intermediate reconcile passes may report a mix of old and new per-broker addresses, all valid at the time.
```
scholzj: Isn't that the case already now?

scholzj (commented):
Thanks for the proposal.

TBH, I have a bit mixed feelings about it. I think you describe the existing behaviour well and I do not have problem to acknowledge that this creates a gap that could cause a problem.

However, it is a rare problem because I would not expect users to reconfigure the listeners like this regularly. So saying that you should expect some unavailability and plan it properly is for me not a problem. I also think there is a simple workaround:
* Roll out the new listener on a new port
* Once the roll out is complete, move the clients
* Once clients are moved, remove the old listener

Why does that not work for you?

Given the above, the proposal feels like a huge change for a very niche benefit. So I think in order to accept the change, I would need to believe that it has also other benefits to the code and it is worth it in general. But I think it would bring a lot of other problems as well and there are many things where the current model fits better and there would be many edge cases where the other model does not work. So that is why I'm mixed on this ... I can see some benefits, but also a lot of problems.

Also with all the changes, I think you would just minimize the window where the issues happen. But not solve it completely. In addition, I think it would be quite complicated to keep the consistency and make sure the cluster is not in 5 different states as someone was changing the configurations in the custom resources.

Out of curiosity ... do you have some prototype of this? Or just the proposal?

acgtun:
> Thanks for the proposal.
> 
> TBH, I have a bit mixed feelings about it. I think you describe the existing behaviour well and I do not have problem to acknowledge that this creates a gap that could cause a problem.
> 
> However, it is a rare problem because I would not expect users to reconfigure the listeners like this regularly. So saying that you should expect some unavailability and plan it properly is for me not a problem. I also think there is a simple workaround:
> 
> * Roll out the new listener on a new port
> * Once the roll out is complete, move the clients
> * Once clients are moved, remove the old listener
> 
> Why does that not work for you?
> 
> Given the above, the proposal feels like a huge change for a very niche benefit. So I think in order to accept the change, I would need to believe that it has also other benefits to the code and it is worth it in general. But I think it would bring a lot of other problems as well and there are many things where the current model fits better and there would be many edge cases where the other model does not work. So that is why I'm mixed on this ... I can see some benefits, but also a lot of problems.
> 
> Also with all the changes, I think you would just minimize the window where the issues happen. But not solve it completely. In addition, I think it would be quite complicated to keep the consistency and make sure the cluster is not in 5 different states as someone was changing the configurations in the custom resources.
> 
> Out of curiosity ... do you have some prototype of this? Or just the proposal?

Thanks @scholzj for your detailed reply. It is a proposal, and we had this issue in production a few months ago. Yes, indeed, we should follow the steps to rollout a new listener first, after it completed, then remove the old one. We were new users for strimzi at the time, which didn't awre of this and faced this issue.

acgtun:
> Please split the proposal by having one line per sentence and also change the description if you leveraged AI help to write it (which is pretty clear to me). Thanks!

Thanks @ppatierno. Sorry about that, now i know we need one line per sentence.

148-safe-listener-connectivity-rollout.md
acgtun: Agreed, that note is moot — I'll drop the ZooKeeper scoping.

scholzj:
I thought about this a bit more. While this model might work fine during the regular operations, I wonder how would it handle startup and various recovery situations. E.g. when the Pods or Services are missing, recovering the cluster one by one might not work well and might not be sufficient to recover the cluster (e.g. because the pods being recovered one by one will crash-loop and never find each other etc.).

acgtun:
> I thought about this a bit more. While this model might work fine during the regular operations, I wonder how would it handle startup and various recovery situations. E.g. when the Pods or Services are missing, recovering the cluster one by one might not work well and might not be sufficient to recover the cluster (e.g. because the pods being recovered one by one will crash-loop and never find each other etc.).

Thanks @scholzj, that's a fair concern.

To clarify the intended scope: the per-broker coupled path only kicks in for an in-place change to a listener that brokers are already serving. Startup and recovery are different - the per-broker resources are just missing, not being switched under a running broker, so there is no mismatch window to protect. In those cases the operator should keep the current up-front reconciliation and create all the listener resources before the pods come up.

Also, brokers don't rely on the external client listener to find each other. Quorum/replication go over the internal control-plane and replication listeners on the headless `brokers` service, which are always reconciled up front and are never part of a connectivity change. So bringing brokers up one by one during startup/recovery isn't affected by this.

That said, you're right that the operator then has to reliably tell "in-place change on a serving listener" apart from "resources absent during creation/recovery", and the combined case (a recovery happening while a change is still pending in the CR) is exactly the kind of multi-state edge case you flagged earlier. That's real added complexity for a fairly narrow benefit, so I share your hesitation. I've updated the proposal to spell out the startup/recovery handling, but I'm also fine closing this if you don't think it's worth pursuing.

scholzj:
Yeah, to be honest, I do not think this makes really sense as some special code path for just changes to listeners. I do not think the maintenance effort for that is really worth what we would get out of it. I find the idea interesting as a general approach. But not sure how easy it would be to make it actually work.

148-safe-listener-connectivity-rollout.md
```diff
@@ -0,0 +1,321 @@
+# Safe rollout of listener connectivity changes
+
+Make connectivity-affecting listener changes safe to roll out and roll back.
+These changes are listener type changes, port changes, and adding or removing a listener.
+This covers all listener types that provision per-broker Kubernetes resources: `nodeport`, `loadbalancer`, `cluster-ip`, `route`, and `ingress`.
```
ppatierno, line 5: tlsroute instead of ingress which is deprecated?

148-safe-listener-connectivity-rollout.md
```diff
@@ -0,0 +1,321 @@
...
+.compose(i -> listeners())                       // (1) reconcile ALL Services + compute advertised.listeners
+...
+.compose(i -> brokerConfigurationConfigMaps())   // (2) write broker configs (incl. advertised.listeners)
+...
+.compose(i -> podSet())
+.compose(podSetDiffs -> rollingUpdate(podSetDiffs)) // (3) restart pods ONE BY ONE
+```
+
+In phase (1), `listeners()` reconciles the shared bootstrap resources and every per-broker listener resource at once.
+Each external listener provisions per-broker Kubernetes objects: a `Service` for `nodeport`, `loadbalancer`, and `cluster-ip`, plus a `Route` for `route` or an `Ingress` for `ingress`.
```
ppatierno, line 28: TLSRoute missing?

ppatierno (commented):
I left a couple of comments about missing reference to TLSRoute and how this proposal impacts the support for Gateway API.
Said that I agree with Jakub I see this proposal adding too much complexity for no additional big values.

acgtun:
Thanks @scholzj and @ppatierno for the feedbacks and suggestion. Based on the above discussion, looks like this added a lot of complexity but not much value and there is a way to get around it. 
I close the proposal for now. thanks again.

PR #232, opened 2026-06-18
Add AGENTS.md file and update PR template

scholzj:
### Type of Change

- Other

### Description

Given how common the discussion around one sentence per line and numbering is, this PR introduces two changes:
* Adds an AGENTS.md file that tries to make that clear to AI tools
* Adds the one-sensente-per-line to the PR template

### Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -17,7 +17,7 @@
 ### Proposal Format
 
 - Proposal files must be Markdown
-* You MUST follow a strict one sentence per line format
+- You MUST follow a strict one sentence per line format
 - Do NOT number proposal sections (e.g., "1. Motivation", "2. Proposal"), use Markdown headers only (`##`, `###`)
 - DO use numbered lists within sections when showing sequential steps
 - DO use bullet points for non-sequential items
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,25 @@
+# Strimzi Proposals
+
+## Scope
+
+- This repo is a Markdown proposal archive, not a codebase. Proposal files live at the root as `NNN-title.md`; proposal assets live under `images/`.
+- There is no root build, test, lint, formatter, or package manifest.
+
+## Proposal Workflow
+
+- Start new proposals by copying the `000-template.md`, but the template explicitly allows adding or removing sections to fit the proposal.
+- Use the next free three-digit sequence number for a new proposal and all related assets before merge, matching `.github/PULL_REQUEST_TEMPLATE.md`.
+- When opening a PR with an AI tool, always use `.github/PULL_REQUEST_TEMPLATE.md`.
+- Do not update `README.md` when opening a proposal PR; update the index only at the end before maintainers merge the PR.
+- `README.md` is the proposal index and is ordered newest/highest number first.
+- Proposal files must be Markdown and follow a strict one sentence per line format.
+- Name proposal-specific images/assets with the same numeric prefix as the proposal, following existing files such as `images/144-...svg`.
+
+## Verification
+
+- With no configured checks, verify Markdown edits by reviewing the rendered structure, relative links, and image paths.
+- Run `git diff --check` before handing off if whitespace-sensitive Markdown changes were made.
+
+## Local Files
+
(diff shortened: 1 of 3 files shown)


im-konge (approved):

AGENTS.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Proposal Workflow
+
+- Start new proposals by copying the `000-template.md`, but the template explicitly allows adding or removing sections to fit the proposal.
+- Use the next free three-digit sequence number for a new proposal and all related assets before merge, matching `.github/PULL_REQUEST_TEMPLATE.md`.
+- When opening a PR with an AI tool, always use `.github/PULL_REQUEST_TEMPLATE.md`.
+- Do not update `README.md` when opening a proposal PR; update the index only at the end before maintainers merge the PR.
+- `README.md` is the proposal index and is ordered newest/highest number first.
+- Proposal files must be Markdown and follow a strict one sentence per line format.
+- Name proposal-specific images/assets with the same numeric prefix as the proposal, following existing files such as `images/144-...svg`.
```
ppatierno, line 15: Not sure we need a dedicated section but I would add something like:

```md
## Numbering and Lists

  - Do NOT number proposal sections (e.g., "1. Motivation", "2. Proposal")
  - Use Markdown headers only (`##`, `###`)
  - DO use numbered lists within sections when showing sequential steps
  - DO use bullet points for non-sequential items
```

AGENTS.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Verification
+
+- With no configured checks, verify Markdown edits by reviewing the rendered structure, relative links, and image paths.
+- Run `git diff --check` before handing off if whitespace-sensitive Markdown changes were made.
```
ppatierno, line 29: should we add this as check within the template?

AGENTS.md
```diff
@@ -0,0 +1,25 @@
...
+- There is no root build, test, lint, formatter, or package manifest.
+
+## Proposal Workflow
+
+- Start new proposals by copying the `000-template.md`, but the template explicitly allows adding or removing sections to fit the proposal.
+- Use the next free three-digit sequence number for a new proposal and all related assets before merge, matching `.github/PULL_REQUEST_TEMPLATE.md`.
+- When opening a PR with an AI tool, always use `.github/PULL_REQUEST_TEMPLATE.md`.
+- Do not update `README.md` when opening a proposal PR; update the index only at the end before maintainers merge the PR.
+- `README.md` is the proposal index and is ordered newest/highest number first.
+- Proposal files must be Markdown and follow a strict one sentence per line format.
```
ppatierno: Usually agent are good at following examples maybe we could add something like:

**Bad example:**
 ```markdown
This is sentence one. This is sentence two. And this is sentence three.
```
**Good example:**
```markdown
This is sentence one.
This is sentence two.
And this is sentence three.
```

ppatierno (commented):
Overall LGTM but I left a few comments for improvements.

see-quick (approved):
Thanks 👍 .

AGENTS.md
scholzj: I think this goes a bit too far. It should understand what one sentence per line means I think.

AGENTS.md
scholzj: I strucgtured it a bit differently, but I added these points.

AGENTS.md
scholzj: TBH, AI added this for itself on its own ... so who am I to delete it :-). But I'm not 100% sure how would I translate it into the PR template. If you have any suggestion, I'm happy to add it.

ppatierno: Maybe something like "Verify whitespace-sensitive Markdown changes by running `git diff --check`". But if you think it's useless we can keep this out.

ppatierno (approved):

AGENTS.md
ppatierno: > It should understand what one sentence per line means I think.

I don't have much faith in that but fine to not adding examples.

AGENTS.md
scholzj: I don't know. I do not really remember having problems with that. I would say let's keep t and see if we have issued with that like with the one-sentence-per-line.

PR #233, opened 2026-06-22
Templating Additional Volumes for Pods with Stable Identities

scholzj:
### Type of Change

- New Proposal

### Description

This proposal suggests how Strimzi can allow users to _template_ additional volumes and configure different volumes for different Pods / nodes.

### Checklist


scholzj:
@strimzi/maintainers It would be great if you can find some time to review this proposal. We need to be able to maintain the project velocity and show the ability to add new features if we want to move towards graduation.

Frawless (approved):

im-konge (approved):
I don't have any comments to this. Thanks for the proposal. LGTM.

147-templating-additional-volumes.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Motivation
+
+However, additional volumes cannot be configured at the level of individual Pods or nodes.
+They are configured at the level of `Deployment` (for HTTP Bridge, Cruise Control, Kafka Exporter, and Entity Operator) or `StrimziPodSet` (for Kafka, Kafka Node Pool, Connect, and MirrorMaker 2).
+As a result, the exact same volume would be mounted into all Kafka nodes belonging to a given node pool, all nodes belonging to the same Connect cluster, etc.
+
+However, in some situations, users might want to mount different volumes into different Pods.
+This might include:
+* Mounting Secrets or ConfigMaps with per-broker configuration or credentials.
```
ppatierno, line 16: which kind of "per-broker configuration" are you referring to?

scholzj: Some possible examples are described right in the next sentence:

```
The credentials or configuration from a third-party service, such as authorization or authentication, might be stored in different Secrets or ConfigMaps or in the same Secret or ConfigMap under different keys.
```

For example imagine some API keys that are per broker. Or per-broker certificates for Prometheus metrics if you CA doesn't do wildcard certificates etc.

scholzj: @ppatierno Does this help?

146-templating-additional-volumes.md
```diff
@@ -0,0 +1,139 @@
...
+Using a separate class allows Strimzi to clearly define the volume types supported for templating in the API.
+
+Currently, all components use the existing `PodTemplate` class, whether they are stateless Pods managed through `Deployment` or stateful Pods managed through `StrimziPodSet`.
+However, only the stateful Pods should have the new templated volumes field.
+Therefore, a new class, `StatefulPodTemplate`, will be added as well.
+This class will extend the original `PodTemplate` class and will add the templated volumes as a new field.
+The existing `PodTemplate` class will continue to be used for stateless Pod templates.
+The new `StatefulPodTemplate` class will be used for stateful Pod templates.
+
+Corresponding model classes and `TemplateUtils` will be updated to add the templated volumes to the generated Pods.
```
tinaselenge: So in the places that we pass PodTemplate to get volumes, we would pass StatefulPodTemplate instead? And then I guess addAdditionalVolumes() method will then use this to resolve the right volumes?

tinaselenge (commented):
Thanks Jakub. It looks good to me, just left comment on the implementation.

146-templating-additional-volumes.md
scholzj: The `TemplateUtils` class will be extended and new method `addAdditionalTemplatedVolumes(...)` will be added. This method be responsible for templating and creating the additional templated volumes. The existing `TemplateUtils.addAdditionalVolumes(...)` method will remain unchanged. In the places where we prepare the volumes for non-stateful Pods, `TemplateUtils.addAdditionalVolumes(...)` will be called as today without any change. In the places where we prepare the volumes for stateful Pods, the code will call both `TemplateUtils.addAdditionalVolumes(...)` and `TemplateUtils.addAdditionalTemplatedVolumes(...)` to add both the regular as well as the templated volumes.

`TemplateUtils.addAdditionalVolumes(...)` has already validation for duplicate volume names. When duplicate volume name is found, an `InvalidResourceException` exception is raised. The same will be done also in `TemplateUtils.addAdditionalTemplatedVolumes(...)`. As we fail the reconciliation with exception in case of duplicate volume names, we can just run both methods one after each other and do not need to decide which one has priority.

I added this to the proposal as well. If needed, here is the prototype implementation: https://github.com/strimzi/strimzi-kafka-operator/compare/main...scholzj:strimzi-kafka-operator:templating-additional-volumes?expand=1

Please let me know if you have any concerns with this.

scholzj:
@ppatierno @tinaselenge Anything more on this proposal? Thanks.

146-templating-additional-volumes.md
tinaselenge: thanks for the clarification. I had a look at the POC too, it makes sense to me.

tinaselenge (approved):
LGMT, thanks!

scholzj:
This has now 3 binding +1 votes (+ mine implicit). If there are no more comments I will proceed and merge it as approved on Wednesday EOB.

ppatierno (approved):

scholzj:

PR #234, opened 2026-06-26
Added proposal about broker cordoning on scale down

ppatierno:
### Type of Change

- New Proposal

### Description

This proposals aims to address https://github.com/strimzi/strimzi-kafka-operator/issues/12665.
It's about leveraging the cordoning log dirs between brokers when auto-rebalancing is used during scale down operations.

### Checklist


146-broker-cordoning-scale-down.md
```diff
@@ -0,0 +1,380 @@
...
+The operator needs exclusive control over this property to manage the scale-down lifecycle correctly (cordoning before rebalancing, uncordoning on cancelled scale down).
+
+### Detailed workflow
+
+The cordoning logic is implemented in a new `KafkaClusterCreator.brokerCordoningCheck()` method, following the same pattern as the existing `brokerRemovalCheck()` method.
+The method sets a `cordoningCheckFailed` instance variable that indicates whether cordoning/uncordoning operations were performed and propagation is needed, similar to how `brokerRemovalCheck()` sets `scaleDownCheckFailed`.
+It also takes into account the fact that cordoning (or uncordoning) brokers needs some time for the configuration change to propagate across the cluster.
+This means that the auto-rebalancing, together with the actual scale-down, could be delayed across reconciliations, until the cordoning of brokers being removed is confirmed.
+The cordoning check runs on every reconciliation to ensure brokers are in the correct cordoning state, for scaling down or cleaning up (in case of cancelled scale down):
```
scholzj: Why exactly would it be delayed?

146-broker-cordoning-scale-down.md
```diff
@@ -0,0 +1,380 @@
...
+
+## Proposal
+
+### Overview
+
+Extend the auto-rebalancing mechanism to automatically cordon brokers during the scale down process.
+The operator will:
+
+1. Detect brokers scheduled for removal and block the scale down
+2. Cordon **all log directories** on those brokers by setting `cordoned.log.dirs=*` via the Kafka Admin API (using wildcard `*` to cordon all directories)
```
scholzj: Is it dynamic or read-only option?

146-broker-cordoning-scale-down.md
```diff
@@ -0,0 +1,380 @@
...
+- **Always check metadata version first**: Determines if cordoning is supported (Kafka 4.3+ with IBP_4_3_IV0+)
+- **Query all broker cordoning states**: Uses `describeLogDirs()` to check which brokers are currently cordoned
+- **Analyze and reconcile**: Compares current state against desired state (brokers being removed should be cordoned, others should not)
+- **Apply operations if needed**: Cordons brokers being removed, uncordons brokers no longer being removed (cleanup)
+- **Wait for cordoning propagation**: If operations were performed, revert scale down, and retry on next reconciliation
+
+Following a more detailed flow of what happens within the `KafkaClusterCreator` class as cordoning check.
+
+```
+├─ Check metadata version via describeFeatures()
```
scholzj: Doesn't the operator know the metadata version already?

scholzj (commented):
Thanks for the proposal. This would be a useful feature. However, I do not follow why the logic has to be some overcomplicated?

* The KafkaAssemblyOperator already knows any non-empty brokers that are being scaled-down.
* It can pass this information to KafkaReconciler, which configures the `cordoned.log.dirs` for the brokers that should be scaled-down (and does not configure it from the brokers that are ot being scaled down)

146-broker-cordoning-scale-down.md
ppatierno: That's true, the `KafkaCluster` instance exposes it but it's going to be the "desired" metadata version and not necessary the "current" one. I was thinking about a situation where the current cluster is Kafka 4.3 but metadata version is still 4.2 so no cordoning is supported. Then the user wants to upgrade the metadata and set metadata version to 4.3 in the Kafka CR. At this point the cordoning logic will think that cordoning is enabled because it's using the "desired" metadata version (4.3) and not the current one (4.2).
But it's also true that during an upgrade there would be nothing to cordon and the removed nodes collection will be empty, unless users start doing things at same time (upgrade + scale down) which sounds not so reasonable. I think we can simplify and use the metadata version we already know from the `KafkaCluster` instance.

146-broker-cordoning-scale-down.md
ppatierno: It's dynamic per broker, you don't need to roll the broker to take effect.

146-broker-cordoning-scale-down.md
ppatierno: AFAIU, when you update a broker with the `cordoned.log.dirs` dynamic configuration, it has to be propagated to the entire cluster first. While it's propagating, if you query the controller if the broker is cordoned, it could return false (even if you set the cordoning a few seconds ago) because propagation didn't complete. This is the reasoning why right after cordoning a broker we can't be sure it's really cordoned. At least it's my understanding but @mimaison can correct me on this.

scholzj: Is that really a blocker? We would not scale down the node immediately unless it is empty. Also, even if we do thing we need to check that it needs to be cordoned, I still don't follow the logic.

* You still need to integrate this into the regular configuration and rolling process otherwise it unsets your cordoning
* If you need to check for the node to be cordoned, wouldn't the check be better placed before the auto-balancing?

146-broker-cordoning-scale-down.md
scholzj: Fair enough. That said, I do not remember if the proposal suggested how to handle the different versions.

146-broker-cordoning-scale-down.md
scholzj: Ok, worth adding to the proposal I guess.

146-broker-cordoning-scale-down.md
ppatierno: >  I do not remember if the proposal suggested how to handle the different versions

If you are referring to the metadata versions, as you can see in the flow described above, if the metadata version is < 4.3, the cordoning logic is totally skipped while if >= 4.3, cordoning is supported and the operator proceed with the flow I described above.

The main point here is now deciding if we want do another admin request to the Kafka cluster to get the "real" metadata version or not (as I explained in the previous comment). Taking also into account that it would fail when the cluster is new and it's just starting up. In such case, we can check if the `status` section is missing and then using the metadata version exposed by the `KafkaCluster` model as you were suggesting.
But again reiterating what I said above:

> But it's also true that during an upgrade there would be nothing to cordon and the removed nodes collection will be empty, unless users start doing things at same time (upgrade + scale down) which sounds not so reasonable. I think we can simplify and use the metadata version we already know from the KafkaCluster instance.

Maybe we can just leave with the "desired" metadata version and avoid any other additional admin call to the cluster?

146-broker-cordoning-scale-down.md
ppatierno: Yeah, agreed. I added it in the motivation section when I mentioned the property for the first time.

146-broker-cordoning-scale-down.md
ppatierno: > Is that really a blocker? We would not scale down the node immediately unless it is empty.

Yeah but the goal was to avoid starting a rebalancing until the broker was fully cordoned because while propagation is happening, if we start rebalancing, some new partitions could be still added to the broker.
On the other side, maybe I was going to be overly cautious because the rebalancing was going to move such new partition and then the broker is cordoned eventually (after full propagation). Maybe it's better than adding some extra reconciliation cycles to be sure that cordoning happening.

> You still need to integrate this into the regular configuration and rolling process otherwise it unsets your cordoning

That's a good point, I missed. It should also auto-resolve the issue of "cleaning" cordoned brokers which are not removed anymore (the user cancel the scale down) so that should be "un-cordoned" back.

> If you need to check for the node to be cordoned, wouldn't the check be better placed before the auto-balancing?

If we want to be less cautious and not considering the cordoning propagation time as a big deal, this could be another solution. I will back to you with additional thoughts.

146-broker-cordoning-scale-down.md
ppatierno: Going to resolve this comment, because based on other feedback, I changed the approach and this part is not in place anymore.

146-broker-cordoning-scale-down.md
ppatierno: Going to resolve this comment, because based on other feedback, I changed the approach and this part is not in place anymore.

ppatierno:
@scholzj I totally changed the approach based on your comments. The proposal is now ready for another pass.

scholzj:
Sorry, I was busy with other stuff. I should be able to get to it today or latest during the weekend.

146-broker-cordoning-scale-down.md
```diff
@@ -0,0 +1,343 @@
...
+By automatically cordoning the brokers being removed, before initiating the rebalance, Strimzi gets the following:
+
+- Once cordoned, no new partitions can be assigned to brokers being removed
+- Cruise Control only needs to move existing partitions, not chase newly assigned ones
+- The scale down process would have an end when all existing partitions are moved
+
+This enhancement is particularly valuable for:
+
+- Large clusters with high topic creation rates
+- Production environments requiring predictable maintenance windows
```
scholzj: What exactly does this mean? I'm not sure how the rebalancing at scale-down relates to maintenance windows.

146-broker-cordoning-scale-down.md
```diff
@@ -0,0 +1,343 @@
...
+- Automated scaling scenarios where reliability is critical
```
scholzj: Not just here, but in the whole proposal, it might be clearer to talk about scale-down as that is only where it relates I assume? Or did I misunderstood something?

146-broker-cordoning-scale-down.md
```diff
@@ -0,0 +1,343 @@
...
+This time window can be considered negligible compared to the duration of a cluster rebalancing.
+
+The `cordoned.log.dirs` broker configuration property will be considered as **forbidden**.
+The users can't set it within the `Kafka` custom resource `spec.kafka.config` to prevent conflicts with operator-managed cordoning.
+Since broker configuration in Strimzi is common across all Kafka node pools, users cannot target specific brokers or log directories through the `Kafka` custom resource.
+The operator needs exclusive control over this property to manage the scale-down lifecycle correctly.
+
+Despite the cordoning feature is available from Apache Kafka 4.3, there is no need for explicit Kafka version gating.
+The existing `KafkaConfigurationDiff` relies on the Kafka config model, which is selected based on the Kafka binary version.
+On Kafka < 4.3, `cordoned.log.dirs` is not in the config model, so the diff treats it as a "custom config" and silently skips it so the property appears in the desired configuration but is never applied to the broker.
```
scholzj: Is that true? Unknown option in general trigger rolling updates AFAIK.

146-broker-cordoning-scale-down.md
```diff
@@ -0,0 +1,343 @@
...
+  ├─ brokerConfigurationConfigMaps(): generates per-broker ConfigMaps
+  │   └─ For brokers in scalingDownBlockedNodes, the desired config includes cordoned.log.dirs=*
+  └─ rollingUpdate(): KafkaRoller diffs current vs desired config
+      └─ Applies cordoned.log.dirs=* via incrementalAlterConfigs()
+
+KafkaAutoRebalancingReconciler.reconcile()   (runs after KafkaReconciler completes)
+  └─ Sees scalingDownBlockedNodes, creates KafkaRebalance with remove-brokers mode
+```
+
+#### Detailed step-by-step workflow
```
scholzj: I find the structure with:
* Overview
* Detailed workflow
* Even more detailes step-by-step workflow a a bit strange.  I wonder it the second and third steps can be merged?

(I personally can probably even live with the Detailed workflow and just have the step-by-step workflow removed. Not sure it adds much more detail really. But I guess it might for others.)

scholzj (commented):
Thanks for the changes. I like the current proposal and its simplicity. I left some comments about the proposal structure and some questions. But I think the direction is good.

146-broker-cordoning-scale-down.md
scholzj: I tried it and it does seem to roll the broker. I think you will need some gating in `KafkaBrokerConfigurationBuilder` or wait for Kafka 4.4.0 to drop. If we at east do not need to check the metadata version, it should no be super hard.

146-broker-cordoning-scale-down.md
ppatierno: Ok. I merged the two sections by moving some detailed information into the first one but removing others (for example the auto-rebalancing and scale down part are well known and there are no changes related to them within this proposal).

146-broker-cordoning-scale-down.md
ppatierno: You are right, I lost that. The `KafkaReconciler` is actually skipping it because as custom property it doesn't appear in the diff but we also have the broker configuration hash on the pod and it includes everything is custom, so it's going to change and causing the pod rolling. I changed by gating the version within the `KafkaBrokerConfigurationBuilder`.

146-broker-cordoning-scale-down.md
ppatierno: It's only about scale down and I fixed here. I was not able to find other places you are referring to. The proposal is specifically referring to scale-down operations. If other ambiguous "scaling" wording, please let me know where.

146-broker-cordoning-scale-down.md
ppatierno: Using "maintenance window" was a bad wording. I meant something like "bounded time" (which can't be predictable either). I changed the sentence accordingly.

scholzj (approved):
LGTM. Thanks. This would be useful feature. +1 binding.

im-konge (approved):
LGTM, thanks for the proposal. +1 binding

ShubhamRwt (approved):
This feature seems like a great addition to me

Frawless (approved):

tinaselenge (approved):
LGTM, thanks for the proposal.

ppatierno:
This proposal has now 4 +1 binding votes, and 1 +1 non-binding. I will merge it if no additional feedback by EOD of Monday 27th July.

ppatierno:
This proposal is now approved with 4 +1 (binding) votes, and 1 +1 (non binding) vote. Thanks all people who reviewed it.

PR #235, opened 2026-06-29
Add proposal for scheduled mode in auto-rebalance

BK202503:
Adds a `schedule` mode to `spec.cruiseControl.autoRebalance`. Complements the upcoming `imbalance` mode (#211): same list, same template / finalizer machinery, different trigger (cron expression instead of detected goal violation).

Motivation: time-driven prophylactic rebalances are still external today — `CronJob` + `kubectl apply` + cleanup duplicated per team.

Out of scope and alternatives are documented in the proposal body.

### DIFF
--- a/149-scheduled-rebalance.md
+++ b/149-scheduled-rebalance.md
@@ -1,0 +1,94 @@
+# Scheduled mode for auto-rebalance
+
+Add a `schedule` mode to `spec.cruiseControl.autoRebalance` so the Cluster Operator can run `KafkaRebalance` on a recurring cron expression. Opt-in.
+
+## Motivation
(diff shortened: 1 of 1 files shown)


scholzj (commented):
Thanks for the proposal. However, I think this is very premature. The imbalance proposal has still some way to go before it is finished and accepted. Then we need to implement it. And then we might want to first see how it actually works in practice before building on top of it.

BK202503:
Fair point — withdrawing this until #211 lands and we see how it behaves in practice. Will revisit then.

scholzj:
Great, thanks 👍

PR #236, opened 2026-07-02
Add proposal for waiting for pod scheduling before rolling update

acgtun:
### Type of Change

- New Proposal

### Description

This proposal adds an observation-based scheduling barrier between `StrimziPodSet` reconciliation and `KafkaRoller`: before the operator deletes any pod for a rolling restart, every pod desired by the cluster's `StrimziPodSets` must exist and be scheduled.

It closes a race in which a reconciliation that combines scale-up with a revision change (reopened by an operator-internal retry after a transient failure) lets `KafkaRoller` delete a pod while newly created pods are still unscheduled.
On clusters with node-pinned local storage (`WaitForFirstConsumer` local volumes plus hard anti-affinity), a new pod can permanently take the rolled pod's node, leaving it `Pending` forever and failing every subsequent reconciliation until manual PVC/pod deletion.
This was hit in production on Strimzi 0.47.0.

### Checklist

### DIFF
--- a/149-wait-for-pod-scheduling-before-rolling.md
+++ b/149-wait-for-pod-scheduling-before-rolling.md
@@ -1,0 +1,87 @@
+# Wait for all desired pods to be scheduled before starting a rolling update
+
+This proposal adds an observation-based scheduling barrier between `StrimziPodSet` reconciliation and `KafkaRoller`: before the operator deletes any pod for a rolling restart, every pod desired by the cluster's `StrimziPodSets` must exist and be scheduled.
+It closes a race in which a rolled pod using node-pinned local storage permanently loses its node to a pod created moments earlier, leaving it `Pending` forever and failing every subsequent reconciliation.
+
+## Current situation
+
+A single reconciliation of a `Kafka` custom resource can both create new pods (scale-up) and roll existing ones (revision change).
+Since [PR #10746](https://github.com/strimzi/strimzi-kafka-operator/pull/10746), `KafkaReconciler.podSet()` waits for the readiness of all pods in `KafkaCluster.addedNodes()` before `rollingUpdate()` starts `KafkaRoller`.
+
+This guard is derived from intent, not observation: `addedNodes()` comes from `NodeIdAssignment.toBeAdded()`, which compares `KafkaNodePool.status.nodeIds` with the desired replicas, and `updateNodePoolStatuses()` persists the *desired* node IDs early in the reconcile pipeline, before `podSet()` runs.
+That makes it non-idempotent across a failed reconciliation:
+
+1. Reconciliation A persists the new node IDs into `KafkaNodePool.status`, then fails at a later stage (for example a transient `Secret` error) before any pod is created.
+2. Reconciliation B rebuilds its model from the updated status, so `addedNodes()` is empty; it still creates the missing pods via the `StrimziPodSet` diff, but has nothing to wait for.
+3. `KafkaRoller` deletes its first pod while the scheduler is still placing the newly created pods.
+
+For most clusters an unfortunate placement is only a performance nuisance, but with node-local storage it is a deadlock.
+Local volumes (`local-path`, OpenEBS LVM, TopoLVM) with `volumeBindingMode: WaitForFirstConsumer` pin each pod to one node, typically combined with hard one-pod-per-host anti-affinity.
+While a pod is deleted for a roll it has no claim on its node: anti-affinity only counts pods present at scheduling time, and a bound `PersistentVolume` of a non-running pod repels nothing.
+If a new pod lands on the node in that window, its own local volume is provisioned there and the theft is permanent: the rolled pod's replacement is pinned to a taken node and stays `Pending` until a human deletes the squatting pod and its `PersistentVolumeClaim`.
+
+## Motivation
+
+This sequence was hit in production on Strimzi 0.47.0 by a single change that scaled a broker pool from 10 to 40 replicas and changed broker CPU resources.
+A reconciliation persisted the new node IDs, failed on a transient `Secret` error, and the retry created the 30 broker pods and started rolling in the same second.
+One new broker was scheduled onto the node of the just-deleted KRaft controller pod; the recreated controller was pinned there by its metadata volume and stuck `Pending`, the roll timed out with the quorum at two of three controllers, and every reconciliation failed until manual `PersistentVolumeClaim` and pod deletion.
+
+The operator is the right place to fix this.
+Users cannot avoid the trigger: GitOps applies scale-up and revision change as one manifest change, and the race is reopened by operator-internal retries invisible to the user.
+Kubernetes cannot fix this side either, because during the deletion window the replacement pod does not exist, so no scheduler state can protect the node (kubernetes/kubernetes [#128164](https://github.com/kubernetes/kubernetes/issues/128164), closed unresolved, and [#135771](https://github.com/kubernetes/kubernetes/issues/135771) track adjacent variants).
+Only the component that decides *when to delete* a pod can refuse to open the window while competing pods are unscheduled.
+
+## Proposal
+
+Add a scheduling barrier to `KafkaReconciler.podSet()`, after the `StrimziPodSet` reconciliation and before `rollingUpdate()`.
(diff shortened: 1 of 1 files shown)


scholzj (commented):
Thanks for the proposal. However, I do not think this is possible. The roller needs to be able to recover brokers that are not in a ready state. So waiting for everything to be ready before the rolling update would create deadlocks where the Pods never become ready without rolling and rolling never happens because they are not ready. For some situation, even the best-effort waiting for the new nodes is desired given the multilayered architecture of the Kafka cluster. So I do not think we can implement this without breaking way too many other things.

I'm also a bit puzzled by the problem you describe. When using local persistent storage, the volumes should be bound to a particular claims which are bound to a particular Pod. So typically:
* Each node has 1-N local PVs
* The pre-existing Kafka nodes should have these PVs bound to their PVCs which makes sure they are scheduled there again
* New Kafka nodes should be scheduled to the node which has free PVs that are not bound (which with typical setup in these architectures with one Kafka node per worker node would be different worker node)

So I'm not sure where exactly is the problem you describe coming from. If the storage is bound, it should not be taken by some other Pod.

acgtun:
> the



> Thanks for the proposal. However, I do not think this is possible. The roller needs to be able to recover brokers that are not in a ready state. So waiting for everything to be ready before the rolling update would create deadlocks where the Pods never become ready without rolling and rolling never happens because they are not ready. For some situation, even the best-effort waiting for the new nodes is desired given the multilayered architecture of the Kafka cluster. So I do not think we can implement this without breaking way too many other things.
> 
> I'm also a bit puzzled by the problem you describe. When using local persistent storage, the volumes should be bound to a particular claims which are bound to a particular Pod. So typically:
> 
> * Each node has 1-N local PVs
> * The pre-existing Kafka nodes should have these PVs bound to their PVCs which makes sure they are scheduled there again
> * New Kafka nodes should be scheduled to the node which has free PVs that are not bound (which with typical setup in these architectures with one Kafka node per worker node would be different worker node)
> 
> So I'm not sure where exactly is the problem you describe coming from. If the storage is bound, it should not be taken by some other Pod.

Thanks @scholzj for the detailed reply.
the issue happened to a controller and broker pods. the controller pod has much less storage, so the node left a large storage that a broker could potential scheduled (even thought we set every controller and broker are in a independent node, no colocation)

acgtun:
> Thanks for the proposal. However, I do not think this is possible. The roller needs to be able to recover brokers that are not in a ready state. So waiting for everything to be ready before the rolling update would create deadlocks where the Pods never become ready without rolling and rolling never happens because they are not ready. For some situation, even the best-effort waiting for the new nodes is desired given the multilayered architecture of the Kafka cluster. So I do not think we can implement this without breaking way too many other things.
> 
> I'm also a bit puzzled by the problem you describe. When using local persistent storage, the volumes should be bound to a particular claims which are bound to a particular Pod. So typically:
> 
> * Each node has 1-N local PVs
> * The pre-existing Kafka nodes should have these PVs bound to their PVCs which makes sure they are scheduled there again
> * New Kafka nodes should be scheduled to the node which has free PVs that are not bound (which with typical setup in these architectures with one Kafka node per worker node would be different worker node)
> 
> So I'm not sure where exactly is the problem you describe coming from. If the storage is bound, it should not be taken by some other Pod.

Thanks @scholzj for taking the time to review this.

On readiness: sorry if the text was unclear — the proposal does not wait for readiness before rolling, only for scheduling (PodScheduled condition is True, usually a matter of seconds). Waiting for readiness is listed under rejected alternatives for exactly the deadlock reason you describe. An unready-but-scheduled pod never blocks the roll, and Pending pods that the roll itself would restart are exempt and rolled first. The roll is only delayed when the same reconciliation has just created new pods that the scheduler has not placed yet.

On the storage model: you are describing static local PVs, where PVC binding pins existing pods and new pods can only bind free PVs on other nodes — in that model the problem indeed cannot happen. We use dynamic provisioning (OpenEBS LocalPV-LVM with WaitForFirstConsumer; TopoLVM and local-path behave the same). There is no fixed pool of PVs: a new pod's volume is created on whichever node the scheduler picks. The bound PV pins its own pod to the node, but it does not keep other pods away — only the podAntiAffinity rule does that, and it only counts pods that exist at scheduling time.

So while a pod is deleted for the roll, its node briefly looks free to the scheduler. In our incident a new broker was scheduled onto a just-deleted controller's node (the controller's small metadata volume left plenty of free capacity) and got its volume provisioned there. The recreated controller was pinned to that node by its bound PV but now rejected by anti-affinity, so it stayed Pending until we deleted the broker pod and its PVC manually.

scholzj:
> On readiness: sorry if the text was unclear — the proposal does not wait for readiness before rolling, only for scheduling (PodScheduled condition is True, usually a matter of seconds). Waiting for readiness is listed under rejected alternatives for exactly the deadlock reason you describe. An unready-but-scheduled pod never blocks the roll, and Pending pods that the roll itself would restart are exempt and rolled first. The roll is only delayed when the same reconciliation has just created new pods that the scheduler has not placed yet.

Sorry, my misunderstanding. But I do not think it changes much because even for unschedulable Pods, the KafkaRoller is the part that fixes it (imagine wrong resources configuration or some affinity / tollerations config). 

> On the storage model: you are describing static local PVs, where PVC binding pins existing pods and new pods can only bind free PVs on other nodes — in that model the problem indeed cannot happen. We use dynamic provisioning (OpenEBS LocalPV-LVM with WaitForFirstConsumer; TopoLVM and local-path behave the same). There is no fixed pool of PVs: a new pod's volume is created on whichever node the scheduler picks. The bound PV pins its own pod to the node, but it does not keep other pods away — only the podAntiAffinity rule does that, and it only counts pods that exist at scheduling time.
> 
> So while a pod is deleted for the roll, its node briefly looks free to the scheduler. In our incident a new broker was scheduled onto a just-deleted controller's node (the controller's small metadata volume left plenty of free capacity) and got its volume provisioned there. The recreated controller was pinned to that node by its bound PV but now rejected by anti-affinity, so it stayed Pending until we deleted the broker pod and its PVC manually.

Ok, thanks for the explanation. I can see that happen. But I'm afraid I do not know how to easily prevent it without breaking some other things. :-(

acgtun:
> > On readiness: sorry if the text was unclear — the proposal does not wait for readiness before rolling, only for scheduling (PodScheduled condition is True, usually a matter of seconds). Waiting for readiness is listed under rejected alternatives for exactly the deadlock reason you describe. An unready-but-scheduled pod never blocks the roll, and Pending pods that the roll itself would restart are exempt and rolled first. The roll is only delayed when the same reconciliation has just created new pods that the scheduler has not placed yet.
> 
> Sorry, my misunderstanding. But I do not think it changes much because even for unschedulable Pods, the KafkaRoller is the part that fixes it (imagine wrong resources configuration or some affinity / tollerations config).
> 
> > On the storage model: you are describing static local PVs, where PVC binding pins existing pods and new pods can only bind free PVs on other nodes — in that model the problem indeed cannot happen. We use dynamic provisioning (OpenEBS LocalPV-LVM with WaitForFirstConsumer; TopoLVM and local-path behave the same). There is no fixed pool of PVs: a new pod's volume is created on whichever node the scheduler picks. The bound PV pins its own pod to the node, but it does not keep other pods away — only the podAntiAffinity rule does that, and it only counts pods that exist at scheduling time.
> > So while a pod is deleted for the roll, its node briefly looks free to the scheduler. In our incident a new broker was scheduled onto a just-deleted controller's node (the controller's small metadata volume left plenty of free capacity) and got its volume provisioned there. The recreated controller was pinned to that node by its bound PV but now rejected by anti-affinity, so it stayed Pending until we deleted the broker pod and its PVC manually.
> 
> Ok, thanks for the explanation. I can see that happen. But I'm afraid I do not know how to easily prevent it without breaking some other things. :-(

Thanks @scholzj for the review. yes, this is rare edge case. It is easy to fix manually :)

scholzj:
Sorry. I would love to be able to help more. But to be honest, right now I do not have any good solution that would come to my mind :-/

acgtun:
> Sorry. I would love to be able to help more. But to be honest, right now I do not have any good solution that would come to my mind :-/

Thanks @scholzj . No sorry, thanks for your feedback.

PR #237, opened 2026-07-06
Add proposal for Maven mirror support for Kafka Connect builds

egyedt:
### Type of Change

Addition - New Proposal

### Description

Add proposal for Maven mirror support for Kafka Connect builds.

### Checklist

### DIFF
--- a/149-connect-build-maven-mirror.md
+++ b/149-connect-build-maven-mirror.md
@@ -1,0 +1,248 @@
+# Maven mirror support for Kafka Connect builds
+
+## Objective
+
+Add first-class support for configuring one or more Maven **mirror(s)** for the Kafka Connect build.
+When set, *all* Maven resolution performed during the build - connector artifacts, their transitive dependencies, and the Maven build plugins themselves (notably `maven-dependency-plugin`, used by `dependency:copy-dependencies`) - is routed through the configured internal repository (a corporate Nexus/Artifactory, a proxy, or an air-gapped mirror).
+The mirror is an explicit, optional, opt-in build setting.
+When it is not set, the generated build is identical to today.
+
+## Current situation
+
+Strimzi can build Kafka Connect images with additional connector plugins declared in `KafkaConnect.spec.build`.
+For artifacts of type `maven`, the operator generates a multi-stage `Dockerfile` (see `io.strimzi.operator.cluster.model.KafkaConnectDockerfile`, method `connectorPluginsPreStage`) that runs inside the `maven-builder` image (`quay.io/strimzi/maven-builder:latest` by default).
+For each Maven artifact the operator:
+
+1. Downloads the artifact `pom.xml` with `curl`.
+2. Writes a small Maven `settings.xml` file to `/tmp`.
+3. Runs `mvn dependency:copy-dependencies -s <settings.xml> ...` to fetch the artifact's transitive dependencies.
+4. Downloads the artifact `.jar` with `curl`.
+
+When a user sets `MavenArtifact.repository`, that URL is injected into the generated `settings.xml` as an ordinary `<repository>` inside an active profile.
+If no repository is set, it defaults to Maven Central (`MavenArtifact.DEFAULT_REPOSITORY`, `https://repo1.maven.org/maven2/`).
+The custom repository is added **only** to `<repositories>`:
+
+```xml
+<settings xmlns="https://maven.apache.org/SETTINGS/1.0.0">
+  <profiles>
+    <profile>
+      <id>download</id>
+      <repositories>
+        <repository>
+          <id>custom-repo</id>
+          <url>${customRepositoryUrl}</url>
+        </repository>
+      </repositories>
+    </profile>
+  </profiles>
+  <activeProfiles>
+    <activeProfile>download</activeProfile>
+  </activeProfiles>
+</settings>
+```
+
+Two properties of this design matter for this proposal.
+
+- **The `repository` field is, by design, the source of the connector artifact.**
+  It is not intended to be - and normally is not - a general Maven repository that hosts Maven *plugins*.
+  Conflating "where the connector JAR lives" with "where build tooling comes from" is a category error.
+- **There is no way to keep the build off Maven Central.**
+  Because `<pluginRepositories>` is never emitted, Maven resolves `maven-dependency-plugin` (and its dependencies) from the implicit `central` plugin repository defined in Maven's built-in super POM.
+  Even a user who has pointed every artifact at an internal repository still has a hidden, unavoidable dependency on `repo1.maven.org` for the build machinery.
+
+## Motivation
+
+In enterprise, air-gapped, or heavily mirrored environments, that hidden dependency on Maven Central is a frequent and hard-to-diagnose cause of Connect build failures:
+
+- **HTTP 429 (Too Many Requests):** Maven Central increasingly rate-limits anonymous traffic.
+  Shared egress IPs (corporate NAT/proxy) make this common, and the plugin resolution step fails even though every declared artifact is served locally.
+- **Network restrictions / air-gap:** Clusters with no route to `repo1.maven.org` cannot resolve the plugins at all, so the build never completes despite a fully populated internal mirror.
+- **Connection timeouts:** Locked-down egress and slow proxies cause the plugin download to time out.
+
+The failure is confusing to operators because the artifacts they explicitly configured resolve fine - only the build machinery silently depends on Central.
+This was raised in [strimzi-kafka-operator#8087](https://github.com/strimzi/strimzi-kafka-operator/issues/8087).
+
+The idiomatic Maven mechanism for "route everything through our internal repository" is a **mirror**, not an extra plugin repository.
+A mirror intercepts requests to the repositories it matches - including the implicit Central plugin repository - and redirects them to a single URL.
+This removes Maven Central from the resolution path entirely (rather than merely adding an alternative source and *hoping* the plugin happens to be present), which is what air-gapped and rate-limited environments actually require.
+Giving users a simple, declarative way to configure such a mirror is the clean, complete solution to this class of build failures.
+
+## Proposal
+
+Add an optional list of Maven mirrors to the `maven` artifact type in `KafkaConnect.spec.build`.
+When set, the operator writes each mirror into the `settings.xml` it already generates for the build and passes it to Maven via the existing `mvn -s <settings.xml>` invocation.
+Because a Maven mirror intercepts *every* repository request - including the super POM's implicit `central` used to resolve `maven-dependency-plugin` - this removes Maven Central from the build's resolution path entirely.
+
+Builds exist only in `KafkaConnect`, so this is the only resource affected.
+
+### API
+
+Extend the `io.strimzi.api.kafka.model.connect.build.MavenArtifact` model with a new optional `mirrors` field: a list of mirror objects.
+Each entry is an object (not a bare string) so that further per-mirror settings can be added later without a breaking change (see *Future work*).
+
+- `mirrors` (optional): list of `MavenMirror` objects.
+- `MavenMirror` - fields:
+  - `url` (required): base URL of the mirror.
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaConnect
+spec:
+  build:
+    output:
+      type: docker
+      image: my-registry.io/my-org/my-connect:latest
+    plugins:
+      - name: my-connector
+        artifacts:
+          - type: maven
+            group: com.example
+            artifact: my-connector
+            version: 1.0.0
+            mirrors:
+              - url: https://nexus.internal.example.com/repository/maven-public/
+```
+
+The `mirrors` list is set on the `maven` artifact type, alongside the existing `repository` field, because mirroring is a Maven-specific concern.
+It does not replace `repository`, so the two can be combined (see *Interaction with the existing connector `repository` field*).
+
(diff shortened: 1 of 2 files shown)


146-connect-build-maven-mirror.md
```diff
@@ -0,0 +1,252 @@
...
+This was raised in [strimzi-kafka-operator#8087](https://github.com/strimzi/strimzi-kafka-operator/issues/8087).
+
+The idiomatic Maven mechanism for "route everything through our internal repository" is a **mirror**, not an extra plugin repository.
+A mirror intercepts requests to the repositories it matches — including the implicit Central plugin repository — and redirects them to a single URL.
+This removes Maven Central from the resolution path entirely (rather than merely adding an alternative source and *hoping* the plugin happens to be present), which is what air-gapped and rate-limited environments actually require.
+Giving users a first-class, declarative way to configure such a mirror is the clean, complete solution to this class of build failures.
+
+## Proposal
+
+Add an optional, build-scoped Maven mirror configuration to `KafkaConnect.spec.build` (equivalently `KafkaConnector`/`KafkaMirrorMaker2` builds that reuse the same `Build` model).
```
scholzj: There are no `KafkaConnector`/`KafkaMirrorMaker2` builds. It is only in `KafkaConnect`.

146-connect-build-maven-mirror.md
```diff
@@ -0,0 +1,252 @@
...
+
+```yaml
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaConnect
+spec:
+  build:
+    output:
+      type: docker
+      image: my-registry.io/my-org/my-connect:latest
+    maven:
```
scholzj: TBH, I do not think we want to have Maven propagated to this level. The authentication is also something rather controversial as there are many different authentication methods that the API would need to support.

Why can't this be a `mirror` field or flag inside the Maven artifact?

E.g.:

```yaml
    plugins:
      - name: my-connector
        artifacts:
          - type: maven
            group: com.example
            artifact: my-connector
            version: 1.0.0
            repository: https://nexus.internal.example.com/repository/maven-public/
            mirror: true
```

or if we think mirror and repository might be something to configure in parallel:

```yaml
    plugins:
      - name: my-connector
        artifacts:
          - type: maven
            group: com.example
            artifact: my-connector
            version: 1.0.0
            mirror: https://nexus.internal.example.com/repository/maven-public/
```

That does not really handle the `mirrorOf`, but I think that is an acceptable simplification. It does not handle the authentication either. But that should be IMHO dealt with separately. Also, that is a concern for both repository and mirror.

So I think what I propose above would significantly simplify the effort and maintenance and make it more acceptable. @strimzi/maintainers WDYT?

146-connect-build-maven-mirror.md
```diff
@@ -0,0 +1,252 @@
...
+  </activeProfiles>
+</settings>
+```
+
+### Resolution semantics
+
+- With `mirrorOf: "*"`, Maven redirects **every** repository request — release and snapshot, artifact and *plugin* repositories, including the super POM's implicit `central` — to the mirror URL.
+  `maven-dependency-plugin` and its dependencies are therefore resolved from the internal mirror, and Maven Central is never contacted.
+  This is what makes the mirror a complete fix for both the rate-limit (429) and true air-gap cases.
+- `mirrorOf` is fully configurable, so a user who wants to mirror only Central while leaving other repositories direct can set `mirrorOf: "central"` (or `"external:*"` to exclude file/localhost repos), etc.
```
scholzj: I do not think file or localhost repos apply in this context really, or?

scholzj (commented):
Thanks for the proposal. I left few comments ...

146-connect-build-maven-mirror.md
```diff
@@ -0,0 +1,244 @@
...
+          - type: maven
+            group: com.example
+            artifact: my-connector
+            version: 1.0.0
+            repository: https://nexus.internal.example.com/repository/maven-public/
+            mirror: true
+```
+
+Because a Maven mirror is inherently a property of the whole build rather than of a single artifact, `mirror: true` affects resolution for the entire build, not only for the artifact it is declared on.
+Setting it requires an explicit `repository`: mirroring the default Maven Central URL onto itself would be a no-op, so the operator rejects `mirror: true` when no `repository` is set.
```
im-konge: Maybe something worth to mention - how the user will be notified about this? Will there be some CEL validation?

147-connect-build-maven-mirror.md
```diff
@@ -0,0 +1,244 @@
...
+  </activeProfiles>
+</settings>
+```
+
+The `<repository>` continues to be emitted exactly as today.
+The only addition is the `<mirror>`, whose `<url>` is the same `repository` URL and whose `<mirrorOf>` is `*`.
+
+### Resolution semantics
+
+- With the mirror in place (`mirrorOf: "*"`), Maven redirects **every** repository request — release and snapshot, artifact and *plugin* repositories, including the super POM's implicit `central` — to the mirror URL.
```
im-konge: I know that you want to limit the API changes and the scope just for the one and only mirror, but what will happen in case that some dependency will not be pullable from the `mirror`? Should there be some fallback instead of failing immediately because of missing dependency? Maybe something for different proposal?

README.md
im-konge, line 1: I wanted to say that you should keep the formatting as it was, but TBH even for me when I'm adding some new proposal it gets reformatted. So I wonder if we should keep it changed via IDE every time or we should always revert the changes around format.

im-konge (commented):
Thanks for the proposal, I think this will be useful even for our testing purposes - and what we need in STs, as we are hitting pull limits as well. I left few small comments.

146-connect-build-maven-mirror.md
im-konge: I mentioned the stuff around the simplicity (and about fallback and multiple mirrors) in one of my comments, but I wonder about one thing - in case that we would like to add more possible mirrors in the future, wouldn't it mean that we will need to add more field(s) and deprecate `mirror` which will be unused (but deprecated) until v2? Should we instead think about having a possibility to specify more mirrors now?

README.md
egyedt: ack

146-connect-build-maven-mirror.md
scholzj: Are multiple mirrors common? My understanding is that they typically mirror everything or even work as proxies that pull new things on demand. So it seemed to me that one mirror is fine.

* If we think multiple mirrors are a common thing, should have have instead a `List<String>` mirrors?
* Should we also allow combining mirrors with repositories? E.g. if you have a snapshot repository with your fresh PR build of the connector and mirror next to it pointing to the proxy/mirror with other dependencies?

(Sorry, I guess I suggested the mirror boolean flag. But given it is an API, it is probably worth considering carefully.)

README.md
scholzj: Actually, next time, the best thing to do is just avoid updating the README completely. We can updated it once it is approved and we know the number and the other proposals merged in between etc. (No need to change it now, just for next time)

146-connect-build-maven-mirror.md
ppatierno: I am not sure multiple mirrors is really something common but using a list of mirrors, as Jakub is suggesting, would leave the door open for such a specific request. Maybe in most of the cases it will just be one item in the list.

scholzj: If we would want to completely future proof it we could do something like

```
mirrors:
  - url: ...
```

That would allow us to add more fields in the future.

im-konge: Yeah that was something I was thinking about. I'm not saying that it is something common. I just think that there might be requests for adding multiple mirrors support, which would lead to another round of API deprecation. So I would be for:

```
mirrors:
  - url: ...
```

scholzj: @ppatierno ... Does that sound reasonable to you as well?

ppatierno: Yes, +1 from me

scholzj: @egyedt Could you please work this into the proposal? Sorry, this is how proposals go sometimes and there are changes on the fly.

egyedt: I added this! Thanks for the idea! :)

scholzj: Wait ... my assumption still was that this would be part of the Maven type artifact. I.e.:

```yaml
kind: KafkaConnect
spec:
  build:
    output:
      type: docker
      image: my-registry.io/my-org/my-connect:latest
    plugins:
      - name: my-connector
        artifacts:
          - type: maven
            group: com.example
            artifact: my-connector
            version: 1.0.0
            mirrors:
              - url: https://nexus.internal.example.com/repository/maven-public/
```

Not sure about @im-konge and @ppatierno. Sorry for the confusion.

im-konge: Yeah I would put it to the `artifact` directly - because it's something you want to have just for the `maven` type.
So as @scholzj mentioned.

ppatierno: That was my same expectation.

147-connect-build-maven-mirror.md
```diff
@@ -0,0 +1,259 @@
...
+            artifact: my-connector
+            version: 1.0.0
+            mirrors:
+              - url: https://nexus.internal.example.com/repository/maven-public/
+```
+
+The `mirrors` list is set on the `maven` artifact type, alongside the existing `repository` field, because mirroring is a Maven-specific concern.
+It does not replace `repository`, so the two can be combined (see *Interaction with the existing connector `repository` field*).
+
+`url` is a required field on each `MavenMirror`, enforced by structural CRD validation, with an additional CEL rule (`x-kubernetes-validations`) checking that it is a syntactically valid absolute URL.
```
scholzj: I would probably avoid this. Validating this would be hard and CEL validations should be limited in what they do. Validating something as complex as URL is also risky. We could validate it in the operator code. But we do not validate `repository` either and I do not think there are any issues with it.

147-connect-build-maven-mirror.md
```diff
@@ -0,0 +1,259 @@
...
+      </repositories>
+    </profile>
+  </profiles>
+  <activeProfiles>
+    <activeProfile>download</activeProfile>
+  </activeProfiles>
+</settings>
+```
+
+The `<repositories>` block continues to be emitted exactly as today.
```
scholzj: IIRC, today when the `repository` field is not set, we still configure the repoitory to `https://repo1.maven.org/maven2/`. Does that make sense when the user leaved repository unset and sets mirror to add explicit Maven Central repository? I think we would need to adjust this logic? E.g. If mirrors are set and repository is not, configure only mirrors?

149-connect-build-maven-mirror.md
```diff
@@ -0,0 +1,259 @@
...
+  TLS verification against the mirror follows normal Maven behaviour unless the user separately opts into insecure mode.
+- **Authentication deferred, not weakened.** Because mirror authentication is out of scope, this proposal introduces no new credential handling.
+  When credential support is added later (for both `repository` and mirror), it should follow the existing Strimzi `Secret`-reference pattern and be exposed only to the build step, taking care not to leak credentials into image layers or logs.
+
+## Future work
+
+- **Configurable `mirrorOf`.** An optional `mirrorOf` field on `MavenMirror`, defaulting to `*`, would let users leave selected repositories direct (e.g. `*,!custom-repo` to keep a snapshot repository unmirrored) or mirror only Central.
+- **Mirror (and repository) authentication.** First-class support for authenticated mirrors and repositories, via `Secret` references, handled uniformly for both, added as optional fields on `MavenMirror`.
+- **Proxy configuration.** Some environments require an HTTP(S) proxy rather than (or in addition to) a mirror; a proxy configuration could be added following the same opt-in pattern.
```
scholzj, line 214: So, are you actually planning to work on this? If not, it might make sense to add something like _if needed_. Because this makes it sound like you are planning to work on it.


```suggestion
Possible enhancements that can be added through future proposals if needed:

- **Configurable `mirrorOf`.** An optional `mirrorOf` field on `MavenMirror`, defaulting to `*`, would let users leave selected repositories direct (e.g. `*,!custom-repo` to keep a snapshot repository unmirrored) or mirror only Central.
- **Mirror (and repository) authentication.** First-class support for authenticated mirrors and repositories, via `Secret` references, handled uniformly for both, added as optional fields on `MavenMirror`.
- **Proxy configuration.** Some environments require an HTTP(S) proxy rather than (or in addition to) a mirror; a proxy configuration could be added following the same opt-in pattern.
```

scholzj (commented):
Thanks for the changes. I left some nits. But it looks good otherwise.

egyedt:
> Thanks for the changes. I left some nits. But it looks good otherwise.

Thanks @scholzj, I uploaded a fixed version based on your last NITs.

scholzj (approved):
LGTM. +1 binding. Thanks for the proposal.

PS: No need to squash the commits after every change - it makes it actually harder to see the changes 😉.

im-konge (approved):
LGTM, thanks for the proposal. +1 binding.

scholzj:
@ppatierno Do you habe anything more on this?

ppatierno (approved):

scholzj:
This has now 3 binding +1 votes. If there are no more comments till Tuesday EOB, I will merge it as approved. If anyone needs more time, please let us know.

scholzj:
Approved with 3 binding +1 votes.

scholzj:
Thanks for the proposal @egyedt

egyedt:
Thanks guys, and thanks @scholzj for merging it!

PR #238, opened 2026-07-07
Add proposal for per-topic storage budgets with principal-level enforcement

acgtun:
### Type of Change

- New Proposal

### Description

This proposal extends the kafka-quotas-plugin so that operators can assign storage budgets to individual topics.
When a topic exceeds its budget, only the principals writing to that topic are throttled, while every other tenant continues at full rate.
The existing cluster-wide out-of-disk protection from proposal 047 remains in place as a backstop.

### Checklist

### DIFF
--- a/147-per-topic-storage-budgets.md
+++ b/147-per-topic-storage-budgets.md
@@ -1,0 +1,137 @@
+# Per-topic storage budgets with principal-level enforcement
+
+This proposal extends the [kafka-quotas-plugin](https://github.com/strimzi/kafka-quotas-plugin) so that operators can assign storage budgets to individual topics.
+When a topic exceeds its budget, only the principals writing to that topic are throttled, while every other tenant continues at full rate.
+The existing cluster-wide protection remains in place as a backstop.
+
+## Current situation
+
+Since [proposal 047](https://github.com/strimzi/proposals/blob/main/047-cluster-wide-volume-usage-quota-management.md), the kafka-quotas-plugin provides cluster-wide out-of-disk protection.
+`VolumeSource` polls `Admin#describeLogDirs` (KIP-827) every `storage.check.interval` to collect the total and available bytes of each volume.
+If any volume breaches `storage.per.volume.limit.min.available.{bytes,ratio}`, the throttle factor drops to 0.0 and `StaticQuotaCallback` cuts the produce rate of all non-excluded principals to roughly 1 byte/s (Kafka cannot represent a zero quota).
+
+This design is intentional: proposal 047 scoped the plugin as a last line of defense against running out of disk, and it fills that role well.
+As a consequence, the enforcement decision is binary and global: the plugin does not track which topic is consuming the disk and has no way to act on a subset of clients.
+The per-principal control it offers is a static produce/fetch rate that is identical for all principals and independent of storage state.
+
+## Motivation
+
+Consider a large multi-tenant cluster with hundreds of brokers, multi-terabyte volumes, short retention, and high per-broker ingest, protected only by the cluster-wide volume limit.
+
+The cluster-wide protection reliably prevents the disk from filling up, which is exactly its job, but the blast radius of enforcement is the whole cluster.
+A single tenant ramping up a new topic can cross a volume threshold quickly, at which point produce freezes for every tenant on every broker.
+A single-tenant incident becomes a full-cluster producer outage.
+
+Several existing mechanisms come close, but each was designed for a different problem:
+
+- Per-user quotas (`KafkaUser.spec.quotas` or the built-in callback) cap request rates, but they cannot be combined with this plugin, because Kafka allows only one `client.quota.callback.class` per broker.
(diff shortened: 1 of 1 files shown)


im-konge:
@acgtun thanks for the proposal, I didn't go through it, but I think it would be better to finish the previous proposals first before opening another - given that there are some comments from maintainers on previous proposals. Also, I'm not sure that there will be enough time to review these new proposals - based on how much time maintainers have. We should instead focus on one or two and move them forward. Thanks.

acgtun:
> @acgtun thanks for the proposal, I didn't go through it, but I think it would be better to finish the previous proposals first before opening another - given that there are some comments from maintainers on previous proposals. Also, I'm not sure that there will be enough time to review these new proposals - based on how much time maintainers have. We should instead focus on one or two and move them forward. Thanks.

Thanks @im-konge for the suggestions. I closed some proposals and rework on some others. thanks

im-konge:
> > @acgtun thanks for the proposal, I didn't go through it, but I think it would be better to finish the previous proposals first before opening another - given that there are some comments from maintainers on previous proposals. Also, I'm not sure that there will be enough time to review these new proposals - based on how much time maintainers have. We should instead focus on one or two and move them forward. Thanks.
> 
> Thanks @im-konge for the suggestions. I closed some proposals and rework on some others. thanks

Thanks, just to be clear, I'm not saying that we shouldn't continue with the proposals - depending on opinions from others - but having just one or two and move them forward will get them actually over the line quicker than having more of them. Thanks for the work and your suggestions :) I will have a look on the proposals as soon as I can :)

ppatierno:
I am sorry but I don't plan to review this proposal until the other proposals will be addressed (merged or refused) and implemented eventually. Keep opening proposals is not a good practice in my opinion, even because after a proposal there will be the implementation to review as well.

acgtun:
Hi @im-konge and @ppatierno totally agree with you.Thanks for all your feedbacks, appreciated

147-per-topic-storage-budgets.md
```diff
@@ -0,0 +1,137 @@
...
+A true per-topic rate quota is therefore not achievable within the quota callback contract and would require a new interception point in Apache Kafka, which means a KIP.
+The callback API is also moving away from topic awareness: `updateClusterMetadata(Cluster)`, its only topic-aware hook, was never supported in KRaft mode and is deprecated in Kafka 4.4 for removal in 5.0 ([KIP-1200](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=373887083)).
+This proposal therefore detects per topic using the log dir data, but enforces per principal using the existing quota path, and does not depend on `updateClusterMetadata`.
+
+### Principal-to-topic mapping
+
+The plugin needs to know which principals write to a topic, and supports two sources for this mapping:
+
+- `config` (the initial mode): the operator declares `topicPattern → principals` explicitly.
+  This is deterministic and auditable, and provisioning systems (as well as Strimzi `KafkaUser`/`KafkaTopic` ownership) already hold this data.
```
scholzj, line 57: I do not think they really hold the data.
* They are optional
* Their management is completely decoupled from the Kafka cluster or quota configuration

So I do not think they provide any useful information.

147-per-topic-storage-budgets.md
```diff
@@ -0,0 +1,137 @@
...
+- `acl` (optional, additive): the plugin resolves the principals with `WRITE` access on the topic via `describeAcls`, on the same poll cadence.
```
scholzj, line 58: Same as above, ACLs are often hidden or using 3rd party authorizers, so their data live somewhere else. I do not think you can build anything reliable based on them.

scholzj (commented):
Thanks for the proposal. I left some comments. In general:
* I'm not sure you can that easily use ACLs or the Strimzi custom resources to track this. You would need to have some detailed configuration. 
* I'm not sure I really follow the per-topic configuration. Topic ultimately consists of partitions. The partitions are distributed across nodes. And different clients might write to different partitions, different partitions might have different sizes, etc. So the logic of blocking based on topic size is not really clear to me. It seems like there would be a lot of edge-cases where this would be not desired.
* I do not really follow why the `retention.bytes` is not a good measure for this. It applies per-partition. But given how Kafka works, that does not seem like a significantly worse choice then trying to enforce the quota per-topic. (I'm not necessarily saying you are wrong ... just that I do not understand it from the proposal)

acgtun:
> * do not really follow why the `retention.bytes`

Thanks @scholzj for the feedback. I will close the proposal for now and reopen after I have more thoughts. thanks again.

scholzj:
Just one more thought I had about this. Sadly, we did not have much time to spend on the Quotas plugin. So it has not been touched for a long time. Something to consider would be also adding a good support for custom Quotas plugins - something we are currently missing if I remember correctly. That would also allow you to have your own Quotas plugin that makes sense to you regardless whether it makes sense to the whole community or not.

PR #240, opened 2026-07-14
Migrate Strimzi images from ubi9-minimal to ubi10-micro

Frawless:
### Type of Change

- New Proposal

### Description

This proposal covers migration from ubi9-minimal to ubi10-micro in Strimzi images.

### Checklist


scholzj (commented):
@Frawless Any chance you have somewhere some implementation code to see how the changes look like?

147-use-ubi10-as-base-image.md
```diff
@@ -0,0 +1,188 @@
...
+COPY --from=downloader /usr/bin/tini /usr/bin/tini
+```
+
+**base/Dockerfile — after (ubi10-micro):**
+```dockerfile
+FROM registry.access.redhat.com/ubi10/ubi-minimal:latest AS downloader
+# ... download Tini (unchanged) ...
+
+# Install runtime dependencies into a chroot
+FROM registry.access.redhat.com/ubi10/ubi-minimal:latest AS builder
```
scholzj: It is really more of an implementation detail. But do we need separate downloader and builder images? Can't it be just one stage?

Frawless:
> @Frawless Any chance you have somewhere some implementation code to see how the changes look like?

You can check it here - https://github.com/Frawless/strimzi-kafka-operator/tree/use-ubi10  please ignore the build script as it was just for pushing multi-arch images and manifests with podman, this won't be part of the changes.

147-use-ubi10-as-base-image.md
```diff
@@ -0,0 +1,188 @@
...
+### Wolfi base image
+
+[Wolfi OS](https://edu.chainguard.dev/open-source/wolfi/overview/) is used as the base in images produced by [Chainguard](https://edu.chainguard.dev/chainguard/chainguard-images/overview/).
+They also offer Strimzi hardened images.
+Wolfi is an open source project [licensed under Apache 2.0](https://edu.chainguard.dev/open-source/wolfi/faq/) which is fine for us.
+Chainguard describes their images as distroless, specially curated to run in cloud-native environments.
+
+However, there are a couple of differences that make it not suitable for us:
+- it uses `apk` instead of `dnf`/`microdnf` so we would need to rewrite most of our Dockerfiles
+- it is [supported only on `amd64` and `arm64`](https://edu.chainguard.dev/chainguard/chainguard-images/overview/#architecture/)
```
scholzj: My understanding also is that the free tier allows us to use only the latest version of the images. We cannot pick the version. So if new version breaks things, we have a problem. That is IMHO worth mentioning and is bigger blocker than `apk`.

scholzj (commented):
I left some nits. Also:
* Can you please cover what we would do with the outlier images?
    * Kaniko is heading to recycle bin, so nothing to do
    * Buildah I assume no major change either
    * (Although even that might be worth mentioning)
    * But the Maven Builder matters in this context and should be covered

152-use-ubi10-as-base-image.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Minimal vs Micro
+
+The key difference is that `ubi-micro` excludes the package manager (`microdnf`) and all of its dependencies.
+This makes it Red Hat's [distroless](https://www.redhat.com/en/blog/introduction-ubi-micro) container image - built from the same RHEL packages but without the packaging tools.
+
+| Feature                      |        `ubi-minimal`         |              `ubi-micro`               |
+|------------------------------|:----------------------------:|:--------------------------------------:|
+| Package manager              |          `microdnf`          |                  None                  |
+| Shell (`bash`)               |             Yes              |                   No                   |
```
ppatierno, line 33: Can we get bash in our images anyway?

Frawless: Check line 78. It is installed in builder and copied to runner.

147-use-ubi10-as-base-image.md
Frawless: Yes, I updated the proposal with just one stage for download and install.

147-use-ubi10-as-base-image.md
Frawless: I added a bullet about it.

Frawless:
> I left some nits. Also:
> 
> * Can you please cover what we would do with the outlier images?
>   
>   * Kaniko is heading to recycle bin, so nothing to do
>   * Buildah I assume no major change either
>   * (Although even that might be worth mentioning)
>   * But the Maven Builder matters in this context and should be covered

I tried use `ubi10-micro` also for maven-builder, it seems to be working fine with just JRE so I updated the proposal with that as well + buildah/kaniko.

scholzj (approved):
LGTM, this would be useful and the CVE improvements look promising! +1 Binding. Thanks

147-use-ubi10-as-base-image.md
```diff
@@ -0,0 +1,297 @@
...
+More details can be found in the following sources: [RHEL 10 — Types of container images](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/10/html/building_running_and_managing_containers/types-of-container-images), [Introduction to UBI Micro](https://www.redhat.com/en/blog/introduction-ubi-micro), [UBI 10 Micro catalog](https://catalog.redhat.com/en/software/containers/ubi10-micro/66f2abd91123095c735db44f)
+
+### Changes in Dockerfiles
+
+Because `micro` does not contain `microdnf` we need to handle the package installation in a builder stage and then copy the installed packages to the runtime image.
+This is done using `microdnf --installroot` which installs packages into a chroot directory, and then the chroot contents are copied into the `ubi-micro` runtime stage.
+
+#### Base image + operators
+
+**base/Dockerfile — before (ubi9-minimal):**
```
see-quick: Can you make it here more explicit (i.e., title saying `before` and another for `after`)? Just a minor suggestion, but it would improve readability.

152-use-ubi10-as-base-image.md
```diff
@@ -0,0 +1,297 @@
...
+| Base image size (compressed) |            ~33 MB            |                ~7.5 MB                 |
```
see-quick, line 34: Out of curiosity, but with those changes with builder our modified `ubi-micro` image would be possibly same size or near ubi-minimal I assume?

see-quick (approved):
LGTM 👍, just a few questions. Also, I am wondering whether it makes sense to move the ubi9 images to ubi10 in test-containers.

152-use-ubi10-as-base-image.md
Frawless: I think the final images are around ~20 MB smaller than current ones (according Quay, feel free to check it, the links are in the proposal), but the CVE surface is much smaller. Maybe this metrics is useless and should be then removed from the proposal?

147-use-ubi10-as-base-image.md
Frawless: I updated the wording a little bit.

Frawless:
> LGTM 👍, just a few questions. Also, I am wondering whether it makes sense to move the ubi9 images to ubi10 in test-containers.

I think it depends on how much work will be needed, because I don't think test container is something that will run in production and it requires to do CVE respins in case of new CVEs within base image.

ShubhamRwt (approved):
LGTM

152-use-ubi10-as-base-image.md
see-quick: > Maybe this metrics is useless and should be then removed from the proposal?

No need. I was just curious about the sizing.

152-use-ubi10-as-base-image.md
```diff
@@ -0,0 +1,297 @@
...
+
+All projects that produce images are affected:
+- `strimzi-kafka-operator`
+- `strimzi-kafka-bridge`
+- `drain-cleaner`
+- `test-clients`
+- `test-container`
+- `client-examples`
+- `kafka-access-operator`
+- `mqtt-bridge`
```
ppatierno, line 277: but the proposal doesn't provide any details about the images outside the operator repository. I think we should have sections for them with the proposal covering them all.

ppatierno (changes requested):

152-use-ubi10-as-base-image.md
Frawless: I pushed small change. I don't want to add section for each image we have in the org, it would basically means I will implement everything as part of this proposal. All projects can use same mechanism and same code changes that are shown for operators repo as examples. I assume that operators repo will migrate as first so then the other projects can just copy/paste stuff from there (operators images will have much bigger testing coverage so the changes will be well tested I think). The changes I pushed should be fine form my POV, but if you think more details are needed, we can talk more about it.

147-use-ubi10-as-base-image.md
```diff
@@ -0,0 +1,314 @@
...
+
+Images in the other projects can use the same way how to migrate from their current base images to `ubi10-micro`.
+The mechanism will consist from multi-stage build, and it could look like the following:
+- use `ubi10-minimal` as a base and name it as `builder`
+- download tini (if used)
+- install all required packages with `--installroot /mnt/rootfs` option
+- use `ubi10-micro` as a base for final image
+- copy all dependencies from `/mnt/rootfs`
+- copy built Java artifacts
+- create strimzi user and set proper rights
```
ppatierno: I like this addition but at this point why not putting this section at the top? I mean, the proposal provides this section as the overall approach of moving all Strimzi images to use ubi10-micro and following it, you provide more details about the Strimzi operator image as an example. Also not sure if we should go into details of Kafka and Maven builder images instead of the operator one only as an example.
For me the proposal could be:
* this is the general approach and changes to images build
* take the Strimzi operator as an example.

Frawless: I moved the section above the examples. I would keep kafka and maven builder examples, there are small differences and it makes sense to me to have it there.

ppatierno: > there are small differences and it makes sense to me to have it there.

Are there such small differences with other projects as well? Then why Kafka and Maven images should be special and being here compared to the images from other projects? For me, the proposal should have just one example (the operator) and the overall process you moved at the beginning (thanks!). I am not sure what makes some images to be "special" compared to others to be included or not in the proposal. It looks like we want to cover the operator (as the main repo) but leaving the other projects as "relegated".

Frawless: Maven builder currently doesn't use ubi so it makes sense to me to have it there. Kafka extends operators base image and then do the copy part. I expect that all other projects will use one of these tree ways - change to multistage build, migrate from non-ubi image, extend base image and then build final one (guess this is only for operators probably).

152-use-ubi10-as-base-image.md
```diff
@@ -0,0 +1,317 @@
...
+
+However, it should be evaluated if it makes sense to use `ubi10-micro` in testing projects like `test-clients`, `test-container`, or in `client-examples`.
```
im-konge, line 279: Just a nit to this, I don't think that we need to have `micro` for any of these three projects. `test-clients` and `test-container` are just something used in testing, nothing that would be used in prod -> we are not doing any CVE respins or something like that.

And `client-examples` is just set of examples, which anyway nobody updated for quite a long time.

im-konge (approved):
LGTM, thanks for the proposal. +1 binding.

152-use-ubi10-as-base-image.md
Frawless: I agree, it is basically up to project owners from my pov. In case anyone will want to implement it and it won't have any additional maintenance burden then why not.

ppatierno (approved):

Frawless:
This has now 5 binding +1 votes. If there are no more comments till Tuesday EOB, I will merge it as approved. If anyone needs more time, please let me know.

PR #241, opened 2026-07-31
Add proposal for Strimzi-native configurable RSA key sizes

kyguy:
### Type of Change

- New Proposal

### Description

This proposal adds a `keySize` property to the `CertificateAuthority` configuration in the `Kafka` custom resource allowing users to configure the RSA key size used by Strimzi's internal CAs for both CA keys (`clusterCa` and `clientsCa`) and the certificate keys they issue.

Addresses: https://github.com/strimzi/strimzi-kafka-operator/issues/12769

### Checklist


152-configurable-ca-key-size.md
```diff
@@ -0,0 +1,200 @@
...
+
+* **Default value:** `4096`, the current root certificate key size, preserving backward compatibility for root certificate keys.
+  This will change the effective key size for leaf certificates from 2048 bits, OpenSSL's implicit default, to 4096 bits.
+  This is a deliberate change: relying on an implicit OpenSSL default is fragile and aligning leaf certificate keys with the root certificate key size is a safer baseline.
+  For users who need the previous 2048-bit leaf key size, they can set `keySize: 2048` explicitly.
+
+* **Allowed values:** `2048`, `3072`, `4096`, `8192` which are standard RSA key sizes that are well-supported by OpenSSL and Java's `java.security` libraries.
+  These values will be enforced by CRD schema validation using an `enum` constraint.
+
+* **Minimum value:** `2048` since smaller RSA keys do not meet any current security standard and therefore should not be supported.
```
scholzj: Is there any actual technical reason why to control these so tightly? The regulations you listed earlier - some of them suggest `≥3000` as the size. So I assume 3000 bit size would work technically?

152-configurable-ca-key-size.md
```diff
@@ -0,0 +1,200 @@
...
+
+Following the existing pattern established by `validityDays` we would add a `keySize` field to `CertificateAuthority.java` like this:
+
+```java
+public class CertificateAuthority implements UnknownPropertyPreserving {
+    //... existing fields
+    public static final int DEFAULT_KEY_SIZE = 4096;
+
+    // ... existing fields 
+    private int keySize;
```
scholzj, line 121: Didn't you said earlier that it is an Enum? To be clear, I would probably prefer int without restrictions (see the previous comment). But trying to figure out the inconsistency.

152-configurable-ca-key-size.md
```diff
@@ -0,0 +1,200 @@
...
+When a user changes the `keySize` on an existing cluster, the behavior depends on the `certificateExpirationPolicy`:
+
+* **No immediate key regeneration.** Changing `keySize` alone does not trigger an immediate root key replacement or leaf certificate regeneration.
+  The new key size takes effect the next time a key is generated during root key replacement (triggered by `certificateExpirationPolicy: replace-key` at renewal time or a manual force-replace annotation) or when new leaf certificates are issued (e.g. scaling up, adding a new `KafkaUser`).
+* **Existing certificates remain valid.** Certificates generated with the previous key size continue to function until they are naturally renewed or replaced.
+  This avoids unnecessary rolling restarts.
+* **Gradual rollout.** To immediately apply a new key size to all certificates, the user should:
+  1. Set the desired `keySize` in the `Kafka` resource.
+  2. Trigger a root key replacement using the `strimzi.io/force-replace` annotation on the root key Secret.
+  This will generate a new root key with the configured size and trigger re-issuance of all leaf certificates followed by rolling restarts.
```
scholzj: I assume the same applies to the changed default during upgrade when this proposal is implemented? Should be clear from the proposal.

152-configurable-ca-key-size.md
```diff
@@ -0,0 +1,200 @@
...
+## Compatibility
+
+This change is **backward compatible**:
+
+* The `keySize` property defaults to `4096` when not set, matching the current root key size.
+* For leaf keys, the effective default changes from 2048 (OpenSSL implicit) to 4096 (Strimzi explicit).
+  This is a behavioral change, but it is in the direction of stronger security and only affects newly generated leaf keys, the existing certificates are not regenerated.
+  Users who need to preserve 2048-bit leaf keys can set `keySize: 2048` explicitly.
+* No existing `Kafka` custom resources need to be modified.
+* No rolling restarts are triggered by the introduction of this feature alone.
```
scholzj, line 184: What happens when downgrading from Strimzi with default 4096 bit key to older version with default 2048? I assume the old 4096 bit keys will continue to be used until the renewal? Worth clarifying in the proposal.

scholzj (commented):
Thanks for the proposal @kyguy. I left some nits, but it looks mostly good.

152-configurable-ca-key-size.md
kyguy: > Is there any actual technical reason why to control these so tightly?

I was initially thinking of it as a guardrail for users. That being said, if users are changing this configuration, they probably know what they are doing and would appreciate the flexibility.

> So I assume 3000 bit size would work technically?

That is a fair point. 3072 would make a better default as it meets security compliance requirements and is better for performance.

I have just removed the enum restriction and updated the default keySize to `3072` in the latest commit.

152-configurable-ca-key-size.md
kyguy: Yes, that was oversight, I intended it as an enum as a guardrail for users. That being said, if users are changing this configuration, they probably know what they are doing and would appreciate the flexibility.

I have just removed the enum restriction in the latest commit.

152-configurable-ca-key-size.md
kyguy: Yes, thanks for noticing, you are right, I have just added a line for this in the latest commit

152-configurable-ca-key-size.md
kyguy: Yes, thanks for noticing, you are right, I have just added a line for this in the latest commit.

152-configurable-ca-key-size.md
scholzj: You can use documentation to point users to the common sizes (frankly, never heard about anyone actually using 3072)

kyguy: That sounds good to me, Just added an additional note for this in the latest commit.

152-configurable-ca-key-size.md
```diff
@@ -0,0 +1,200 @@
+# Strimzi-native Configurable Certificate Key Strength
+
+This proposal adds a `keySize` property to the `CertificateAuthority` configuration in the `Kafka` custom resource allowing users to configure the RSA key size used by Strimzi's internal CAs for both the CA keys themselves and the certificate keys they issue.
+For differentiating between CA certificates and certificates **issued** by a CA, the proposal uses the following terms: 
+- **Root certificates**: The CA certificates.
+- **Leaf certificates**: The certificates **issued** by a CA, also known as end-entity (EE) certificates, the certificates used to identify the Kafka brokers, Strimzi Operators, Cruise Control, and KafkaUser components.
```
ppatierno: "KafkaUser components" ... what are they? I guess CO, TO and UO are included in "Strimzi Operators"?

152-configurable-ca-key-size.md
```diff
@@ -0,0 +1,200 @@
...
+
+The `keySize` property controls the RSA key size in bits for **all** keys generated by Strimzi, both the root certificates and the leaf certificates.
+
+This follows the same pattern as `validityDays` which also applies to all certificates managed under the CA.
+
+#### Defaults and validation
+
+* **Default value:** `3072`, which meets all current and upcoming regulatory requirements (≥3000-bit RSA keys).
+  This changes the effective key size for root keys from `4096` (Strimzi's current hardcoded value) to `3072`.
+  This will change the effective key size for leaf certificates from `2048` bits, OpenSSL's implicit default, to `3072` bits.
```
ppatierno: I can understand that 3072 is big enough for the requirements but to be honest using it as default means reducing the current security of the root certificates (while increasing the security for leaf certificates only). I would default to 4096. AFAICS the mathematical max key lenght for RSA seems to be 15360 so users could be over or less but for me default should be 4096.

152-configurable-ca-key-size.md
kyguy: > "KafkaUser components" ... what are they?

Let me use "clients" here instead of components

>  I guess CO, TO and UO are included in "Strimzi Operators"?

Yes!

Just updated the text in latest commit to make this more clear

152-configurable-ca-key-size.md
kyguy: Just updated default to 4096 bits in latest commit.

152-configurable-ca-key-size.md
```diff
@@ -0,0 +1,195 @@
...
+* For leaf keys, the effective default changes from `2048` (OpenSSL implicit) to `4096` (Strimzi explicit).
+  This is a behavioral change, but it is in the direction of stronger security and only affects newly generated leaf keys, the existing certificates are not regenerated. Users who need to preserve 2048-bit leaf keys can set keySize: 2048 explicitly.
+* No existing `Kafka` custom resources need to be modified.
+* No rolling restarts are triggered by the introduction of this feature alone.
+
+## Rejected alternatives
+
+### Separate key sizes for root and leaf certificates
+
+Having two properties (e.g., `caKeySize` and `certKeySize`) was considered to allow independent control over root and leaf key sizes.
```
tinaselenge, line 190: We have this [issue](https://github.com/strimzi/strimzi-kafka-operator/issues/12765) raised to address a problem with validityDays and renewalDays by separating these internally for at least KafkaUser certificates. This might be something to consider for the implementation but also makes me wonder if we should reconsider separating CA cert and end-entity cert configurations in general.

152-configurable-ca-key-size.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+The `keySize` property controls the RSA key size in bits for **all** keys generated by Strimzi, both the root certificates and the leaf certificates.
+
+This follows the same pattern as `validityDays` which also applies to all certificates managed under the CA.
+
+#### Defaults and validation
+
+* **Default value:** `4096`, the current root certificate key size, preserving backward compatibility for root certificate keys.
+  This will change the effective key size for leaf certificates from `2048` bits, OpenSSL's implicit default, to `4096` bits.
```
tinaselenge, line 89: I am assuming that the reason we had 2 different keySize is that root CA keys are long-lived and high-impact if compromised, so they should be the strongest. Whereas, leaf keys are short-lived and is likely to be renewed frequently, so smaller keys are a reasonable trade-off for better TLS handshake performance. The default keySize would be increased to 4096, even though 3072 typically meets security requirements, so that we don't compromise root CA key strength. However, leaf keys are used on every TLS handshake, so this could have a performance impact for clients that may reconnect frequently? Also, are there scenarios where compliance regulations apply differently to CA keys vs leaf keys, requiring stronger keys for the CA but allowing smaller keys for leaf certs?

This make me wonder if we should add a per-KafkaUser keySize override for clients CA, similar to how validityDays and renewalDays work today after #214. This way users can configure smaller keySize for user certificates without affecting the CA key strength. What do you think?

tinaselenge (commented):
Thanks @kyguy for the proposal. It mostly looks good to me but I've left a couple of comments.

152-configurable-ca-key-size.md
kyguy: > This might be something to consider for the implementation

Thanks for the heads up. I'll pay close attention to this during the implementation regardless of whether the implementation lands before or after the CA refactor work. (I'll coordinate with you and Kate on this)

> also makes me wonder if we should reconsider separating CA cert and end-entity cert configurations in general.

I was thinking about this as well while putting this proposal together.  Separating the configurations would give users more flexibility; however, I'm not sure there has been much ask for it. Given it would require an API redesign and add complexity to the codebase, and there hasn't been much demand for separate configurations (as far as I know) I would leave it as it is for now.

152-configurable-ca-key-size.md
kyguy: > I am assuming that the reason we had 2 different keySize is that root CA keys are long-lived and high-impact if compromised, so they should be the strongest. Whereas, leaf keys are short-lived and is likely to be renewed frequently, so smaller keys are a reasonable trade-off for better TLS handshake performance. 

I believe the CA keys were set to `4096` for security impact reasons, but I'm not sure the leaf keys were kept smaller for performance reasons. Since the leaf keys use OpenSSL's default, I think the decision was effectively outsourced to OpenSSL rather than being a deliberate performance trade-off.

> The default keySize would be increased to 4096, even though 3072 typically meets security requirements, so that we don't compromise root CA key strength. However, leaf keys are used on every TLS handshake, so this could have a performance impact for clients that may reconnect frequently?

It could but I think for most Kafka clusters the impact would be minimal since Kafka connections are long-lived and reconnections should be infrequent. Users could always set the key size lower and bring their own CA with a higher key size if this was a requirement. I imagine if a user really needed tighter security for the CA, they would bring their own CA anyway.

> Also, are there scenarios where compliance regulations apply differently to CA keys vs leaf keys, requiring stronger keys for the CA but allowing smaller keys for leaf certs?

Most of the security authorities mentioned in the `Motivation` section of the proposal (all except the CA/Browser Forum) set a uniform minimum key size and don't differentiate between CA and leaf key requirements. Even though the CA/Browser Forum does specify different key sizes, it's to ensure stronger protection for root CAs due to their higher security impact not to allow weaker leaf keys for performance reasons.

> This make me wonder if we should add a per-KafkaUser keySize override for clients CA, similar to how validityDays and renewalDays work today after https://github.com/strimzi/proposals/pull/214. This way users can configure smaller keySize for user certificates without affecting the CA key strength. What do you think?

For the leaf certificates used by KafkaUser clients correct? If so, yes I think that is reasonable. However, I think it is something we should keep out of scope for this proposal for simplicity's sake and add the feature in a follow-up proposal or PR. 

How does that sound to you?

scholzj: I never heard about the cert size being a performance issue. The cipher suites would have much bigger performance impact. Unless we have some actual demand for different key sizes I would stick with one size.

152-configurable-ca-key-size.md
```diff
@@ -0,0 +1,195 @@
...
+  # ...
+```
+
+The `keySize` property controls the RSA key size in bits for **all** keys generated by Strimzi, both the root certificates and the leaf certificates.
+
+This follows the same pattern as `validityDays` which also applies to all certificates managed under the CA.
+
+#### Defaults and validation
+
+* **Default value:** `4096`, the current root certificate key size, preserving backward compatibility for root certificate keys.
```
scholzj: Not sure the key size has anything to do with backwards compatibility as I assume we and all clients accept the usual key sizes.

152-configurable-ca-key-size.md
```diff
@@ -0,0 +1,195 @@
...
+  This will change the effective key size for leaf certificates from `2048` bits, OpenSSL's implicit default, to `4096` bits.
+  This is a deliberate change: relying on an implicit OpenSSL default is fragile and aligning leaf certificate keys with the root certificate key size is a safer baseline. For users who need the previous 2048-bit leaf key size, they can set keySize: `2048` explicitly.
+
+* **Minimum value:** `2048` since smaller RSA keys do not meet any current security standard and therefore should not be supported.
```
scholzj: Why do we need to enforce minimum value? Is there some technical issue preventing smaller sizes?

152-configurable-ca-key-size.md
```diff
@@ -0,0 +1,195 @@
...
+The `keySize` field does not affect the root certificate keys since those are configured by the user through the `Certificate` resources that the user creates, which are outside Strimzi's control.
+
+### Implementation
+
+Following the existing pattern established by `validityDays` we would add a `keySize` field to `CertificateAuthority.java` like this:
+
+```java
+public class CertificateAuthority implements UnknownPropertyPreserving {
+    //... existing fields
+    public static final int DEFAULT_KEY_SIZE = 4096;
```
scholzj: Why do we need to keep the default here?

152-configurable-ca-key-size.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Behavior on key size change
+
+When a user changes the `keySize` on an existing cluster`:
+
+* **No immediate key regeneration.** Changing `keySize` alone does not trigger an immediate root key replacement or leaf certificate regeneration.
+  The new key size takes effect the next time a key is generated during root key replacement (triggered by `certificateExpirationPolicy: replace-key` at renewal time or a manual force-replace annotation) or when new leaf certificates are issued (e.g. scaling up, adding a new `KafkaUser`).
+* **Existing certificates remain valid.** Certificates generated with the previous key size continue to function until they are naturally renewed or replaced.
+  This avoids unnecessary rolling restarts.
+* **Gradual rollout.** To immediately apply a new key size to all certificates, the user should:
```
scholzj: It is not really gradual rollout if the user has to do something?

152-configurable-ca-key-size.md
```diff
@@ -0,0 +1,195 @@
...
+  1. Set the desired `keySize` in the `Kafka` resource.
+  2. Trigger a root key replacement using the `strimzi.io/force-replace` annotation on the root key Secret.
```
scholzj:
```suggestion
  2. Trigger a CA private key replacement using the `strimzi.io/force-replace` annotation on the `<cluster-name>-cluster-ca` Secret.
```

152-configurable-ca-key-size.md
```diff
@@ -0,0 +1,195 @@
...
+  This will generate a new root key with the configured size and trigger re-issuance of all leaf certificates followed by rolling restarts.
```
scholzj:
```suggestion
  This will generate and rollout a new CA private key with the configured size and trigger re-issuance of all leaf certificates followed by rolling restarts.
```

152-configurable-ca-key-size.md
scholzj: Why do we need to keep the default here?

kyguy: We don't, that should live in the operator code, removing.

kyguy: We don't, that should live in the operator code, removing.

152-configurable-ca-key-size.md
kyguy: Yes you are right, that should say "Immediate" rollout, fixing.

152-configurable-ca-key-size.md
kyguy: Fixed.

152-configurable-ca-key-size.md
kyguy: We don't need to, from what I understood, keys under 2048 bit keys are considered insecure by most current standards. That being said, you could argue it is not necessarily Strimzi's job to enforce key size and I would agree with you on that. 

I removed the minimum requirement in the latest commit.

152-configurable-ca-key-size.md
kyguy: Fixed.

152-configurable-ca-key-size.md
kyguy: Fixed.

scholzj (approved):
LGTM. +1 binding. Thanks for the proposal.

ppatierno (approved):
LGTM. +1 from me (binding).

152-configurable-ca-key-size.md
ppatierno: I tried to run the openssl command with several key sizes and it doesn't allow to go under 512, giving the following error:

```shell
Error setting RSA length
00930224877F0000:error:1C8000AB:Provider routines:rsa_gen_set_params:key size too small:providers/implementations/keymgmt/rsa_kmgmt.c:520:
```

So maybe we should have a minimum to avoid this failure.

scholzj: That might be an actual technical limitation. But what about Cert Manager where this would apply as well?

ppatierno: I found this within the cert-manager codebase https://github.com/cert-manager/cert-manager/blob/master/pkg/util/pki/generate.go#L36 It seems to force 2048 bits as minimum.

ppatierno: Also in the doc https://cert-manager.io/docs/reference/api-docs/ if you go down to `CertificatePrivateKey` it states:

> Size is the key bit size of the corresponding private key for this certificate.
If algorithm is set to RSA, valid values are 2048, 4096 or 8192, and will default to 2048 if not specified. If algorithm is set to ECDSA, valid values are 256, 384 or 521, and will default to 256 if not specified. If algorithm is set to Ed25519, Size is ignored. No other values are allowed.

Which sounds like it doesn't allow anything else like weird values i.e. 1346 which is allowed by openssl instead.

152-configurable-ca-key-size.md
tinaselenge: ok thanks for the explanation. 
> For the leaf certificates used by KafkaUser clients correct? If so, yes I think that is reasonable. However, I think it is something we should keep out of scope for this proposal for simplicity's sake and add the feature in a follow-up proposal or PR.

I'm happy with this suggestion. We can add it if users ask for it in the feature.

tinaselenge (approved):

152-configurable-ca-key-size.md
scholzj: Hmm, so how do we best handle it? 🤔

kyguy: I could add the CRD validation annotation back in: `@Minimum(2048)`. From what I understand, values under `2048` are not considered to be cryptographically safe by any security authority anyway.

How does everyone feel about that?

scholzj: Well, it sounds like OpenSSL supports lover values without any issues.

kyguy: Alternatively, what if we referenced the minimum supported bits of OpenSSL and cert-manager in our documentation, with links to their respective documentation as a guide for users configuring `keySize`?

Would that be suitable?

scholzj: Yeah, I guess we can keep the minimum as 512 and document the rest.

ppatierno: I think we don't have many solutions here. Setting the minimum to what we provide as built-in mechanism (openssl) sounds reasonable to me. The rest can be just documented (even for future other cert managers integration, if any).

kyguy: Let me add a commit with this information and then we will be good to go

kyguy: Alright, added a minimum annotation for CRD validation and a note about adding documentation as part of the validation.

If that change looks ok, we should be clear to move forward

scholzj: I think that looks good.

scholzj: Thanks.

kyguy: Just updated the proposal sequence number and index in README.md.

Once the most recent changes look good to you @ppatierno we should be ready to merge!

scholzj:
This proposal now has 3 binding +1 votes. If there are no more comments by Friday EOB, it will be merged as approved.

@kyguy This is how you do it ^^^ 😉

153-configurable-ca-key-size.md
```diff
@@ -0,0 +1,195 @@
...
+    // ... existing methods 
+}
+```
+
+#### User Operator considerations
+
+The `keySize` for the clients CA must be forwarded to the User Operator so it can generate `KafkaUser` certificates with the correct key size.
+This will be done by adding a new environment variable `STRIMZI_CLIENTS_CA_KEY_SIZE` to the Entity Operator deployment set by the Cluster Operator based on the `spec.clientsCa.keySize` value in the `Kafka` resource.
+
+### Behavior on key size change
```
katheris, line 144: I think some thought is needed for the cert-manager part of this. If the key size is changed in the Certificate resource then cert-manager will immediately issue a new certificate.

In the PR for cert-manager there are two ways a user can trigger a certificate renewal, either by changing something in the Kafka config so that strimzi updates something in the Certificate resource (e.g. something that changes the subject), or by using the cert-manager CLI to request cert-manager renews the cert. Ideally the user should never be changing fields in the Certificate resource, and Strimzi does not know when cert-manager will renew certificates, so there is no way for Strimzi to know when to update the Certificate resource.

So I think for cert-manager the simplest implementation would be to have Strimzi immediately apply the new key size to the Certificate resource, and therefore it would immediately trigger new certificates to be issued.
@tinaselenge @kyguy @scholzj @ppatierno what do you think?

katheris (commented):
Thanks @kyguy. I've added one comment in relation to the behaviour on key size change for cert-manager. But otherwise I'm happy with this proposal.

We can always discuss that on the cert-manager PR instead if you don't want that discussion to hold up this proposal.

153-configurable-ca-key-size.md
kyguy: > So I think for cert-manager the simplest implementation would be to have Strimzi immediately apply the new key size to the Certificate resource, and therefore it would immediately trigger new certificates to be issued.

This sounds fine to me, so basically, the behavior on key size change would be different between Strimzi-managed CA and Cert-Manager-managed CA where the former would use lazy regeneration (new key size takes effect at next renewal or replacement) and the latter would use immediate regeneration. Am I understanding that correctly?

If that is suitable to everyone I can add a note about this difference in behavior with Cert-Manager-managed CA and mention that the behavior will be documented.

How does that sound?

kyguy: Just added a commit to show what this would look like, let me know what you think!

scholzj: I think the Cert Manager behaviour is the more desired one. So I do not think we should prevent it. We should just document it properly that the behaviour differ.

@katheris What does Cert Manager do when the the expiration configs are changed?

ppatierno: I also agree with that. Changing the key size is usually done to increase security (I think most of users would change to use a bigger key size, not a smaller one) so I would expect users would love to have the certificates re-generated immediately. I was wondering if we should think about changing the behavior with the self-generated certificates by Strimzi to align with cert manager.

katheris: Yes @kyguy I’m happy with us just documenting the difference.

@scholzj my understanding is it works the same so cert-manager immediately issues a new certificate if the duration changes but I haven’t tested it

@ppatierno I think I can see pros and cons either way, it may be that people want to bring in new key sizes but not force early renewal and it’s pretty simple to trigger a renewal. So I don’t have a strong opinion on whether we align them or not.

ppatierno: > @ppatierno I think I can see pros and cons either way, it may be that people want to bring in new key sizes but not force early renewal and it’s pretty simple to trigger a renewal. So I don’t have a strong opinion on whether we align them or not.

We could argue why cert-manager chose this path of immediate renewal when these things change. It's a project focused on certificates handling so maybe the contributors/maintainers made such a decision based on real use cases? But I don't know.

scholzj: I think it is complicated. I did not mind the idea of not renewing the certificates immediately from the original proposal. However, I wonder how confusing it would be if it behaves differently depending on the issuer. That said, per my question and Kate's response, we already have the way we treat changes to the lifetime options behaving differently. So I think we have the confusion one way or another.

So my take would be to just document how the Strimzi issuer behaves and how the Cert Manager issue behaves. It is not great, but at least we maintain continuity and the same behaviour per issuer.

But not a strong opinion, so if someone feels strongly about it I might be easy to convince 😉.

kyguy: I don't have a strong opinion either, there are merits to both paths.

A couple of questions:
-  If we kept the lazy regeneration method for Strimzi-managed CA, is it something we could easily change in the future?
- If we did opt for immediate generation for `keySize` changes, I assume we would have the update Strimzi so any other CertificatyAuthority configuration changes would trigger a immediate generation of new CA as well, is that the correct understanding?

katheris: > If we kept the lazy regeneration method for Strimzi-managed CA, is it something we could easily change in the future?

yes I think so

> If we did opt for immediate generation for keySize changes, I assume we would have the update Strimzi so any other CertificatyAuthority configuration changes would trigger a immediate generation of new CA as well, is that the correct understanding?

code wise I don’t think so, but we might want to for consistency I guess

scholzj: > * If we kept the lazy regeneration method for Strimzi-managed CA, is it something we could easily change in the future?

Technically probably yes. But the behavioural change is harder as you change something users might be used to. You might trigger situations such as _"The renewal will be next month. Let's update the key size to be ready for it ... wait, WTF is it rolling now in the middle of a busy working day?"_. So I think we should aim to avoid it.

> * If we did opt for immediate generation for `keySize` changes, I assume we would have the update Strimzi so any other CertificatyAuthority configuration changes would trigger a immediate generation of new CA as well, is that the correct understanding?

That is part of the problem. As this behaviour already exists in Strimzi for other options. So by renewing immediately for key size you introduce inconsistency. And by changing everything to renew immediately, you break the previous behaviour people might be used to.

I'm not saying it is not possible. But it is not an obvious choice without any downsides.

kyguy: Those are fair points, this has me leaning toward keeping the lazy generation for the Strimzi-managed CA. 

I would rather have the inconsistency flagged by users than an unexpected change for existing users.  We could always change the behavior of the `CertificateAuthority` API to use immediate generation in the future if users are complaining about the inconsistency (and make the change loudly to reduce user surprise)

ppatierno: Ok let's stay with the current behaviour with the improved documentation.

katheris (approved):
Thanks @kyguy, I'm happy with what we have here

+1 binding

153-configurable-ca-key-size.md
katheris: I'm happy with that, it keeps consistency with what we have today and means we can then see what feedback we get from users

kyguy:
This proposal now has 4 binding +1 votes. If there are no more comments by today EOB, can we merge as approved?

scholzj:
> This proposal now has 4 binding +1 votes. If there are no more comments by today EOB, can we merge as approved?

@kyguy Please ping me EOB and I will merge it. Thanks.

kyguy:
Sorry @scholzj I forgot to ping yesterday evening, this can be merged whenever it is most convenient for you!

scholzj:

PR #242, opened 2026-08-14
Projected Service Account Tokens in Additional Volumes

scholzj:
### Type of Change

- New Proposal

### Description

This proposal suggests adding support for projected Service Account token volumes to the additional volumes section of Pod templates. This would allow users to use the same short-lived token-based authentication mechanism that was approved for internal communication in [SEP-150](https://github.com/strimzi/proposals/blob/main/150-configurable-security-of-internal-communication.md).

### Checklist

### DIFF
--- a/153-projected-service-account-tokens-in-additional-volumes.md
+++ b/153-projected-service-account-tokens-in-additional-volumes.md
@@ -1,0 +1,187 @@
+# Projected Service Account Tokens in Additional Volumes
+
+This proposal suggests adding support for projected Service Account token volumes to the additional volumes section of Pod templates.
+This would allow users to use the same short-lived token-based authentication mechanism that was approved for internal communication in [SEP-150](https://github.com/strimzi/proposals/blob/main/150-configurable-security-of-internal-communication.md).
+
+## Current Situation
+
+Strimzi currently supports three authentication mechanisms for user-defined Kafka listeners and Kafka clients in client-based operands (Kafka Connect, MirrorMaker 2, and Kafka Bridge):
+
+* mTLS authentication based on TLS client certificates
+* Authentication based on usernames and passwords (SASL SCRAM-SHA-512)
+* OAuth authentication
+
+Each of these mechanisms has its own advantages and disadvantages.
+For example:
+* mTLS requires TLS encryption and is therefore unsuitable for environments where the infrastructure provides TLS encryption, such as through Istio or another service mesh.
+* SCRAM-SHA-512 requires the management and distribution of passwords.
+  It also uses long-lived credentials.
+
+OAuth authentication also has drawbacks.
+It provides secure token-based authentication, but a common obstacle is that it requires an OAuth server to manage credentials and issue tokens.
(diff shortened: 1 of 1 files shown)


tinaselenge (commented):
Thanks Jakub. The proposal looks good to me. I guess a doc update is given but is it worth mentioning what sort of tests we would add? All the client operands will be tested with this I'm guessing?

scholzj:
@tinaselenge In general, we cover the additional volumes in unit tests only. So to be honest I did not plan to add any specific system tests for it. But I can add system tests for it using it for authentication which is the main motivation for the proposal. I added that as well as the documentation to the proposal. Thanks.

im-konge (approved):
LGTM, thanks.

Frawless (approved):

tinaselenge (approved):
Thanks Jakub. I thought a system test would be useful and we already have client operand tested with different authentication settings, so this makes sense I think.

scholzj:
Thanks for the reviews @tinaselenge @im-konge @Frawless 

This proposal has now 3 +1 votes (plus mine as an author). If there are no more comments by Friday EOB I will merge it as approved.

ppatierno (approved):
LGTM.
Just a quick question for my understanding ...
We need to support projected volumes because it's the only way we can get service account tokens (which we can't get through normal volumes, apart the default Service Account token, which is mounted at `/var/run/secrets/kubernetes.io/serviceaccount/`). Is that right?

scholzj:
> We need to support projected volumes because it's the only way we can get service account tokens (which we can't get through normal volumes, apart the default Service Account token, which is mounted at `/var/run/secrets/kubernetes.io/serviceaccount/`). Is that right?

@ppatierno The token in `/var/run/secrets/kubernetes.io/serviceaccount/` is minted for communication with the Kubernetes API server and has its audience. The projected volume allows us to get another token for the same service account but with a different audience. The different audience is important, because:
* You can configure Kafka to accept only tokens with a specific audience
* If this token leaks in Kafka logs, it cannot be used to talk with the Kubernetes API as it has different audience
* Regular service account tokens with the Kubernetes API server audience cannot be used to connect to Kafka as they do not have the custom audience.

So the audience is what differentiates the two tokens and makes both Kube API server as well as Kafka more secure.

ppatierno:
@scholzj but what makes the need for "projected" volumes is the fact that it's the only way to get a service account token with the characteristics you mentioned. There is no way to get it via "normal" volumes. This is what I meant.

scholzj:
> @scholzj but what makes the need for "projected" volumes is the fact that it's the only way to get a service account token with the characteristics you mentioned. There is no way to get it via "normal" volumes. This is what I meant.

There are two ways how to do it: 
* The projected volume is the simple and secure solution -> this is the current proposal
* You can request the token directly from the Kubernetes API -> but that is much more complicated, you need support for it in the authenticator, you need the RBAC for it, the RBAC is a security risk if not configured properly etc.

But no, you cannot mount it though normal volumes or anything like that. So from the two options, this proposal uses the easier and more secure one.

scholzj:

PR #243, opened 2026-08-16
Proposal for new Gateway API TCPRoute listener type

williamhaw:
### Type of Change

_Select the type of your PR and delete the other items_

- New Proposal

### Description

This proposal introduces a new listener type `tcproute` based on the Gateway API and on TCPRoute and ListenerSet resources. Much inspiration is taken from the existing `tlsroute` listener implementation.

### Checklist

### DIFF
--- a/153-tcp-route-listener.md
+++ b/153-tcp-route-listener.md
@@ -1,7 +1,7 @@
 # Gateway API-based `type: tcproute` listener
 
 This proposal suggests adding a new listener type (`type: tcproute`) based on the Gateway API and its `TCPRoute` resources.
-It complements the `type: tlsroute` listener introduced in [SEP-136](https://github.com/strimzi/proposals/blob/main/136-tls-route-listener.md) by exposing Apache Kafka through a single shared gateway using one port per node instead of one hostname per node.
+It is an alternative to the `type: tlsroute` listener introduced in [SEP-136](https://github.com/strimzi/proposals/blob/main/136-tls-route-listener.md): both expose Apache Kafka through a single shared gateway, but `type: tcproute` distinguishes brokers by port rather than by hostname.
 
 ## Current situation
 
@@ -22,50 +22,62 @@ They differ in _how_ the addresses are made distinct:
 
 The first option is expensive: a 10-broker cluster provisions 11 cloud load balancers.
 The second option requires exposing the Kubernetes nodes themselves.
-The third option requires TLS on the wire, one DNS name per broker, and certificates that cover every broker hostname.
+The third option requires one resolvable DNS name per broker, and a certificate that covers every broker hostname.
+Scaling a node pool then means the new DNS records must exist before the new broker is reachable, either through wildcard DNS or through another moving part such as external-dns and its propagation delay.
 
 The `TCPRoute` API is the missing combination: a single shared gateway, and therefore a single cloud load balancer, where the brokers are distinguished by port rather than by hostname or by IP address.
+Scaling adds a port on an address that already exists.
+It does not add a DNS name, and the certificate already covers that name.
 
 ## Motivation
 
-`TCPRoute` resources route raw TCP connections from a gateway listener to a backend service based on protocol and port alone, with no L7 or TLS awareness.
-They graduated to the _Standard_ channel and to the `v1` API version in Gateway API 1.6.0.
+The operational problem that motivates this listener is the same class of problem as [strimzi-kafka-operator#11123](https://github.com/strimzi/strimzi-kafka-operator/issues/11123): a newly added broker is unreachable until configuration that Strimzi does not create is already in place.
 
-Adding a `type: tcproute` listener brings three benefits over the listener types Strimzi has today:
+With `type: tlsroute`, that configuration is a DNS record for the new broker's hostname, and a certificate that includes it as a SAN.
+Every broker needs its own resolvable hostname.
+With `type: tcproute`, brokers share one address and are distinguished by port.
+A node pool can grow without new DNS records, without wildcard DNS, and without waiting for an external DNS controller.
 
-1. **One load balancer instead of N+1.**
-   All bootstrap and per-broker `TCPRoute` resources attach to the same `Gateway`.
-   Gateway API implementations back a `Gateway` with a single load balancer, for example one AWS NLB, that exposes one port per gateway listener.
-   Scaling a cluster from 3 to 30 brokers adds 27 ports to an existing load balancer instead of provisioning 27 new ones.
-   This reduces cost, avoids cloud load balancer quotas on the number of load balancers, and removes the provisioning latency that makes broker scale-up slow with `type: loadbalancer` listeners.
-2. **No dependency on TLS-SNI.**
-   `type: tlsroute` listeners can only carry TLS traffic, because SNI is what tells the gateway which broker a connection belongs to.
-   `TCPRoute` resources have no such requirement, so a `type: tcproute` listener works with `tls: true` for TLS passthrough all the way to the broker, and with `tls: false` for example for a `SASL_PLAINTEXT` listener inside a trusted network.
-3. **A single DNS name and a single certificate SAN.**
-   Because the brokers are distinguished by port, all of them share one address.
-   Users do not need wildcard DNS or one DNS record per broker, and the broker certificates need to cover only one name.
+The other properties follow from routing on port rather than on hostname:
 
-Users can already achieve this manually with a `type: cluster-ip` listener, self-managed `TCPRoute` resources, and hand-maintained `advertisedHost` and `advertisedPort` overrides.
-That is exactly the situation described in [strimzi-kafka-operator#11123](https://github.com/strimzi/strimzi-kafka-operator/issues/11123): the routing resources have to be kept in sync with the node IDs that Strimzi assigns, and brokers that come up before their routes exist break producers and consumers.
+* The cluster uses one gateway, and therefore one cloud load balancer, instead of the N+1 load balancers of `type: loadbalancer`.
+  Gateway API implementations back a `Gateway` with a single load balancer that exposes one port per gateway listener.
+  Scaling from 3 to 30 brokers adds 27 ports to an existing load balancer instead of provisioning 27 new ones.
+* TLS is orthogonal to the routing.
+  A gateway listener with `protocol: TCP` forwards the connection untouched, so `tls: true` still gives end-to-end TLS terminated at the broker, and mTLS is available.
+  `tls: false` works as well, in the same way it already does for `type: loadbalancer` and `type: nodeport`, but it is not the reason to add the listener.
 
-## Relationship to the `type: tlsroute` listener
+Users can already achieve this manually with a `type: cluster-ip` listener, self-managed `TCPRoute` resources, and hand-maintained `advertisedHost` and `advertisedPort` overrides. However, the routing resources have to be kept in sync with the node IDs that Strimzi assigns, and brokers that come up before their routes exist break producers and consumers.
 
-The `type: tcproute` and `type: tlsroute` listeners are complementary rather than competing, and neither replaces the other.
-They make opposite trade-offs, and the deciding factors are the size of the cluster and whether TLS-SNI can be required from clients.
+This proposal automates the N+1 `TCPRoute` resources and the advertised addresses.
+It does not provision the ports on the gateway.
+Users create the gateway listeners, and the recommended way to keep scale-up from racing with that step is to pre-provision a range of listeners with headroom, as described below.
+
+`TCPRoute` resources graduated to the _Standard_ channel and to the `v1` API version in Gateway API 1.6.0.
+
+## Choosing between `type: tcproute` and `type: tlsroute`
+
+The `type: tcproute` and `type: tlsroute` listeners compete in the same way that `type: loadbalancer` and `type: nodeport` already do.
+A given client-facing listener uses one of them, not both in parallel for the same purpose.
+They make opposite trade-offs.
 
 |  | `type: tlsroute` | `type: tcproute` |
 |---|---|---|
 | Brokers distinguished by | Hostname (TLS-SNI) | Port |
 | Gateway listeners needed | One, shared by all brokers | One per broker plus one for the bootstrap |
-| TLS | Required on the wire | Optional |
-| mTLS authentication | With TLS passthrough | With `tls: true` |
 | DNS records | One per broker, or a wildcard | One, shared |
 | Certificate SANs | One per broker | One, shared |
+| Scale-up requires | New DNS names, or wildcard DNS | A gateway listener that already exists on the new port |
+| TLS | Required on the wire (SNI) | Orthogonal; `tls: true` is passthrough to the broker |
+| mTLS authentication | With TLS passthrough | With `tls: true` |
 | Practical cluster size | Unbounded | Bounded by the gateway and load balancer listener limits |
 
 Because a `TLSRoute` listener is shared, `type: tlsroute` scales to any number of brokers on a single gateway listener.
 Because `type: tcproute` consumes one listener per broker, it runs into the limits described below.
-The documentation will recommend `type: tlsroute` as the default for large clusters, and `type: tcproute` when clients cannot use TLS-SNI, when the listener should not require TLS at all, or when a single DNS name and certificate is worth more than unbounded scale.
+
+The documentation will treat this as a choice between listener types, not as a pair that should be combined.
+`type: tlsroute` is the better fit when unbounded scale on one gateway listener matters more than orchestrating the creation of per-broker DNS records and certificates.
+`type: tcproute` is the better fit when one load balancer for the whole cluster is preferable to one per broker, and when brokers can share a hostname so adding a broker does not require a new DNS record or certificate SAN.
 
 ## Proposal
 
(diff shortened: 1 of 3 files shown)


153-tcp-route-listener.md
```diff
@@ -0,0 +1,516 @@
...
+3. **A single DNS name and a single certificate SAN.**
+   Because the brokers are distinguished by port, all of them share one address.
+   Users do not need wildcard DNS or one DNS record per broker, and the broker certificates need to cover only one name.
+
+Users can already achieve this manually with a `type: cluster-ip` listener, self-managed `TCPRoute` resources, and hand-maintained `advertisedHost` and `advertisedPort` overrides.
+That is exactly the situation described in [strimzi-kafka-operator#11123](https://github.com/strimzi/strimzi-kafka-operator/issues/11123): the routing resources have to be kept in sync with the node IDs that Strimzi assigns, and brokers that come up before their routes exist break producers and consumers.
+
+## Relationship to the `type: tlsroute` listener
+
+The `type: tcproute` and `type: tlsroute` listeners are complementary rather than competing, and neither replaces the other.
```
scholzj: I think saying that they are complementary suggests you would use both of them in parallel. That is not the case. They are competing (in exactly the same way they compete with load balancer, node port, and other listeners).

153-tcp-route-listener.md
```diff
@@ -0,0 +1,516 @@
...
+
+Strimzi does not manage `Gateway` resources, and it should not start now.
+Gateways are usually owned by an infrastructure team, live in a different namespace, and are shared between many applications.
+
+The Gateway API solves this with the `ListenerSet` resource ([GEP-1713](https://gateway-api.sigs.k8s.io/geps/gep-1713/)), which reached the _Standard_ channel as `v1` in Gateway API 1.5.0.
+A `ListenerSet` is a namespaced resource that contributes listeners to a `Gateway` owned by someone else.
+The gateway owner opts in by setting `.spec.allowedListeners` on the `Gateway`, and from then on the application namespace can add and remove listeners on its own.
+Listeners contributed through a `ListenerSet` share the parent gateway's address and infrastructure, which is precisely the "one load balancer, many ports" model this proposal is after.
+
+This proposal suggests that Strimzi manages one `ListenerSet` per `type: tcproute` listener, containing one TCP listener entry for the bootstrap and one for each broker.
```
scholzj: I'm not convinced Strimzi should be managing this. I think that should be user managed. Not all users would want / be able to use `ListenerSet`. They might instead want to set it in the `Gateway` resource. Having the users manage it also clears the way for making sure the right ports are configured and free.

153-tcp-route-listener.md
```diff
@@ -0,0 +1,516 @@
...
+### Cluster size limits
+
+One `type: tcproute` listener maps to exactly one gateway, and therefore to one load balancer.
+A cluster with N brokers consumes N+1 listeners on that gateway, and both the Gateway API and the underlying infrastructure impose limits on how many listeners a gateway can have:
+
+* A `Gateway` resource is limited to 64 listeners.
+  Lifting this limit is one of the reasons `ListenerSet` exists, and it is another argument for preferring `createListenerSet: true`.
+* Cloud providers apply their own limits.
+  An AWS NLB supports at most 50 listeners and that quota is not adjustable, so `1 + N <= 50` and a single gateway supports at most 49 brokers.
+  Any listeners the parent `Gateway` defines itself, if they have routes attached, count against the same budget.
```
scholzj: Again, a reason why the users should manage this on their own because Strimzi cannot keep control over this.

153-tcp-route-listener.md
```diff
@@ -0,0 +1,516 @@
...
+      type: tls
+    configuration:
+      parentRefs:
+        - name: kafka-gateway
+          namespace: infra
+      createListenerSet: true
+      bootstrap:
+        host: kafka.example.com
+        gatewayPort: 9192
+      gatewayPortTemplate: "9200 + {nodeId}"
```
scholzj: Why does this need a new field?

153-tcp-route-listener.md
```diff
@@ -0,0 +1,516 @@
...
+For example, `9200 + {nodeId}` assigns port `9200` to node 0, port `9201` to node 1, and so on.
+Reusing that syntax means the implementation can reuse the existing template rendering code, and it makes the ports deterministic: a given node ID always maps to the same port, so DNS, firewall rules, and client configuration stay stable across restarts and rescheduling.
+
+The bootstrap `gatewayPort` is mandatory.
+For the brokers, either `gatewayPortTemplate` or a `gatewayPort` for every broker must be configured, following the same pattern as `host` and `hostTemplate` for the other route-based listeners.
+
+#### Addresses
+
+`TCPRoute` resources contain no hostname, so the `host` fields have a slightly different meaning than for the other route-based listeners:
+they do not influence the generated resources at all and are only used as the address that Strimzi advertises to clients, publishes in the `Kafka` CR status, and adds to the broker certificates.
```
scholzj: Then it should use the advertisedHost field I guess?

153-tcp-route-listener.md
```diff
@@ -0,0 +1,516 @@
...
+
+* `.configuration.bootstrap.host` sets the bootstrap address.
```
scholzj: Why is this relevant if it is not needed by Gateway API?

153-tcp-route-listener.md
```diff
@@ -0,0 +1,516 @@
...
+* `.configuration.hostTemplate` and `.configuration.brokers[].host` set the per-broker addresses.
+  When neither is configured, the brokers use the bootstrap host, since with `TCPRoute` all brokers are reached through the same gateway address.
+* `advertisedHost`, `advertisedHostTemplate`, `advertisedPort`, and `advertisedPortTemplate` keep their usual meaning as overrides of what the brokers put into `advertised.listeners`.
+* `.configuration.bootstrap.alternativeNames` keeps its usual meaning and can be used to add further names to the bootstrap certificate.
+
+Unlike for the other route-based listeners, the `host` fields are optional.
+When `.configuration.bootstrap.host` is not set and the parent reference points to a `Gateway`, Strimzi reads the first address from the gateway's `.status.addresses` and uses it as the address for the bootstrap and for all brokers.
```
scholzj: I do not think we want to do this. We do not touch the Gateway resources and I do not think we should get involved with them as that might mean RBAC issues. AFAIK this is also often source of compatibility issues.

scholzj (commented):
Thanks for the proposal. I left some comments. Historically, there was never much interest in the TCP routing because of the complexity of the setup and because most users anyway need to use TLS. So I think we need to carefully think if this is worth it or not.

williamhaw:
@scholzj thanks for the review comments, I have now simplified the proposal.

For additional context, we currently run 16 and counting small clusters (<15 brokers) inside our customers' cloud accounts and expose them behind internal NLBs (only reachable inside the VPC). Clients connect over SASL_PLAINTEXT. Since these Kafka clusters are provisioned in our customers' accounts, it would involve much back-and-forth permission changes to support the DNS record and certificate machinery with `tlsroute`. Hence we'd rather keep it simple and decouple the networking gain (single Gateway/LB) from TLS.

154-tcp-route-listener.md
```diff
@@ -0,0 +1,518 @@
...
+Strimzi does not manage `Gateway` resources, and it will not manage the TCP listeners on them either.
+Gateways are usually owned by an infrastructure team, live in a different namespace, and are shared between many applications.
+Strimzi cannot know which ports are free on a shared gateway, or how much listener capacity is left.
+Users may also prefer to define the listeners on the `Gateway` itself, or to contribute them through a `ListenerSet` they manage, and the operator should not pick that for them.
+
+The user is therefore responsible for making sure the parent gateway has a TCP listener on every advertised port, allowing `TCPRoute` attachment from the Kafka namespace.
+Strimzi creates only the `TCPRoute` resources and the advertised addresses.
+When a broker is added whose advertised port has no gateway listener yet, that broker is not reachable until the listener exists, which is the same class of race as in the original issue.
+
+The recommended operational pattern is to pre-provision a range of gateway listeners with headroom, and let Strimzi attach and detach routes as brokers come and go:
```
scholzj: MAybe example would be better then recommended.

154-tcp-route-listener.md
```diff
@@ -0,0 +1,518 @@
...
+
+- One per-listener bootstrap `Service` pointing to all Kafka brokers
+- One `Service` per broker
+- One bootstrap `TCPRoute` and one `TCPRoute` per broker, each pointing to the corresponding service
+
+The important difference is how a route is bound to the gateway.
+A `TLSRoute` is matched by hostname, so all TLS routes can share one gateway listener.
+A `TCPRoute` has nothing to match on, so each route needs its own gateway listener.
+The Gateway API specification is explicit about this: if several `TCPRoute` resources attach to the same listener, all of them are `Accepted` but only the oldest one receives traffic.
+If a `TCPRoute` sets neither `sectionName` nor `port` on its parent reference, it attaches to every TCP listener on the gateway.
```
scholzj, line 97: What if the port it sets is not configured in Gateway/ListenerSet?

154-tcp-route-listener.md
```diff
@@ -0,0 +1,518 @@
...
+Because Strimzi does not manage the gateway listeners, it also cannot keep these limits from being hit.
+Conflicts and exhausted capacity surface as routes that are never accepted, and therefore as a failed reconciliation after the reconciliation timeout.
+
+### Implementation support
+
+`TCPRoute` is an _Extended_ support feature, so support has to be verified per implementation.
+At the time of writing, the AWS Load Balancer Controller supports what this proposal needs:
+
+- Version 3.5.0 requires Gateway API 1.6 CRDs, serves `TCPRoute` at `gateway.networking.k8s.io/v1`, and passes Gateway API 1.6.0 conformance.
+- The `gateway.k8s.aws/nlb` GatewayClass provisions one NLB per `Gateway` and materialises one NLB listener for each gateway listener that has a route attached, which is exactly the model described above.
```
scholzj: I do not think we want to target any particular implementation. I would stick with saying something like _TCPRoute needs to be supported by the Gateway implementation_.

154-tcp-route-listener.md
```diff
@@ -0,0 +1,518 @@
...
+    tls: true
+    authentication:
+      type: tls
+    configuration:
+      parentRefs:
+        - name: kafka-gateway
+          namespace: infra
+      bootstrap:
+        host: kafka.example.com
+      advertisedPortTemplate: "9200 + {nodeId}"
```
scholzj, line 176: Sorry. I might have incorrectly told you that the `advertisedPortTemplate` is all what is needed. But I don't think so. I think we need to configure it separately - through some port and port template field. Also, we need port configuration for the bootstrap TCPRoute, or?

I'm not sure how to name it. `port` would be probably too generic. And we want to avoid the name to be too specific such as `tcpRoutePort`. But we will need something, or?

154-tcp-route-listener.md
```diff
@@ -0,0 +1,518 @@
...
+
+Configuring more than one parent reference is allowed for consistency with `type: tlsroute`, but as explained above it creates additional paths to the same brokers rather than distributing them.
+
+#### Addresses
+
+`TCPRoute` resources contain no hostname, so the `host` and `hostTemplate` fields used by `type: route`, `type: ingress`, and `type: tlsroute` are not used.
+Those fields configure the hostname in the generated route, and there is no such hostname here.
+Kafka still needs an advertised address, and the broker certificates still need SANs, which is why the existing advertised-address fields are the right ones:
+
+- `.configuration.bootstrap.host` is required and sets the bootstrap address published to clients, stored in the `Kafka` CR status, and added to the broker certificates.
```
scholzj, line 243: There is nothing like `bootstrap address published to clients`. I'm not sure how to set the address in the status. But the host field has different semantic. As for SANs, you should use the additional names for that.

154-tcp-route-listener.md
```diff
@@ -0,0 +1,518 @@
...
+- `.configuration.advertisedHostTemplate` and `.configuration.brokers[].advertisedHost` set the per-broker advertised hosts.
+  When neither is configured, the brokers use the bootstrap host, since with `TCPRoute` all brokers are reached through the same gateway address.
+- The bootstrap advertised port is the listener `port`.
+  Strimzi also uses it as the gateway port on the bootstrap `TCPRoute` parent reference.
+  This is the same default as for `type: loadbalancer`.
```
scholzj, line 248: That is not the same. Here, the actual port will likely not be the listener port. As said earlier, we need some field for configuring the ports.

154-tcp-route-listener.md
```diff
@@ -0,0 +1,518 @@
...
+Once that is released, the implementation needs only a Fabric8 version bump, in the same way the `type: tlsroute` implementation followed the bump to 7.7.0.
+Building on `v1alpha2` instead is rejected, because that version was deprecated in Gateway API 1.6 and will be removed.
+If the Fabric8 release lags, the fallback is for Strimzi to carry the four `TCPRoute` model classes itself, reusing the existing Fabric8 `v1.ParentReference` and `v1.BackendRef` types, since the `v1alpha2` and `v1` schemas are identical.
+
+### Testing strategy
+
+The listener will be covered by unit tests and manual testing.
+System tests are not covered by this proposal, for the same reasons given in SEP-136: several existing listener types rely on unit and manual testing only, and the value of system tests is limited when the behaviour depends on which Gateway API implementation is installed.
```
scholzj: I know this might sound very unfair. Especially given I wrote SEP-136 and the Ingress listener type before. But who is going to test this manually? I test the TLSRoutes and Ingress. But I will not do it, it is not a feature I'm particularly interested in and I do not have time to test everything. And I do not think you ever tested any Strimzi releases. So if nothing else, I think the manual testing should be removed.

scholzj (commented):
Thanks. I left some more comments. Sorry, I think in the last review round I might have mislead you about the port configurations.

154-tcp-route-listener.md
```diff
@@ -0,0 +1,518 @@
...
+        namespaces:
+          from: Selector
+          selector:
+            matchLabels:
+              kubernetes.io/metadata.name: myproject
+```
+
+#### `parentRefs`
+
+The `parentRefs` field is the same field, with the same schema, as the one used by `type: tlsroute` listeners.
```
tinaselenge: can you please provide an example for this as well?

154-tcp-route-listener.md
```diff
@@ -0,0 +1,518 @@
...
+- `.configuration.advertisedHostTemplate` and `.configuration.brokers[].advertisedHost` set the per-broker advertised hosts.
+  When neither is configured, the brokers use the bootstrap host, since with `TCPRoute` all brokers are reached through the same gateway address.
+- The bootstrap advertised port is the listener `port`.
+  Strimzi also uses it as the gateway port on the bootstrap `TCPRoute` parent reference.
+  This is the same default as for `type: loadbalancer`.
+- `.configuration.advertisedPortTemplate` and `.configuration.brokers[].advertisedPort` set the per-broker advertised ports, using the same template syntax as [SEP-135](https://github.com/strimzi/proposals/blob/main/135-templating-advertised-port-fields.md).
+  Either the template or a per-broker `advertisedPort` for every broker must be configured.
+  Strimzi also uses these values as the gateway ports on the per-broker `TCPRoute` parent references.
+- `.configuration.bootstrap.alternativeNames` keeps its usual meaning and can be used to add further names to the bootstrap certificate.
```
tinaselenge: Sorry, I'm having a hard time following this section. I think some examples based on the previous one would be useful, so when Kafka CR is configured with this listener type, what Strimzi would be doing, how the created TLSRoute would look like, what implications these advertisedPorts and alternativeNames have on the resource.

154-tcp-route-listener.md
```diff
@@ -0,0 +1,518 @@
...
+          port: 9094
+```
+
+The naming of all generated resources follows the same rules as for the existing route and ingress based listeners.
+
+With this configuration, the brokers advertise `kafka.example.com:9200`, `kafka.example.com:9201`, and `kafka.example.com:9202`, and the `Kafka` CR status reports the bootstrap address `kafka.example.com:9094`.
+
+### Readiness
+
+After creating the resources, Strimzi will wait for the `.status` section of every generated `TCPRoute` to contain at least one parent reference, exactly as it does for `TLSRoute` resources.
```
tinaselenge: please show an example as well.

154-tcp-route-listener.md
```diff
@@ -0,0 +1,518 @@
...
+The listener validation will be extended to check that:
+
+- `parentRefs` is configured
+- `.configuration.bootstrap.host` is configured
+- Either `advertisedPortTemplate` is configured, or every broker has an `advertisedPort`
+- All advertised ports within one listener are unique, within the valid port range, and distinct from the listener `port` used for bootstrap
+- `host` and `hostTemplate` are not configured, because they do not apply to `TCPRoute` resources
+
+Strimzi cannot validate that the configured ports are free on the gateway, or that the gateway has capacity for them, because the gateway may be shared with other applications and other Kafka clusters.
+Conflicts and exhausted capacity surface as routes that are never accepted, and therefore as a failed reconciliation after the reconciliation timeout.
```
tinaselenge: If a route is not accepted and a reconciliation fails, what error users would see, and how can they find out easily what caused it, e.g. there is no capacity left etc.

154-tcp-route-listener.md
williamhaw: fair enough!

154-tcp-route-listener.md
williamhaw: That's a good question. The user will have to get the load balancer address with an out-of-band method. Or if the Gateway has an external-dns annotation they can use that value.

154-tcp-route-listener.md
williamhaw: I'm thinking of `gatewayListenerPort`: https://gateway-api.sigs.k8s.io/guides/user-guides/tcp/

williamhaw:
Thanks for the comments, I will take some time to refine this proposal.

154-tcp-route-listener.md
```diff
@@ -0,0 +1,511 @@
...
+With `type: tlsroute`, that configuration is a DNS record for the new broker's hostname, and a certificate that includes it as a SAN.
+Every broker needs its own resolvable hostname.
+With `type: tcproute`, brokers share one address and are distinguished by port.
+A node pool can grow without new DNS records, without wildcard DNS, and without waiting for an external DNS controller.
+
+The other properties follow from routing on port rather than on hostname:
+
+- The cluster uses one gateway, and therefore one cloud load balancer, instead of the N+1 load balancers of `type: loadbalancer`.
+  Gateway API implementations back a `Gateway` with a single load balancer that exposes one port per gateway listener.
+  Scaling from 3 to 30 brokers adds 27 ports to an existing load balancer instead of provisioning 27 new ones.
```
scholzj, line 45: You make it sound like that happens automatically. But I think you actually need to prepare the ports manually, or?

154-tcp-route-listener.md
```diff
@@ -0,0 +1,511 @@
...
+The `type: tcproute` and `type: tlsroute` listeners compete in the same way that `type: loadbalancer` and `type: nodeport` already do.
+A given client-facing listener uses one of them, not both in parallel for the same purpose.
+They make opposite trade-offs.
+
+|                          | `type: tlsroute`               | `type: tcproute`                                         |
+| ------------------------ | ------------------------------ | -------------------------------------------------------- |
+| Brokers distinguished by | Hostname (TLS-SNI)             | Port                                                     |
+| Gateway listeners needed | One, shared by all brokers     | One per broker plus one for the bootstrap                |
+| DNS records              | One per broker, or a wildcard  | One, shared                                              |
+| Certificate SANs         | One per broker                 | One, shared                                              |
```
scholzj, line 70: I think typically you would use a wildcard certificate here.

```suggestion
| Certificate SANs         | One per broker, or a wildcard                 | One, shared                                              |
```

154-tcp-route-listener.md
```diff
@@ -0,0 +1,511 @@
...
+
+#### Addresses
+
+`TCPRoute` resources contain no hostname, so the `host` and `hostTemplate` fields used by `type: route`, `type: ingress`, and `type: tlsroute` are not used.
+Those fields configure the hostname in the generated route, and there is no such hostname here.
+Kafka still needs an advertised address, and the broker certificates still need SANs, which is why the existing advertised-address fields are the right ones:
+
+- `.configuration.bootstrap.host` is required and sets the bootstrap address published to clients, stored in the `Kafka` CR status, and added to the broker certificates.
+- `.configuration.advertisedHostTemplate` and `.configuration.brokers[].advertisedHost` set the per-broker advertised hosts.
+  When neither is configured, the brokers use the bootstrap host, since with `TCPRoute` all brokers are reached through the same gateway address.
```
scholzj, line 245: No, it should definitely not use anything from the bootstrap configuration. It should be required for this listener type and validated in the validation.

154-tcp-route-listener.md
```diff
@@ -0,0 +1,511 @@
...
+  This is the same default as for `type: loadbalancer`.
+- `.configuration.advertisedPortTemplate` and `.configuration.brokers[].advertisedPort` set the per-broker advertised ports, using the same template syntax as [SEP-135](https://github.com/strimzi/proposals/blob/main/135-templating-advertised-port-fields.md).
+  Either the template or a per-broker `advertisedPort` for every broker must be configured.
+  Strimzi also uses these values as the gateway ports on the per-broker `TCPRoute` parent references.
+- `.configuration.bootstrap.alternativeNames` keeps its usual meaning and can be used to add further names to the bootstrap certificate.
+
+The advertised port and the gateway port are the same value.
+That means this listener cannot advertise a different port than the gateway listens on, which would only matter with port translation in front of the load balancer.
+
+Unlike `type: tlsroute`, the per-broker advertised ports have no default such as 443, because they have to be distinct from the bootstrap and from each other.
```
scholzj, line 257: They should default to the _port_ used for the TCP Route. So they have a default.

154-tcp-route-listener.md
```diff
@@ -0,0 +1,511 @@
...
+Fabric8 7.8.0 was released one day before Gateway API 1.6.0, and still pins `sigs.k8s.io/gateway-api` at 1.5.1, where `TCPRoute` exists only as the now-deprecated `v1alpha2`.
+This is being addressed upstream in [fabric8io/kubernetes-client#8032](https://github.com/fabric8io/kubernetes-client/issues/8032) and [fabric8io/kubernetes-client#8033](https://github.com/fabric8io/kubernetes-client/pull/8033), which bump the pin to 1.6.1 and regenerate the model, adding `v1.TCPRoute` and `v1.UDPRoute` while keeping the `v1alpha2` types.
+Once that is released, the implementation needs only a Fabric8 version bump, in the same way the `type: tlsroute` implementation followed the bump to 7.7.0.
+Building on `v1alpha2` instead is rejected, because that version was deprecated in Gateway API 1.6 and will be removed.
+If the Fabric8 release lags, the fallback is for Strimzi to carry the four `TCPRoute` model classes itself, reusing the existing Fabric8 `v1.ParentReference` and `v1.BackendRef` types, since the `v1alpha2` and `v1` schemas are identical.
```
scholzj, line 395: Fabric8 7.9.0 is already out and integrated in Strimzi main. I think it has the new Gateway `v1` APIs. But you should probably doublecheck and update this section.

154-tcp-route-listener.md
```diff
@@ -0,0 +1,511 @@
...
+
+### Sharding one listener across multiple gateways
+
+As described in the limits section, a `type: tcproute` listener is bounded by the number of listeners its gateway supports.
+Spreading the brokers of one listener across several gateways, so that a cluster can exceed that limit, is out of scope.
+
+The natural unit for such sharding in Strimzi would be the node pool, with different pools attaching to different gateways, and the bootstrap living on any one of them.
+Kafka does not require the bootstrap and the brokers to share an address, so this would work from a protocol point of view.
+It would, however, mean moving part of the listener configuration into the `KafkaNodePool` CR, which is a larger API change than this proposal wants to make, and it should be evaluated on its own merits in a future proposal.
+Until then, clusters that outgrow a single gateway should use a `type: tlsroute` listener.
```
scholzj, line 423: It is worth noting that this can be still done manually by using the `cluster-ip` listener and managing your own TCPRoutes, right?

154-tcp-route-listener.md
scholzj: I think we should try to find some neutral name. `port` might be too confusing as we use `port` already in the `.spec.kafka.listeners`. Maybe `networkPort`? `gatewayListenerPort` is too specific and cannot be reused. We did that mistake already with `nodePort` and that is why we cannot reuse that now.

154-tcp-route-listener.md
scholzj: I guess we can always construct the address for the status from the broker addresses. I don't know. Let's see what other maintainers think.

154-tcp-route-listener.md
```diff
@@ -0,0 +1,511 @@
...
+- OpenShift Routes (OpenShift only) (`type: route`)
+- Ingress with TLS passthrough (`type: ingress`, deprecated)
+- Gateway API TLS routes (`type: tlsroute`)
+
+Because the Kafka protocol requires clients to connect to every broker individually, each of these listener types has to expose N+1 distinct addresses for a cluster with N brokers.
+They differ in _how_ the addresses are made distinct:
+
+- `type: loadbalancer` gives each broker its own load balancer, so each address is a different IP address or hostname.
+- `type: nodeport` gives each broker a different port on every Kubernetes node.
+- `type: route`, `type: ingress`, and `type: tlsroute` give each broker a different hostname on a shared address and use TLS-SNI or HTTP host headers to demultiplex the traffic.
```
ppatierno, line 21: I guess we should not mention HTTP host headers because Kafka hasn't an HTTP-based protocol.

154-tcp-route-listener.md
```diff
@@ -0,0 +1,511 @@
...
+With this configuration, the brokers advertise `kafka.example.com:9200`, `kafka.example.com:9201`, and `kafka.example.com:9202`, and the `Kafka` CR status reports the bootstrap address `kafka.example.com:9094`.
+
+### Readiness
+
+After creating the resources, Strimzi will wait for the `.status` section of every generated `TCPRoute` to contain at least one parent reference, exactly as it does for `TLSRoute` resources.
+This indicates that the route was accepted.
+Strimzi will not evaluate individual conditions and will not try to detect warnings, errors, or failures, because implementations report them differently.
+If a route has no parent references after the Cluster Operator _reconciliation timeout_, the reconciliation fails with a corresponding error.
+
+If a gateway listener is missing, if a port is already used by another listener, or if the gateway has run out of listener capacity, the routes attached to that port will not be accepted either.
```
ppatierno, line 358: If I got it right, scaling up a Kafka cluster won't be automatic anymore, because the user has to add a new Gateway listener manually to match the corresponding TCPRoute automatically created by the Strimzi operator. Is my understanding right?

PR #108, opened 2024-02-13
JBOD support in KRaft mode

scholzj:
This proposal discusses the JBOD support in KRaft mode.

PR #245, opened 2026-09-01
Add proposal for Kafka Exporter re-implementation within Strimzi org

Frawless:
### Type of Change

- New Proposal

### Description

This PR proposes new implementation of Kafka Exporter tool within Strimzi org in Java.

There is existing PoC - https://github.com/Frawless/franz-spediter/tree/pure-java (note that I didn't update the readme  from original Quarkus implementation that is in `quarkus` branch).

TODO - add links for Strimzi images with new KE

### Checklist

### DIFF
--- a/157-kafka-exporter-re-implementation.md
+++ b/157-kafka-exporter-re-implementation.md
@@ -25,14 +25,29 @@ We will create new repository under Strimzi organization - `strimzi/kafka-export
 ### Implementation
 
 Because Strimzi mostly consists of Java projects, the proposal is to implement the tool in Java.
-The implementation will use the minimal required dependencies and will contain all the existing features of the latest version of Kafka Exporter.
+The implementation will use the minimal required dependencies and will cover all functionality currently used by Strimzi.
+Features not used by Strimzi are explicitly out of scope for the initial implementation, including Helm charts, Kerberos authentication, and SASL mechanisms other than OAUTHBEARER.
 
 All the metrics that the tool will export will follow the same naming as the existing Kafka Exporter, but this can change in the future based on community feedback and needs.
 
 To keep the minimal dependency tree we will use the following:
 - `kafka-clients` for `AdminClient`
 - `prometheus-metrics-core` and `prometheus-metrics-exposition-formats` for Prometheus endpoint
-- JDK built-in http server for `/metrics` and `/healthz/ready` endpoints
+- JDK built-in HTTP server for the `/metrics` and health check endpoints
+
+#### HTTP Server
+
+The tool exposes two HTTP listeners on separate ports, following the same pattern used by the Kafka Bridge:
+
+- **Management port** (default: `9404`) — serves `/healthz/ready` over plain HTTP only.
+This port is never TLS-enabled and is used exclusively by the operator's liveness and readiness probes.
+Keeping health check endpoints on a dedicated plain-HTTP port means probe behaviour is stable regardless of the TLS configuration of the metrics endpoint.
+- **Metrics port** (default: `9403`) — serves `/metrics`.
+In the initial release this port also uses plain HTTP.
+
+Adding TLS support to the metrics port is tracked in [strimzi-kafka-operator#12556](https://github.com/strimzi/strimzi-kafka-operator/issues/12556) and is planned as a follow-up.
+When implemented, it will follow the same CRD pattern used by Kafka Bridge's `spec.http.tls` — a `tls` block under `spec.kafkaExporter` containing a `certificateAndKey` reference and an optional `trustedCertificates` list for mTLS.
+The two-port architecture ensures that enabling TLS on the metrics port will not require any changes to the probe configuration in the operator.
 
 ### CI/CD
 
@@ -42,8 +57,40 @@ As an output of the build and release process, we will produce a tarball with a 
 ### Versioning
 
 The tool will follow Strimzi versioning `<major>.<minor>.<micro>` as other projects do.
-The first version will be `0.1.0`, even though the tool already covers all existing Kafka Exporter functionality.
-We will keep the tool in the 0.x line for one or two releases to gather feedback and confirm parity with the Go implementation in production before releasing the `1.0.0`.
+The first version will be `0.1.0`, even though the tool already covers all Strimzi-required functionality.
+We will keep the tool in the 0.x line for two releases to gather feedback and validate the implementation covers all Strimzi use cases in production before releasing the `1.0.0`.
+
+### Strimzi Kafka Operator changes
+
+#### Feature Gate for Strimzi Kafka Operator
+
+The switch from the Go binary to the Java implementation will be gated behind a new feature gate — `KafkaExporterJavaImplementation` — to allow a safe, opt-in transition for users.
+
+The gate will progress through the standard Strimzi feature gate lifecycle:
+
+1. Introduced as **alpha** (disabled by default) - users can opt in to the new Java implementation while the Go binary remains the default.
+2. Promoted to **beta** (enabled by default) after two releases (in 1.4.0) once sufficient feedback and production validation has been gathered - users can still opt out by explicitly disabling the gate.
+3. Promoted to **GA** and the feature gate removed after four releases (in 1.6.0) - the Go binary is dropped and the Java implementation becomes the only option.
+
+When the feature gate is disabled, the operator continues to use the existing Go binary and `kafka_exporter_run.sh` launch script unchanged.
+When the feature gate is enabled, the operator copies the fat-jar artifact from the `strimzi/kafka-exporter` repository into the Kafka image and uses an updated launch script to start the Java implementation.
+
+#### CRD and API Changes
+
+Moving from a Go binary to a Java implementation requires normalising some fields in `KafkaExporterSpec` that are Go-specific.
+
+The following changes will be made to the `kafkaExporter` section of the Kafka CR:
+
+- `logging` — currently accepts Go-style log levels (`info`, `debug`, `trace`).
+  The existing field is kept and its value is mapped internally to the equivalent Java log level, so existing CRs continue to work without changes.
+- `enableSaramaLogging` — this field controls logging of the Sarama Go client library, which has no equivalent in the Java implementation.
+  The field will be deprecated and ignored when the Java implementation is active.
+  When the feature gate is enabled and this field is set, the operator will emit a warning in its log.
+- `jvmOptions` — a new field following the standard Strimzi `JvmOptions` type will be added to allow users to configure JVM heap, GC options, and other JVM flags, consistent with how other Java-based Strimzi components expose this.
+  This field is only meaningful when the Java implementation is active.
+  When the feature gate is disabled and this field is set, the operator will emit a warning in its log that `jvmOptions` is ignored by the Go binary.
+
+The deprecated Go-specific fields will be removed in a future API version once the feature gate reaches GA and the Go binary is fully retired.
 
 ### Documentation
 
@@ -52,27 +99,24 @@ The Strimzi documentation will be updated to reflect the switch from the upstrea
 
 ### Testing
 
-The new repository will include unit tests covering the metrics collection and registry logic, and integration tests running against a real Kafka instance.
+The new repository will include unit tests covering the metrics collection and registry logic, and integration tests running against a real Kafka instance using `strimzi-test-container`.
 Existing system tests in `strimzi-kafka-operator` will ensure that new implementation of Kafka Exporter continues to work end-to-end after the swap.
 No additional e2e scenarios will be needed.
 
 ### Security
 
-The new implementation will keep the same TLS and SASL/OAuth handling that Kafka Exporter has today, so the Kafka Exporter container keeps the same permissions and credential sources it already has, mounted certificates and Secrets provided by the Strimzi operator.
+The new implementation preserves the TLS/mTLS configuration used by Strimzi today. 
+The tool continues to use the cluster CA certificate and client certificate/key mounted by the operator via Secrets.
 No credentials will be logged or persisted by the tool itself.
-Moving to a Java implementation removes the current need to trust and verify pre-built third-party Go binaries and their checksums, replacing them with well known JVM dependencies that already go through Strimzi's existing CVE scanning and patching process.
-The AdminClient will only need the same read-level Kafka permissions the current tool requires today, describing topics and consumer groups, so no privilege changes are introduced.
+Moving to a Java implementation removes the current need to trust and verify pre-built third-party Go binaries and their checksums, replacing them with well-known JVM dependencies that already go through Strimzi's existing CVE scanning and patching process.
 
-Any security improvements will be added once we will create repo with initial implementation.
+SASL OAUTHBEARER support will be added as a follow-up when required by Cluster Security.
+Other SASL mechanisms are out of scope and will not be implemented.
 
 ## Affected Projects
 
 This proposal affects only `strimzi-kafka-operator` repo.
 Other sub-projects are not affected.
-
-In the operator repository, we will need to change the make targets to copy the new artifacts from the `strimzi/kafka-exporter` repository into the Kafka image.
-The existing `kafka_exporter_run.sh` script will be enhanced to launch the fat-jar instead of the Go binary with all necessary configs.
-The possibilities to configure Kafka Exporter through Kafka CR will remain the same as today.
 
 ## Backwards Compatibility
 
@@ -96,4 +140,4 @@ Kafka Exporter is a small tool and Go is great for such tools, however, we do no
 
 Quarkus provides an easy way to write small tools that expose Prometheus metrics.
 However, it also brings additional dependencies that can be easily avoided with pure Java.
-The main advantage of Quarkus for a tool like this would be native image builds, but since Strimzi images uses the JVM, there is no benefit in native images now.
+The main advantage of Quarkus for a tool like this would be native image builds, but native images are not viable for Strimzi because they only support x86-64 and aarch64, while Strimzi ships images for s390x and ppc64le as well.
--- a/157-kafka-exporter-re-implementation.md
+++ b/157-kafka-exporter-re-implementation.md
@@ -20,7 +20,7 @@ Currently, there is 215 open issues and 58 open pull requests from the community
 
 To mitigate security problems and allow us to better maintain the tool and provide new features, we should re-implement Kafka Exporter tool.
 
-We will create new repository under Strimzi organization - `strimzi/kafka-exporter` - that will contain the implementation code.
+We will create new repository under Strimzi organization - `strimzi/lag-reporter` - that will contain the implementation code.
 
 ### Implementation
 
@@ -39,41 +39,39 @@ To keep the minimal dependency tree we will use the following:
 
 The tool exposes two HTTP listeners on separate ports, following the same pattern used by the Kafka Bridge:
 
-- **Management port** (default: `9404`) — serves `/healthz/ready` over plain HTTP only.
+- **Management port** (default: `8080`) — serves `/healthz/ready` over plain HTTP only.
 This port is never TLS-enabled and is used exclusively by the operator's liveness and readiness probes.
 Keeping health check endpoints on a dedicated plain-HTTP port means probe behaviour is stable regardless of the TLS configuration of the metrics endpoint.
-- **Metrics port** (default: `9403`) — serves `/metrics`.
+- **Metrics port** (default: `9404`) — serves `/metrics`.
 In the initial release this port also uses plain HTTP.
 
-Adding TLS support to the metrics port is tracked in [strimzi-kafka-operator#12556](https://github.com/strimzi/strimzi-kafka-operator/issues/12556) and is planned as a follow-up.
-When implemented, it will follow the same CRD pattern used by Kafka Bridge's `spec.http.tls` — a `tls` block under `spec.kafkaExporter` containing a `certificateAndKey` reference and an optional `trustedCertificates` list for mTLS.
-The two-port architecture ensures that enabling TLS on the metrics port will not require any changes to the probe configuration in the operator.
+There is community demand for having TLS enabled on metrics endpoint ([strimzi-kafka-operator#12556](https://github.com/strimzi/strimzi-kafka-operator/issues/12556)).
+The implementation and integration into Strimzi Kafka Operator will require new proposal as it will need API changes for the tool.
 
 ### CI/CD
 
 The tool will follow Strimzi standards and will adopt the same CI/CD workflow we use for other projects.
-As an output of the build and release process, we will produce a tarball with a fat-jar that can be used in Strimzi images or in standalone distributions connected to Kafka.
+As an output of the build and release process, we will produce a zip file that we use across Strimzi org and that can be used in Strimzi images or in standalone distributions connected to Kafka.
 
 ### Versioning
 
 The tool will follow Strimzi versioning `<major>.<minor>.<micro>` as other projects do.
 The first version will be `0.1.0`, even though the tool already covers all Strimzi-required functionality.
-We will keep the tool in the 0.x line for two releases to gather feedback and validate the implementation covers all Strimzi use cases in production before releasing the `1.0.0`.
 
 ### Strimzi Kafka Operator changes
 
 #### Feature Gate for Strimzi Kafka Operator
 
-The switch from the Go binary to the Java implementation will be gated behind a new feature gate — `KafkaExporterJavaImplementation` — to allow a safe, opt-in transition for users.
+The switch from the Go binary to the Java implementation will be gated behind a new feature gate — `StrimziLagReporter` — to allow a safe, opt-in transition for users.
 
 The gate will progress through the standard Strimzi feature gate lifecycle:
 
-1. Introduced as **alpha** (disabled by default) - users can opt in to the new Java implementation while the Go binary remains the default.
-2. Promoted to **beta** (enabled by default) after two releases (in 1.4.0) once sufficient feedback and production validation has been gathered - users can still opt out by explicitly disabling the gate.
-3. Promoted to **GA** and the feature gate removed after four releases (in 1.6.0) - the Go binary is dropped and the Java implementation becomes the only option.
+1. Introduced as **alpha** (disabled by default) - users can opt in to the new Java implementation while the Go binary remains the default (introduced in 1.4.0).
+2. Promoted to **beta** (enabled by default) after two releases (in 1.6.0) once sufficient feedback and production validation has been gathered - users can still opt out by explicitly disabling the gate.
+3. Promoted to **GA** and the feature gate removed after four releases (in 1.8.0) - the Go binary is dropped and the Java implementation becomes the only option.
 
 When the feature gate is disabled, the operator continues to use the existing Go binary and `kafka_exporter_run.sh` launch script unchanged.
-When the feature gate is enabled, the operator copies the fat-jar artifact from the `strimzi/kafka-exporter` repository into the Kafka image and uses an updated launch script to start the Java implementation.
+When the feature gate is enabled, the operator will use Java binary baked in the Kafka image and `lag_reporter_run.sh` to launch the application.
 
 #### CRD and API Changes
 
@@ -94,13 +92,13 @@ The deprecated Go-specific fields will be removed in a future API version once t
 
 ### Documentation
 
-The `strimzi/kafka-exporter` repository will include a README covering all configuration options, the full list of exported metrics, and instructions for running the tool standalone.
+The `strimzi/lag-reporter` repository will include a README covering all configuration options, the full list of exported metrics, and instructions for running the tool standalone.
 The Strimzi documentation will be updated to reflect the switch from the upstream Go binary to the new Java implementation.
 
 ### Testing
 
 The new repository will include unit tests covering the metrics collection and registry logic, and integration tests running against a real Kafka instance using `strimzi-test-container`.
-Existing system tests in `strimzi-kafka-operator` will ensure that new implementation of Kafka Exporter continues to work end-to-end after the swap.
+Existing system tests in `strimzi-kafka-operator` will ensure that implementation of Lag Reporter continues to work end-to-end after the swap.
 No additional e2e scenarios will be needed.
 
 ### Security
@@ -110,8 +108,8 @@ The tool continues to use the cluster CA certificate and client certificate/key 
 No credentials will be logged or persisted by the tool itself.
 Moving to a Java implementation removes the current need to trust and verify pre-built third-party Go binaries and their checksums, replacing them with well-known JVM dependencies that already go through Strimzi's existing CVE scanning and patching process.
 
-SASL OAUTHBEARER support will be added as a follow-up when required by Cluster Security.
-Other SASL mechanisms are out of scope and will not be implemented.
+SASL OAUTHBEARER support will be added as a follow-up.
+Other SASL mechanisms are currently out of scope.
 
 ## Affected Projects
 
--- a/157-kafka-exporter-re-implementation.md
+++ b/157-kafka-exporter-re-implementation.md
@@ -14,6 +14,7 @@ The latest release, v1.10.0, was published on 2026-09-08 after a period of minim
 This opens space for potential security issues as CVEs may not be fixed promptly.
 
 This situation is also not helpful for new features that we would like to add into the exporter.
+Additionally, the upstream project relies on a third-party Go Kafka client rather than the official Apache Kafka client, meaning it does not benefit from the same level of support and compatibility guarantees that the official client provides.
 
 ## Proposal
 
@@ -33,6 +34,17 @@ To keep the minimal dependency tree we will use the following:
 - `kafka-clients` for `AdminClient`
 - `prometheus-metrics-core` and `prometheus-metrics-exposition-formats` for Prometheus endpoint
 - JDK built-in HTTP server for the `/metrics` and health check endpoints
+
+#### Collection Model
+
+Metrics collection is decoupled from the Prometheus scrape.
+A background scheduler runs a collection cycle on a configurable interval (default 30 seconds).
+Each cycle executes a sequence of batched `AdminClient` calls — cluster description, topic listing and description, offset fetching, consumer group listing, group description, and committed offset fetching — and assembles the results into an immutable snapshot.
+The snapshot is atomically swapped into the metrics registry.
+The `/metrics` endpoint only reads the latest snapshot; no Kafka calls happen in the scrape path and HTTP responses are always fast regardless of cluster size.
+
+On collection failure the previous snapshot continues to be served and the readiness probe reflects the unhealthy state.
+This avoids the `up == 0` failure mode seen in the upstream `kafka_exporter`, where scrapes time out under cluster load and dashboards go blind exactly when the cluster is most stressed.
 
 #### HTTP Server
 
@@ -79,7 +91,6 @@ The section will include the following fields from the start:
 - `groupExcludeRegex` — consumer group exclude regex.
 - `topicRegex` — topic include regex (default: `.*`).
 - `topicExcludeRegex` — topic exclude regex.
-- `showAllOffsets` — whether to report offsets for all partitions the group has ever committed to (default: `true`).
 - `logging` — standard Strimzi `Logging` type (Log4j 2), replacing the plain string `logging` field from `kafkaExporter`.
 - `jvmOptions` — standard Strimzi `JvmOptions` type for JVM heap, GC options, and flags.
 - `resources` — CPU and memory resource requirements.
@@ -110,14 +121,16 @@ The tool continues to use the cluster CA certificate and client certificate/key 
 No credentials will be logged or persisted by the tool itself.
 Moving to a Java implementation removes the current need to trust and verify pre-built third-party Go binaries and their checksums, replacing them with well-known JVM dependencies that already go through Strimzi's existing CVE scanning and patching process.
 
-SASL OAUTHBEARER support will be added as a near-term follow-up.
+SASL OAUTHBEARER support will be part of initial implementation.
 It is required for Strimzi Cluster Security and is straightforward to implement by including Strimzi OAuth as a dependency and adding the relevant Kafka client configuration.
 Other SASL mechanisms are currently out of scope.
 
 ## Affected Projects
 
-This proposal affects only `strimzi-kafka-operator` repo.
-Other sub-projects are not affected.
+This proposal affects the following projects:
+
+- `strimzi/insights-reporter` — new repository created by this proposal.
+- `strimzi-kafka-operator` — operator changes to introduce `spec.insightsReporter`, deprecate `spec.kafkaExporter`, and update system tests.
 
 ## Backwards Compatibility
 
--- a/157-kafka-exporter-re-implementation.md
+++ b/157-kafka-exporter-re-implementation.md
@@ -1,0 +1,99 @@
+# Kafka Exporter reimplementation
+
+This proposal suggests to re-implement Kafka Exporter tool under Strimzi organization in Java.
+
+## Current Situation
+
+Strimzi currently downloads binaries from released version of [kafka_exporter](https://github.com/danielqsj/kafka_exporter) tool on GitHub.
+The binaries are baked into Strimzi images and shipped as an optional tool for extending Kafka consumer metrics.
+
+## Motivation
+
+The project is widely used by Kafka community, however, it is not well maintained.
+Latest release is from 2025-02-17 which makes it almost year and half old.
+This opens space for potential security issues as there are CVEs that are not fixed.
+
+This situation is also not much helpful for new features that we would like to add into the exporter.
+Currently, there is 215 open issues and 58 open pull requests from the community with not much attention from the code owner.
+
+## Proposal
+
+To mitigate security problems and allow us to better maintain the tool and provide new features, we should re-implement Kafka Exporter tool.
+
+We will create new repository under Strimzi organization - `strimzi/kafka-exporter` - that will contain the implementation code.
+
+### Implementation
+
+Because Strimzi mostly consists of Java projects, the proposal is to implement the tool in Java.
+The implementation will use the minimal required dependencies and will contain all the existing features of the latest version of Kafka Exporter.
+
+All the metrics that the tool will export will follow the same naming as the existing Kafka Exporter, but this can change in the future based on community feedback and needs.
+
+To keep the minimal dependency tree we will use the following:
+- `kafka-clients` for `AdminClient`
+- `prometheus-metrics-core` and `prometheus-metrics-exposition-formats` for Prometheus endpoint
+- JDK built-in http server for `/metrics` and `/healthz/ready` endpoints
+
+### CI/CD
+
+The tool will follow Strimzi standards and will adopt the same CI/CD workflow we use for other projects.
+As an output of the build and release process, we will produce a tarball with a fat-jar that can be used in Strimzi images or in standalone distributions connected to Kafka.
+
+### Versioning
+
+The tool will follow Strimzi versioning `<major>.<minor>.<micro>` as other projects do.
+The first version will be `0.1.0`, even though the tool already covers all existing Kafka Exporter functionality.
+We will keep the tool in the 0.x line for one or two releases to gather feedback and confirm parity with the Go implementation in production before releasing the `1.0.0`.
+
+### Documentation
+
+The `strimzi/kafka-exporter` repository will include a README covering all configuration options, the full list of exported metrics, and instructions for running the tool standalone.
+The Strimzi documentation will be updated to reflect the switch from the upstream Go binary to the new Java implementation.
+
+### Testing
+
+The new repository will include unit tests covering the metrics collection and registry logic, and integration tests running against a real Kafka instance.
+Existing system tests in `strimzi-kafka-operator` will ensure that new implementation of Kafka Exporter continues to work end-to-end after the swap.
+No additional e2e scenarios will be needed.
+
+### Security
+
+The new implementation will keep the same TLS and SASL/OAuth handling that Kafka Exporter has today, so the Kafka Exporter container keeps the same permissions and credential sources it already has, mounted certificates and Secrets provided by the Strimzi operator.
+No credentials will be logged or persisted by the tool itself.
+Moving to a Java implementation removes the current need to trust and verify pre-built third-party Go binaries and their checksums, replacing them with well known JVM dependencies that already go through Strimzi's existing CVE scanning and patching process.
+The AdminClient will only need the same read-level Kafka permissions the current tool requires today, describing topics and consumer groups, so no privilege changes are introduced.
+
+Any security improvements will be added once we will create repo with initial implementation.
+
+## Affected Projects
+
+This proposal affects only `strimzi-kafka-operator` repo.
+Other sub-projects are not affected.
+
+In the operator repository, we will need to change the make targets to copy the new artifacts from the `strimzi/kafka-exporter` repository into the Kafka image.
+The existing `kafka_exporter_run.sh` script will be enhanced to launch the fat-jar instead of the Go binary with all necessary configs.
+The possibilities to configure Kafka Exporter through Kafka CR will remain the same as today.
+
+## Backwards Compatibility
+
+This proposal is fully compatible with previous versions as new implementation will export all the same metrics as the original implementation.
+The tool will also follow the same environment variables and parameters that are used for configuration of current Kafka Exporter.
+Breaking removals of configuration options will only be possible in upcoming releases.
+
+## Rejected Alternatives
+
+### Fork Kafka Exporter to Strimzi org
+
+We could simply fork the original Kafka Exporter and just fix CVEs.
+However, this would require onboarding it to our build mechanisms and removing things we don't want, such as Helm charts.
+These changes are straightforward, but we would still need to maintain Go code, which is not our team's expertise.
+
+### Re-write Kafka Exporter in Go
+
+Kafka Exporter is a small tool and Go is great for such tools, however, we do not have much experience across the maintainers team, and it might be hard to find an owner for it with proper experience.
+
+### Re-write Kafka Exporter in Quarkus
+
+Quarkus provides an easy way to write small tools that expose Prometheus metrics.
+However, it also brings additional dependencies that can be easily avoided with pure Java.
+The main advantage of Quarkus for a tool like this would be native image builds, but since Strimzi images uses the JVM, there is no benefit in native images now.
--- a/157-kafka-exporter-re-implementation.md
+++ b/157-kafka-exporter-re-implementation.md
@@ -46,6 +46,7 @@ In the initial release this port also uses plain HTTP.
 
 There is community demand for having TLS enabled on metrics endpoint ([strimzi-kafka-operator#12556](https://github.com/strimzi/strimzi-kafka-operator/issues/12556)).
 The implementation and integration of TLS configuration into Strimzi Kafka Operator will require a new proposal as it will need API changes for the tool.
+When TLS support is added, the implementation will use PEM files and Strimzi config providers, consistent with the approach used by other Strimzi components.
 
 ### CI/CD
 
@@ -65,7 +66,8 @@ Insights Reporter will be introduced as a new, parallel section in the Kafka CR 
 This allows users to opt in to the new Java implementation without any changes to their existing `kafkaExporter` configuration.
 
 The `spec.kafkaExporter` section will be set as **deprecated** from the moment `spec.insightsReporter` is introduced.
-The upstream `kafka_exporter` binary will be updated to v1.10.0 and continue to be supported and bundled in Strimzi images until new API version or until the tool will work without significant required changes on our side.
+The upstream `kafka_exporter` binary will be updated to v1.10.0 and continue to be supported and bundled in Strimzi images.
+We will keep to update Kafka Exporter until new API version or until the tool will work without significant required changes on our side.
 The `spec.kafkaExporter` section and the Go binary will be removed in a future API version once the deprecation period ends.
 
 The new `spec.insightsReporter` section follows the same conventions as other Strimzi components such as CruiseControl.
@@ -87,7 +89,9 @@ The section will include the following fields from the start:
 
 Fields that exist in `kafkaExporter` but are Go-specific, such as `enableSaramaLogging`, will not be carried over to `spec.insightsReporter`.
 
-When both `spec.kafkaExporter` and `spec.insightsReporter` are set, the operator will emit a warning and use `spec.insightsReporter`, ignoring `spec.kafkaExporter`.
+Both `spec.kafkaExporter` and `spec.insightsReporter` can be set at the same time.
+This allows users to run both implementations in parallel during migration — for example to compare metrics output or verify parity before switching over.
+When both are configured, the operator will deploy both components independently and emit a warning recommending migration to `spec.insightsReporter`.
 
 ### Documentation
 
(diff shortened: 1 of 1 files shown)


157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Testing
+
+The new repository will include unit tests covering the metrics collection and registry logic, and integration tests running against a real Kafka instance.
+Existing system tests in `strimzi-kafka-operator` will ensure that new implementation of Kafka Exporter continues to work end-to-end after the swap.
+No additional e2e scenarios will be needed.
+
+### Security
+
+The new implementation will keep the same TLS and SASL/OAuth handling that Kafka Exporter has today, so the Kafka Exporter container keeps the same permissions and credential sources it already has, mounted certificates and Secrets provided by the Strimzi operator.
```
scholzj: We currently use only TLS/mTLS. So I think we do not need to maintain any backwards compatibility for SASL.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,99 @@
...
+No credentials will be logged or persisted by the tool itself.
+Moving to a Java implementation removes the current need to trust and verify pre-built third-party Go binaries and their checksums, replacing them with well known JVM dependencies that already go through Strimzi's existing CVE scanning and patching process.
+The AdminClient will only need the same read-level Kafka permissions the current tool requires today, describing topics and consumer groups, so no privilege changes are introduced.
```
scholzj: That is good. But mostly irrelevant as we run it as super user?

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,99 @@
...
+## Proposal
+
+To mitigate security problems and allow us to better maintain the tool and provide new features, we should re-implement Kafka Exporter tool.
+
+We will create new repository under Strimzi organization - `strimzi/kafka-exporter` - that will contain the implementation code.
+
+### Implementation
+
+Because Strimzi mostly consists of Java projects, the proposal is to implement the tool in Java.
+The implementation will use the minimal required dependencies and will contain all the existing features of the latest version of Kafka Exporter.
```
scholzj: I would do this differently ...
* We want to maintain the compatibility where it matters: on the metrics interface (metric types, naming, etc.)
* We want to maintain parity with the features used by Strimzi

But I do not think we should aim to replace the Kafka Exporter one-to-one. We should be (at least initially) very focused on what Strimzi needs and ignore the rest. For example:
* We do not need Helm Chart
* We do not need Kerberos or various SASL stuff (we will need SASL OAuthbearer for the Cluster Security, but not necessarily PLAIN)

So we should drop everything what saves us some effort. The SASL might be trivial to do given it is Kafka client inside. But the Helm Chart, Kerberos ... screw that.

I will comment more on the CLI interface separately.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,99 @@
...
+### CI/CD
+
+The tool will follow Strimzi standards and will adopt the same CI/CD workflow we use for other projects.
+As an output of the build and release process, we will produce a tarball with a fat-jar that can be used in Strimzi images or in standalone distributions connected to Kafka.
+
+### Versioning
+
+The tool will follow Strimzi versioning `<major>.<minor>.<micro>` as other projects do.
+The first version will be `0.1.0`, even though the tool already covers all existing Kafka Exporter functionality.
+We will keep the tool in the 0.x line for one or two releases to gather feedback and confirm parity with the Go implementation in production before releasing the `1.0.0`.
```
scholzj: Should we use some feature gate to handle the transition? You suggest few releases to gather feedback and confirm parity. But without feature gate, that does not seem to be useful as we need to get it right on the first try.

scholzj (commented):
Thanks for the proposal. I think in general it goes in the right direction. However, I wonder about few things (and to some extent I mentioned them in separate comments):

* Should this be done through feature gate to facilitate smooth transition?
* I think this needs a bit more thinking about the CRD API as we need to _normalize_ it for Java. So I think some parts will need to be deprecated. Others added, etc. E.g.:
    * `logging`, `saramaLogging` deprecated and ignored?
    * Our standard logging for Java added under some new name?
    * JVM options added?
    * ...
* I assume technically we could hide it under the old API and map `logging` to general Java log level and `saramaLogging` to Kafka client log level. But is it worth it? I assume would wound need to anyway change tings one day. (That said this is a discussion point 😉)
* I'm in general fine with using Java's HTTP server if it can handle the below (TLS, multiple listeners, etc.). But for the record, we use Jetty in operators. So I think you could consider that as well.

Please think also about this thing - https://github.com/strimzi/strimzi-kafka-operator/issues/12556 - does not need to be done from day 0. But I expect it to be one of the futures to be added. So we need to make sure we can add support for TLS and we can have non-TLS listener (for healthchecks) and TLS for metrics when desired.

157-kafka-exporter-re-implementation.md
Frawless: Fair point.

157-kafka-exporter-re-implementation.md
Frawless: Right, I didn't mention it, but by `one-to-one` I mean mostly just exported metrics and overall functionality of the exporter. I do not count with helm charts, I think not even kerberos. I will double check that and update it in the proposal.

157-kafka-exporter-re-implementation.md
Frawless: Ye that came to my mind during writing the proposal. Might be better to have feature gate for it. Especially when we consider it will be hard switch without feature gate that could cause troubles to the users. I will update it in the proposal and we will see what others think about it.

Frawless:
> * I think this needs a bit more thinking about the CRD API as we need to _normalize_ it for Java. So I think some parts will need to be deprecated. Others added, etc. E.g.:
>   
>   * `logging`, `saramaLogging` deprecated and ignored?
>   * Our standard logging for Java added under some new name?
>   * JVM options added?
>   * ...
> * I assume technically we could hide it under the old API and map `logging` to general Java log level and `saramaLogging` to Kafka client log level. But is it worth it? I assume would wound need to anyway change tings one day. (That said this is a discussion point 😉)

In my PoC I basically just hide it under the old API, but I will check what we have here and think about changes around that.

> * I'm in general fine with using Java's HTTP server if it can handle the below (TLS, multiple listeners, etc.). But for the record, we use Jetty in operators. So I think you could consider that as well.
>
> Please think also about this thing - [strimzi/strimzi-kafka-operator#12556](https://github.com/strimzi/strimzi-kafka-operator/issues/12556) - does not need to be done from day 0. But I expect it to be one of the futures to be added. So we need to make sure we can add support for TLS and we can have non-TLS listener (for healthchecks) and TLS for metrics when desired.

I will check both of the points.

scholzj:
> In my PoC I basically just hide it under the old API, but I will check what we have here and think about changes around that.

I'm not against doing that initially if we can do it in some reasonable way. But it probably deserves more details on how would you do it. Also, we would still need to add the extensions (e.g. JVM options or proper logging support) sooner  or later. So the question is what is better I guess.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,99 @@
...
+This opens space for potential security issues as there are CVEs that are not fixed.
+
+This situation is also not much helpful for new features that we would like to add into the exporter.
+Currently, there is 215 open issues and 58 open pull requests from the community with not much attention from the code owner.
+
+## Proposal
+
+To mitigate security problems and allow us to better maintain the tool and provide new features, we should re-implement Kafka Exporter tool.
+
+We will create new repository under Strimzi organization - `strimzi/kafka-exporter` - that will contain the implementation code.
```
scholzj: As it is not really fork anymore, should we come up with some more original / less copycat name? (I know it is not an exact copy as it uses `-` instead of `_`, but still). Strimzi Lag Metrics? Strimzi Additional Metrics? Strimzi Lag Exporter? It is of course not only lag it exports ...

see-quick: If this doesn't cause any problems I would rather stick with `kafka-exporter`. Why have multiple names for more/less same behavior.

157-kafka-exporter-re-implementation.md
see-quick: I agree with feature gates.

btw is it still parity?? I mean, if our approach would implement just as much as only Strimzi needs, then we can't have parity with Go implementation at all (maybe on that specific parts ...) but I would rather exclude that parity word to not confuse anybody.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,99 @@
...
+We will keep the tool in the 0.x line for one or two releases to gather feedback and confirm parity with the Go implementation in production before releasing the `1.0.0`.
+
+### Documentation
+
+The `strimzi/kafka-exporter` repository will include a README covering all configuration options, the full list of exported metrics, and instructions for running the tool standalone.
+The Strimzi documentation will be updated to reflect the switch from the upstream Go binary to the new Java implementation.
+
+### Testing
+
+The new repository will include unit tests covering the metrics collection and registry logic, and integration tests running against a real Kafka instance.
```
see-quick: Hmmm, I assume you plan to use `strimzi-test-container` within ITs?

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Rejected Alternatives
+
+### Fork Kafka Exporter to Strimzi org
+
+We could simply fork the original Kafka Exporter and just fix CVEs.
+However, this would require onboarding it to our build mechanisms and removing things we don't want, such as Helm charts.
+These changes are straightforward, but we would still need to maintain Go code, which is not our team's expertise.
+
+### Re-write Kafka Exporter in Go
```
see-quick, line 137: I am not sure I understand the title? Kafka exporter is in Go or?

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,99 @@
...
+
+Kafka Exporter is a small tool and Go is great for such tools, however, we do not have much experience across the maintainers team, and it might be hard to find an owner for it with proper experience.
+
+### Re-write Kafka Exporter in Quarkus
+
+Quarkus provides an easy way to write small tools that expose Prometheus metrics.
+However, it also brings additional dependencies that can be easily avoided with pure Java.
+The main advantage of Quarkus for a tool like this would be native image builds, but since Strimzi images uses the JVM, there is no benefit in native images now.
```
see-quick: Also native images are not great for us because they only support x86-64 and aarch64 and not s390x and also I think ppc64le.

157-kafka-exporter-re-implementation.md
scholzj: Well, it is also a name of someone's project. You wouldn't want someone to write a Kafka operator and call it Strimzi because it does more or less the same, or? At least I would not.

157-kafka-exporter-re-implementation.md
Frawless: I changed the wording.

157-kafka-exporter-re-implementation.md
Frawless: Maybe `Lag Reporter` ?

157-kafka-exporter-re-implementation.md
Frawless: Yes, I added a word about it.

157-kafka-exporter-re-implementation.md
Frawless: Yes, this is for completely new implementation in Go, not forking it.

157-kafka-exporter-re-implementation.md
Frawless: Added a note.

Frawless:
I pushed couple of changes mostly related to feature gate and CRD API changes. I didn't rename the project name yet so once we agree on this or new one, I will change it across the whole proposal.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### HTTP Server
+
+The tool exposes two HTTP listeners on separate ports, following the same pattern used by the Kafka Bridge:
+
+- **Management port** (default: `9404`) — serves `/healthz/ready` over plain HTTP only.
+This port is never TLS-enabled and is used exclusively by the operator's liveness and readiness probes.
+Keeping health check endpoints on a dedicated plain-HTTP port means probe behaviour is stable regardless of the TLS configuration of the metrics endpoint.
+- **Metrics port** (default: `9403`) — serves `/metrics`.
+In the initial release this port also uses plain HTTP.
```
scholzj: Should the ports be the other way around? The metrics use today 9404 and changing that would require users to change their metric collectors etc. So moving the management port to another port and keeping metrics on 9404 makes more sense. Also, would something like 8080 make more sense for the management port then 9403? We use 8080 in various other projects.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,143 @@
...
+
+Adding TLS support to the metrics port is tracked in [strimzi-kafka-operator#12556](https://github.com/strimzi/strimzi-kafka-operator/issues/12556) and is planned as a follow-up.
+When implemented, it will follow the same CRD pattern used by Kafka Bridge's `spec.http.tls` — a `tls` block under `spec.kafkaExporter` containing a `certificateAndKey` reference and an optional `trustedCertificates` list for mTLS.
+The two-port architecture ensures that enabling TLS on the metrics port will not require any changes to the probe configuration in the operator.
```
scholzj: Should we instead leave it to a separate proposal? I'm not sure this would be the right API to build here just for the TLS in metrics given it is a rare thing.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,143 @@
...
+
+The tool will follow Strimzi versioning `<major>.<minor>.<micro>` as other projects do.
+The first version will be `0.1.0`, even though the tool already covers all Strimzi-required functionality.
+We will keep the tool in the 0.x line for two releases to gather feedback and validate the implementation covers all Strimzi use cases in production before releasing the `1.0.0`.
+
+### Strimzi Kafka Operator changes
+
+#### Feature Gate for Strimzi Kafka Operator
+
+The switch from the Go binary to the Java implementation will be gated behind a new feature gate — `KafkaExporterJavaImplementation` — to allow a safe, opt-in transition for users.
```
scholzj: Maybe we should adjust the name of the feature gate after we agree on the name ... `StrimziLagReporter` might be more use friendly without the Java ...

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,143 @@
...
+
+The gate will progress through the standard Strimzi feature gate lifecycle:
+
+1. Introduced as **alpha** (disabled by default) - users can opt in to the new Java implementation while the Go binary remains the default.
+2. Promoted to **beta** (enabled by default) after two releases (in 1.4.0) once sufficient feedback and production validation has been gathered - users can still opt out by explicitly disabling the gate.
+3. Promoted to **GA** and the feature gate removed after four releases (in 1.6.0) - the Go binary is dropped and the Java implementation becomes the only option.
```
scholzj: In the best case this would be released in 1.3.0, but more likely in 1.4.0. So the versions should be 1.5.0 or 1.7.0 or 1.6.0 and 1.8.0?

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,143 @@
...
+
+When the feature gate is disabled, the operator continues to use the existing Go binary and `kafka_exporter_run.sh` launch script unchanged.
+When the feature gate is enabled, the operator copies the fat-jar artifact from the `strimzi/kafka-exporter` repository into the Kafka image and uses an updated launch script to start the Java implementation.
```
scholzj: I guess this is wrongly worded. The JARs needs to be copied in any case at build time. the feature gate is configured in runtime. So it will likely just switch between two run scripts ... `kafka_exporter_run.sh` and `lag_reporter_run.sh`.

Also, do we want to use fat-jars? I think for most other projects we do not use fat JARs. might be better if the lag reporter repo produces a ZIP archive that we use.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Security
+
+The new implementation preserves the TLS/mTLS configuration used by Strimzi today. 
+The tool continues to use the cluster CA certificate and client certificate/key mounted by the operator via Secrets.
+No credentials will be logged or persisted by the tool itself.
+Moving to a Java implementation removes the current need to trust and verify pre-built third-party Go binaries and their checksums, replacing them with well-known JVM dependencies that already go through Strimzi's existing CVE scanning and patching process.
+
+SASL OAUTHBEARER support will be added as a follow-up when required by Cluster Security.
+Other SASL mechanisms are out of scope and will not be implemented.
```
scholzj: I would slightly reword it as essentially Cluster Security requires it already today.

```suggestion
SASL OAUTHBEARER support will be added as a follow-up.
Other SASL mechanisms are currently out of scope.
```

scholzj (commented):
Thanks for the changes. So more nits ...

157-kafka-exporter-re-implementation.md
scholzj: Lag Reporter works for me. Goes well with Metrics Reporter we already have.

ppatierno: But what if, in the future, it's not going to report only lag?
I know we now have Strimzi Metrics Reporter but maybe in this case we can use Strimzi Kafka Exporter, even if it doesn't work at repository level because it would still be `kafka-exporter` but under the `strimzi` organization. I can't come up with a better name but I think that lag reporter is too restrictive.

157-kafka-exporter-re-implementation.md
ppatierno: I agree with Jakub about keeping metrics on the current 9404 port. For the management one we use 8081 on the bridge today, but I am fine if we go with 8080.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,143 @@
...
+### CI/CD
+
+The tool will follow Strimzi standards and will adopt the same CI/CD workflow we use for other projects.
+As an output of the build and release process, we will produce a tarball with a fat-jar that can be used in Strimzi images or in standalone distributions connected to Kafka.
+
+### Versioning
+
+The tool will follow Strimzi versioning `<major>.<minor>.<micro>` as other projects do.
+The first version will be `0.1.0`, even though the tool already covers all Strimzi-required functionality.
+We will keep the tool in the 0.x line for two releases to gather feedback and validate the implementation covers all Strimzi use cases in production before releasing the `1.0.0`.
```
ppatierno: I am not sure we need this explicit mention about when we'll have a 1.0.0 release. I think this can be kept out of the proposal.

157-kafka-exporter-re-implementation.md
ppatierno: +1

157-kafka-exporter-re-implementation.md
scholzj: Well ... maybe then something like ...
* _Additional Metrics Reporter_
* _Extra Metrics Reporter_
* _Offset Reporter_

I would really prefer NOT using the name of someone else's project.

Frawless:
```
Additional Metrics Reporter
Extra Metrics Reporter
```

These sounds a little bit strange...like why it is not part of metrics-reporter at all

Frawless: actually would that be an option? Have a separated module as part of metrics-reporter? Even if it has a little bit different way of how it works

ppatierno: > like why it is not part of metrics-reporter at all

Which was my first gut feeling about not having a new component but putting these metrics within the Strimzi metrics reporter.

Frawless: The question is whether we want to group metrics related tooling under one repository. In such case it might makes sense to put it into metrics-reporter. However, it has different architecture than metrics-reporter. ME basically export existing metrics from clients/server. KE polls the metrics via Kafka Admin API and expose them to users. So probably there won't be much things that can be shared? Maybe HTTP server configs, but we will need to unite the approach for both tools (no big deal I guess). Also we will probably commit ourself to support it as standalone tool as well. If it will be in separated repo, we can more easily say it is just tool for Strimzi images.

scholzj: I don't think it makes sense to have those in the same project. They are very different in terms of the distribution mechanism (Maven Central JARs versus archive), their nature (Plugin versus standalone application), and likely also the resulting lifecycle (one is integrated into the clients / brokers) while others is a separate application relying on a fully public APIs. So to be honest I do not see how you could put it into a single repository without making things more complicated.

Frawless: I agree with Jakub after brief checking of the metrics-reporter, it's purpose, usage, etc. So let's get back to picking different name. 

- cluster-metrics-reporter
- kafka-cluster-metrics-reporter
- kafka-admin-metrics-reporter

maybe?

scholzj: TBH, I think ht sugestions I made earlier are better ... they are shorter and I think they also better explain what it does. 😉

Frawless:
When I wanted to apply latest changes I realized that with changing the name we will also need to change the API. Like `KafkaExporterSpec` won't make sense for new tool. I had a short chat with @scholzj about it and we can basically skip FG part as LagReporter will be gated by different API anyway. We will just deprecate KE API and let it live until next major version or until it will not break itself (best-effort support with direct replacement in form of LagReporter).

New API will not influence existing users so it can be experimental for a while (we do not need to deprecate KE right after we will add LagReporter integration) and we will have time to think about TLS section for example.

@strimzi/maintainers wdyt?

157-kafka-exporter-re-implementation.md
Frawless: So I am for `Lag Reporter` or `Offset Reporter`. Renaming it sounds reasonable to me.

ppatierno:
> New API will not influence existing users so it can be experimental for a while (we do not need to deprecate KE right after we will add LagReporter integration) and we will have time to think about TLS section for example.

I think this makes sense but ... at least to me ... the name is not good. Let's continue the conversation on the dedicated comment, I think we need more maintainers to jump into it.

157-kafka-exporter-re-implementation.md
ppatierno: My point is, what if this tool will report something which is not a lag or offset in the future? Even KE was not named lag/<specific metric> exporter but it had a more generic name and I think we should go in the same direction. Maybe we can still use the "exporter" suffix instead of "reporter" and trying different variations of the names suggested by @Frawless.
Some additional ones trying to show the derived nature of the metrics:

* kafka-insights-reporter (or exporter)
* kafka-stats-reporter (or exporter)

Of course, I won't be -1 and I won't block the proposal if others will agree on lag/offset reporter. 
@katheris @tinaselenge @im-konge can we get your opinion here?

scholzj: I think we should drop the `kafka-` prefix. But I can probably live with these. Although I think insights is a bit more fancy than what it does. 😃

Frawless: `insights-reporter` sounds good, little bit fancy, but good :D

im-konge: `insights-reporter` looks good to me.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,145 @@
...
+- **Metrics port** (default: `9404`) — serves `/metrics`.
+In the initial release this port also uses plain HTTP.
+
+There is community demand for having TLS enabled on metrics endpoint ([strimzi-kafka-operator#12556](https://github.com/strimzi/strimzi-kafka-operator/issues/12556)).
+The implementation and integration of TLS configuration into Strimzi Kafka Operator will require a new proposal as it will need API changes for the tool.
+
+### CI/CD
+
+The tool will follow Strimzi standards and will adopt the same CI/CD workflow we use for other projects.
+As an output of the build and release process, we will produce a zip file that we use across Strimzi org and that can be used in Strimzi images or in standalone distributions connected to Kafka.
```
scholzj, line 53: One thing for consideration ... if it has separate repo, separate release lifecycle, etc. Should it has its own image like Bridge does? 🤔 

No strong opinions from my side, just something that came to my mind.

I think the pros would be:
* Separate images are cleaner approach. The reason KE is in the Kafka image is mainly that it does not have a separate repo in Strimzi and we did not wanted to have a separate image for it due to Strimzi Operator CIs.
* With separate images, less people who do not even use it would be bitching about CVEs in the insight reporter dependencies etc.

The cons would be that any base image CVE fixing would need to happen separately from the operators with separate respins.

As I said, just something what came to my mind, no strong opinions from myself.

@ppatierno As you do most Bridge stuff, any thoughts on this? Do you think the current Bridge approach is good? Or not?

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Strimzi Kafka Operator changes
+
+#### CRD and API Changes
+
+Insights Reporter will be introduced as a new, parallel section in the Kafka CR — `spec.insightsReporter` — alongside the existing `spec.kafkaExporter` section.
+This allows users to opt in to the new Java implementation without any changes to their existing `kafkaExporter` configuration.
+
+The `spec.kafkaExporter` section will be set as **deprecated** from the moment `spec.insightsReporter` is introduced.
+The upstream `kafka_exporter` binary will be updated to v1.10.0 and continue to be supported and bundled in Strimzi images until new API version or until the tool will work without significant required changes on our side.
```
scholzj, line 68: * The 1.10 update already happened
* I would probably mention that we will keep updating it `until new API version or until the tool will work without significant required changes on our side`.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,145 @@
...
+- `logging` — standard Strimzi `Logging` type (Log4j 2), replacing the plain string `logging` field from `kafkaExporter`.
+- `jvmOptions` — standard Strimzi `JvmOptions` type for JVM heap, GC options, and flags.
+- `resources` — CPU and memory resource requirements.
+- `livenessProbe` — liveness probe configuration.
+- `readinessProbe` — readiness probe configuration.
+- `template` — pod and container template overrides.
+
+Fields that exist in `kafkaExporter` but are Go-specific, such as `enableSaramaLogging`, will not be carried over to `spec.insightsReporter`.
+
+When both `spec.kafkaExporter` and `spec.insightsReporter` are set, the operator will emit a warning and use `spec.insightsReporter`, ignoring `spec.kafkaExporter`.
```
scholzj, line 90: Is there any reason for this? Users might want to run them in parallel to migrate, compare the metrics, etc., or?

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Testing
+
+The new repository will include unit tests covering the metrics collection and registry logic, and integration tests running against a real Kafka instance using `strimzi-test-container`.
+Existing system tests in `strimzi-kafka-operator` will be extended to cover Insights Reporter and will continue to cover the existing Kafka Exporter for the duration of the deprecation period.
+
+### Security
+
+The new implementation preserves the TLS/mTLS configuration used by Strimzi today.
+The tool continues to use the cluster CA certificate and client certificate/key mounted by the operator via Secrets.
```
scholzj, line 105: I expect we would want to use PEM files and config providers straight away?

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,145 @@
...
+No credentials will be logged or persisted by the tool itself.
+Moving to a Java implementation removes the current need to trust and verify pre-built third-party Go binaries and their checksums, replacing them with well-known JVM dependencies that already go through Strimzi's existing CVE scanning and patching process.
+
+SASL OAUTHBEARER support will be added as a follow-up.
+Other SASL mechanisms are currently out of scope.
```
scholzj, line 110: I guess I originally introduced this when pointing out what to not port from KE. That said:
* The SA authentication is merged now. So we would need to introduce some checks if OAUTHBEARER is not supported initially. I planned that for KE originally, but the new 1.10 release saved me.
* This is actually super simple:
    * Include Strimzi OAuth as dependency
    * 3 lines in the Kafka client configuration

Similarly, if we consider standalone use, the other SASL mechanisms are basically supported out of the box.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,145 @@
...
+
+We will create new repository under Strimzi organization - `strimzi/insights-reporter` - that will contain the implementation code.
+
+### Implementation
+
+Because Strimzi mostly consists of Java projects, the proposal is to implement the tool in Java.
+The implementation will use the minimal required dependencies and will cover all functionality currently used by Strimzi.
+Features not used by Strimzi are explicitly out of scope for the initial implementation, including Helm charts, Kerberos authentication, and SASL mechanisms other than OAUTHBEARER.
+
+All the metrics that the tool will export will follow the same naming as the existing Kafka Exporter, but this can change in the future based on community feedback and needs.
```
scholzj, line 30: Just want to make sure we considered this ... are we fine with the metrics as they are?

The original approach required them to be identical. But as we now introduce it as a separate tool, we can actually change things if we want. But we would need to do it right now. Remove some metrics, rename some metrics, etc.

It is not like I have any particular changes in mind. Just want to make sure we _considered_ it, because we will not have the chance again.

scholzj (commented):
Sorry, few more things to consider ...

157-kafka-exporter-re-implementation.md
Frawless: I change the ports, 8080 for management and 9404 for metrics.

157-kafka-exporter-re-implementation.md
scholzj: Just to clarify ... we obviously do not need to do it in the same PR. But we should not leave that for another proposal as that would take possibly weeks to approve etc.

157-kafka-exporter-re-implementation.md
ppatierno: That's a good point. I remember that back in the days we had several more images within Strimzi and then we decided to consolidate within fewer ones. IIRC for example Cruise Control had its own but then we moved to include it within Kafka? (but my memories could be wrong). Said that to avoid increasing the number of images it would be fine for me having the insights-reporter within the Kafka image because in the end belongs to Kafka. It was different for the HTTP bridge which is more a totally different thing component-wise. But I am open to any discussion if we think a new image could bring advantages.

Frawless: Given that we could have possible troubles with ubi10 and "fips-agent", I would go with having `insights-reporter` inside Kafka images and in case we will find it will make sense to put it into separeted image we can do it later.

scholzj: Ok, sounds good.

ppatierno: Fine with me as well.

157-kafka-exporter-re-implementation.md
Frawless: Fair point, I was thinking about it just a little bit. Question is - if users will run both in parallel, wouldn't it cause some metrics name collisions? Like we will probably need to prefix all of the metrics?

157-kafka-exporter-re-implementation.md
Frawless: Makes sense, I will update the wording

157-kafka-exporter-re-implementation.md
Frawless: I think it will not be that hard to rename the metrics until we will integrate it to Strimzi, but this might be possible problem https://github.com/strimzi/proposals/pull/245#discussion_r3988397478 . We will also need to create new dashboards for `insights-reporter`, right?

157-kafka-exporter-re-implementation.md
Frawless: Yes, I will update the proposal

157-kafka-exporter-re-implementation.md
Frawless: Valid point

157-kafka-exporter-re-implementation.md
scholzj: It might. But you can scrape only one of them. Or scrape them with different Prometheus. I think warning is good. But ignoring one of them might not work for every situation.

157-kafka-exporter-re-implementation.md
scholzj: Yes, we would need a new dashboard. As I said, it is not like I have any ideas what and how should be renamed. Or that some metric should be dropped. It just that with the new way its not impossible anymore. But maybe we just do not have any ideas and want to keep things as they are because that is easiest. (I mean, remember the discussion on prpject name ... renaming metrics might not be easier 😉)

Frawless: Ye...I think changing stuff will be possible even in the future. Would it make sense to define some kind of tech-preview phase during which we will evaluate stuff and possibly change the names, etc?

scholzj: I guess you can call it _early access_ as we do with the Cert Manager or cluster security.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Proposal
+
+To mitigate security problems and allow us to better maintain the tool and provide new features, we should re-implement Kafka Exporter tool.
+
+We will create new repository under Strimzi organization - `strimzi/insights-reporter` - that will contain the implementation code.
+
+### Implementation
+
+Because Strimzi mostly consists of Java projects, the proposal is to implement the tool in Java.
```
tomncooper, line 26: Not to mention being able to base the tool on the supported Admin client from the Apache Kafka project itself rather than a third-party go client.

157-kafka-exporter-re-implementation.md
Frawless: Ye...I did changes where I kept the warning but let them both run alongside. It will be part of the next push.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,145 @@
...
+The latest release, v1.10.0, was published on 2026-09-08 after a period of minimal activity, but the project still has 216 open issues and 56 open pull requests with limited attention from the code owner.
+This opens space for potential security issues as CVEs may not be fixed promptly.
+
+This situation is also not helpful for new features that we would like to add into the exporter.
+
+## Proposal
+
+To mitigate security problems and allow us to better maintain the tool and provide new features, we should re-implement Kafka Exporter tool.
+
+We will create new repository under Strimzi organization - `strimzi/insights-reporter` - that will contain the implementation code.
```
tomncooper: Insights-reporter feels very generic/broad, if we are planning to exand the tool to do provide many more custom metrics and other features in future then that seems appropriate. 

But if it going to stay as a CG lag reporter then maybe `client-insights-reporter` or `consumer-insights-reporter` would be more appropriate?

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Implementation
+
+Because Strimzi mostly consists of Java projects, the proposal is to implement the tool in Java.
+The implementation will use the minimal required dependencies and will cover all functionality currently used by Strimzi.
+Features not used by Strimzi are explicitly out of scope for the initial implementation, including Helm charts, Kerberos authentication, and SASL mechanisms other than OAUTHBEARER.
```
tomncooper, line 28: Line 109 says OAUTHBEARER will be done in a follow-up not the initial implementation?

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,145 @@
...
+The new `spec.insightsReporter` section follows the same conventions as other Strimzi components such as CruiseControl.
+The component is enabled by including the section in the CR and disabled by omitting it — there is no separate `enabled` field, consistent with `spec.kafkaExporter` and `spec.cruiseControl`.
+The section will include the following fields from the start:
+
+- `image` — container image override.
+- `groupRegex` — consumer group include regex (default: `.*`).
+- `groupExcludeRegex` — consumer group exclude regex.
+- `topicRegex` — topic include regex (default: `.*`).
+- `topicExcludeRegex` — topic exclude regex.
+- `showAllOffsets` — whether to report offsets for all partitions the group has ever committed to (default: `true`).
```
tomncooper, line 80: I know this is a high level proposal, but as you have mentioned replicating this setting, I think it would be good to discuss the collection model that the exporter uses.

In the kafka_exporter this setting default is what makes each Prometheus scrape fetch committed offsets for _every_ partition of _every_ group, synchronously, inside the scrape request. That's the root of the most common type of upstream issues:  scrapes timing out on large clusters and the exporter effectively going blind under load (e.g. [#9](https://github.com/danielqsj/kafka_exporter/issues/9), [#108](https://github.com/danielqsj/kafka_exporter/issues/108), [#377](https://github.com/danielqsj/kafka_exporter/issues/377), [#497](https://github.com/danielqsj/kafka_exporter/issues/497), most still open). Because Prometheus drops a scrape entirely when it exceeds the timeout, this shows up as `up == 0` right when the cluster is busiest.

A ground-up re-implementation gives us the chance to address this issue. We could decouple collection from the scrape, refresh a cached snapshot on a background timer and have `/metrics` just serve the latest snapshot. This is the pattern the official Prometheus [exporter guidance](https://prometheus.io/docs/instrumenting/writing_exporters/) recommends for expensive collectors, and which `prometheus-metrics-core` supports directly.

I'm not saying we need low-level implementation discussion here, but the proposal does sound like a like-for-like reimplementation (with a few parts removed). I think it would be worth this proposal mentioning improving the underlying collection model to address scaling issues with the existing kafka_exporter.

tomncooper (commented):
I think this new CG lag exporter is long overdue. My only comment was around using this reimplementation as an opportunity to use a more scalable collection model to help with Kafka clusters with large numbers of topics/partitions/CGs.

157-kafka-exporter-re-implementation.md
Frawless: It was just wrong leftover, I should be fixed now.

157-kafka-exporter-re-implementation.md
Frawless: I removed `showAllOffsets` from API as it is not used at all in my PoC and I think it doesn't make much sense. PoC implementation basically counts with what you described. I added a few lines about the into the proposal.

scholzj: We definitely **do want** it to show all offsets. Showing only active consumers makes only little sense. That has nothing to do with the implementation behind it - whether it is synchronous or not.

scholzj (approved):
LGTM. Thanks.

157-kafka-exporter-re-implementation.md
mimaison: Dos this mean users won't be able to use this with a Kafka cluster not using SASL OAuth?

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,163 @@
...
+### Implementation
+
+Because Strimzi mostly consists of Java projects, the proposal is to implement the tool in Java.
+The implementation will use the minimal required dependencies and will cover all functionality currently used by Strimzi.
+Features not used by Strimzi are explicitly out of scope for the initial implementation, including Helm charts, Kerberos authentication, and SASL mechanisms other than OAUTHBEARER.
+
+All the metrics that the tool will export will follow the same naming as the existing Kafka Exporter, but this can change in the future based on community feedback and needs.
+
+To keep the minimal dependency tree we will use the following:
+- `kafka-clients` for `AdminClient`
```
mimaison: `AdminClient` -> `Admin`. `AdminClient` should not be used.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,163 @@
...
+- `prometheus-metrics-core` and `prometheus-metrics-exposition-formats` for Prometheus endpoint
+- JDK built-in HTTP server for the `/metrics` and health check endpoints
```
mimaison, line 36: metrics-reporter uses `prometheus-metrics-exporter-httpserver` for exposing metrics

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,163 @@
...
+
+#### Collection Model
+
+Metrics collection is decoupled from the Prometheus scrape.
+A background scheduler runs a collection cycle on a configurable interval (default 30 seconds).
+Each cycle executes a sequence of batched `AdminClient` calls — cluster description, topic listing and description, offset fetching, consumer group listing, group description, and committed offset fetching — and assembles the results into an immutable snapshot.
```
mimaison: Again it should be `Admin`

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,163 @@
...
+
+#### HTTP Server
+
+The tool exposes two HTTP listeners on separate ports, following the same pattern used by the Kafka Bridge:
+
+- **Management port** (default: `8080`) — serves `/healthz/ready` over plain HTTP only.
+This port is never TLS-enabled and is used exclusively by the operator's liveness and readiness probes.
+Keeping health check endpoints on a dedicated plain-HTTP port means probe behaviour is stable regardless of the TLS configuration of the metrics endpoint.
+- **Metrics port** (default: `9404`) — serves `/metrics`.
+In the initial release this port also uses plain HTTP.
```
mimaison, line 57: We're adding https to all the components, shouldn't we include it by default in this new one. We say below that there's demand for it and it's pretty much already descroibed in the following paragraph.

mimaison (commented):
Thanks for the proposal. I left a few comments

157-kafka-exporter-re-implementation.md
scholzj: The _demand_ is one user who opened an issue. And SSL will bring whole set of questions and discussions around the API etc. So I think leaving it out makes sense.

157-kafka-exporter-re-implementation.md
Frawless: > Dos this mean users won't be able to use this with a Kafka cluster not using SASL OAuth?

Isn't this supposed to be on different line? 

This is meant for internal Kafka <-> Insights Reporter communication that Strimzi supports after changes described in [cluster security proposal. ](https://github.com/strimzi/proposals/blob/main/150-configurable-security-of-internal-communication.md) .  This match with current KE + Strimzi integration implementation.

mimaison: > Isn't this supposed to be on different line?

It's unfortunate that this will only be usable with Strimzi. I think it would be worth supporting any Kafka clusters.

157-kafka-exporter-re-implementation.md
mimaison: We've had proposals with metrics-reporter and strimzi-kafka-bridge that have established a pattern. I'd expect the configurations to be identical for this new component. In my mind, having it in the proposal does not force it to be included in the initial commit or release.

scholzj: I do not think there is any common pattern between the Bridge and metrics reporter TLS. Also, none of those components had it included in their initial proposal.

mimaison: > I do not think there is any common pattern between the Bridge and metrics reporter TLS

The components have different prefixes (`http` for `strimzi-kafka-bridge`, and `prometheus.metrics.reporter.listener` for `metrics-reporter`) but otherwise the configurations pretty much match.

| strimzi-kafka-bridge | metrics-reporter |
| - | - |
| http.ssl.enable | Enabled if prometheus.metrics.reporter.listener has an entry with https |
| http.ssl.certificate.location | prometheus.metrics.reporter.listener.ssl.certificate.location |
| http.ssl.key.location | prometheus.metrics.reporter.listener.ssl.key.location |
| http.ssl.certificate | prometheus.metrics.reporter.listener.ssl.certificate |
| http.ssl.key | prometheus.metrics.reporter.listener.ssl.key |
| http.ssl.enabled.protocols | prometheus.metrics.reporter.listener.ssl.enabled.protocols |
| http.ssl.enabled.cipher.suites | prometheus.metrics.reporter.listener.ssl.enabled.cipher.suites |

Am I missing something?

scholzj: Yes. You are missing the actual Strimzi API. The CRDs. That is what really matters here.

ppatierno: > I do not think there is any common pattern between the Bridge and metrics reporter TLS.

They seem to be pretty similar as described by Mickael and the same kind of configuration is coming with the MQTT bridge as well.

> Also, none of those components had it included in their initial proposal.

Which doesn't mean we can't have SSL since the beginning for it. At some point we could have metrics reporter having SSL soon but not the insights-reporter, so actually using an encrypted channel for some metrics but not for others.

> Yes. You are missing the actual Strimzi API. The CRDs. That is what really matters here.

I see your point but why we can't envisage the API shape within the new `spec.insightsReporter` by supporting SSL configuration via some additional fields through this proposal?

scholzj: Well, Kafka Exporter is out now with the support I needed for cluster security. So I'm actually not in a hurry with this anymore. 🤷 

So we can keep discussing this proposal for months as with other proposals. Lets then include also OTLP which is way more important for the future than something like HTTPS. Let's go into the technical details about whether it will have a configuration file like Bridge does (I actually assumed it would not) and how will the different APIs look like, etc.

----

But I think the main motivation for this proposal is to address the issues with the Kafka Exporter (bus factor, release frequency, CVEs, ...). If that is the case, then the focus of the proposal should be what we need to have to replace Kafka Exporter. Things such as HTTPS or OTLP do not belong there and should be handled in separate proposals.

And if that helps us to keep things more lean, agile, and moving forward instead of being planned for months, that is only an added benefit.

157-kafka-exporter-re-implementation.md
Frawless: There will be space for improvements in the future. Such feature can be proposed by additional proposal.

157-kafka-exporter-re-implementation.md
Frawless: We are more and more slipping from original idea - just write what we need ASAP to have quick replacement. In case we will go with fork, will we add https, etc. to initial fork as well?

I think adding new features into initial proposal might slow-down not just proposal review process, but also reviews of the implementation and integration PRs as they will need to handle more things at the beggining to just full-fill the proposal. I would like to keep it as simple as possible and anyone can open additional proposals right after this one is merged. The additional proposals will be smaller, easier to review and implement.

157-kafka-exporter-re-implementation.md
katheris: I realise that I am late to this thread, but I am not sure on the `insights-reporter` naming. To me it implies interpretation of the data, not just exposing stats, and given we already have a metrics reporter I worry that users would assume it's supplying guidance rather than just metrics.

Looking at the earlier suggestions Jakub had around `additional-metrics-reporter` I wondered about something like `extended-metrics-reporter` or even `derived-metrics-reporter`, making it distinct from the existing metrics reporter and making it clear that it's metrics that are being created based on something else..

naming is always hard!

scholzj: > I realise that I am late to this thread, but I am not sure on the insights-reporter naming. To me it implies interpretation of the data, not just exposing stats, and given we already have a metrics reporter I worry that users would assume it's supplying guidance rather than just metrics.

It does not just expose stats. It does actually interpret the data. That is how it calculates the lag for example. And if we one day add something such as time-based lag for example, you would be putting even more insights into it. So I do not think it just exposes the stats.

> Looking at the earlier suggestions Jakub had around additional-metrics-reporter I wondered about something like extended-metrics-reporter or even derived-metrics-reporter, making it distinct from the existing metrics reporter and making it clear that it's metrics that are being created based on something else..

I think these names are much more confusing now we have `metrics-reporter` project because they directly link to it by just adding a prefix. So they actually suggest that this extends / adds to the metrics reporter metrics which is not the case. (And yes, I know who suggested _additional metrics reporter_ first 😉)

Frawless: how about something like `admin-reporter` or `kafka-admin-reporter` ? Or `telemetry-reporter` , `admin-telemetry-reporter` ? But it could be a little bit confusing with open-telemetry

im-konge (approved):
LGTM, thanks.
I think that the rewrite as it is today in Go to Java is sufficient for the start, things like HTTPS and other can be added later on - as it was done in other projects. I don't think that we need to "have everything day one". These improvements can be discussed in separate proposals (if needed).

Thanks for working on this :)

157-kafka-exporter-re-implementation.md
ppatierno: I got some other fancy suggestions from AI :-) What about `kafka-lens` or `kafka-pulse`?
or without the kafka prefix we can have `metrics-probe`, `metrics-pulse`, `metrics-lens`? or some variations putting "cluster" instead of "metrics"?

scholzj: I think `lens` collides with Lenses. `probe` doesn't really fit for me. `pulse` ... seems strange but probably best from these. But they suggest some kind heartbeat feature. So I would probably associate it mostly with Canary rather than consumer lag for example. 

Honestly, I think the insights are the best we have so far if we do not want to go with the lag focus and have something like `lag-reporter` or `offset-reporter`.

see-quick: IMO, I would call it `strimzi-kafka-exporter`? As it replaces the external `kafka_exporter`. It we want create completely new name it would cause confusion across all of us I think ... but if we **_really_** want different name then I am +1 for the `lag-reporter` as I think that's main think what is does (I think that someone already suggested this previosuly) and I know that in the future we might have more and more metrics ... but I am not sure we can name it better.

ppatierno: I am still not convinced about specific well focused names with "lag" or "offset" because we can have this thing evolving in something more and the name won't make any sense soon but we'll be blocked on it because changes will take more troubles than advantages.

scholzj: That is why I liked the Insights ... it IMHO describes pretty well what it does and what it might do in the future in an open-ended way.

tinaselenge: As @see-quick mentioned, can we keep the original name as `strimzi-kafka-exporter`? Or `strimzi-metrics-exporter` but I guess that might be too similar to `metrics-reporter`. 

I also like @Frawless 's suggestion from earlier, `client-metrics-reporter` or `client-metrics-exporter` since we are exporting metrics from Kafka clients.

im-konge: I wouldn't name it strimzi-kafka-exporter, it doesn't feel right to me.

For me, insights-reporter is a good name and we picked that up before. I also approved the proposal thinking that we keep this name.

Frawless: I think we should avoid anything similar to original name. I completely agree with Jakub that we should came up with something original.

Frawless: So I was thinking a little bit and I found almost perfect name - Strimzi Sova. Sova is a great animal! They have excellent vision in low light and exceptional hearing - pretty great for watching metrics. It is very precise and silent observer that perfectly fits to our use case. And finally, Sovy are always connected with wisdom, which is needed for evaluating metrics. The spec could look like

```
spec:
   sova: {}
```

oh and I almost forgot! Sova looks pretty cool!

image

im-konge: `strimzi-sova` looks good to me 😏 if `insights-reporter` is not the way

im-konge: image
AI generated, but nice 😄

tinaselenge: I'm +1 for strimzi-sova! I prefer this over the names we are trying to come up with that describes what the tool does.

see-quick: +1 to `strimzi-sova`

ppatierno: Not sure how it sounds in your native language but for me `strimzi-sova` sounds pretty cool! +1
PS: I am not suggesting the italian version which is `strimzi-gufo` because I would not like it :-D

scholzj: TBH, I do not think this really solves anything. We can name the project whatever we want. That is fine as nobody really cares about it.

But the CRD needs to have some aspect of readability in order to provide a good UX. `.spec.sova` is something nobody would understand, it would not be clear to anyone as nobody would understand the meaning. It would be also prone to typos because nobody will know what it means.

So we still need some more reasonable name for that. So would that be `.spec.insights`?

tinaselenge: I didn't vote for insights because it might be very common/generic name across projects, not to mention there is Red Hat Insights as well. I understand in terms of typing, because it's not an english word, could be prone to typos but it's a fairly short name.

scholzj: Ok, fine. Let's not go with insights then. That said, I do not think we can go with `.spec.sova`. It needs to be something what makes some sense. So as I said, we still need to decide that. But maybe the naming conflicts are different in the CR from the project name. So maybe `.spec.extraMetrics`, `.spec.additionalMetrics`, or `.spec.offsetMetrics` would work?

ppatierno: yeah I agree that the name of the component doesn't need to match with the API within the spec, but still again to repeat myself I am not for using "offset" but I am fine with the other suggestions from Jakub.

Frawless: Variations of _additional metrics_ was mentioned few times already. Having it in Kafka CRD is fine for me as `.spec.additionalMetrics` .  `.spec. extraMetrics` is good as well....extra looks a little bit as exporter with that `ex` :D

scholzj: `extraMetrics` is also shorter and has less thinking about how many `d` are there etc. :-D

Frawless: fair point, now I like it more :D

ppatierno: @Frawless are you going to update the proposal with the new agreement in terms of name and spec? So I will have another pass.

Frawless: Sure, I will update the proposal during today.

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,163 @@
...
+### Versioning
+
+The tool will follow Strimzi versioning `<major>.<minor>.<micro>` as other projects do.
+The first version will be `0.1.0`, even though the tool already covers all Strimzi-required functionality.
+
+### Strimzi Kafka Operator changes
+
+#### CRD and API Changes
+
+Insights Reporter will be introduced as a new, parallel section in the Kafka CR — `spec.extraMetrics` — alongside the existing `spec.kafkaExporter` section.
```
tinaselenge: `Insights Reporter` ?

157-kafka-exporter-re-implementation.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Documentation
+
+The `strimzi/sova` repository will include a README covering all configuration options, the full list of exported metrics, and instructions for running the tool standalone.
+The Strimzi documentation will be updated to reflect the new component and the deprecation of `spec.kafkaExporter`.
+
+### Testing
+
+The new repository will include unit tests covering the metrics collection and registry logic, and integration tests running against a real Kafka instance using `strimzi-test-container`.
+Existing system tests in `strimzi-kafka-operator` will be extended to cover Insights Reporter and will continue to cover the existing Kafka Exporter for the duration of the deprecation period.
```
tinaselenge:
```suggestion
Existing system tests in `strimzi-kafka-operator` will be extended to cover Sova and will continue to cover the existing Kafka Exporter for the duration of the deprecation period.
```

tinaselenge (commented):
Thanks for the proposal. It looks good to me, just a couple minor of comments.

PR #246, opened 2026-09-15
Donation of StreamsHub MCP servers to Strimzi 

Frawless:
### Type of Change

- New Proposal

### Description

This proposal covers donation of StreamsHub MCP servers to Strimzi organization. 

### Checklist


157-mcp-servers.md
```diff
@@ -0,0 +1,201 @@
...
+The last release under the StreamsHub organisation will be `0.3.0`.
+The first release under the Strimzi organisation will be `0.4.0`, continuing the existing version sequence.
+
+### Repository structure
+
+The repository is a Maven multi-module project.
+Currently, it contains the following parts:
+
+```
+mcp-servers/
```
scholzj: So, this is actually the repository root I guess?

157-mcp-servers.md
```diff
@@ -0,0 +1,201 @@
...
+├── common/                        # Shared SPI interfaces and utilities
+├── metrics-prometheus/            # Prometheus/Thanos/VictoriaMetrics metrics provider
+├── loki-log-provider/             # Grafana Loki log provider
+├── elasticsearch-log-provider/    # Elasticsearch and OpenSearch log provider
+├── strimzi-mcp/                   # Strimzi MCP server
+└── systemtest/                    # System test suite
+```
+
+Future MCP servers can be added as new modules alongside the existing ones.
+As a good example of additional MCP server could be Kafka MCP server.
```
scholzj: Doesn't this go against the `Strimzi is the natural home for a Strimzi-specific MCP server.` idea?

157-mcp-servers.md
```diff
@@ -0,0 +1,201 @@
...
+
+The `common` module defines stable SPI contracts that allow alternative implementations to be selected via configuration without forking the repository:
+
+- `MetricsProvider` — selected via `mcp.metrics.provider`.
+  The default implementation scrapes Prometheus metrics directly from pod HTTP endpoints.
+  The bundled `metrics-prometheus` module provides an alternative that queries an existing Prometheus, Thanos, or VictoriaMetrics instance.
+- `LogCollectorProvider` — selected via `mcp.log.provider`.
+  The default implementation reads logs directly from Kubernetes pod logs via the Fabric8 client.
+  The bundled `loki-log-provider` and `elasticsearch-log-provider` modules provide alternatives for Grafana Loki and Elasticsearch or OpenSearch respectively.
+- `GuardrailFilter` — a CDI filter chain where multiple filter beans apply in priority order, covering input validation, output sanitisation, log redaction, and response size limiting.
```
scholzj: Can you provide more explanation of what this is, what it does, etc.?

157-mcp-servers.md
```diff
@@ -0,0 +1,201 @@
...
+### Pluggable provider architecture
+
+The `common` module defines stable SPI contracts that allow alternative implementations to be selected via configuration without forking the repository:
+
+- `MetricsProvider` — selected via `mcp.metrics.provider`.
+  The default implementation scrapes Prometheus metrics directly from pod HTTP endpoints.
+  The bundled `metrics-prometheus` module provides an alternative that queries an existing Prometheus, Thanos, or VictoriaMetrics instance.
+- `LogCollectorProvider` — selected via `mcp.log.provider`.
+  The default implementation reads logs directly from Kubernetes pod logs via the Fabric8 client.
+  The bundled `loki-log-provider` and `elasticsearch-log-provider` modules provide alternatives for Grafana Loki and Elasticsearch or OpenSearch respectively.
```
scholzj, line 105: Why are these things needed? Why not integrate with the corresponding MCPs instead?

157-mcp-servers.md
```diff
@@ -0,0 +1,201 @@
...
+- `GuardrailFilter` — a CDI filter chain where multiple filter beans apply in priority order, covering input validation, output sanitisation, log redaction, and response size limiting.
+
+### Security and RBAC
+
+The MCP server uses a dedicated ServiceAccount with a minimal ClusterRole that grants only `get`, `list`, and `watch` on Strimzi custom resources, pod resources, and related Kubernetes resources.
+An opt-in per-namespace Role for sensitive resources (Secrets for certificate metadata, and `pods/proxy` for direct metrics scraping) is provided separately.
+Both roles are modelled on the existing `strimzi-view` ClusterRole.
```
scholzj: What is modified on them?

157-mcp-servers.md
```diff
@@ -0,0 +1,201 @@
...
+  The bundled `metrics-prometheus` module provides an alternative that queries an existing Prometheus, Thanos, or VictoriaMetrics instance.
+- `LogCollectorProvider` — selected via `mcp.log.provider`.
+  The default implementation reads logs directly from Kubernetes pod logs via the Fabric8 client.
+  The bundled `loki-log-provider` and `elasticsearch-log-provider` modules provide alternatives for Grafana Loki and Elasticsearch or OpenSearch respectively.
+- `GuardrailFilter` — a CDI filter chain where multiple filter beans apply in priority order, covering input validation, output sanitisation, log redaction, and response size limiting.
+
+### Security and RBAC
+
+The MCP server uses a dedicated ServiceAccount with a minimal ClusterRole that grants only `get`, `list`, and `watch` on Strimzi custom resources, pod resources, and related Kubernetes resources.
+An opt-in per-namespace Role for sensitive resources (Secrets for certificate metadata, and `pods/proxy` for direct metrics scraping) is provided separately.
```
scholzj: How does this fit together? Or how do these work when the Strimzi CRs, Pods, etc. are granted cluster-wide and secrets only namespace wide?

Also, what is the pods/proxy requirement and how does it fit the integration for metrics providers?

157-mcp-servers.md
```diff
@@ -0,0 +1,201 @@
...
+Both roles are modelled on the existing `strimzi-view` ClusterRole.
+
+The initial authorization model relies entirely on Kubernetes RBAC.
+MCP-level authentication and authorization is explicitly out of scope for the initial release.
+It is intended as follow-up work under the Strimzi organisation and will be covered by a separate proposal.
```
scholzj, line 119: So, every user of the MCP has the same access level and the MCP functionality is limited only by the MCP RBAC? Is that what this is trying to say?

157-mcp-servers.md
```diff
@@ -0,0 +1,201 @@
...
+The Maven `groupId` changes from `io.streamshub` to `io.strimzi`.
+The artefact IDs change accordingly:
+
+| Module                     | Old `artifactId`                        | New `artifactId`                      |
+|----------------------------|-----------------------------------------|---------------------------------------|
+| Parent POM                 | `streamshub-mcp`                        | `strimzi-mcp-servers`                 |
+| Shared SPI                 | `streamshub-mcp-common`                 | `strimzi-mcp-common`                  |
+| Prometheus metrics         | `streamshub-metrics-prometheus`         | `strimzi-metrics-prometheus-provider` |
+| Loki log provider          | `streamshub-loki-log-provider`          | `strimzi-loki-log-provider`           |
+| Elasticsearch log provider | `streamshub-elasticsearch-log-provider` | `strimzi-elasticsearch-log-provider`  |
```
scholzj: Should these artifacts have some `mcp` in the name as well? Or sould there be some group IT that incorporates the MCP? E.g. `io.strimzi.mcp` or `strimzi-mcp-metrics-prometheus-provider` etc.?

157-mcp-servers.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Naming and Maven coordinates
+
+The repository name changes from `streamshub-mcp` to `mcp-servers`.
+The Maven `groupId` changes from `io.streamshub` to `io.strimzi`.
+The artefact IDs change accordingly:
+
+| Module                     | Old `artifactId`                        | New `artifactId`                      |
+|----------------------------|-----------------------------------------|---------------------------------------|
+| Parent POM                 | `streamshub-mcp`                        | `strimzi-mcp-servers`                 |
```
scholzj: Why servers in plural? Would it mabe make more sense to call the parent pom `strimzi-mcp` and the current `strimzi-mcp` name `strimzi-mcp-server`?

157-mcp-servers.md
```diff
@@ -0,0 +1,201 @@
...
+| Prometheus metrics         | `streamshub-metrics-prometheus`         | `strimzi-metrics-prometheus-provider` |
+| Loki log provider          | `streamshub-loki-log-provider`          | `strimzi-loki-log-provider`           |
+| Elasticsearch log provider | `streamshub-elasticsearch-log-provider` | `strimzi-elasticsearch-log-provider`  |
+| Strimzi MCP server         | `strimzi-mcp`                           | `strimzi-mcp` (unchanged)             |
+
+The `strimzi-mcp` server module already uses the target artefact ID, so only its `groupId` changes.
+
+Container images move from `quay.io/streamshub/strimzi-mcp` to `quay.io/strimzi/strimzi-mcp`.
+
+### Provider selector values
```
scholzj: What exactly are _provider selector values_?

157-mcp-servers.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Conditions of transfer
+
+The StreamsHub maintainers approved the donation subject to the following conditions, all of which this proposal accepts:
+
+- All released artefacts (container images and Maven artefacts) remain publicly available.
+- David Kornel will be granted component owner access to the repository.
+  Jakub Stejskal is already a Strimzi maintainer and requires no additional access.
+- The shared SPI module (`strimzi-mcp-common`) will be published to Maven Central so that third-party providers can be built without forking.
+- The pluggable SPI interfaces (`MetricsProvider`, `LogCollectorProvider`) remain extensible contracts with possible improvements/replacements in the future.
```
scholzj: While I do not necessarily have any intention to change these right now, we cannot accept any condition like this that bind us forever to some specific structure, names, publishing mechanisms, etc.

157-mcp-servers.md
```diff
@@ -0,0 +1,201 @@
...
+### Governance and CI/CD
+
+The `mcp-servers` repository adopts the standard Strimzi governance model, including the same maintainer and approver structure, code of conduct, and contribution process.
+The repository also adopts Strimzi CI/CD tooling with the minimal necessary changes to accommodate the Maven multi-module build and the system tests that require a Kubernetes cluster.
+
+### Conditions of transfer
+
+The StreamsHub maintainers approved the donation subject to the following conditions, all of which this proposal accepts:
+
+- All released artefacts (container images and Maven artefacts) remain publicly available.
```
scholzj: Same as below. While I do not necessarily have any intention to change these right now, the assumption that the project will forever publish Maven artifacts and container images is IMHO not acceptable. The ownership of the project and its architecture decisions has to be transferred as well with it. Without it is it has no real future.

157-mcp-servers.md
```diff
@@ -0,0 +1,201 @@
...
+- David Kornel will be granted component owner access to the repository.
+  Jakub Stejskal is already a Strimzi maintainer and requires no additional access.
```
scholzj: Putting it into conditions makes it questionable with regards to the future. E.g. does it invalidate the _donation_ if we kick David out a day after this proposal is accepted? I think David should be shifted out of the _conditions_ and added as a separate point of the proposal. Jakub should not be mentioned as he already is a maintainer.

Also, the right wording would be something like ...
```
David Kornel will be names as one of the compoenent owners responsible for the development of the Strimzi MCP compoenent.
```

It is about the community role, not about the access level to the repository. That is not what should be defined in this proposal as that is role of Strimzi governance.

scholzj (commented):
Thanks for the proposal. I left some comments. I think it might be also useful to add more about the plan for:

* The docs
* The website

(i.e. how will it be integrated into the Strimzi docs and website)

157-mcp-servers.md
```diff
@@ -0,0 +1,201 @@
...
+
+**MCP Tools** provide read-only structured access to Strimzi infrastructure.
+The tools cover all supported Strimzi custom resources — Kafka, KafkaNodePool, KafkaTopic, KafkaUser, KafkaConnect, KafkaConnector, KafkaBridge, KafkaMirrorMaker2, and KafkaRebalance — as well as the Strimzi Cluster Operator and the Strimzi Drain Cleaner.
+Every resource type has list and get tools.
+Resource types that own a workload additionally have pod and log retrieval tools.
+Metrics tools expose Prometheus-format metrics from Kafka broker and controller pods, KafkaConnect, KafkaBridge, KafkaExporter, and the Strimzi operators.
+Composite diagnostic tools use MCP Sampling and Elicitation to orchestrate multi-step investigations in a single tool call.
+
+**MCP Resources** expose live Kubernetes state as structured context that clients can attach directly to conversations without explicit tool calls.
+Resource templates follow the Kubernetes API URI hierarchy and cover Kafka cluster status, cluster topology, KafkaNodePool status, KafkaTopic status, KafkaUser status, and Strimzi operators status.
```
ppatierno: are all the other resources status not covered or just missing here? For example, `KafkaConnector` and `KafkaRebalance`?

157-mcp-servers.md
```diff
@@ -0,0 +1,201 @@
...
+| `troubleshoot-connect`             | Diagnoses a KafkaConnect cluster that is not ready or failing to run connectors — checks Connect CR status, pod health, and Connect worker logs.                                                            |
+| `troubleshoot-connector`           | Investigates a specific KafkaConnector that is in a failed or paused state — checks connector status, task failures, and the parent KafkaConnect worker logs.                                               |
+| `troubleshoot-bridge`              | Diagnoses a KafkaBridge deployment that is not ready or returning errors — checks Bridge CR status, pod health, and Bridge pod logs.                                                                        |
+| `troubleshoot-mirror-maker`        | Investigates a KafkaMirrorMaker2 instance with replication problems — checks MM2 CR status, connector states, pod health, and MM2 pod logs.                                                                 |
```
ppatierno, line 81: a `troubleshoot-rebalance` could be also interesting here.

157-mcp-servers.md
ppatierno: I had the same question in mind but I think the answer is in the door open to add additional MCP servers like the mentioned Kafka one. I am not saying it's right, it's just my explanation why they were using the plural.

157-mcp-servers.md
ppatierno: I was also wondering if we should keep `strimzi` in the name, given we are trying to remove it from components (i.e. metrics-reporter).

157-mcp-servers.md
ppatierno: Or maybe this should not be mentioned at all as a condition and later on a vote could start within maintainers to nominate David as component owner? (which I am fine with)

157-mcp-servers.md
Frawless: Yes, I will change the tree a little bit.

157-mcp-servers.md
Frawless: I actually meant it as complementary to it. Like now we have just Strimzi MCP that handles Strimzi specific resources, but for some operations it would be great to have also Kafka MCP that will operate next to Strimzi one and LLM will coordinate both.

157-mcp-servers.md
Frawless: The reason behind this was to not use external CRUD MCPs and let most of the operations in our hands. We wanted to have just read operators (or eventually CRUD jus tin some dev mode). With having our providers for log/metrics collection we can handle sensitve data if there are any. We also filter out unnecessary items, but I guess this can be handled by external MCPs as well as long as Strimzi MCP will query them, otherwise the context will get polluted pretty quickly with a lot of metrics/logs.

157-mcp-servers.md
Frawless: Correct. 

At some point we would like to implement [user impersonation](https://kubernetes.io/docs/reference/access-authn-authz/user-impersonation/) and validate access rights against external oauth server so the MCP could have cluster wide access, but users will get data only about items they have access to.

157-mcp-servers.md
Frawless: We used `streamshub-` prefix in provider name to allow users who will implement their own solutions to easily distinguish between our implementation and their implementation. I have no strong opinion on module names, but I would keep/add prefix in the names. I more like `strimzi-mcp-metrics-prometheus-provider` which means `strimzi-mcp-` prefix.

157-mcp-servers.md
Frawless: Exactly as Paolo saying. That is the idea behind it.

157-mcp-servers.md
kornys: It is basically a filter chain that wraps all our MCP tool calls. Every tool class is annotated with `@Guarded`, which triggers a CDI interceptor that runs a bunch of small filter beans in order, some before the tool executes and some after.
Each filter handles one thing: -> rate limiting (per category, off by default), stripping control characters from inputs, redacting secrets like tokens and passwords from the output, truncating responses that get too big, and recording metrics. They apply in priority order, so e.g. redaction happens before size-limiting.
The main idea is that each concern is its own isolated, testable bean instead of being scattered across the tools. If we want to add another guardrail later, it is just a new filter with a priority.

157-mcp-servers.md
Frawless: Guess it is just strange wording, I will reword it a little bit....in general it is name of the provider in MCP config. Such us `streamshub-metrics-prometheus` for scraping the metrics.

157-mcp-servers.md
Frawless: I will change the wording. It is not meant like it can't change in the future. It is more like `let's do not change it in the first release and then things can change`. I think the only must-have condition is to keep everything public, it doesn't matter how, in which format, and to where it will be published, but this is not a problem.

157-mcp-servers.md
Frawless: I will update the wording + add note about standard voting process we have.

157-mcp-servers.md
Frawless: They just missing in the list. I will update the proposal.

157-mcp-servers.md
Frawless: Definitely. We can open enhancement issue for later.

157-mcp-servers.md
kornys: strimzi-view only covers the Strimzi custom resources: read (get/list/watch) on kafkas, kafkaconnects, kafkaconnectors, kafkamirrormaker2s, kafkabridges, kafkarebalances, kafkatopics, kafkausers, kafkanodepools, and strimzipodsets. Our role keeps all of that unchanged and adds the plain Kubernetes resources the MCP server needs to inspect a cluster.
  - apps/deployments (get/list/watch) — operator deployments
  - pods, services (get/list)
  - pods/log (get) — pod logs
  - events (list) — used with field selectors
  - configmaps (get/list) — logging overrides / external config refs
  - route.openshift.io/routes + networking.k8s.io/ingresses (get/list) — listener endpoints
  - coordination.k8s.io/leases (get/list) — operator leader-election status
  - admissionregistration.k8s.io/validatingwebhookconfigurations (get/list) — Drain Cleaner

157-mcp-servers.md
kornys: The two bindings just stack. Kubernetes RBAC is additive, so the service account ends up with the union of both. We bind the main role cluster wide because it's all read only and nothing in it is sensitive (the CRs, Pods, logs, events), and that's exactly the stuff you want to be able to read anywhere for discovery. The sensitive role is bound per namespace, so secrets and pods/proxy only apply where you actually apply the RoleBinding, normally the namespaces where your Kafka clusters live. There's no conflict between the two. If you don't apply the sensitive role, the cert and pod scraping features just fail closed in that namespace and everything else keeps working.

Secrets are namespace scoped on purpose because that's the only thing with real key material in it. The cert tooling only does a get on two secrets per cluster (the cluster CA and clients CA) to report things like expiry, so keeping it per namespace means you only expose secrets in the namespaces you choose.

pods/proxy is only there for the built in pod scraping metrics provider. It reads each broker's Prometheus endpoint through the API server's pod proxy (/api/v1/namespaces/{ns}/pods/{pod}:{port}/proxy/metrics), and that call needs get on pods/proxy. Since it always targets specific pods it's naturally namespace scoped too. If you point the server at an external Prometheus instead, it never scrapes pods directly, so you don't need pods/proxy at all, which is why it lives in the opt in role rather than the cluster wide one.

157-mcp-servers.md
kornys: well from my POV this sentence means that we do not remove released artifacts.

157-mcp-servers.md
scholzj: So, my main point I was trying to make is that we need some sort of namespacing. That can happen by everything beeing in some shared group (e.g. `io.strimzi.mcp`) or through prefix (e.g. `strimzi-mcp-...` or `mcp-...`).

157-mcp-servers.md
scholzj: But the parent has many children and only one of them seems to be an actual server. So that is why I also suggested to make the parent more neutral and name the children server when it is a server.

157-mcp-servers.md
scholzj: Well ... we do not control the already released artifacts, so we cannot remove them if those were meant. For the future artifacts released under Strimzi, we rely on free services, so it is hard to guarantee much. As a CNCF the guarantees we provide are in general around sources and their license. Not about availability of any binaries.

157-mcp-servers.md
scholzj: That seems pretty pointless as all the package name changes would cause major disruptions in that release anyway. But yes, that can be probably agreed on as it is clearly limited in scope.

157-mcp-servers.md
scholzj: Actually, things such as resource names are sensitive in many cases as well. But I wonder more about how the MCP itself handles it when it has access to different level of detail for different clusters / namespaces.

 > pods/proxy is only there for the built in pod scraping metrics provider. It reads each broker's Prometheus endpoint through the API server's pod proxy (/api/v1/namespaces/{ns}/pods/{pod}:{port}/proxy/metrics), and that call needs get on pods/proxy. Since it always targets specific pods it's naturally namespace scoped too. If you point the server at an external Prometheus instead, it never scrapes pods directly, so you don't need pods/proxy at all, which is why it lives in the opt in role rather than the cluster wide one.
 
I thought it is reading the metrics from the metric collectors that are part of the project.

157-mcp-servers.md
scholzj: I think those are all reasonable permissions. But not something I would imagine as `modelled on the existing `strimzi-view` ClusterRole`. That would meam to me more like removing the `watch` RBAC or something.

157-mcp-servers.md
scholzj: My concern is that this obviously makes testing harder. It also creates an unbounded are for users requesting new plugins to support new things. That is where I think relying on other MCPs instead makes more sense as you can focus on a single thing. I suspect this might also cause issues when a user wants to use Strimzi MCP with their own existing MCP for metrics or logs instead of them writing their own plugin. But I guess we can leave that out of the initial contribution.

157-mcp-servers.md
scholzj: Sounds good. Might be worth adding to the proposal.

157-mcp-servers.md
kornys: Metrics collector is plugable so if someone for testing purposes cant use prometheus then he can use pods scraping in mcp. 
About mcp behavior, if it does not have proper rights for some stuff it skips it and tell LLM that for example secrets cant be read due to missing rights

157-mcp-servers.md
kornys: @Frawless Can you add that into proposal when you will fix other stuff after first round of comments?

Frawless: Yes, I will add it

157-mcp-servers.md
Frawless: Solution for that might be some no-op provider to leave it on external tools. We currently cover all providers by automated e2e tests.

157-mcp-servers.md
Frawless: That sounds reasonable. Having parent as `strimzi-mcp` and server itself as `strimzi-mcp-server` sounds reasonable to me. @kornys wdyt?

kornys: yeah, Im fine with that

Frawless:
I pushed changes for most of the comments, please a take look again once you will some free time.

kornys:
@Frawless Not sure, but shouldn’t the proposal also include the rejected alternatives?

Frawless:
> @Frawless Not sure, but shouldn’t the proposal also include the rejected alternatives?

I can add it, but when I was writing the proposal it sounded me a little bit strange to add rejected alternatives when the proposal is about code donation.

